PR #17, opened 2021-11-15
Support empty and inheritance class declarations (close #14)

### LINKED ISSUE #14: Support inheritance in JS code-gen
```scala
class A: {}
class B: A
```

```
Unexpected error: scala.NotImplementedError: an implementation is missing
```

This should do the obvious thing, using JS class extension, so that it also works in pattern matching.

chengluyu:
Basically, in three lines: https://github.com/hkust-taco/mlscript/blob/6cf565b52cb0fbd31df0d5f23b108449c1d0df49/shared/src/main/scala/mlscript/JSBackend.scala#L164-L169

Question: how do we support inheritance with additional fields. I tried to add something like this.

```scala
      case Union(Record(fields), Primitive(base)) => // bla bla bla
      case Union(Record(base), Primitive(fields)) => // bla bla bla
```

But extending union types is not allowed.


js/src/main/scala/Main.scala
```diff
@@ -42,7 +42,6 @@ object Main {
           "Parse error: " + extra.trace().msg +
             s" at line $lineNum:<BLOCKQUOTE>$lineStr</BLOCKQUOTE>"
         case Success(pgrm, index) =>
-          println(s"Parsed: $pgrm")
```
LPTK: Is this removal on purpose? I find it quite useful to debug what's going on, when a problem occurs.

js/src/main/scala/Main.scala
```diff
@@ -76,7 +75,6 @@ object Main {
                 |    }
                 |  }
                 |})()""".stripMargin
-              println("Running code: " + code)
```
LPTK: Same question here.

shared/src/main/scala/mlscript/Codegen.scala
```diff
@@ -250,7 +250,7 @@ object JSBinary {
       "instanceof" -> 12,
       "==" -> 11,
       "&&" -> 7,
-      "||" -> 7,
+      "||" -> 7
```
LPTK: Can we leave ending commas? I find them more convenient to edit and consistent. If you're using scalafmt: `trailingCommas =  always`.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -62,13 +62,17 @@ class JSBackend {
...
+      "add" -> "+",
+      "sub" -> "-",
+      "mul" -> "*",
+      "div" -> "/",
       "lt" -> "<",
       "le" -> "<=",
       "gt" -> ">",
       "ge" -> ">=",
       "eq" -> "==",
-      "ne" -> "!=",
```
LPTK, line 71: Same remark as above.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -198,17 +209,16 @@ class JSBackend {
...
           // Generate something like `exprs.push(<expr>)`.
-          .concat(otherStmts.zipWithIndex.collect {
-            case (term: Term, index) =>
-              new JSExprStmt(
-                JSInvoke(
-                  JSMember(JSIdent(exprResultObjName), "push"),
-                  translateTerm(term)
-                )
+          .concat(otherStmts.zipWithIndex.collect { case (term: Term, index) =>
+            new JSExprStmt(
```
LPTK: Nitpick: from a code style point of view, it would be good to remove all these useless `new` keywords.

LPTK (changes requested):
Inheritance with additional fields is done through type intersection, not union!
Also, the way to properly implement this is to have a recursive function that traverses the given types while accumulating the list of fields. Indeed, the parent type may be an arbitrarily nested structure, such as `{ x: int } & A & { y: string }`.
Moreover, the current inheritance semantics is to require for the class constructor to inherit field required in parent classes. And one can also use aliases in parent specifications. So your function should also traverse parent classes and aliases to find the fields defined in them. **[EDIT: now that I think of it, you only need to traverse aliases and traits; base classes will handle setting their own fields in the super constructor.]**
For example, consider:
```scala
type Helper = { y: int }
class A: { x: int }
class B: A & Helper
class C: B
// here constructor `C` requires both fields `x` and `y`
C { x = 1; y = 2 }
```

chengluyu:
Yes. I was just realized that I didn't handle initialization of base classes. Wait for my more commits!

chengluyu:
> Should probably pass the fields to `super`?

Yes. I’m doing it.

js/src/main/scala/Main.scala
chengluyu: It’s accidental. I was confused by a lot of outputs in console.

js/src/main/scala/Main.scala
chengluyu: Also accidental.

shared/src/main/scala/mlscript/Codegen.scala
chengluyu: Sure.

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: Yes. They looks ugly. I was thinking about adding `apply` to all classes. Now it’s time to do it.

LPTK: You don't need to! The `apply` method is auto-generated for case classes!

chengluyu: 🥲 I thought I should implement them. Thanks for pointing out!

shared/src/main/scala/mlscript/Codegen.scala
```diff
@@ -382,27 +382,42 @@ final case class JSConstDecl(pattern: Str, body: JSExpr) extends JSStmt {
...
+    base match {
+      case Some(baseCls) =>
+        val inheritedFields = baseCls.fields.filterNot { case name =>
+          fieldsSet contains name
+        }
+        val items = baseCls.fields.map(x => s"$x: fields.$x").mkString(", ")
+        val prologue =
+          s"class $name extends ${baseCls.name} {" ::
+            "  constructor(fields) {" ::
+            s"    super({ $items });" ::
```
LPTK: I think you can just pass `fields` directly. There should be no need to create another record object.

chengluyu:
I’m trying support extending intersection types. I added these lines.

```scala
      case Inter(Primitive(clsName), Record(fields)) => 
        nameClsMap get clsName match {
          case N      => ???
          case S(cls) => JSClassDecl(name, fields map { _._1 }, S(cls))
        }
```

And it can translate

```
class A: { x: int }
class B: A & { y: str }
```

into

```js
class A {
  constructor(fields) {
    this.x = fields.x;
  }
}

class B extends A {
  constructor(fields) {
    super(fields);
    this.y = fields.y;
  }
}
```

I’m trying to support something like `class B: { a: int } & A & { y: str }`.

shared/src/main/scala/mlscript/Codegen.scala
chengluyu: I set `trailingCommas` to `keep` in order to prevent introducing too many changes.

LPTK: Ok 👍

chengluyu:
Now it is possible to extend a series of intersection types.

```
class A: {}
class B: A & { x: int } & { y: int }
```

will generates

```js
class A {}
class B extends A {
  constructor(fields) {
    super(fields);
    this.x = fields.x;
    this.y = fields.y;
  }
}
```

LPTK:
Cool, but it doesn't work with parameterized types, such as the example in the `local_testing.html` file:

```scala
class Left[A]: { value: A }
class Right[A]: { value: A }
```

chengluyu:
Now we can translate

```
class A
type C = B 
type B = A
class D: { y: int } & C & { x: int }
```

into

```js
class A {}
// // type alias C
// // type alias B
class D extends A {
  constructor(fields) {
    super(fields);
    this.y = fields.y;
    this.x = fields.x;
  }
}
```

Note: that force-push is caused by rebase.

LPTK:
And don't forget that you need to support the `AppliedType` type form for things like `Left[A]`! Just throw away the type variable if `Left`  is a class type, but if it's a type alias or trait you need to actually substitute `A` in its body!

This is legal:
```scala
class A
type Id[T] = T
class B: Id[A]
```

To do this, add a map parameter to `expandRecordInter` to keep track of what type variables are referring  to.

LPTK:
Sorry this keeps getting bigger, like a rabbit hole :^)

chengluyu:
> Sorry this keeps getting bigger, like a rabbit hole :^)

Yes. 😅 I’m working on substitution. I will push the change later.

chengluyu:
Now we can translate

```
class A
type B = Id[A]
type Id[T] = T
type Identity[T] = Id[T]
type C = Identity[Identity[B]]
class Z: Id[C]  
```

into

```js
class A {}
// type alias B
// type alias Id
// type alias Identity
// type alias C
class Z extends A {
  constructor(fields) {
    super(fields);
  }
}
```

Cheers!

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -224,48 +286,43 @@ class JSBackend {
...
-      case _ => ???
     }
+  }
 
   def apply(pgrm: Pgrm): Ls[Str] = {
-
     val (diags, (typeDefs, otherStmts)) = pgrm.desugared
-    resolveAliases(typeDefs)
+
+    // Collect type aliases into a map so we can evaluate them.
```
LPTK: I don't think `evaluate` is really the right terminology. Maybe just say "normalize"?
```suggestion
    // Collect type aliases into a map so we can normalize them.
```

chengluyu:
@LPTK Do you have idea why CI complains about the use of `Applied`? Tests and build work fine on my local computer.

fo5for:
> @LPTK Do you have idea why CI complains about the use of `Applied`? Tests and build work fine on my local computer.

Probably because `Applied` is removed in bc8e42f017403cfa47448466b6cb42953cdb4d2a and CI is merging automatically before testing.

chengluyu:
> > @LPTK Do you have idea why CI complains about the use of `Applied`? Tests and build work fine on my local computer.
> 
> Probably because `Applied` is removed in [bc8e42f](https://github.com/hkust-taco/mlscript/commit/bc8e42f017403cfa47448466b6cb42953cdb4d2a) and CI is merging automatically before testing.

Thanks!

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -164,52 +164,114 @@ class JSBackend {
...
+                )
             }
-          }
-          name -> current
-      }
-    )
+          case S(ty) => ty
+        }
+      // For `Bot`, `Literal`, and `Top`, we don't need to substitute.
+      case _ => body
```
LPTK: See the comment I made to Tony a few days ago here: https://github.com/hkust-taco/mlscript/commit/e49c62e958224b9b1532a30fd45ac0909086fcbd#r60500415

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -164,52 +164,114 @@ class JSBackend {
...
+  // This function evaluates something like `T[A, B]`.
+  private def applyTypeAlias(name: Str, targs: Ls[Type]): Type =
+    typeAliasMap get name match {
+      case S(tparams -> body) =>
+        if (targs.length === tparams.length) {
+          val subs = collection.immutable.HashMap(
+            tparams zip targs map { case (k, v) => k -> v }: _*
+          )
+          substitute(body, subs)
+        } else {
```
LPTK: This whole verbose conditional should probably just be an `assert(targs.length === tparams.length, targs -> tparams)` at the beginning.

LPTK (approved):
Okay! Besides the minor comments I made, this looks good to me. Feel free to merge after making the minor changes ;^)

LPTK:
PS: use the "squash and merge" method.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -219,8 +216,12 @@ class JSBackend {
             }
           case S(ty) => ty
         }
-      // For `Bot`, `Literal`, and `Top`, we don't need to substitute.
-      case _ => body
+      case Bot => body
```
LPTK: Would be better as: `case Bot | Top | _: Literal | _: TypeVar => body`

PR #18, opened 2021-11-17
Fix JavaScript runtime errors when shadowing names (close #13)

### LINKED ISSUE #13: Support shadowing in JS code-gen
```scala
def tmp = 1
def tmp = succ tmp
```

```
Runtime error occurred:
SyntaxError: Identifier 'tmp' has already been declared
```

chengluyu:
This PR proposes a simple change resolving name shadowing problems in generated JavaScript code.

But this introduce another problem on showing evaluation results. For example, code

```
def tmp = 1
def tmp = add tmp 1
```

will have results

```
val tmp: 1 = 2
val tmp: int = 2
```

I will introduce another fix or refactor evaluation result collection.


chengluyu:
This PR is ready for review.

js/src/main/scala/Main.scala
```diff
@@ -100,9 +100,26 @@ object Main {
...
+                        nameShadowingCount += name -> (count + 1)
+                        s"$name@$count"
+                      // This is the first time we meet this name.
+                      case None =>
+                        nameShadowingCount += name -> 1
+                        name
+                    }
+                    // Get the evaluation results.
+                    defResults get realName
+                  // This type definition is a expression. (No name)
```
LPTK:
```suggestion
                  // This type definition is an expression. (No name)
```

js/src/main/scala/Main.scala
```diff
@@ -144,7 +161,7 @@ object Main {
 
   def checkProgramType(pgrm: Pgrm): Ls[Option[Str] -> Str] -> Option[Str] = {
     val (diags, (typeDefs, stmts)) = pgrm.desugared
-
```
LPTK: Is there a scalafmt option not to remove trailing indentation spaces? Removing them makes editing harder for those not using an automated formatter.

js/src/main/scala/Main.scala
```diff
@@ -289,14 +306,24 @@ object Main {
           ctx += nme -> ty_sch
           declared.get(nme).foreach { sign =>
             // ctx += nme -> sign  // override with less precise declared type?
-            subsume(ty_sch, sign)(ctx, raise, TypeProvenance(d.toLoc, "def definition"))
+            subsume(ty_sch, sign)(
```
LPTK: If you're using scalafmt, please change your column width setting to at leas 100. Reformatting this needlessly is a bit ridiculous.

js/src/main/scala/Main.scala
```diff
@@ -289,14 +306,24 @@ object Main {
...
+              TypeProvenance(d.toLoc, "def definition")
+            )
           }
           res ++= formatBinding(d.nme, ty_sch)
           results append S(d.nme) -> (getType(ty_sch).show)
         case d @ Def(isrec, nme, R(PolyType(tps, rhs))) =>
           val errProv = TypeProvenance(rhs.toLoc, "def signature")
-          val ty_sch = PolymorphicType(0, typeType(rhs)(ctx.nextLevel, raise,
-            tps.map(tp => tp.name -> freshVar(noProv/*FIXME*/)(1)).toMap))
+          val ty_sch = PolymorphicType(
```
LPTK: Same remark as above.

LPTK (approved):
Okay, this looks good enough to me. Please merge after having made the requested minor changes.

It's unfortunate that you had to duplicate the renaming logic. This is a prime spot where bugs can creep in. It would be a better design to rename things only once and keep a mapping to their real names around, or to encapsulate the renaming logic into a single renamer abstraction that you would use twice, ensuring it really does its job the same each time, even in the face of future changes.

But let's merge this for now, as it's quite urgently needed.

js/src/main/scala/Main.scala
chengluyu: Reverted.

js/src/main/scala/Main.scala
chengluyu: Reverted.

js/src/main/scala/Main.scala
chengluyu: I didn't find related options. Reverted.

js/src/main/scala/Main.scala
chengluyu: Corrected.

chengluyu:
> It's unfortunate that you had to duplicate the renaming logic. This is a prime spot where bugs can creep in. It would be a better design to rename things only once and keep a mapping to their real names around, or to encapsulate the renaming logic into a single renamer abstraction that you would use twice, ensuring it really does its job the same each time, even in the face of future changes.

I extract the renaming logic into a class `ShadowNameResolver`. Not sure if this is a good name.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -342,13 +344,36 @@ class JSBackend {
...
+              } else {
+                // Re-assign and record the value as a new name:
+                // ```
+                // <tempName> = <expr>;
+                // defs["<name>@<number of shadow>"] = <tempName>;
+                // ```
+                JSExprStmt(JSAssignExpr(JSIdent(tempName), translatedBody)) ::
+                  JSExprStmt(
+                    JSAssignExpr(
+                      // We use @ because it is not valid in identifiers.
```
```suggestion
```

shared/src/main/scala/mlscript/utils/CodegenHelpers.scala
```diff
@@ -0,0 +1,20 @@
...
+
+class ShadowNameResolver {
+  private val nameShadowingCount = collection.mutable.HashMap[Str, Int]()
+
+  def apply(name: Str): Str = {
+    nameShadowingCount get name match {
+      // It is a shadowed name.
+      case Some(count) =>
+        nameShadowingCount += name -> (count + 1)
+        s"$name@$count"
```
LPTK, line 14:
```suggestion
        // We use @ because it is not valid in identifiers.
        s"$name@$count"
```

chengluyu:
Just saw the update. Can we squash and merge now?

LPTK:
Yes. I had to fix a conflict because you changed the indentation in the other PR 😅

chengluyu:
> Yes. I had to fix a conflict because you changed the indentation in the other PR 😅

Ah. Sorry about this. I will tune my scalafmt, making it doing less dramatic changes next time.

LPTK:
No worries, these things happen.

PR #22, opened 2021-11-24
Implement method parsing and type checking (closes #9)

### LINKED ISSUE #9: Implement method parsing and type checking
As in the later versions of the paper. This is just an issue to track the progress of the current implementation in https://github.com/hkust-taco/mlscript/tree/methods

fo5for:
Implemented:
- Method declaration and definition processing, including typing and subsumption checking
- Explicit and implicit method calls
- Shadowing for implicit method calls in definitions
- `this` identifier with shadowing semantics in method definitions

Issues:
- `ConstraintSolver.extrude` does not handle the polarity of type variables in `TypeRef`s
- Type simplification ignores type parameters of classes
- `DiffTests` outputs the defined method type for methods declaration if the same method is both declared and defined in the class
- To prohibit construction of abstract classes


shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -646,10 +819,10 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
           case ((N, t), i) =>
             // err("Missing name for record field", t.prov.loco)
             warn("Missing name for record field", t.prov.loco)
-            ("_" + (i + 1), t)
+            (Var("_" + (i + 1)), t)
         }.toList
         RecordType.mk(fs)(prov)
-      } else TupleType(fields.reverse)(prov)
+      } else TupleType(fields.map { case S(n) -> t => S(n) -> t case N -> t => N -> t }.reverse)(prov)
```
LPTK: Isn't this `map` doing nothing?

shared/src/test/diff/mlscript/Methods.mls
```diff
@@ -0,0 +1,126 @@
+//TODO: - ConstraintSolver.extrude TypeRef polarity
+//      - ConstraintSolver.constrain.annoyingImpl missing cases?
```
LPTK: So, have you found any of these missing cases again? Would be good to add a test exercising it to keep track.

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -396,12 +396,15 @@ class ConstraintSolver extends NormalForms { self: Typer =>
       case e @ ExtrType(_) => e
       case p @ ProxyType(und) => ProxyType(extrude(und, lvl, pol))(p.prov)
       case _: ClassTag | _: TraitTag => ty
-      // case TypeRef(d, ts) => TypeRef(d, ts.map(extrude(_, lvl, pol))) // FIXME pol...
+      case tr @ TypeRef(d, ts) => TypeRef(d, ts.map(extrude(_, lvl, pol)))(tr.prov, tr.ctx) // FIXME pol...
```
LPTK, line 399: Let's open an issue to keep track of this wrong part as it could later cause bugs.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -169,7 +169,7 @@ class JSBackend {
       JSConstDecl(defResultObjName, JSRecord(Nil)) ::
         JSConstDecl(exprResultObjName, JSArray(Nil)) ::
         typeDefs
-          .map { case TypeDef(kind, Primitive(name), typeParams, actualType) =>
+          .map { case TypeDef(kind, Primitive(name), typeParams, actualType, _, _) => // TODO: handle methods
```
LPTK, line 324: Let's open an issue for this too. Can you do it, plz?

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -272,16 +297,131 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+            }
+        }
+        val thisTy = TypeRef(td, td.tparams.map(p => allTargsMaps(tn.name)("")(p.name)))(prov, ctx)
+        newEnv ++= (decls ++ defs).flatMap{ case mn -> PolymorphicType(level, body) => 
+          val m_ty = PolymorphicType(level, FunctionType(thisTy, body)(prov))
+          s"${tn.name}.${mn}" -> m_ty ::
+            (ctx.methodBase.get(mn) match {
+              case S(S(v)) if ctx.tyDefs.get(tn.name).exists(_.allBaseClasses(ctx)(Set.empty).contains(v)) => Nil
+              case S(S(v)) if ctx.tyDefs.get(v.name.toList.mapHead(_.toUpper).mkString)
+                  .exists(_.allBaseClasses(ctx)(Set.empty).contains(Var(tn.name.toList.mapHead(_.toLower).mkString))) =>
```
LPTK: A remark: this logic is a little excessively inefficient. `allBaseClasses` traverses the whole type every time, and we may call it a lot here. Would be good to cache it in the contexts. But we could do that later.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -272,16 +297,131 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+        newMthDefs += tn.name ->
+          ((td.baseClasses.flatMap(t => newMthDefs.getOrElse(t.name, Map.empty))).toMap ++ defs.toMap)
+      }
+      ctx.copy(mthDecls = newMthDecls.toMap, mthDefs = newMthDefs.toMap)
+    } |> { implicit ctx =>
+      val newEnv = MutMap.from(ctx.env)
+      val newMthDecls = ctx.mthDecls
+      val newMthDefs = ctx.mthDefs
+      newDefs.foreach { td =>
+        implicit val prov: TypeProvenance = tp(td.toLoc, "data definition")
```
LPTK: Shouldn't this say "type definition"?

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -145,8 +159,19 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
       case _ => Set.empty // TODO TupleType?
     }
   
+  def targsMapOf(ty: Type)(implicit ctx: Ctx, raise: Raise, lookup: Option[(Str, Option[Str])] = N,
+      vars: Map[Str, SimpleType] = Map.empty): Map[Str, SimpleType] = ty match {
+    case Inter(l, r) => targsMapOf(l) ++ targsMapOf(r)
+    case AppliedType(b, ts) => ctx.tyDefs.get(b.name).fold(Map.empty[Str, SimpleType]) {
+      td => td.tparams.map(_.name).zip(ts.map(typeType(_)(ctx.nextLevel, raise, lookup, vars))).toMap
+    }
+    case _ => Map.empty
```
LPTK: Same remark as before. It's not clear why ignoring the other cases is okay here. An exhaustive list of the other cases as well as a comment documenting why we can ignore them would be warranted.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -272,16 +297,131 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+      newDefs.foreach { td =>
+        implicit val prov: TypeProvenance = tp(td.toLoc, "data definition")
+        val tn = td.nme
+        val decls = newMthDecls(tn.name)
+        val defs = newMthDefs(tn.name)
+        val rigidSubsMap = ctx.getTargsMaps(td.nme.name, "").getOrElse(Map.empty).collect { 
+            case s -> (tv: TypeVariable) => tv -> freshenAbove(ctx.lvl, tv, true)
+          }.toMap[SimpleType, SimpleType]
+        val targsMap = targsMapOf(td.body)(ctx, raise, S((tn.name, N)))
+        val subsMap = td.baseClasses.flatMap(b => ctx.getTargsMaps(b.name, "")).flatten
```
LPTK: Same remark as below. Flattening collections of Map-s may override keys.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -145,8 +159,19 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
       case _ => Set.empty // TODO TupleType?
     }
   
+  def targsMapOf(ty: Type)(implicit ctx: Ctx, raise: Raise, lookup: Option[(Str, Option[Str])] = N,
+      vars: Map[Str, SimpleType] = Map.empty): Map[Str, SimpleType] = ty match {
+    case Inter(l, r) => targsMapOf(l) ++ targsMapOf(r)
```
LPTK: Is it intended that keys in the output of `targsMapOf(l)` may be overridden by `targsMapOf(r) `? Map's `++` is a little dangerous because of this.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -24,30 +24,44 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
     def get(name: Str): Opt[TypeScheme] = env.get(name) orElse parent.dlof(_.get(name))(N)
     def contains(name: Str): Bool = env.contains(name) || parent.exists(_.contains(name))
-    def nest: Ctx = copy(Some(this), MutMap.empty)
+    def nest: Ctx = copy(Some(this), MutMap.empty, targsMaps = MutMap.empty)
     def nextLevel: Ctx = copy(lvl = lvl + 1)
+    def getTargsMaps(cls: Str): Option[MutMap[Str, Map[Str, SimpleType]]] =
+      targsMaps.get(cls) orElse parent.dlof(_.getTargsMaps(cls))(N)
+    def getTargsMaps(cls: Str, mth: Str): Option[Map[Str, SimpleType]] =
+      getTargsMaps(cls).dlof(_.get(mth) orElse parent.dlof(_.getTargsMaps(cls, mth))(N))(N)
+    def getTargsMaps(cls: Str, mth: Str, targ: Str): Option[SimpleType] =
```
LPTK: It seems this overload is not used.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -155,21 +180,21 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
       ctx.copy(tyDefs = oldDefs ++ newDefs.flatMap { td =>
         implicit val prov: TypeProvenance = tp(td.toLoc, "data definition")
         val n = td.nme
-        val dummyTargs =
-          td.tparams.map(p => freshVar(noProv/*TODO*/)(ctx.lvl + 1))
-        val targsMap: Map[Str, SimpleType] = td.tparams.map(_.name).zip(dummyTargs).toMap
-        val body_ty = typeType(td.body, simplify = false)(ctx.nextLevel, raise, targsMap)
-        var fields = SortedMap.empty[Str, SimpleType]
+        val dummyTargs = td.tparams.map(p => freshVar(noProv/*TODO*/)(ctx.lvl + 1))
+        allTargsMaps += n.name -> MutMap("" -> td.tparams.map(_.name).zip(dummyTargs).toMap)
```
LPTK: There seems to be something unholy going on with **empty strings** here... can you document it? I guess it's for storing mappings for class parameters, associated with no methods. So if it's indeed used as a sentinel value, this would be better expressed with an option type.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -272,16 +297,131 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
           td.nme.name -> td :: Nil
         else Nil
       })
+    } |> { implicit ctx =>
```
LPTK: Okay, my original code was already borderline with its use of `} |> { implicit ctx =>`, but this is getting out of hands 😅 
We should split all these typing phases into properly-named local `def`s, ideally with one or two lines of comment each documenting what it's doing.

shared/src/main/scala/mlscript/Typer.scala
LPTK: I realize "data definition" was used before in this function, but that was probably a mistake.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -272,16 +297,131 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+      val newEnv = MutMap.from(ctx.env)
+      val newMthDecls = ctx.mthDecls
+      val newMthDefs = ctx.mthDefs
+      newDefs.foreach { td =>
+        implicit val prov: TypeProvenance = tp(td.toLoc, "data definition")
+        val tn = td.nme
+        val decls = newMthDecls(tn.name)
+        val defs = newMthDefs(tn.name)
+        val rigidSubsMap = ctx.getTargsMaps(td.nme.name, "").getOrElse(Map.empty).collect { 
+            case s -> (tv: TypeVariable) => tv -> freshenAbove(ctx.lvl, tv, true)
```
LPTK: How can it be okay to filter out only the cases where the mapped value is a `TypeVariable`? That's very fishy. Couldn't it be a `ProxyType` at some point (maybe not now, but if the code base changes in the future)? Or an application `Id[X]` of the `type Id[T] = T` type? If we're indeed only interested in type variables here, this should be expressed in the type, to make sure we're not messing up, and for documentation.

LPTK (changes requested):
This is amazing, thanks!

However, I've outlined several places where the code seems unclear or needs some documentation/refactoring. Let's make sure to fix these before the merge.

shared/src/main/scala/mlscript/Typer.scala
fo5for: ... yes. Originally when I changed records (and thus tuples) to used `Var`s for field names, I only changed them in `Type` but not `SimpleType`. So I needed to map the `Var`s back to strings. I later decided to also change them in `SimpleType`, and I only removed the `.name` here, leaving an identity map 😅 

Addressed in 193deba9ab7b3358683405b91fdcb2f2f4dc1c54

shared/src/test/diff/mlscript/Methods.mls
fo5for: No... I believe they appeared in ascriptions before they were properly handled. So they could be due to mishandling them, or simply a typo that has since been fixed.

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in 193deba9ab7b3358683405b91fdcb2f2f4dc1c54

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in 193deba9ab7b3358683405b91fdcb2f2f4dc1c54

shared/src/main/scala/mlscript/Typer.scala
fo5for: This function extracts the type arguments applied to the base class from the class body. Classes can only inherit from a conjunction of `Primitive`s, `AppliedType`s and `Record`s, and the only ones with applied type arguments are `AppliedType`s. I guess it wasn't clear that this is only supposed to be used on the body of a type definition.

The recursive function is now wrapped in a function taking a `TypeDef`, and the legal cases where an empty map should be returned are now listed.

shared/src/main/scala/mlscript/Typer.scala
fo5for: I wrote most of this PR without having traits in mind, so I was basically assuming single inheritance, and treating `TypeDef.baseClasses` more like an option than a set. If we decide to continue supporting traits, this can be changed into a nested map, one for each class / trait.

shared/src/main/scala/mlscript/Typer.scala
fo5for: Same as above. This can be changed to a nested map to support traits.

shared/src/main/scala/mlscript/Typer.scala
fo5for: Updated direct references to `Ctx.targsMaps` to use this instead in 193deba9ab7b3358683405b91fdcb2f2f4dc1c54

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in 193deba9ab7b3358683405b91fdcb2f2f4dc1c54

shared/src/main/scala/mlscript/Typer.scala
fo5for: `targsMaps` store the mapping from type parameters to the dummy `TypeVariable`s, and to the rigidified `TraitTag`a when typing methods. It was originally just a map to `TypeVariable`s before I realized the class type parameters should be rigidified when typing methods.

I couldn't come up with a way to express this invariant in the type. I tried to use `TypeVariable \/ TraitTag`, but it would require a use of `asInstanceOf`, as `freshenAbove` returns a `SimpleType` even when called directly with a `TypeVariable`. For now I have added 3 `die` statements before I can come up with a better solution.

I think your worry about the other cases came from nested `TypeRef`s or `ProxyType`s in the body. They are in fact not handled by this map, but by the `subsMap` a few lines below. The `TypeVariable`s we are collecting are the dummy type arguments created when typing the body of the `TypeDef`. The actually applied type arguments are on the rhs of `subsMap`. I believe it is handling nested `AppliedType`s properly, as they would be typed as `TypeRef`s by `targsAppOf` (formerly `targsMapOf`. I am running out of names...).

However, as I was writing tests to test this out, I discovered that nested type arguments are not typed correctly, even in the `mlscript` branch:
```scala
class Foo[A]: { value: A }
class Bar[A]: Foo[Foo[A]]
//│ Defined class Foo
```
The `TypeDef` of `Bar` is not registered to the context, even though no error are emitted. I believe this is due to `checkRegular` returning `false`.

shared/src/test/diff/mlscript/Methods.mls
LPTK: Cool, let's remove the comment then :^)

```suggestion
```

shared/src/main/scala/mlscript/Typer.scala
LPTK: Okay, but that doesn't work, even without traits, because types need to be inspected semantically, not syntactically:

```ts
// FIXME should raise error
:e
class Class3A[A]
    method MtdA: A
type Type3A = Class3A[string]
class Class3B: Type3A
    method MtdA = 1
//│ Defined class Class3A
//│ Declared Class3A.MtdA: class3A -> nothing
//│ Defined type Type3A
//│ Defined class Class3B
//│ Defined Class3B.MtdA: nothing -> 1
```
(On an unrelated note, the `Class3B.MtdA: nothing -> 1` is strange.)

The _source of truth_ for figuring out class parents is the `checkParents` function, which is the one that checks what is legal or illegal inheritance. It currently only returns a boolean because at the time we did not ned anything else, but if we need more info, we should get it from there. It's better than duplicating the parents-analysis logic and making non-local assumptions that are not obvious and may thus break in the future.

LPTK: On the other hand, `checkParents` does not traverse inherited parents. So I feel like extending `checkCycleComputeFields` to also glean method declarations and definitions from parents (both classes and traits) would be the better way to go. This performs a proper tree traversal, which is more appropriate for detecting problems like this: https://github.com/hkust-taco/mlscript/pull/22/commits/a1c3b47d22ee3fc62332250c5cfa96e3f1a3e694#diff-271f81eabb0cf04b1f9fe33d0620031626c2c2cb6c311ebafa625a54caf951cbR268

I think in the extended `checkCycleComputeFields` traversal, we could in a single go:

- accumulate a set of method declaration signatures from parent types, transitively, using the proper shadowing semantics; there might be several signatures coming from several class and trait parents, and the method definition will have to satisfy them all

- accumulate method definitions that do not have a signature, if there are any – these  will be used to block further overriding of the method

- compute the type arguments mapping on the fly: when traversing a typedef, we pass a map that we will use to `typeType` the method signatures we find, so that they are immediately instantiated at the right types

I understand this is quite a refactoring, and I actually think we should do this in a later PR, after merging this one, which has been pending long enough (I need to merge it to unblock some of my own changes, too).

shared/src/main/scala/mlscript/Typer.scala
LPTK: Okay, thanks for the explanation. Now I understand what you are doing, although I don't really agree with how it's done. See my other comment https://github.com/hkust-taco/mlscript/pull/22#discussion_r757856454 for an idea to make the whole thing more functional, relying on computing the relevant information once, at the right place. We'd simply pass that relevant info along with each type definition being processed (instead of storing it in the context through the weird `targsMaps` mutable map), so that we can later perform subsumption checking after typing methods. Does that make sense to you?

LPTK (approved):

LPTK:
I think we can merge this,  after you've addressed https://github.com/hkust-taco/mlscript/pull/22#discussion_r756714295 and https://github.com/hkust-taco/mlscript/pull/22#discussion_r757838926. Not sure why the CI is failing, though.

LPTK:
The CI was failing because of my parallel bug fix in https://github.com/hkust-taco/mlscript/commit/199e9371934a5fc5747fe033fcf474e28dbe06b5

LPTK:
Oh and also the selection-typing part should be improved as we discussed: https://github.com/hkust-taco/mlscript/pull/22/files#diff-09ce6a8849c3c07173c504d6966f025cfa3338fa80afedecb4e77c6c013d50d3R660
Notably, code like `err(msg"Class ${name} has no method ${fieldName.name}", term.toLoc)` should be deduplicated.

shared/src/test/diff/mlscript/Methods.mls
fo5for: Addressed in 68be5ec98a994c8360d5c552793a8fac415d8e88

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in f51f6f231a75d9ca2f52f1485b37b99d580eaaac

fo5for:
> Oh and also the selection-typing part should be improved as we discussed: https://github.com/hkust-taco/mlscript/pull/22/files#diff-09ce6a8849c3c07173c504d6966f025cfa3338fa80afedecb4e77c6c013d50d3R660 Notably, code like `err(msg"Class ${name} has no method ${fieldName.name}", term.toLoc)` should be deduplicated.

Addressed in bc397a97795021e356962da3ff03b6fb266dfc52

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in 6f66ec2b249323177e94b61b3cfc5e8d8dc43015

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in 6f66ec2b249323177e94b61b3cfc5e8d8dc43015

LPTK:
Okay thanks, that's really good work! The `processTypeDefs` implementation is a lot clearer now. I just pushed a few improvements, and we're ready to merge.

PR #29, opened 2021-11-30
Support binary operators in MLParser and web demo (finally!)

### LINKED ISSUE #15: Support basic operators in MLParser
Such as `2 + 2` instead of `add 2 2`.

LPTK:
Fixes #15

Still need the corresponding code-gen update. Cc @chengluyu.

### DIFF
--- a/shared/src/main/scala/mlscript/Codegen.scala
+++ b/shared/src/main/scala/mlscript/Codegen.scala
@@ -31,8 +31,7 @@ object SourceLine {
 
 class SourceCode(val lines: Ls[SourceLine]) {
 
-  /** Concat two parts of source code vertically. "1 \\ 2" + "3 \\ 4" == "1 \\ 2
-    * \\ 3 \\ 4"
+  /** Concat two parts of source code vertically. "1 \\ 2" + "3 \\ 4" == "1 \\ 2 \\ 3 \\ 4"
     *
     * @param that
     *   another part of source code
@@ -40,8 +39,7 @@ class SourceCode(val lines: Ls[SourceLine]) {
     */
   def +(that: SourceCode): SourceCode = new SourceCode(lines ++ that.lines)
 
-  /** Concat two parts of source code horizontally. "1 \\ 2" ++ "3 \\ 4" == "1
-    * \\ 2 3 \\ 4"
+  /** Concat two parts of source code horizontally. "1 \\ 2" ++ "3 \\ 4" == "1 \\ 2 3 \\ 4"
     *
     * @param that
     *   another part of source code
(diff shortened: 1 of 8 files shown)


shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -37,17 +37,16 @@ class JSBackend {
...
+        value match {
+          case Var(_) => code.replace("$", s"$key, ") -> (names :+ key)
+          case rcd: Rcd => {
+            val (subCode, subNames) = translateLetPattern(rcd)
+            code.replace("$", s"$key: ${subCode}, ") -> (names ++ subNames)
           }
+          // Is there any other cases?
+          case _ => ???
+        }
       }
```
LPTK: @chengluyu Ugh, I just noticed this part with the `$` substitutions. That's _very_ funky logic, let's not do that!! There are much cleaner ways of building this string (for example you can accumulate a _function_, instead of a plain string). BTW, `$` is a valid character in JS and Scala identifier names... Also, let's replace the YOLO `case _ => ???` with an enumeration of all terms form that can be here and a proper error reading "unsupported pattern shape" 🙂

chengluyu: Hi, I replaced the string replacement logic with AST construction.

LPTK:
@chengluyu Note for the future: it would have been better if you'd have made your own PR against my branch. This way I could have properly reviewed and squashed your changes. Here I can't do that because I can't review my own PR.

chengluyu:
Thanks for your suggestions! I’ve applied them.

1. Get rid of all `reduceLeft`.
2. Make code concise with `{ _ + _ }`.
3. Add `JSWildcardPattern` and translate literals in parameters to wildcards.
4. Enumerate all possible term types in the last case branch.

I’ve tested with cherry-picking commits for another PR. After this PR is merged I can merge that PR immediately.

shared/src/main/scala/mlscript/Codegen.scala
```diff
@@ -200,9 +205,15 @@ final case class JSPlaceholderExpr() extends JSExpr {
 final case class JSArrowFn(params: Ls[JSPattern], body: JSExpr) extends JSExpr {
   def precedence: Int = 22
   def toSourceCode: SourceCode =
-    ((params map { _.toSourceCode }) reduceLeft [SourceCode] { case (x, y) =>
-      x ++ SourceCode.from(", ") ++ y
-    }).parenthesized ++ SourceCode.from(" => ") ++ body.toSourceCode
+    params
+      .foldLeft(SourceCode.empty) { case (x, y) =>
+        x ++ SourceCode.from(", ") ++ (y match {
+          // JavaScript does not allow empty arguments, e.g. `function (,x) {}`.
```
LPTK: I don't understand what that comment is trying to say.
Won't your function wrongly generate `(,x) => ...` for 1-parameter lambdas?
Instead, it should generate either `(x) => x + 1` or `(x,) => x + 1` (both are valid).

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -63,8 +63,12 @@ class JSBackend {
     }
     // This branch supports `def f (x: int) = x`.
     case Asc(trm, _) => translatePattern(trm)
+    // Replace literals with wildcards.
+    case _: Lit => JSWildcardPattern()
     // Others are not supported yet.
-    case _ => throw new Error(s"term $t is not a valid pattern")
+    case _: Lam | _: App | _: Tup | _: Sel | _: Let | _: Blk | _: Bra | _: Asc | _: Bind | _: Test |
```
LPTK: `Asc` is already handled above.

LPTK (commented):
(Funnily, now GitHub lets me make a review, though it won't let me approve the PR.)

Okay, please make the necessary minor fixes and also fix the git conflict. Then I'll squash your commits and push them again.

shared/src/main/scala/mlscript/Codegen.scala
chengluyu: That comment is for the line below `case JSWildcardPattern() => SourceCode.from("_")`. Because we translate literals to wildcards, and JavaScript does not allow holes in parameter list, e.g. `function f(,,y)`.

But I made a mistake during refactoring. `SourceCode.from(", ")` should be after `y match { ... }`. Thanks for pointing out. And unfortunately I didn't test functions. (I just tested `def x = 1 + 1`)

LPTK: Okay. The generated code will not be valid though if there is more than one wildcard or literal: JS rejects `(x,_,_) => x + 1`. We may need to zip the parameters with their index (`.zipWithIndex`) and generate `(x,_1,_2) => x + 1` instead. And do something similar for `let` patterns.

shared/src/main/scala/mlscript/JSBackend.scala
LPTK: @chengluyu don't forget to mark as resolved when it is. I think this is resolved, right?

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -328,16 +327,24 @@ class JSBackend {
...
-            }
+          .map {
+            case TypeDef(
+                  kind,
+                  Primitive(name),
+                  typeParams,
+                  actualType,
+                  _,
+                  _
+                ) => // TODO: handle methods
```
LPTK: This comment caused the entire block to be refactored in an awkward way... 😅

chengluyu: 😅 Yes. I merged an old `mlscript` branch. Now this version of `mlscript` branch was merged.

shared/src/main/scala/mlscript/Codegen.scala
chengluyu: Good suggestion! Just updated!

PR #32, opened 2021-12-01
Bug fixes for method typing

fo5for:
Issue: subsumption checking was not performed on propagated method definitions, allowing this to be typed:
```scala
class Test4A
    method Mth4A: anything
    method Mth4A = true
//│ Defined class Test4A
//│ Declared Test4A.Mth4A: test4A -> true
//│ Defined Test4A.Mth4A: test4A -> true
:e
class Test4B: Test4A
    method Mth4A: int
//│ Defined class Test4B
//│ Declared Test4B.Mth4A: test4B -> true
```

It now emits an error:
https://github.com/hkust-taco/mlscript/blob/1e7cee08b7a7ae7de95b52249cc1aea5da33d855/shared/src/test/diff/mlscript/BadMethods.mls#L373-L384

Side effect: the error message are improved in some cases:
https://github.com/hkust-taco/mlscript/compare/048e0a1..1e7cee0#diff-271f81eabb0cf04b1f9fe33d0620031626c2c2cb6c311ebafa625a54caf951cbL144-R150

### DIFF
--- a/shared/src/test/diff/mlscript/Methods.mls
+++ b/shared/src/test/diff/mlscript/Methods.mls
@@ -211,9 +211,15 @@ class Class3C: Class2C
 //│ ╟── expression of type `42` does not match type `bool`
 //│ ║  l.207: 	    method MtdB = 42
 //│ ║         	                  ^^
+//│ ╟── but it flows into method definition with expected type `bool`
+//│ ║  l.207: 	    method MtdB = 42
+//│ ║         	           ^^^^^^^^^
 //│ ╟── Note: constraint arises from type reference:
 //│ ║  l.191: 	class Class2C: Class2B[int, bool]
-//│ ╙──       	                            ^^^^
+//│ ║         	                            ^^^^
+//│ ╟── from method declaration:
+//│ ║  l.190: 	    method MtdB: B
+//│ ╙──       	           ^^^^^^^
 //│ Defined class Class3C
 //│ Defined Class3C.MtdA: (class3C & {Class2B#A = int, Class1A#A = int, Class2B#B = bool, Trait2A#A = bool} & trait2A) -> 42
 //│ Defined Class3C.MtdB: (class3C & {Class2B#A = int, Class1A#A = int, Class2B#B = bool, Trait2A#A = bool} & trait2A) -> 42
--- a/shared/src/test/diff/mlscript/Methods.mls
+++ b/shared/src/test/diff/mlscript/Methods.mls
@@ -1,7 +1,9 @@
 class Foo[A, B]: { x: A; y: B }
     method Fun[C]: (A -> B -> C) -> (A -> B -> C)
+    method Fun[C, D] (f: C -> D) = f
 //│ Defined class Foo
 //│ Declared Foo.Fun: (foo & {Foo#B = 'B, Foo#A = 'A, y: 'B, x: 'A}) -> ('A -> 'B -> 'C) -> 'A -> 'B -> 'C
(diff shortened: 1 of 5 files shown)


LPTK:
Good! Though on a related note, I'm surprised that the type of `Test4A.Mth4A` in your example is `true` instead of `anything`. Explicit method signatures should take precedence over inferred types, at least when viewed from the outside of the class (probably from the inside too). Otherwise requiring explicit signatures when overriding makes little sense, as the signature is ignored anyway.

LPTK:
@fo5for maybe you want to address that in this PR to kill two birds with one stone?

fo5for:
The way I intended for it to work is that implicit calls uses the declared signature and explicit calls uses the inferred signature, so the caller can choose which one to use. (It's not working in the current version due to a bug...)

```scala
o = Test4A {}
o.Mth4A
o.(Test4A.Mth4A)
//│ o: test4A
//│ res: anything
//│ res: true
```

I realized that this may not be possible if the class inherits from another class that declared the same method. So I'll change it to always use the declared signature.


> @fo5for maybe you want to address that in this PR to kill two birds with one stone?

Sure. I will actually kill 3 birds by making diffTests print the declared and defined signature correctly. It was printing the defined signature twice if they are both present.

shared/src/test/diff/mlscript/Methods.mls
```diff
@@ -10,20 +12,20 @@ class Bar: Foo[int, bool]
...
-//│ res: (foo & {Foo#B: 'B, Foo#A: 'A, y: 'B, x: 'A}) -> ('A -> 'B -> 'C) -> 'A -> 'B -> 'C
+//│ res: (foo & {Foo#B: 'B, Foo#A: 'A, y: 'B, x: 'A}) -> (nothing -> anything & 'a) -> 'a
 //│ res: (bar & {Foo#B: bool, Foo#A: int, y: bool, x: int}) -> 'a -> 'a
 
 f = Foo { x = 1; y = 2 }
 //│ f: foo & {Foo#B: 'B | 2 .. 'B, Foo#A: 'A | 1 .. 'A, y: 2, x: 1}
 
 f.(Foo.Fun)
-//│ res: (('A | 1) -> ('B | 2) -> 'C) -> 'A -> 'B -> 'C
+//│ res: (nothing -> anything & 'a) -> 'a
```
LPTK: These diffs are very fishy. I wonder what's happening here.

fo5for: I reverted the precedence fix to test the other fixes independently, then applied it back manually before committing. Turns out I applied it in the wrong place 😅

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -253,9 +253,12 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite {
                   // ^ it may not end up being defined if there's an error
                 output(s"Defined " + td.kind.str + " " + td.nme.name)
                 val ttd = ctx.tyDefs(td.nme.name)
+                val thisTy = typer.TypeRef(ttd, ttd.tparams.flatMap(p =>
```
LPTK: Is there an easy way of avoiding too much logic duplication in `DiffTests`? It is very likely that code manipulating type provenance etc. will change in future Typer updates, and this would create a discrepancy. The logic should at least be moved and shared in a Typer method.

LPTK:
@fo5for Let me know when the PR is ready for review and merge.

Issue #34, opened 2021-12-02
Add JS tests


LPTK:
The JS code-gen has become quite non-trivial and needs to be tested. We should update the diff-tests to also compile and run the JS, probably by shelling out to `node`.

- [x] Make sure to import the tests documented in https://github.com/hkust-taco/mlscript/pull/30

LPTK:
While we're at it, there are some cases where JS code-gen results in wrong JS code, such as: `fun {x = 1} -> 0`. Would be good to solve them and add tests for them (it's probably not worth opening an issue for them at this point).

EDIT: also this one: `fun ((x, y)) -> x`

LPTK:
> * [ ]  Make sure to import the tests documented in [JavaScript code generation for methods #30](https://github.com/hkust-taco/mlscript/pull/30)

@chengluyu You had forgotten to add tests for this stuff. I've added them in https://github.com/hkust-taco/mlscript/commit/7c4b70cda9d3ed0326e2199dc6a79646ad387504.

LPTK:
> While we're at it, there are some cases where JS code-gen results in wrong JS code, such as: `fun {x = 1} -> 0`. Would be good to solve them and add tests for them (it's probably not worth opening an issue for them at this point).
> 
> EDIT: also this one: `fun ((x, y)) -> x`

@chengluyu You also forgot to add tests for that. Added in https://github.com/hkust-taco/mlscript/commit/7d04016c9c702ccc8a896728b28e39e638b5839f.

You should be more careful when closing an issue, especially when it clearly has uncompleted items/tasks.

chengluyu:
> > While we're at it, there are some cases where JS code-gen results in wrong JS code, such as: `fun {x = 1} -> 0`. Would be good to solve them and add tests for them (it's probably not worth opening an issue for them at this point).
> > EDIT: also this one: `fun ((x, y)) -> x`
> 
> @chengluyu You also forgot to add tests for that. Added in [7d04016](https://github.com/hkust-taco/mlscript/commit/7d04016c9c702ccc8a896728b28e39e638b5839f).
> 
> You should be more careful when closing an issue, especially when it clearly has uncompleted items/tasks.

Thanks for pointing this out! It was a big oversight!

PR #19, opened 2021-11-18
Implement multi-parameter functions

LPTK:
This is not yet ready to merge, as the JS code-gen is broken. @chengluyu said he'd have a look and fix it.

PR #35, opened 2021-12-06
Support `with` construction (close #33)

### LINKED ISSUE #33: Support `with` construct in JS
For instance:

```ts
class C[A]: { x: A }
//│ Defined class C

c = C{x = 1}
//│ c: c & {C#A: 'A | 1 .. 'A, x: 1}

d = c with { x = "hi"; y = 2 }
//│ d: c & {C#A: 'A | 1 .. 'A, y: 2, x: "hi"}

d.x
//│ res: "hi"
```

To do this, I guess we should create a new object, copy over all the fields of the previous object (there is a JS method to do that but I forget its name), and set the new fields in the new object, potentially overriding old ones. We may have to also explicitly set the prototype of the new object to the same thing as the old one, so we retain class instance-ness.

chengluyu:


shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -143,6 +143,17 @@ class JSBackend {
     case StrLit(value) => JSLit(JSLit.makeStringLiteral(value))
     // `Asc(x, ty)` <== `x: Type`
     case Asc(trm, _) => translateTerm(trm)
+    // `c with { x = "hi"; y = 2 }` ==> `Object.assign(c, { x: "hi", y: 2 })`
```
LPTK: As discussed in the meeting, this is not sound as it should create a new object, not update the old one in place.

LPTK (changes requested):

chengluyu:
Now we translate

```
class A
def a = A{}
def x = a with { y = 0 }
```

to

```js
class A {}
let a = new A({});
let x = (function (target, fields) {
  const copy = Object.assign({}, target, fields);
  Object.setPrototypeOf(copy, Object.getPrototypeOf(target));
  return copy;
})(a, { y: 0 });
```

The reason why we use IIFE here is that we don’t want to evaluate `target` twice if `target` is an expression. So I’m making this function a prelude function. It will be included in generated code if and only if `with` construct is used.

LPTK:
@chengluyu Ok, sounds good! Be sure to document this expansion in the code.

chengluyu:
> [...] Be sure to document this expansion in the code.

Sure.

chengluyu:
Now we translate

```
class Test
def withConstruct = Test{} with { y = 0 } with { z = 0 }
```

to 

```js
function withConstruct0(target, fields) {
  const copy = Object.assign({}, target, fields);
  Object.setPrototypeOf(copy, Object.getPrototypeOf(target));
  return copy;
}
class Test {}
let withConstruct = withConstruct0(withConstruct0(new Test({}), { y: 0 }), { z: 0 });
```

and outputs

```
val withConstruct: test & {y: 0, z: 0} = Test { "y": 0, "z": 0 }
```

Note that the auxiliary function `withConstruct0` avoids name conflicts with user-defined names.

Another point I want to mention is that the renaming logic is in a mess right now and I use a lot of mutable states during code generation. But this is not on the first priority. I want to focus on code generation tests after this PR and #19 are merged. After that we can consider making the logic clearer and better.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -344,11 +359,19 @@ class JSBackend {
...
+    }
+
+    withConstructFnName = getTemporaryName(withConstructFnName)
+
     val defResultObjName = getTemporaryName("defs")
     val exprResultObjName = getTemporaryName("exprs")
     // This hash map counts how many times a name has been used.
     val resolveShadowName = new ShadowNameResolver
-    val stmts: Ls[JSStmt] =
+    var stmts: Ls[JSStmt] =
```
LPTK: Let's not use a needless `var` here. Just add the required statement at the end, like a civilized person 😛

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -70,6 +73,9 @@ class JSBackend {
       throw new Error(s"term $t is not a valid pattern")
   }
 
+  private var hasWithConstruct = false
+  private var withConstructFnName = "withConstruct"
```
LPTK: Why a `var`? This should be initialized to the right value directly. Initializing it with a wrong value is begging for errors to happen.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -371,7 +394,10 @@ class JSBackend {
           .concat(otherStmts.flatMap {
             case Def(isRecursive, name, L(body)) =>
               val translatedBody = translateTerm(body)
-              val tempName = getTemporaryName(name)
+              // Get a name that not conflicts with class names.
+              val tempName = (name #:: LazyList
+                .from(0)
```
LPTK, line 397: Does this really work with defs that have the same name? Looks like you're going to start from 0 every time.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -459,4 +498,26 @@ object JSBackend {
     case DecLit(value) => s"DecLit($value)"
     case StrLit(value) => s"StrLit($value)"
   }
+
+  /** Make a function for `with` construct.
+    *
+    * @param name
```
LPTK: Remove this, it's completely useless :^)

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -459,4 +498,26 @@ object JSBackend {
     case DecLit(value) => s"DecLit($value)"
     case StrLit(value) => s"StrLit($value)"
   }
+
+  /** Make a function for `with` construct.
```
LPTK: Can you document what the JS looks like?

chengluyu: Yes. I will document it in Scaladoc comments.

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: Okay. Instead I will put the generated code in this comment.

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: Oh you're right! I’m fixing it soon.

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: They will be modified during the code generation, so I’m using `var`. We don’t know if we need the `withConstruct` helper function ahead of time.

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: This `var` could be avoided for sure. :)

shared/src/main/scala/mlscript/JSBackend.scala
LPTK: In this case you probably want to initialize it with `None` and later make it `Some(usedName)`. Anything you can to avoid having a variable that temporarily holds **_a wrong information_**.

chengluyu: I can make `withConstructFnName` a `val`. But for `hasWithConstruct`, it means whether `With` is presented in the source code and will be modified during the traversal, so it must be `var` for now.

There are two ways making `hasWithConstruct` a `val`.

1. One is add a recursive function which checks if `With` presents in the tree. The down side is we’re actually doing two-pass traversal.
2. Another one is to do the check during translation and append the value to return value.

Both of them introduce more changes to this PR.

I’m really willing to improve the code quality concerning code generation. But I prefer doing it after I add code generation tests (#34) so that refactoring will not accidentally bring much regression.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -136,7 +142,7 @@ class JSBackend {
       )
     case CaseOf(term, cases) => {
       val argument = translateTerm(term)
-      val parameter = getTemporaryName("x")
+      val parameter = "x"
```
LPTK, line 145: This does not seem sound. Every time you introduce a name you have to ask yourself if it may clash with another name already in the program.

For `def f x = case 1 of { 1 -> x }` you currently generate `(x) => (function (x) { if (x == 1) { return x; } throw new Error("non-exhaustive case expression"); })(1)`, which is wrong.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -450,24 +448,32 @@ class JSBackend {
...
     *
     * @param name
     *   the base name
     * @return
     *   identifier like `$name$n` where `n` is an integer
     */
-  private def getTemporaryName(name: Str): Str = (name #:: LazyList
-    .from(0)
-    .map { name + _.toString }).dropWhile { defNames contains _ }.head
+  private def getAlternativeName(name: Str): Str = {
```
LPTK, line 470: Just a note: this seems like a pretty wasteful way of computing the information (it recreates a lazy list every time), but I guess for typical programs it won't matter much. Still, a simple iterator would be more appropriate (it does not retain intermediate results), or even a good old while loop.

shared/src/main/scala/mlscript/JSBackend.scala
LPTK: It's a bit better yes. But what I was suggesting was simpler and closer to the original intent (which was to avoid computing a name if it won't be needed).

To be clear what I was suggesting was this:
```diff
-  private var hasWithConstruct = false
-  private var withConstructFnName = "withConstruct"
+  private var withConstructFnName: Opt[Str] = N
```
and later, upon use, if `withConstructFnName` is empty just initialize it with the right value.
This way you restrict the state space to only valid situations.

Yet another, perhaps even better way would be:
```diff
  private var hasWithConstruct = false
-  private var withConstructFnName = "withConstruct"
+  private lazy val withConstructFnName = {
+    hasWithConstruct = true
+    getAlternativeName("withConstruct")
+  }
```

LPTK (approved):
Okay, thanks for finishing this. There are still problems as I've pointed out, but I'll open an issue about it and we'll look into it once we have proper tests.

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: Yes. I noticed that `getTemporaryName` only concerns names declared at top-level. So using `getTemporaryName` or not won’t affect the generated code in this case. I will fix this next.

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: Yes.

PR #36, opened 2021-12-06
Forbid instantiating abstract classes

### LINKED ISSUE #31: Forbid instantiating abstract classes
For example, this should be rejected:
```ts
class C: {}
  method M: int
c = C {}
c.M
```

fo5for:
Closes #31

The error message lists the methods with missing implementations, which could potentially become too long if many methods are abstract and the class is instantiated many times. May consider listing only at the first time when it raises an error, and only keep the first line in subsequent errors. (But I don't think there is a mechanism to do that currently.)

### DIFF
--- a/shared/src/main/scala/mlscript/Typer.scala
+++ b/shared/src/main/scala/mlscript/Typer.scala
@@ -420,10 +420,10 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
             err(msg"Method names must start with a capital letter", nme.toLoc)
           if (defs.isDefinedAt(nme.name))
             err(msg"Method '${tn}.${nme.name}' is already defined.", nme.toLoc)
-          (targsMap.keySet & targsMap2.keySet).foreach { s =>
-            warn((s"Method type parameter ${targsMap2(s).nameHint.getOrElse("")}" -> targsMap2(s).prov.loco)
-              :: (s"shadows class type parameter ${td.targsMap(s).nameHint.getOrElse("")}" -> td.targsMap(s).prov.loco)
-              :: Nil map (x => Message.fromStr(x._1) -> x._2))
+          td.tparams.flatMap(tp1 => tparams.find(_.name === tp1.name).map(tp1 -> _)).foreach { case tp1 -> tp2 =>
+            warn((msg"Method type parameter ${tp1}" -> tp1.toLoc)
+              :: (msg"shadows class type parameter ${tp2}" -> tp2.toLoc)
+              :: Nil)
           }
           val bodyTy = rhs.fold(
             term => subst(typeLetRhs(rec, nme.name, term)(thisCtx, raise, targsMap ++ targsMap2), reverseRigid),
--- a/shared/src/main/scala/mlscript/Typer.scala
+++ b/shared/src/main/scala/mlscript/Typer.scala
@@ -416,12 +416,16 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
           implicit val prov: TypeProvenance = tp(md.toLoc, rhs.fold(_ => "method definition", _ => "method declaration"))
           val dummyTargs2 = tparams.map(p => freshVar(originProv(p.toLoc, "method type parameter"), S(p.name))(ctx.lvl + 2))
           val targsMap2 = tparams.map(_.name).zip(dummyTargs2).toMap
-          // ^ overwrites. fine for now as the maps are not used afterwards. may pass vars instead
           if (!nme.name.head.isUpper)
             err(msg"Method names must start with a capital letter", nme.toLoc)
           if (defs.isDefinedAt(nme.name))
             err(msg"Method '${tn}.${nme.name}' is already defined.", nme.toLoc)
-          val bodyTy = rhs.fold( // TODO: warn about shadowing?
+          (targsMap.keySet & targsMap2.keySet).foreach { s =>
+            warn((s"Method type parameter ${targsMap2(s).nameHint.getOrElse("")}" -> targsMap2(s).prov.loco)
+              :: (s"shadows class type parameter ${td.targsMap(s).nameHint.getOrElse("")}" -> td.targsMap(s).prov.loco)
+              :: Nil map (x => Message.fromStr(x._1) -> x._2))
+          }
+          val bodyTy = rhs.fold(
             term => subst(typeLetRhs(rec, nme.name, term)(thisCtx, raise, targsMap ++ targsMap2), reverseRigid),
             ty => PolymorphicType(thisCtx.lvl, subst(typeType(ty)(thisCtx, raise, targsMap ++ targsMap2), reverseRigid))
           )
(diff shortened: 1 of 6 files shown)


shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -476,6 +476,8 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
         decls.foreach { case nme -> mt => 
           implicit val prov: TypeProvenance = mt.prov
           td.baseClasses.foreach { case Var(bn) => ctx.mthDecls.get(bn).foreach(_.get(nme).foreach(ss(mt, _, S(bn)))) }
+          if (!defs.isDefinedAt(nme))
+            ctx.env -= tn.name
```
LPTK: It is pretty hacky to remove the ctor from the context after the fact, instead of not registering it in the first place. In fact, it leads to false positives, as this should be rejected:

```ts
class Test7A
    method Mth7A: nothing
class Test7B
    method Mth7B = Test7A
```

but is currently accepted!

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -676,7 +678,13 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
       case v @ ValidVar(name) =>
         val ty = ctx.get(name).getOrElse {
           // TODO: delay type expansion to message display and show the expected type here!
-          err("identifier not found: " + name, term.toLoc)
+          (ctx.tyDefs.get(name), ctx.mthDecls.get(name).map(_ -- ctx.mthDefs(name).keys)) match {
+            case (S(td), S(abstMths)) if abstMths.nonEmpty => err(
+              (msg"Instantiation of an abstract type is forbidden" -> term.toLoc)
+              :: (msg"${td.kind.str} ${td.nme} is abstract:" -> td.toLoc)
```
```suggestion
              :: (msg"Indeed, ${td.kind.str} ${td.nme} is abstract:" -> td.toLoc)
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -676,7 +678,13 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+              :: abstMths.map { case mn -> mthTy => msg"method $mn is abstract" -> mthTy.prov.loco }.toList)
```
```suggestion
              :: abstMths.map { case mn -> mthTy => msg"Hint: method $mn is abstract" -> mthTy.prov.loco }.toList)
```

LPTK (changes requested):
I understand why you implemented it this way to minimize changes, but as I showed it does not work.
To make it work, we need to go the long way around and implement the check _before_ methods are type checked. The good thing is that this is a perfect opportunity to implement what I alluded to in https://github.com/hkust-taco/mlscript/pull/22#discussion_r757856454 😄
Basically, `checkCycleComputeFields` should be adapted to also accumulate the inherited method declarations and/or implementations along with their signatures. For now, in this PR, we can ignore the signatures, but later on I think this will a the better way of implementing method type checking without relying on error-prone global mutable state (we should discuss this in the next meeting).

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -681,8 +668,8 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
           (ctx.tyDefs.get(name), ctx.mthDecls.get(name).map(_ -- ctx.mthDefs(name).keys)) match {
             case (S(td), S(abstMths)) if abstMths.nonEmpty => err(
               (msg"Instantiation of an abstract type is forbidden" -> term.toLoc)
-              :: (msg"${td.kind.str} ${td.nme} is abstract:" -> td.toLoc)
-              :: abstMths.map { case mn -> mthTy => msg"method $mn is abstract" -> mthTy.prov.loco }.toList)
+              :: (msg"Indeed, ${td.kind.str} ${td.nme} is abstract:" -> td.toLoc)
```
LPTK: Actually it looks worse than I thought, sorry xP
How about this?
```suggestion
              :: (msg"Note that ${td.kind.str} ${td.nme} is abstract:" -> td.toLoc)
```

LPTK (approved):

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -393,30 +391,30 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
           val dummyTargs2 = tparams.map(p => freshVar(originProv(p.toLoc, "method type parameter"), S(p.name))(ctx.lvl + 2))
-          ctx.targsMaps(tn.name) += S(nme.name) -> tparams.map(_.name).zip(dummyTargs2).toMap
-          // ^ overwrites. fine for now as the maps are not used afterwards. may pass vars instead
+          val targsMap2 = tparams.map(_.name).zip(dummyTargs2).toMap
           if (!nme.name.head.isUpper)
             err(msg"Method names must start with a capital letter", nme.toLoc)
           if (defs.isDefinedAt(nme.name))
             err(msg"Method '${tn}.${nme.name}' is already defined.", nme.toLoc)
+          (targsMap.keySet & targsMap2.keySet).foreach { s =>
+            warn((s"Method type parameter ${targsMap2(s).nameHint.getOrElse("")}" -> targsMap2(s).prov.loco)
```
LPTK: This warning may be better reported by just looking at the syntactic class definition tree.
But in any case this logic is weird. You're doing unorthodox acrobatics with things like `targsMap2(s).nameHint.getOrElse("")`. Why not just `s`? If you stored Var instead of Str in targsMap you'd actually get a proper provenance as well as the name.

LPTK (changes requested):

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -407,10 +407,10 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
             err(msg"Method names must start with a capital letter", nme.toLoc)
           if (defs.isDefinedAt(nme.name))
             err(msg"Method '${tn}.${nme.name}' is already defined.", nme.toLoc)
-          (targsMap.keySet & targsMap2.keySet).foreach { s =>
-            warn((s"Method type parameter ${targsMap2(s).nameHint.getOrElse("")}" -> targsMap2(s).prov.loco)
-              :: (s"shadows class type parameter ${td.targsMap(s).nameHint.getOrElse("")}" -> td.targsMap(s).prov.loco)
-              :: Nil map (x => Message.fromStr(x._1) -> x._2))
+          td.tparams.flatMap(tp1 => tparams.find(_.name === tp1.name).map(tp1 -> _)).foreach { case tp1 -> tp2 =>
```
LPTK: Let's not create unnecessary lists in memory! 😛 
Also, better not use a quadratic algorithm when it's so easily avoided...
```suggestion
          val tp1s = td.tparams.toSet
          tparams.foreach { tp2 => tp1s.get(tp2) match ...
```

LPTK (approved):
Otherwise LGTM. Let's squash the changes and merge.

PR #38, opened 2021-12-09
Reconstruct class types for better type pretty-printing

LPTK:
For example, for this code:

```scala
class Wrapper[A]: { x: A }
class Psyduck[B]: Wrapper[B]
    method Apply[A]: (B -> A) -> Psyduck[A]
```

We used to infer:

```scala
//│ Declared Psyduck.Apply: (psyduck & {Psyduck#B = 'B, Wrapper#A = 'B, x: 'B}) -> ('B -> 'A) -> (psyduck & {Psyduck#B = 'A, Wrapper#A = 'A, x: 'A})
```

and we now infer:

```scala
//│ Declared Psyduck.Apply: Psyduck['B] -> ('B -> 'A) -> Psyduck['A]
```

---

Note that for this code:

```scala
class Wrapper[A]: { x: A }
class Psyduck[B]: Wrapper[B]
    method Apply f = Psyduck { x = f this.x }
```

we used to infer

```scala
//│ Defined Psyduck.Apply: (psyduck & {Psyduck#B = 'B, Wrapper#A = 'B, x: 'B}) -> ('B -> ('a & 'B0)) -> (psyduck & {Psyduck#B = 'B0, Wrapper#A = 'B0, x: 'a})
```

and we now infer:

```scala
Defined Psyduck.Apply: Psyduck['B] -> ('B -> ('a & 'B0)) -> (Psyduck['B0] with {x: 'a})
```

Notice the refinement `with {x: 'a}` in the type `Psyduck['B0] with {x: 'a}`. It is due to the fact that we do not perform any variance analysis, so the engine considers the `Psyduck` invariant by default, and in this case `B0` may later be taken to be something too general, for example `anything`, but MLscript still gives us access to the precise type of the payload, `{x: 'a}`.


shared/src/main/scala/mlscript/TypeSimplifier.scala
```diff
@@ -282,4 +282,107 @@ trait TypeSimplifier { self: Typer =>
...
+  def reconstructClassTypes(st: SimpleType, pol: Bool, ctx: Ctx): SimpleType = {
+    
+    val renewed = MutMap.empty[TypeVariable, TypeVariable]
+    
+    def renew(tv: TypeVariable): TypeVariable =
+      if (tv.lowerBounds.isEmpty && tv.upperBounds.isEmpty) tv
+      else renewed.get(tv) match {
+        case S(tv2) => tv2
+        case N =>
+          val tv2 = freshVar(tv.prov, tv.nameHint)(0)
```
fo5for: Shouldn't this be `tv.level` instead of `0`? It doesn't break anything right now since it's only used in printing, but I think we should make it correct.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -219,6 +221,26 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+  {
+    ty match {
+      case tr @ TypeRef(td, targs) => fieldsOf(tr.expandWith(paramTags), paramTags)
+      case ComposedType(false, l, r) =>
+        mergeMap(fieldsOf(l, paramTags), fieldsOf(r, paramTags))(_ & _)
+      case RecordType(fs) => fs.toMap
+      case p: ProxyType => fieldsOf(p.underlying, paramTags)
+      case Without(base, ns) => fieldsOf(base, paramTags).filter(ns contains _._1)
+      case TypeBounds(lb, ub) => fieldsOf(ub, paramTags)
+      case _: ObjectTag | _: FunctionType | _: TupleType | ExtrType(false) | _: TypeVariable
```
fo5for: `ExtrType(false)` seems to be redundant here.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -260,38 +282,31 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
-        }}()
+          case tr @ TypeRef(td, targs) => checkCycle(tr.expand)(travsersed + L(td))
+          case ComposedType(_, l, r) => checkCycle(l) && checkCycle(r)
+          case NegType(u) => checkCycle(u)
+          case p: ProxyType => checkCycle(p.underlying)
+          case Without(base, _) => checkCycle(base)
+          case TypeBounds(lb, ub) => checkCycle(lb) && checkCycle(ub)
+          case tv: TypeVariable => travsersed(R(tv)) || {
+            val t2 = travsersed + R(tv)
+            tv.lowerBounds.forall(checkCycle) && tv.upperBounds.forall(checkCycle)
```
fo5for: `t2` should be pass along here?

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -551,7 +566,7 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
               } else TypeRef(td, Nil)(tpr, ctx)
             case L(e) =>
               if (name.isEmpty || !name.head.isLower) e()
-              else typeNamed(tyLoc, name.head.toUpper.toString + name.tail) match {
+              else typeNamed(tyLoc, name.capitalize) match {
                 case R(td) => td.kind match {
                   case Cls => clsNameToNomTag(td)(tp(tyLoc, "class tag"), ctx)
                   case Trt => ???
```
fo5for: This is from the last commit but it should just be `trtNameToNomTag` here. Might as well implement it.

shared/src/main/scala/mlscript/TypeSimplifier.scala
```diff
@@ -282,4 +282,107 @@ trait TypeSimplifier { self: Typer =>
...
+      case ty @ ComposedType(true, l, r) => go(l, pol) | go(r, pol)
+      case ty @ (ComposedType(false, _, _) | _: ObjectTag) =>
+        val dnf @ DNF(cs) = DNF.mk(ty, pol)
+        cs.sorted.map { c =>
+          c.copy(vars = c.vars.map(renew), nvars = c.nvars.map(renew)).toTypeWith(_ match {
+            case LhsRefined(bo, tts, rcd) =>
+              lazy val newRcd = rcd.copy(
+                rcd.fields.filterNot(_._1.name.isCapitalized).mapValues(go(_, pol)))(rcd.prov)
+              bo match {
+                case S(cls @ ClassTag(Var(tagNme), ps)) if !primitiveTypes.contains(tagNme) =>
```
fo5for, line 324: Should we also handle traits?

shared/src/main/scala/mlscript/TypeSimplifier.scala
LPTK: Yes you're right. Currently the type simplifier uses `0` everywhere because as you say it's just for printing, but it's true we should be using `tv.level`.

shared/src/main/scala/mlscript/Typer.scala
LPTK: Whoops, good catch!

shared/src/main/scala/mlscript/TypeSimplifier.scala
LPTK: In principle, it would be nice. But it would be a lot of work and may not be that important. The thing that makes the situation with traits less straightforward is that we don't have a single-inheritance hierarchy for traits, so there could be many traits mixed in, some related with other. For example if trait `T` inherits from class `C[int]` then we should probably reconstruct `c & t & { x: 1 }` as `T & { x: 1 }`. I'm not interested in dealing with that at the moment 😛

fo5for (approved):
👍🏻

PR #41, opened 2021-12-10
Fix #39 (rigid method params, pattern typing, name hints...)

### LINKED ISSUE #39: `TypeVariable.nameHint` lost in constraint solving and type simplification
```scala
class Test[A]: { x: A }
    method Mth[B]: (A -> B) -> B
    method Mth[B] (f: A -> B) = f this.x
//│ Defined class Test
//│ Declared Test.Mth: Test['A] -> ('A -> 'B) -> 'B
//│ Defined Test.Mth: Test['A] -> ('A -> 'a) -> 'a
```

The name hint for the type parameter `B` in the method definition is lost. I have identified two sources of this issue:

1. `nameHint` is not preserved in extrusion
https://github.com/hkust-taco/mlscript/blob/9930721e118bb60940e244c8c682308f5b27582e/shared/src/main/scala/mlscript/ConstraintSolver.scala#L392
(That's a downside of providing default arguments. You are not forced to review every invocation and provide the appropriate argument...)

2. `nameHint` is not unified when simplifying polar variables
Looking at the debug output of the above code snippet (after fixing 1.), the name hint is still present in the canonical inferred type:
`//│ Canon: ((test<> & {Test#A: (A31 -> A31), x: A31}) -> ((((A31 | α32) -> (α33 & B34)) & α35) -> α33))`
But it is later simplified away since it only cooccurs with α33 in one polarity:
`//│ [sub] α32 -> None, B34 -> None, α35 -> None`

LPTK:
* Fix explicit method type parameters (they were not rigid)
 * Fix method signature typing level
 * Fix typing of type ascriptions in patterns – constrain in both directions
 * Propagates name hints more

Let's merge this after https://github.com/hkust-taco/mlscript/pull/38 as it's based on it.

### DIFF
--- a/shared/src/main/scala/mlscript/ConstraintSolver.scala
+++ b/shared/src/main/scala/mlscript/ConstraintSolver.scala
@@ -389,7 +389,7 @@ class ConstraintSolver extends NormalForms { self: Typer =>
       case t @ TupleType(fs) => TupleType(fs.map(nt => nt._1 -> extrude(nt._2, lvl, pol)))(t.prov)
       case w @ Without(b, ns) => Without(extrude(b, lvl, pol), ns)(w.prov)
       case tv: TypeVariable => cache.getOrElse(tv, {
-        val nv = freshVar(tv.prov)(lvl)
+        val nv = freshVar(tv.prov, tv.nameHint)(lvl)
         cache += (tv -> nv)
         if (pol) { tv.upperBounds ::= nv
           nv.lowerBounds = tv.lowerBounds.map(extrude(_, lvl, pol)) }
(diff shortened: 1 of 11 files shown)


fo5for (approved):
I'm surprised we didn't notice that explicit type parameters in method definitions have to be rigid... 😅 I guess it's partly because `def`s don't support explicit type parameters so we didn't have any tests for it. May look into supporting that in `def`s later.

fo5for:
On a side note, `fix` keywords don't work in the title. They have to appear in the comment for it to link automatically...

LPTK:
> we didn't notice that explicit type parameters in method definitions have to be rigid

Do you mean that we didn't notice they weren't actually rigid in the current implementation? Because we did know they had to be rigid, as I explained to you some days ago on Whatsapp, but I didn't thoroughly check this part of the logic when merging your PR, as we were in a hurry.

LPTK:
> On a side note, `fix` keywords don't work in the title. They have to appear in the comment for it to link automatically...

I think they do work. Whenever a commit with `Fix #N` in the title is merged into the main branch, issue `#N` is closed automatically.

fo5for:
> > we didn't notice that explicit type parameters in method definitions have to be rigid
> 
> Do you mean that we didn't notice they weren't actually rigid in the current implementation? Because we did know they had to be rigid, as I explained to you some days ago on Whatsapp, but I didn't thoroughly check this part of the logic when merging your PR, as we were in a hurry.

Right... Back then we were mainly talking about the class type parameters since that was the problem exposed by the tests. I thought method type parameters didn't have to be rigid since they are mostly used in declarations, and no constraining is done when typing declarations. But I forgot they can also be used in definitions 😅

PR #46, opened 2021-12-14
Add `Scope` which tracks and allocates identifiers (close #45)

### LINKED ISSUE #45: JS code-gen is unhygienic
Example:
```ts
def f x = case 1 of { 1 -> x }
```

Generates, for the body of `f`:
```js
(x) => (function (x) { if (x == 1) { return x; } throw new Error("non-exhaustive case expression"); })(1)
```

The normal way of dealing with this is the "gensym" approach, for "generating symbols". For every name x you want to use, you call gensym("x"), and a new name distinct with every other name in the program is picked.

We could also have a more local version that only picks names different from names currently in scope, which would avoid generating too many harder-to-read unique names.

chengluyu:
This PR adds class `Scope` which tracks and allocates identifiers. The `Scope` class also removes infamous `getAlternativeName` (previously `getTemporaryName`). Now the code generation transforms

```
def f x = case 1 of { 1 -> x }
def defs = 0
def exprs = "blabla"
```

to

```js
(() => {
  let [defs, exprs] = (() => {
    const defs1 = {};
    const exprs1 = [];
    let f = (x) => (function (a) {
      if (a == 1) {
        return x;
      }
      throw new Error("non-exhaustive case expression");
    })(1);
    defs1.f = f;
    let defs = 0;
    defs1.defs = defs;
    let exprs = "blabla";
    defs1.exprs = exprs;
    return ([
      defs1,
      exprs1,
    ]);
  })();
  for (const key of Object.keys(defs)) {
    defs[key] = prettyPrint(defs[key]);
  }
  exprs = exprs.map(prettyPrint);
  return { defs, exprs };
  function prettyPrint(value) {
    switch (typeof value) {
      case "number":
      case "boolean":
      case "function":
        return value.toString();
      case "string":
        return '"' + value + '"';
      default:
        return value.constructor.name + " " + JSON.stringify(value, undefined, 2);
    }
  }
})()
```

The `Scope` class organises in a hierarchy and `translateTerm` now accepts an implicit parameter `scope`, so allocated names are unique in its context.


chengluyu:
Force-pushed twice because of a typo and format problems. No functional changes. 😅

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -141,20 +148,29 @@ class JSBackend(pgrm: Pgrm) {
...
     case Let(isRec, name, value, body) =>
-      JSImmEvalFn(name :: Nil, Left(translateTerm(body)), translateTerm(value) :: Nil)
+      val letScope = Scope(name :: Nil, scope)
+      JSImmEvalFn(
+        name :: Nil,
+        Left(translateTerm(body)(letScope)),
+        translateTerm(value)(letScope) :: Nil
+      )
     case Blk(stmts) =>
+      val blockScope = Scope(Nil, scope)
```
LPTK: Is it useful to create a new scope here? What does it achieve?

LPTK (approved):

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: We will translate `Let` to real `let`-declaration instead of IIFE, which needs to allocate names in the scope of current block. So creating scopes for blocks is necessary.

LPTK: Will `let x = 1 in let x = x in x` be translated correctly?

chengluyu: > Will `let x = 1 in let x = x in x` be translated correctly?

Yes. The scope will be able to rename identifiers, which will address this kind of problems.

LPTK: You keep saying "will", as this is not currently done, right? So for now this `blockScope` is useless and is better removed, or at least a comment stating its future purpose should be added.

chengluyu: OK.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -29,7 +29,8 @@ class JSBackend(pgrm: Pgrm) {
 
   // Sometimes, identifiers declared by `let` use the same name as class names.
   // JavaScript does not allow this. So, we need to replace them.
-  val letLhsAliasMap: HashMap[Str, Str] = HashMap()
```
chengluyu, line 29: `ctorAliasMap` is a more meaningful than `letLhsAliasMap`.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -425,55 +426,45 @@ class JSBackend(pgrm: Pgrm) {
...
-              )
-            )
+            // `exprs.push(<expr>)`.
+            case term: Term =>
+              JSExprStmt(
+                JSInvoke(
+                  JSMember(JSIdent(exprResultObjName), "push"),
+                  translateTerm(term)(topLevelScope) :: Nil
+                )
+              ) :: Nil
```
chengluyu: Here I fixed a misbehaviour. Declarations and expressions should generated in order. I used to generate declarations first, and then generate expressions.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -425,55 +426,45 @@ class JSBackend(pgrm: Pgrm) {
...
+                 // Not first time: `<declName> = <expr>;`
+                 JSExprStmt(JSAssignExpr(JSIdent(declName), translatedBody))
+               }) :: // Save the value: `defs.<snapshotName> = <declName>;`
+                JSExprStmt(
+                  JSAssignExpr(
+                    JSMember(JSIdent(defResultObjName), snapshotName),
+                    JSIdent(declName)
+                  )
+                ) :: Nil
+            // Ignore type declarations.
```
chengluyu: The refactor here removes the use of `LazyList`. I also renamed variables (`originalName`, `declName`, and `snapshotName`) to make code more readable.

chengluyu:
@LPTK I think this PR is ready. `Scope` is the basis of upcoming improvements.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -356,16 +370,19 @@ class JSBackend(pgrm: Pgrm) {
...
     val name = method.nme.name
     method.rhs.value match {
-      case Lam(Var(param), rhs)         => JSClassMethod(name, JSNamePattern(param) :: Nil, L(translateTerm(rhs)))
-      case Lam(Asc(Var(param), _), rhs) => JSClassMethod(name, JSNamePattern(param) :: Nil, L(translateTerm(rhs)))
-      case Lam(Tup(params), body) => JSClassMethod(
-        name,
-        params map { case _ -> param => translatePattern(param) },
-        L(translateTerm(body))
-      )
+      case Lam(Var(param), rhs) =>
```
LPTK: Does this case really occur? I thought we were always wrapping parameter lists in tuples now?

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -356,16 +370,19 @@ class JSBackend(pgrm: Pgrm) {
...
+        JSClassMethod(name, JSNamePattern(param) :: Nil, L(translateTerm(rhs)))
+      case Lam(Asc(Var(param), _), rhs) =>
```
LPTK: I don't understand why this case and the one above needs special handling. Why not handle everything with `translatePattern`? There is nothing special about ascribing _one_ parameter (as opposed to any other number of parameters).

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -406,55 +423,45 @@ class JSBackend(pgrm: Pgrm) {
           // defs.name = name;
           // ```
           .concat(otherStmts.flatMap {
-            case Def(isRecursive, name, L(body)) =>
-              val translatedBody = translateTerm(body)
+            case Def(isRecursive, originalName, L(body)) =>
+              val translatedBody = translateTerm(body)(topLevelScope)
+              // `declName` menas the name used in the declaration.
```
```suggestion
              // `declName` means the name used in the declaration.
```

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -406,55 +423,45 @@ class JSBackend(pgrm: Pgrm) {
...
+              // We need to save a snapshot after each computation.
+              // The snapshot name will be same as the original name, or in the
+              // format of `originalName@n` if the original name is used.
+              val snapshotName = resolveShadowName(originalName)
+              (if (snapshotName === originalName) {
+                 // First time: `let <declName> = <expr>;`
+                 JSLetDecl(declName, translatedBody)
+               } else {
+                 // Not first time: `<declName> = <expr>;`
+                 JSExprStmt(JSAssignExpr(JSIdent(declName), translatedBody))
```
LPTK, line 439: Ok now I see that's what's causing https://github.com/hkust-taco/mlscript/issues/49#issuecomment-994556548. Incorrect logic here – a renaming should be used instead. I guess you'll fix it later and that's okay.

chengluyu: Sure.

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: I just realized parameters are always wrapped in tuples. I will remove it.

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: `translatePattern` treats single parameter. But I do found similar patterns in  `translateTerm` and `translateClassMember`. I extract a new function for translating parameter lists.

PR #48, opened 2021-12-14
Fix explicit method type with inherited declaration

fo5for:
Small update to correct the semantics before we work on refactoring the mess 😅

### DIFF
--- a/shared/src/test/diff/mlscript/Methods2.mls
+++ b/shared/src/test/diff/mlscript/Methods2.mls
@@ -55,9 +55,8 @@ Cons.Size
 List.HeadOption
 //│ res: List['A] -> Option['A]
 
-// TODO: Strangely, this explicit call does not have the inferred refined type:
 Cons.HeadOption
-//│ res: (Cons['A] with {tail: List['A]}) -> Option['A]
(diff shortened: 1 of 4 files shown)


shared/src/test/diff/mlscript/Methods2.mls
```diff
@@ -68,22 +67,22 @@ l.Size
 l.Map (fun x -> mul x 2)
 //│ res: List['B .. 'B | int]
 
-o = l.HeadOption
```
LPTK, line 71: Small detail: let's not swap these needlessly.

LPTK (approved):
LGTM. Please go ahead and do the squash-merge.

shared/src/test/diff/mlscript/Methods2.mls
fo5for: The pattern matching below only fails with an `Option`. If `Cons.HeadOption` is called explicitly, the result would be a `Some` and the match would succeed. So I swapped them to preserve the error below.

LPTK: Oh yes good point, I had missed that 👍

PR #50, opened 2021-12-17
Better constraint solving

### LINKED ISSUE #37: Solve annoying constraints in a smarter way
Currently, the `annoying` resolution routine is way too eager to decompose subtyping problems into smaller ones. For example, it turns anything of the shape `A <: B & C` into `A <: B` and `A <: C`. However, that results in [exponential complexity](https://github.com/hkust-taco/mlscript/blob/mlscript/shared/src/test/diff/mlscript/Stress.mls#L191) for solving fairly common patterns, such as pattern matching:

```scala
class A[T]: { fA: T }
class B[T]: { fB: T }
class C[T]: { fC: T }
class D[T]: { fD: T }
class E[T]: { fE: T }
class F[T]: { fF: T }
class G[T]: { fG: T }
//│ Defined class A
//│ Defined class B
//│ Defined class C
//│ Defined class D
//│ Defined class E
//│ Defined class F
//│ Defined class G

:stats
def foo x = case x of {
  | A -> x.fA
  | B -> x.fB
  | C -> x.fC
  | D -> x.fD
  | E -> x.fE
  | F -> x.fF
  | G -> x.fG
  }
def arg: A[int] | B[int] | C[int] | D[int] | E[int] | F[int] | G[int]
foo arg
//│ foo: (A[?] & {fA: 'a} | B[?] & {fB: 'a} | C[?] & {fC: 'a} | D[?] & {fD: 'a} | E[?] & {fE: 'a} | F[?] & {fF: 'a} | G[?] & {fG: 'a}) -> 'a
//│ arg: A[int] & {fA: int} | B[int] & {fB: int} | C[int] & {fC: int} | D[int] & {fD: int} | E[int] & {fE: int} | F[int] & {fF: int} | G[int] & {fG: int}
//│ res: int
//│ constrain calls: 11627
//│ annoying  calls: 77290
```
(This took about 10s on my machine.)

This is because pattern matching infers types of the form `A | B & ~A | C & ~A & ~B | ...` (internally, before the type is simplified for showing to the user), which are implicitly put into conjunctive normal form during constraint solving, as in `(A | B | C | ...) & (A | ~A | C | ...) & (A | B | ~A | ...) & (A | ~A | ~A | ...) & (A | B | ~B | ...) & (A | ~A | ~B | ...) | ...`.

Two things to note here:

 * inferred types like `B & ~A` can actually usually be reduced to just `B` thanks to single-inheritance of classes, but the solver does not currently do that;

 * if we were matching traits, we could't actually reduce these (lack of single-inheritance restriction for traits).

Thankfully, it is fairly easy to solve these constraints in polynomial time. We just need to put both sides in DNF first to whittle down the possible matches, and only decompose constraints when tricky things appear, such as type variables in only one of the RHS conjuncts.

LPTK:
This PR fixes https://github.com/hkust-taco/mlscript/issues/37.

It is probably best reviewed commit by commit (the commits are atomic).

@fo5for @chengluyu I added you as reviewers so you get a chance to look at the constraint solver. But I don't actually expect you to understand it in depth yet. So it's fine if your reviews are only superficial.


shared/src/test/diff/mlscript/RecursiveTypes.mls
```diff
@@ -307,39 +307,58 @@ f { a = 1 }
...
+//│ ║  l.342: 	f infina
 //│ ║         	  ^^^^^^
 //│ ╟── Note: constraint arises from field selection:
 //│ ║  l.275: 	rec def f x = if x > 0 then (f (x with { a = x - 1 })).a else x
 //│ ╙──       	                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 //│ res: error | int | (('b | (0 & {a: 'c} as 'c) | 'a)\a & {a: int} as 'a)
 
 
-// FIXME why does type printing show `nothing`?
+// TODO why do we get this weird `'a as 'a`? (Note: it used to show `nothing`.)
```
fo5for: I think this is because empty `Without`s are created in `LhsNf.&`, and they are not simplified away. So it's actually `Without('a, Set.empty) as 'a`. I think simplifying them away in `SimplifyType.transform` fixes them without any side-effects. (_I think..._ I'm not too familiar with the constraining and simplification parts.)

LPTK: Aha, we have a winner 😄 

Good catch!

fo5for (approved):
Looks good to me. There are some bugs in `RecursiveTypes.mls` but they were most likely also present before, just shadowed by other bugs.

chengluyu (approved):
I basically understand what was done in this PR but have no comments for now.

PR #52, opened 2021-12-21
Fix duplicated trailing commas for records and arrays

chengluyu:
This PR fixes duplicated trailing commas for records and arrays. This PR also makes it easier to create `SourceCode` and `SourceLine` by implementing `.apply`.

Before the fix,

```
def nested = { a = { value = "a"; level = 0 }; b = { value = "b"; level = 1 } }
```

generates

```js
let nested = {
  a: {,
    value: "a",,
    level: 0,,
  },
  b: {,
    value: "b",,
    level: 1,,
  },
};
```

After the fix, it generates

```js
let nested = {
  a: {
    value: "a",
    level: 0
  },
  b: {
    value: "b",
    level: 1
  }
};
```

### DIFF
--- a/shared/src/main/scala/mlscript/codegen/Codegen.scala
+++ b/shared/src/main/scala/mlscript/codegen/Codegen.scala
@@ -59,7 +59,7 @@ class SourceCode(val lines: Ls[SourceLine]) {
   def parenthesized(implicit run: Bool = true): SourceCode =
     if (run) {
       lines.length match {
-        case 0 => SourceCode.from("()")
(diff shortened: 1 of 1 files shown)


LPTK (approved):
Ok cool. Please go ahead and squash-merge it. Though we're still sorely missing JS tests, so that should be the next thing on the radar.

chengluyu:
> Ok cool. Please go ahead and squash-merge it. Though we're still sorely missing JS tests, so that should be the next thing on the radar.

I’m working on that! It’s near the horizon!

PR #53, opened 2021-12-21
Test JavaScript code generation

### LINKED ISSUE #34: Add JS tests

chengluyu:
# Summary

This PR add test suite and test cases for JavaScript code generation. It also has some improvements about code generation.

## Test pipeline

- Spawn a Node.js process for each test file.
- Create a `JSBackend` for each test file. The `JSBackend` will memorize declared symbols and prelude code.
- For each test block, `JSBackend` translates it into several queries. Then, we send each query to Node.js, fetch the output, and write it down.
 
By the way, only files ending with `.mls` have code generation tests.

## Test flags and their behaviors

I added one global flag and three local flags.

- `:NoJS` disables all code generation and execution.
- `:ne` disables execution but also generates JavaScript code.
- `:ng` disables code generation and execution.
- `:js` shows the generated JavaScript code.   

Some existing flags have further meanings now.

- `:e` also disables code generation and execution because it ill-typed code may crashes code generation.
- `:pe` also disables code generation and execution for the same reason as above.
- `:AllowTypeErrors` does the same.
- `// FIXME` and `// TODO` does the same.

Here is a tabular explanation of their usage.

| Flag                     | Generation         | Execution | Notes          |
|--------------------------|--------------------|-----------|----------------|
| `:NoJS`                  | :x:                | :x:       | New and global |
| `:ng`                    | :x:                | :x:       | New            |
| `:ne`                    | :white_check_mark: | :x:       | New            |
| `:e`                     | :x:                | :x:       | Extant         |
| `:pe`                    | :x:                | :x:       | Extant         |
| `:AllowTypeErrors`       | :x:                | :x:       | Extant         |
| `// FIXME` and `// TODO` | :x:                | :x:       | Extant         |

## Features and fixes for code generation

To make most test cases work, I implemented some code generation features.

- Translate `Tup(terms)` to array literals (`JSArray`). For example, `def t = (0, 0)` will be translated to `[0, 0]`. This corresponds with TypeScript.
- Translate `Tup(fields)` in arguments to array pattern. For example, `def f((x, y)) = x + y` will be translated to `([ x, y ]) => (x + y)`.
- Inheriting traits and a class at the same time will not cause error `"Cannot have two base classes"`.

### Learn from tests

- [x] Pure definitions referenced in later tests will cause the error `Uncaught ReferenceError: <identifier> is not defined`. A possible solution is to generate function stubs for them. https://github.com/hkust-taco/mlscript/blob/08d53d1d13507303a8fbfaa9e0edfb1d700c2634/shared/src/test/diff/mlscript/RecursiveTypes.mls#L151-L157 **Follow-up**: let’s ignore them for now.
- [x] Lack in built-in functions such as `succ`, `id`. https://github.com/hkust-taco/mlscript/blob/08d53d1d13507303a8fbfaa9e0edfb1d700c2634/shared/src/test/diff/mlscript/Seqs.mls#L75-L78 **Follow-up**: implemented in a65be87.
- [x] In DiffTests, `res` stands for the last evaluation result. I have not implemented it yet. https://github.com/hkust-taco/mlscript/blob/08d53d1d13507303a8fbfaa9e0edfb1d700c2634/shared/src/test/diff/mlscript/GenericClasses.mls#L79-L81 **Follow-up**: fixed in 3c22732.
- [x] Wrap object literals with parentheses when turn it into a statement. https://github.com/hkust-taco/mlscript/blob/08d53d1d13507303a8fbfaa9e0edfb1d700c2634/shared/src/test/diff/mlscript/Ascribe.mls#L17-L19 **Follow-up**: fixed in 6c22ad2.

### Solved issues

- [x] Fix the failed CI.
- [x] The test process waits forever after sending multiple lines to Node.js. Solution: to fit Node.js’ multiple line mode.
- [x] The test cannot parse multi-line reply from Node.js. I need find a way to drain the `stdout` of child process.
  **Solution** I use a most primitive way of draining `stdout` of Node.js: consume until meet a sequence of `"\n >"`, which is the prompt.
- [x] Test cases in the same files are not running together.

## Call for suggestions

- [x] The format of execution results. Suggestion from @LPTK: interleaving the results with the type definitions.
- [x] The test mode names. I created two modes: `r` for executing generated code and `g` for showing the generated code. How do you like that?


LPTK:
Nice! Looking forward to having proper JS tests in the test suite.

> I created two modes: `r` for executing generated code and `g` for showing the generated code. How do you like that?

I think the default should be to execute _all_ well-typed code through the JS backend. The optional commands should be to opt out of JS code execution (for example call it `:nr` or `:no-run`) as well as to opt out of code-gen altogether (for example call it `:ng` or `:no-gen`). Opting out would be useful when some features are not yet implemented or when the code would get into an infinite loop for example.

There should also be a `:re` option for specifying that a runtime exception is expected.

Finally, there could be a global `:NoJS` option to completely opt out of the JS backend for a given file. Similar to `AllowTypeErrors`: https://github.com/hkust-taco/mlscript/blob/8b3df1dc25e2279e44a8440e8e85dbfc8bf08c30/shared/src/test/scala/mlscript/DiffTests.scala#L100

chengluyu:
Update: Node.js may replies multiple lines for some code. I need to exhaust child process’ `stdout` before sending more lines. However, I didn’t find a way to check if there is more lines. `p.stdout.available` does not always work. I’m figuring it out.

chengluyu:
Because I haven’t push anything new to this PR in past few days. I decided to write a sum up about the status of quo. 

In a word, the REPL method is not feasible for code generation test (I’ll explain later) so I turn to another approach. But this approach requires some latest code generation features, which is included in #54. Therefore, I’m going to make #54 a *mega PR* and merge PR into #54. Also, I’ll rename #54 when this PR is ready.

Now the basic tests are ready. But existing test files contains some features that current code generation doesn’t cover. For example, tuple constructions and destructions. I’m currently working on them. Finally, this PR will contain:

- code generation tests,
- bug fixes of code generations,
- features covered in the test cases.

chengluyu:
# Why not use the REPL approach?


An ideal execution loop should be similar to this.

```scala
node.stdin.writeLine("console.log('Hello, world!')")
node.stdin.flush()
val buffer = new StringBuilder()
while (node.stdout.wrapped.available() > 0) {
  buffer.append(node.stdout.wrapped.read().toChar)
}
val result = buffer.toString
```


Note that we’re trying to drain the stream because Node.js might reply with multiple lines.

The problem is that when the while loop begins, the process may have not respond yet. A solution is to `Thread.sleep` a fair amount of time before reading anything. But it is not feasible for two reasons.


1. The “fair amount” is sensitive to many factors. For example, sometimes waiting 20 ms is enough for reading the hello message, but sometimes it is not.

	Moreover, the time might be variable on different machines. I didn’t test on other machine but I estimated so. So it is hard to find a minimal time.

	Longer waiting time will definitely read all lines but would increase the test time. For example, if we wait 50 ms after each `writeLine`, one test case would cost more than one second.


2. We will never know whether Node.js is ready for next input because it might reply with multiple lines and it may have different intervals between these lines.

By the way, it is not feasible to write send generated code to Node.js and waits for the output because we don’t know the correspondence between input and output.

Then I have other two solutions.


1. Communicate with Node.js via IPC. This should works fine but requires a lot of efforts.
2. Give up the idea of interact with Node.js REPL. Instead, we execute a curated code by command `node --print <script>` which reports results line by line.

I implemented solution 2.

LPTK:
Thanks for the detailed rundown. As discussed in the meeting, no need to turn this into a mega PR. We can just wait until #54 is merged.

shared/src/test/diff/mlscript/Ascribe.mls
```diff
@@ -1,61 +1,71 @@
...
 //│ ╟── expression of type `"hello"` is not a function
 //│ ╟── Note: constraint arises from function type:
-//│ ║  l.6: 	"hello": 'a -> 'a
+//│ ║  l.7: 	"hello": 'a -> 'a
 //│ ╙──     	         ^^^^^^^^
 //│ res: 'a -> ('a | error)
 
 { name = "Bob" } : { name: 'a }
 //│ res: {name: "Bob"}
+//│    = 'Bob'
```
LPTK: Interesting. This should evaluate to a record, not just a string.

shared/src/test/diff/mlscript/ExprProb.mls
```diff
@@ -138,25 +160,49 @@ eval1 done e1
...
+//│    = Uncaught ReferenceError: intToString is not defined
+//│          at REPL23:1:69
+//│          at REPL24:1:235
 
 e1 = add (lit 1) (add (lit 2) (lit 3))
 eval1 done e1
 pretty1 done e1
 prettier1 done (eval1 done) e1
 //│ e1: Add['E .. 'E | (Add['E0 .. 'E0 | Lit & {val: 2} | Lit & {val: 3}] with {lhs: Lit & {val: 2}, rhs: Lit & {val: 3}}) | Lit & {val: 1}] with {lhs: Lit & {val: 1}, rhs: Add['E0 .. 'E0 | Lit & {val: 2} | Lit & {val: 3}] with {lhs: Lit & {val: 2}, rhs: Lit & {val: 3}}}
+//│   = '[object Object][object Object][object Object]'
```
LPTK: Whoops... some objects were concatenated?

shared/src/test/diff/mlscript/Ascribe.mls
chengluyu: Generated code is `{  name: "Bob"  };`. So I should wrap object literals with parentheses otherwise the parser interprets them as blocks with a label followed by a string literal.

shared/src/test/diff/mlscript/ExprProb.mls
chengluyu: In the era without binary expressions, I hard coded the translation from `add x y` to `x + y`. So the definition here does not work. I’ll remove this conversion. https://github.com/hkust-taco/mlscript/blob/08d53d1d13507303a8fbfaa9e0edfb1d700c2634/shared/src/test/diff/mlscript/ExprProb.mls#L13

shared/src/test/diff/mlscript/Ascribe.mls
LPTK: Wow, JavaScript's parser never ceases to amaze me 😂

chengluyu: By the way, I believe some generated code trigger bad [semicolon insertion](https://stackoverflow.com/questions/2846283/what-are-the-rules-for-javascripts-automatic-semicolon-insertion-asi) position somewhere ... until we find it.

LPTK:
This is shaping up quite nicely, cool work!

Some remarks:

### 1

> [ ] Pure definitions referenced in later tests will cause the error `Uncaught ReferenceError: <identifier> is not defined`. A possible solution is to generate function stubs for them.

There is no good thing to do here. The binding is indeed missing. I think these test cases should be marked `:ne` on an individual basis (e.g., for those in ExprProb) or if they're all like that just enter `:NoJS` mode (e.g., for the end of RecursiveTypes).

### 2

You should probably add one or two files that print JS code-gen for representative examples, just to have a trace of what it does and more easily catch regressions.

### 3

Consider this program:

```ts
:js
n = 123 with { x = 1 }
//│ // Prelude
//│ function prettyPrint(value) {
//│   switch (typeof prettyPrint) {
//│     case "number": 
//│     case "boolean": 
//│     case "function": 
//│       return "[Function: " + (value.name ?? "<anonymous>") + "]";
//│     case "string": 
//│       return "\"" + value + "\"";
//│     default:
//│       return value.constructor.name + " " + JSON.stringify(value);
//│   }
//│ }
//│ function withConstruct(target, fields) {
//│   const copy = Object.assign({}, target, fields);
//│   Object.setPrototypeOf(copy, Object.getPrototypeOf(target));
//│   return copy;
//│ }
//│ // Query 0
//│ const n = withConstruct(123, {  x: 1  }); n;
//│ // End of generated code
//│ n: 123 & {x: 1}
//│  = Number { x: 1 }

:js
n + 1
//│ // Prelude
//│ function withConstruct(target, fields) {
//│   const copy = Object.assign({}, target, fields);
//│   Object.setPrototypeOf(copy, Object.getPrototypeOf(target));
//│   return copy;
//│ }
//│ // Query 0
//│ n + 1;
//│ // End of generated code
//│ res: int
//│    = Uncaught TypeError: Number.prototype.valueOf requires that 'this' be a Number
//│          at Number.valueOf (<anonymous>)
```

I wonder why `withConstruct` is re-inserted in the second prelude.

Also, is there a way to make `with` work with numbers?

### 5

Is there an easy way of detecting crashes (e.g., exceptions being raised)? These should incur test case failures. And there should be a way of denoting expected crash (call it `:ee` for "execution error"), so it's not reported as a failure.

### 4

Maybe we should still execute code that does not type check, but just expect and ignore any resulting runtime errors.

chengluyu:
The commit 5bcc563 removes the conversion which caused the concatenated object problems. It also makes some test cases failed because `add`, `mul` are not defined.

chengluyu:
Reply to the remarks.

- [x] In reply to
   > There is no good thing to do here. The binding is indeed missing. I think these test cases should be marked `:ne` on an individual basis (e.g., for those in ExprProb) or if they're all like that just enter `:NoJS` mode (e.g., for the end of RecursiveTypes).

   Let’s mark the test cases as no execution for now. I sketched an update for `Scope`, which could track attributes of defined symbols. With the update, when reference unimplemented identifiers, the code generation will emit nothing and warn. The update will be introduced in #55.

   **Follow-up**
   
   Since this is very far and does not affect current tests. I marked it as future work.
- [x] In reply to
   > You should probably add one or two files that print JS code-gen for representative examples, just to have a trace of what it does and more easily catch regressions.

   I’m adding some basic cases as described in `translateTerm`.

   **Follow-up**

   - 9fae057 Add more test blocks and support more binary operators
   - c8396dc Merge branch 'regression-prevention' into codegen-test
   - c62b5b3 Add some cases as described in `translateTerm` and fix parens
   - 286d3fc Preserve lines of generated code and print them with `:js` on
- [x] In reply to
  > Consider this program: [...]
 
  Duplicated prelude is a bug. I’ll fix it. For the question, the answer is yes. Just `Object.assign(0, { /* ... */ }`. My implementation of `withConstruct` doesn’t cover this case. Thanks for pointing out!

  **Follow-up**
  
  - ed08872 handled extension on extended primitives.
  - 4e3e29e handled primitive types in `withConstruct`.
  - 3990551 fixes the duplication problem.
- [x] In reply to
   > Maybe we should still execute code that does not type check, but just expect and ignore any resulting runtime errors.

   Let me have a try.

   **Follow-up**

   Since ed2b155, test cases which failed to type check will be execute anyway. Some `Uncaught ReferenceError: res is not defined` disappeared. I also added detection of code generation crashes.
- [ ] In reply to
  > Is there an easy way of detecting crashes (e.g., exceptions being raised)? These should incur test case failures. And there should be a way of denoting expected crash (call it `:ee` for "execution error"), so it's not reported as a failure.

  Parsing output from Node.js seems feasible. I have no better idea about this for now. I also want to add timeout to `stream.read()` otherwise we will occasionally suffer from infinite loops.

LPTK:
> * Parsing output from Node.js seems feasible.

That seems messy. Can you just generate a try/catch that produces something which can be easily recognized and displayed as an execution error?

chengluyu:
> > * Parsing output from Node.js seems feasible.
> 
> That seems messy. Can you just generate a try/catch that produces something which can be easily recognized and displayed as an execution error?

Good idea! Let me try it!

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -383,6 +377,14 @@ class JSBackend {
       case term => JSClassGetter(name, L(translateTerm(term)))
     }
   }
+}
+
+class JSWebBackend extends JSBackend {
+  // Name of the array that contains execution results
+  val resultsName: Str = topLevelScope allocateJavaScriptName "results"
+
+  // TODO: remove this the prelude code manager is done
```
```suggestion
  // TODO: remove this when the prelude code manager is done
```

shared/src/test/diff/codegen/Polyfill.mls
```diff
@@ -0,0 +1,26 @@
...
+//│     Object.setPrototypeOf(copy, Object.getPrototypeOf(target));
+//│     return copy;
+//│   }
+//│ }
+//│ let res;
+//│ // Query 0
+//│ const n = withConstruct(123, {  x: 1  }); res = n;
+//│ // End of generated code
+//│ n: 123 & {x: 1}
+//│  = [Number: 123] { x: 1 }
```
LPTK, line 78: I wanted to forbid this in the type system eventually, but it turns out JavaScript can do it. Thankfully I had not added the restriction yet! 🤪 

So, can we also extend things like numbers and booleans in classes?

shared/src/test/diff/mlscript/BadAlias.mls
```diff
@@ -14,7 +14,7 @@ type A = A
 
 res: "oops"
 //│ res: "oops"
-//│    = Uncaught ReferenceError: res is not defined
+//│    = undefined
```
LPTK: Why `undefined`?

LPTK:
This does not work:

```ts
n = 123 with { x = 1 }
//│ n: 123 & {x: 1}
//│  = [Number: 123] { x: 1 }

m = n with { y = 2 }
//│ m: 123 & {x: 1, y: 2}
//│  = Number { x: 1, y: 2 }

m + 1
//│ res: int
//│    = Uncaught TypeError: Number.prototype.valueOf requires that 'this' be a Number
//│          at Number.valueOf (<anonymous>)
```

shared/src/test/diff/mlscript/BadAlias.mls
chengluyu: 🧐 Because previous block is marked as `:e`.

> Maybe we should still execute code that does not type check, but just expect and ignore any resulting runtime errors.

Yes, you’re right. By doing so, here will be `"oops"`.

LPTK: Oh right, I see.

shared/src/test/diff/codegen/Polyfill.mls
chengluyu: Yes, you can!

```js
class Test extends Number {
  constructor() {
    super();
    this.x = 0;
  }
}
// undefined
new Test()
// [Number (Test): 0] { x: 0 }
new Test() + 1
// 1
new Test() + 100
// 100
```

Note that `[Number (Test): 0] { x: 0 }` indicates it’s actually a number.

chengluyu:
> This does not work:
> ```ts
> 
> n = 123 with { x = 1 }
> //│ n: 123 & {x: 1}
> //│  = [Number: 123] { x: 1 }
> 
> m = n with { y = 2 }
> //│ m: 123 & {x: 1, y: 2}
> //│  = Number { x: 1, y: 2 }
> 
> m + 1
> //│ res: int
> //│    = Uncaught TypeError: Number.prototype.valueOf requires that 'this' be a Number
> //│          at Number.valueOf (<anonymous>)
> ```

Another uncovered case! I'd improved the `withConstruct` tomorrow.

shared/src/test/diff/mlscript/BadAlias.mls
chengluyu: Changed since ed2b155: even marked as `:e`, test cases will be executed anyway.

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -370,6 +371,14 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite {
                 }
                 showResult(prefixLength)
             }
+
+            // If code generation fails, show the error message.
+            results match {
+              case L(err) =>
+                output("Code generation crashed with following error:")
```
```suggestion
                output("Code generation crashed with the following error:")
```

shared/src/test/diff/mlscript/BadInherit.mls
```diff
@@ -26,42 +33,64 @@ def f(x:Child) = x: nothing
...
 :e
 type Sly1 = Parent2
 class Child2: Parent1 & Sly1
 //│ ╔══[ERROR] cannot inherit from a type alias
-//│ ║  l.46: 	class Child2: Parent1 & Sly1
+//│ ║  l.54: 	class Child2: Parent1 & Sly1
 //│ ╙──      	      ^^^^^^^^^^^^^^^^^^^^^^
 //│ Defined type alias Sly1
+//│ Code generation crashed with following error:
+//│   Cannot have two base classes: Parent1, Parent2
```
LPTK, line 56: Note: for "expected" errors like this, it would be cleaner to raise a proper CodeGenError object, and not display the stack trace. Displaying stack traces in test output is annoying as the line numbers often change.

LPTK: Note note: We still want to catch proper code-gen _crashes_ and report traces for them. These should just not appear in tests.

chengluyu: Ah I see. I was doubting if I should include stack traces here.

chengluyu: So should I hide stack traces from Node.js REPL?

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -265,6 +274,57 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite {
...
+                  output("// End of generated code")
+                }
+                // Execute code.
+                if (!mode.noExecution) {
+                  testCode.prelude match {
+                    case Nil => ()
+                    case lines => host.execute(lines mkString " ")
+                  }
+                  S(testCode.queries map { q =>
+                    // Useful for find out what are really happening.
```
```suggestion
                    // Useful for find out what is really happening.
```

chengluyu: I should install a spell checker. 🤡

shared/src/test/diff/mlscript/BadInherit.mls
LPTK: I don't think we need to hide Node stack traces. These refer to line numbers in test code. Whereas code-gen error stack traces refer to line numbers in our implementation!

chengluyu: Ah I see the logic. Thanks!

chengluyu: Resolved in 823fc99.

shared/src/test/diff/codegen/Polyfill.mls
```diff
@@ -41,15 +41,18 @@ n = 123 with { x = 1 }
 //│   return !x;
 //│ }
 //│ let res;
-//│ // Query 0
-//│ const n = withConstruct(123, {  x: 1  }); res = n;
+//│ // Query 1
+//│ const n = withConstruct(123, { 
+//│ x: 1
```
LPTK: Why the strange indentation?

chengluyu:
I created a separated test file [shared/src/test/diff/codegen/Terms.mls](https://github.com/hkust-taco/mlscript/blob/9fae057a2d9b6621e3f313238a5c852e0f743b82/shared/src/test/diff/codegen/Terms.mls#L1). All blocks in this file are marked with `:js` so that we can track progression of code generation. 

**Related Commits**

- 9fae057 Add more test blocks and support more binary operators
- c8396dc Merge branch 'regression-prevention' into codegen-test
- c62b5b3 Add some cases as described in `translateTerm` and fix parens
- 286d3fc Preserve lines of generated code and print them with `:js` on

shared/src/test/diff/codegen/Terms.mls
```diff
@@ -0,0 +1,307 @@
...
+//│ res = (function (x) {
+//│   return x + 1;
+//│ })(0);
+//│ // End of generated code
+//│ res: int
+//│    = 1
+
+// Blk: Unused in MLParser
+
+// CaseOf: Just the wildcard
```
LPTK, line 234: You should also test `def f x = case x of { }`.

shared/src/test/diff/codegen/Polyfill.mls
chengluyu: I’m investigating it. 🧐

shared/src/test/diff/codegen/Terms.mls
chengluyu: Sure.

shared/src/test/diff/mlscript/BadInherit.mls
LPTK: Actually, it looks like I wasn't clear enough in my messages. You should distinguish actual _crashes_ (unexpected errors, such as stack overflow or `None.get`) from expected errors. Only the latter should be reported without a stack trace. The former is not supposed to happen ever, but if it does we want to see the stack trace.

LPTK:
> I created a separated test file [shared/src/test/diff/codegen/Terms.mls](https://github.com/hkust-taco/mlscript/blob/9fae057a2d9b6621e3f313238a5c852e0f743b82/shared/src/test/diff/codegen/Terms.mls#L1). All blocks in this file are marked with `:js` so that we can track progression of code generation.

Great!

shared/src/test/diff/codegen/Polyfill.mls
```diff
@@ -50,10 +52,50 @@ n = 123 with { x = 1 }
 //│ n: 123 & {x: 1}
 //│  = [Number: 123] { x: 1 }
 
-:js
-n + 1
```
LPTK: Why remove the `n + 1` test? I think it's also useful.

shared/src/test/diff/codegen/Polyfill.mls
chengluyu: Fixed! https://github.com/hkust-taco/mlscript/blob/80d13bbcc89cf501557aeb152c1ada3c8e93cd93/shared/src/test/diff/codegen/Polyfill.mls#L47

shared/src/test/diff/mlscript/BadInherit.mls
chengluyu: Crashes and errors are distinguishable since f81cc67!

shared/src/test/diff/codegen/Polyfill.mls
chengluyu: It was back!

shared/src/test/diff/mlscript/Ascribe.mls
```diff
@@ -59,8 +59,8 @@ def foo({ a = "hey" }: { a: string}) = 0
 //│ ║  l.51: 	def foo({ a = "hey" }: { a: string}) = 0
 //│ ╙──      	              ^^^^^
 //│ foo: {a: string} -> 0
-//│    = const foo4 = ({  a:   }) => 0; res = foo4;
-//│                            ^
+//│    = const foo4 = ({ a:  }) => 0; res = foo4;
```
LPTK: This is a code-gen bug that should be fixed, right?

chengluyu: Just fixed!

chengluyu:
Now I’m working on detecting runtime errors. There remains several issues.

- From `Ascribe.mls` and `Ops.mls`: unable to detect `SyntaxError` because it cannot be caught.
- From `BadInherit.mls`, `BadTraits.mls`, `RecursiveTypes.mls` and `Subsume.mls`: to use the result, we need to declare variables before the try statement, then assign the variable with the expression. So if the initialization of a variable met errors, it will be defined anyway and cause confusion if the variable is referenced in following expressions.

For example, we defined `t` first, then use `t`.

```
:js
def t = T error
//│ // Query 1
//│ let t;
//│ t = T(error);
//│ res = t;
//│ // End of generated code
//│ t: nothing
//│ Unexpected runtime error
//│   ReferenceError: T is not defined

:js
t: nothing
//│ // Query 1
//│ res = t;
//│ // End of generated code
//│ res: nothing
//│    = undefined
```

I’m considering fixing this by writing variables directly to `globalThis`. For example,

```
def t = T error
```

will be translated into

```js
globalThis["t"] = <expr>;
```

so that if there is anything wrong inside `<expr>`, `t` will be not defined.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -6,6 +6,7 @@ import collection.mutable.{HashMap, HashSet, Stack}
 import collection.immutable.LazyList
 import scala.collection.immutable
 import mlscript.codegen.{CodeGenError, Scope}
+import javax.xml.transform.Source
```
LPTK: Probably added by mistake.

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -275,7 +277,7 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite {
               }
             )
 
-            var results: (Str, Bool) \/ Opt[Ls[Ls[Str]]] = if (!allowTypeErrors &&
+            var results: (Str, Bool) \/ Opt[Ls[Str \/ Str]] = if (!allowTypeErrors &&
```
LPTK: `Str \/ Str` could just be a `(Bool, Str)`

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -77,8 +77,10 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite {
       expectTypeErrors: Bool, expectWarnings: Bool, expectParseErrors: Bool,
       fixme: Bool, showParse: Bool, verbose: Bool, noSimplification: Bool,
       explainErrors: Bool, dbg: Bool, fullExceptionStack: Bool, stats: Bool,
-      noExecution: Bool, noGeneration: Bool, showGeneratedJS: Bool)
-    val defaultMode = Mode(false, false, false, false, false, false, false, false, false, false, false, false, false, false)
+      noExecution: Bool, noGeneration: Bool, showGeneratedJS: Bool,
+      expectRuntimeError: Bool)
```
LPTK: This seems unused.

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -318,9 +320,15 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite {
             def showResult(prefixLength: Int) = {
               results match {
                 case R(S(head :: next)) =>
-                  head.zipWithIndex foreach { case (s, i) =>
-                    if (i == 0) output(" " * prefixLength + "= " + s)
-                    else output(" " * (prefixLength + 2) + s)
+                  val text = head match {
+                    case L(err) =>
+                      output("Unexpected runtime error")
```
LPTK: I'd change this to just `"Runtime error"` when the code block has already failed to type check. The error isn't exactly unexpected 😛

LPTK:
> * From `Ascribe.mls` and `Ops.mls`: unable to detect `SyntaxError` because it cannot be caught.

The only solution is to make sure we only produce syntactically-valid code! :^P
BTW, why is the syntax error now swallowed and not printed at all? The previous behavior for these cases was more desirable, as now it fails totally silently.

> * From `BadInherit.mls`, `BadTraits.mls`, `RecursiveTypes.mls` and `Subsume.mls`: to use the result, we need to declare variables before the try statement, then assign the variable with the expression. So if the initialization of a variable met errors, it will be defined anyway and cause confusion if the variable is referenced in following expressions.

So the only difference is this, right?
```diff
-//│    = Uncaught ReferenceError: c2 is not defined
+//│    = undefined
```

If so, I don't think it matters much. It may even be better this way, to clear any previously defined value for this variable name.

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: Oops, it seems auto-import malfunctioned.

shared/src/test/scala/mlscript/DiffTests.scala
chengluyu: I see.

shared/src/test/scala/mlscript/DiffTests.scala
chengluyu: Ah, it's used in the following commits.

LPTK:
> It may even be better this way, to clear any previously defined value for this variable name.

It would also be good to clear the value of `res` before reassigning it, so the old value is not carried forward and printed in case the `res` reassignment fails for some reason, as in here: https://github.com/hkust-taco/mlscript/pull/53/commits/d0aae5f8e10965853a2cdaff84fc8fcf4c2a7721#diff-6b36259a266bc85a957033c8083ee116942641d9f6b1e632137a7de9c41200f5R45

chengluyu:
> > * From `Ascribe.mls` and `Ops.mls`: unable to detect `SyntaxError` because it cannot be caught.
> 
> The only solution is to make sure we only produce syntactically-valid code! :^P BTW, why is the syntax error now swallowed and not printed at all? The previous behavior for these cases was more desirable, as now it fails totally silently.

I’m investigating this.

**Follow-up** Because the parenthesis issue, which cause the syntax error, has been fixed.

> 
> > * From `BadInherit.mls`, `BadTraits.mls`, `RecursiveTypes.mls` and `Subsume.mls`: to use the result, we need to declare variables before the try statement, then assign the variable with the expression. So if the initialization of a variable met errors, it will be defined anyway and cause confusion if the variable is referenced in following expressions.
> 
> So the only difference is this, right?
> 
> ```diff
> -//│    = Uncaught ReferenceError: c2 is not defined
> +//│    = undefined
> ```
> 
> If so, I don't think it matters much. It may even be better this way, to clear any previously defined value for this variable name.

I think we have to show `ReferenceError` rather than showing a `undefined` because other functions may also produce undefined.

chengluyu:
By the way, how to report failures when runtime errors occur?

I’m reviewing of changes from all commits in order to collect problems.

LPTK:
> By the way, how to report failures when runtime errors occur?

Add a new `totalExecFailures` var here: https://github.com/hkust-taco/mlscript/blob/89f7c6f3186c505fcc2a7b09c838e3072417024a/shared/src/test/scala/mlscript/DiffTests.scala#L159-L160

And check it accordingly.

Also update the `failures` buffer accordingly.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -518,8 +518,11 @@ class JSTestBackend extends JSBackend {
           val translatedBody = translateTerm(body)(topLevelScope)
           val jsName = topLevelScope declare mlsName
           S(
-            S(JSLetDecl.from(jsName :: topLevelScope.getTempVars())) ->
-             ((JSIdent(jsName) := translatedBody) ::
+            topLevelScope.emitTempVarDecls() ->
+             ((JSIdent("globalThis").member(jsName) := (translatedBody match {
```
LPTK, line 545: By the way, why `globalThis`? It may contain other definitions that will clash with the ones we register. All we need is a simple local dictionary.

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -265,6 +276,64 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite {
...
+            } else {
+              R(N)
+            }
+
+            def showResult(prefixLength: Int) = {
+              results match {
+                case R(S(head :: next)) =>
+                  val text = head match {
+                    case (false, err) =>
+                      output("Runtime error")
```
```suggestion
                      output("Runtime error:")
```

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: Because I want to make sure that if some errors occur during the initialization of a variable, it will not declare the name. By doing so, future expressions which reference the variable can throw `ReferenceError` rather than evaluate `undefined` and produce some strange values.

For example, with the previous execution strategy, it produces `NaN` for the second test block.

```
def a = <something error>
// Generated code:
// let a;
// try { a = <...>; res = a; } catch (e) { ... }
// Runtime error: ...

a + 1
// Result: NaN
```

If we use the current approach, the reference to `a` causes `ReferenceError`. 

```
def a = <something error>
// Generated code:
// try { someResultDictionary.a = <...>; res = a; } catch (e) { ... }
// Runtime error: ...

a + 1
// Runtime error: ReferenceError: a is not defined
```

The reason why I used `globalThis` are as follows.

1. in Node.js REPL, declare a un-scoped variable is equivalent to writing it to the global object. Here is an example from REPL.
   
   ```
   ❯ node
   Welcome to Node.js v17.2.0.
   Type ".help" for more information.
   > a = 0
   0
   > globalThis.a
   0
   > var b = 5
   undefined
   > globalThis.b
   5
   ```
2. I have to replace all code generation for identifiers by member access of that local dictionaries.

LPTK: Ah, ok sure.

chengluyu: But I agree with you, `globalThis` may cause conflicts. So I'm implementing the 2 in my previous comment.

chengluyu:
Two more steps.

- [x] Some runtime errors are caused by code generation. For example, trait initializations. ~~They could work previously after some commits.~~ Find out why.
- [x] Merge with the `mlscript` branch locally and re-run tests.

chengluyu:
I reviewed all cases required `:re` flag because I’m afraid of if there are some hidden errors.

- [x] `BadInherit.mls`: `Crazy` has no definition, so following blocks have RE. **This is fine.**
- [x] `BadTraits.mls`: Trait initialization with an identifier as the argument are supported. **Fixed in 0b20529.**
- [x] `ExprProb.mls`: Some are caused by the lack of built-in methods (**fixed in ec9de56 and 8e38dfb**), whereas some are caused by missing implementations.
- [x] `GenericClasses.mls`: They are caused by the lack of built-in `error`. **Fixed in ec9de56.**
- [x] `Lists.mls`: The `originalCons` at line 9 leak the constructor function. **Fixed in efdcc18.**
- [x] `MethodAndMatches.mls`: Some functions have no definitions. **This is fine.**
- [x] `Methods.mls`: The calling convention `Class.Member instance` is not supported yet. **Mark as future work.**
- [x] `MLTuples.mls`:
  - [x] Same as `BadTraits.mls`, trait initialization with expressions as the argument is not implemented. **Fixed in 0b20529 respectively.**
  - [x] The last block fails because of the lack of `bool`. **Where does the `bool` comes from?**
- [x] `Paper.mls`: The built-in `div` not implemented. **Fixed in 8e38dfb.**
- [x] `Seqs.mls`: How come about line 32? Should the `Cons` here be used as `Cons { head = 1; tail = Nil }`? **Confirmed. It’s okay.**
- [x] `Subsume.mls`: I don’t understand why the result of expression at line 207 is 1. **Confirmed. It’s okay.**
- [x] `Tony.mls`: Try to call without `new` because the function is an alias to a constructor. **Fixed in efdcc18.**
- [x] `Traits.mls`: The built-in `error` is not implemented and the trait initialization problem. **Fixed in ec9de56 and 0b20529 respectively.**
- [x] `Trio.mls`: The `arg` has no implementation. **This is fine.**
- [x] `TypeDefs.mls`: The built-in `error` is not implemented. **Fixed in ec9de56.**
- [x] `With.mls`: `fooManual` has no implementation. **This is fine.**

I will ignore runtime errors caused by missing implementations and lack in built-in functions and try to figure out a solution for the rest. If no good solution, I will ignore them for now and create a issue for future work.

chengluyu:
Just merged with `mlscript` before merging because I need to re-run tests and make sure no new problems were involved. BTW I added `:NoJS` to all stress tests because they seems full of dummies.

Some new todos.

- [x] Check `Arrays.mls` and other newly created test in the merge commit.
- [x] Confirm if problems mentioned in previous comment still exists.
- [x] Fix minor issues which seems easy to fix.
- [ ] Click the squash and merge button.
- [x] Create an issue for things not finished in this PR.

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: I’ll leave this as future work.

shared/src/test/scala/mlscript/DiffTests.scala
chengluyu: It’s used in runtime error flags. I’m marking this as resolved.

LPTK:
Great work! We're finally nearing the end of this PR 😃 

> * ~`ExprProb.mls`: Some are caused by the lack of built-in methods, whereas some are caused by missing implementations.~

Funny that the recursive functions seem to work here.

> * `Lists.mls`: The `originalCons` at line 9 leak the constructor function.

Same remark as below for `Tony.mls`.

> * `Seqs.mls`: How come about line 32? Should the `Cons` here be used as `Cons { head = 1; tail = Nil }`?

Yes, it should be used this way. The type and runtime errors are correct. I think the `// FIXME` is obsolete.

> * `Subsume.mls`: I don’t understand why the result of expression at line 207 is 1.

`f` has declared type `(int & 'a) -> 'a`, so passing `1` returns `1`. But `f` is implemented wrongly, so the runtime error is expected.

> * `Tony.mls`: Try to call without `new` because the function is an alias to a constructor.

I think in JS code-gen you should eta expand references to class constructors. As in, instead of compiling `None` to `None`, you should compile it to `x => new None(x)`.

> * `Traits.mls`: The built-in `error` is not implemented and the trait initialization problem.

`error` should be implemented easily.

shared/src/test/diff/mlscript/ExprProb.mls
```diff
@@ -48,38 +55,60 @@ rec def eval1 k e = case e of {
   | _ -> k e
   }
 //│ eval1: ('a -> int) -> ((Add[?] with {lhs: 'b, rhs: 'b}) | Lit | 'a & ~add & ~lit as 'b) -> int
+//│      = [Function: eval1]
 //│ constrain calls  : 78
 //│ annoying  calls  : 0
 //│ subtyping calls  : 655
 
 :ns
+:stats
```
LPTK: Why add `:stats` here?

chengluyu: Maybe I was a little bit unconscious when I dealt with merge conflicts here.

chengluyu: Removed.

shared/src/test/diff/codegen/Terms.mls
```diff
@@ -96,7 +96,7 @@ T { x = 0 }
 :js
 A { a = 0 }
 //│ // Query 1
-//│ res = new A({ a: 0 });
+//│ res = ((x) => new A(x))({ a: 0 });
```
LPTK: For performance and readability reasons, we should avoid doing this in applied constructor expressions, as in here!

LPTK:
> Where does the bool comes from?

It's okay to put `:re` here. This `bool` in terms is a remnant from an older prototype and will be removed.

shared/src/test/diff/codegen/Terms.mls
chengluyu: Fixed!

chengluyu:
I think this PR is nearly done! I’m summarizing future works.

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -359,7 +453,72 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite {
     }
     if (testFaield)
       fail(s"Unexpected diagnostics (or lack thereof) at: " + failures.map("l."+_).mkString(", "))
-    
+    host.terminate()
```
LPTK: This should probably be written as:

```scala
val host = ReplHost()
try {

  ...

} finally host.terminate()
```

to avoid leaking processes.

LPTK (approved):
There are warnings in your code: https://github.com/hkust-taco/mlscript/runs/4772836013?check_suite_focus=true#step:4:19
Please make sure to address them.

Trait are not supported, but we don't have anything to track this missing implementation. Please open a new issue or add a task to https://github.com/hkust-taco/mlscript/issues/65.

Otherwise: Awesome, LGTM! Please squash and merge after addressing the above and the comment below.

PR #54, opened 2021-12-24
Fix re-declaration semantics and refine generated code (close #49)

### LINKED ISSUE #49: Wrong semantics of re-declarations
Look at this example first.

Code:

```
def x = 0
def getX () = x
def x = 1
def y = getX ()
```

Types and results:

```
val x: 0 = 0
val getX: () -> 0 = () => x
val x: 1 = 1
val y: 0 = 1 // <---- notice this
```

My current solution is to rebind the re-declared `x` to another name. For example, generated code should be like this:

```js
let x = 0;
let getX = () => x;
let x0 = 1;
let y = getX(); // Evaluates to `0`, which matches the type.
```

chengluyu:
This PR fix #49 and add rename operations to `Scope`. Besides, this PR simplify and improve generated code. 

### Source MLscript

```
class Test: { x: int }
def Test x = Test { x }
def x = Test 0
def Test a = Test (a + 1)
def y = Test 0
def f Test = Test
def f x = 0
```

### Generated JavaScript

```js
(() => {
function prettyPrint(value) {
  switch (typeof value) {
    case "number":
    case "boolean":
    case "function":
      return value.toString();
    case "string":
      return '"' + value + '"';
    default:
      return value.constructor.name + " " + JSON.stringify(value, undefined, 2);
  }
}
const results = [];
class Test {
  constructor(fields) {
    this.x = fields.x;
  }
}
const Test1 = (x) => new Test({ x: x });
results.push(Test1);
const x = Test1(0);
results.push(x);
const Test2 = (a) => Test1(a + 1);
results.push(Test2);
const y = Test2(0);
results.push(y);
const f = (Test) => Test;
results.push(f);
const f1 = (x) => 0;
results.push(f1);
return results.map(prettyPrint);
})()
```

### Source MLscript

```
class A: { a: int }
class B: { b: int }

foo x = case x of { _ -> 1 }
foo x = case x of {
    | A -> x.a
    | _ -> 0
  }
def foo x = case id x of
  { A -> x.a
  | B -> x.b
  | _ -> 1
  }

```

### Generated JavaScript

```js
(() => {
// prettyPrint omitted
const results = [];
class A {
  constructor(fields) {
    this.a = fields.a;
  }
}
class B {
  constructor(fields) {
    this.b = fields.b;
  }
}
const foo = (x) => (x, 1);
results.push(foo);
const foo1 = (x) => (x instanceof A ? x.a : 0);
results.push(foo1);
const foo2 = (x) => {
  let temp;
  return temp = id(x), temp instanceof A ? x.a : temp instanceof B ? x.b : 1;
};
results.push(foo2);
return results.map(prettyPrint);
})()
```

Note that we cache the result with a temporary variable if there are more than two case branches in order to prevent duplicated evaluation.


js/src/main/scala/Main.scala
```diff
@@ -141,69 +125,54 @@ object Main {
...
+        val errorBuilder = new StringBuilder()
+        errorBuilder ++= "<font color=\"red\">Runtime error occurred:</font>"
+        errorBuilder ++= htmlLineBreak + e.getMessage
+        errorBuilder ++= htmlLineBreak
+        errorBuilder ++= htmlLineBreak
+        L(errorBuilder.toString)
     }
-    (defResults, exprResults, sb.toString)
   }
-
```
LPTK, line 186: Did you find a way to configure scalafmt so it does not do this? I'd like you not to commit changes like this. Removing indentation is fine if the whole team uses an automatic formatter, but for me it's a pain.

js/src/main/scala/Main.scala
```diff
@@ -346,8 +315,14 @@ object Main {
           results append S(d.nme) -> (getType(ty_sch).show)
         case d @ Def(isrec, nme, R(PolyType(tps, rhs))) =>
           val errProv = TypeProvenance(rhs.toLoc, "def signature")
-          val ty_sch = PolymorphicType(0, typeType(rhs)(ctx.nextLevel, raise,
-            vars = tps.map(tp => tp.name -> freshVar(noProv/*FIXME*/)(1)).toMap))
```
LPTK: Dude... 🙄

LPTK (changes requested):
You implementation has at least one small oversight:

```scala
def f: (int & 'a) -> (int & 'a) -> 'a
def f a b = if gt a b then a else b

f 1 2

f x y = x

f 1 3
```

yields:

```scala
val f: (int & 'a) -> (int & 'a) -> 'a = (a) => (b) => a > b ? a : b
val f: (int & 'a) -> (int & 'a) -> 'a = (x) => (y) => x
val res: 1 | 2 = 2
val f: 'a -> anything -> 'a = no value
val res: 1 = 1
```

js/src/main/scala/Main.scala
chengluyu: These annoying changes were reverted. I searched a while in their document and found nothing. So I opened an issue requesting their help.

js/src/main/scala/Main.scala
chengluyu: The change has been reverted. I’m making scalafmt ignore this file so it won’t happen again.

chengluyu:
> You implementation has at least one small oversight:
> 
> ```scala
> def f: (int & 'a) -> (int & 'a) -> 'a
> def f a b = if gt a b then a else b
> 
> f 1 2
> 
> f x y = x
> 
> f 1 3
> ```
> 
> yields:
> 
> ```scala
> val f: (int & 'a) -> (int & 'a) -> 'a = (a) => (b) => a > b ? a : b
> val f: (int & 'a) -> (int & 'a) -> 'a = (x) => (y) => x
> val res: 1 | 2 = 2
> val f: 'a -> anything -> 'a = no value
> val res: 1 = 1
> ```

The output on my side is different from yours.

```
val f: (int & 'a) -> (int & 'a) -> 'a = (a) => (b) => (a > b ? a : b)
val f: (int & 'a) -> (int & 'a) -> 'a = 2
val res: 1 | 2 = (x) => (y) => x
val f: 'a -> anything -> 'a = 1
val res: 1 = no value
```

I just pushed the fix. Type check results and execution results should be aligned now.

js/src/main/scala/Main.scala
LPTK: Ok cool.

js/src/main/scala/Main.scala
LPTK: Scalafmt seems way too opinionated about cases like this. No one wants to see code that looks like this, taking 8 lines of mostly empty lines, instead of just letting us neatly break it up in two.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -152,29 +153,36 @@ class JSBackend(pgrm: Pgrm) {
...
       )
-    }
-    case IntLit(value) => {
-      val useBigInt = MinimalSafeInteger <= value && value <= MaximalSafeInteger
-      JSLit(if (useBigInt) { value.toString }
-      else { value.toString + "n" })
-    }
+    // Pattern match with only one branch -> comma expression
+    case CaseOf(trm, Wildcard(default)) =>
+      println("what")
```
LPTK: Please remove all these lines of debug-printing.

LPTK (changes requested):
The example input in `local_testing.html` crashes with:

```
Runtime error occurred:
SyntaxError: Identifier 'temp' has already been declared
```

I minimized it to:

```scala
class Left[A]: { value: A }

def testVal = Left{ value = 1 }
def res = case testVal of
  { Left -> testVal.value
  }
```

PS: You should be doing these basic tests on your own, not wait that I do them myself.

js/src/main/scala/Main.scala
chengluyu: They said they won’t implement this feature. So ignoring this file won’t cause this kind of issue anymore.

js/src/main/scala/Main.scala
chengluyu: Yes. They politely refused my request. But anyway, now I’m reducing my use of formatter.

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: They were removed.

chengluyu:
> The example input in `local_testing.html` crashes with:
> 
> ```
> Runtime error occurred:
> SyntaxError: Identifier 'temp' has already been declared
> ```
> 
> I minimized it to:
> 
> ```scala
> class Left[A]: { value: A }
> 
> def testVal = Left{ value = 1 }
> def res = case testVal of
>   { Left -> testVal.value
>   }
> ```
> 
> PS: You should be doing these basic tests on your own, not wait that I do them myself.

Thanks for pointing it out! I did not consider this case after introducing 8d6800f. I should try to keep this PR atomic. 😅

js/src/main/scala/Main.scala
LPTK: Can you link the issue here please?

js/src/main/scala/Main.scala
LPTK: Is this another issue or the same as the one above? Link?

LPTK (approved):
Okay, this LGTM now. I think we should merge it.

js/src/main/scala/Main.scala
chengluyu: Link: scalameta/scalafmt#3006

PR #55, opened 2021-12-29
Add the prelude code manager (close #47)

### LINKED ISSUE #47: Insert prelude code in a neat way
The code generator often needs to insert some pre-defined code snippets (temporarily named “prelude code”) in the beginning of generated code. One example is `with` construct (#33, #35). Another example is `error` function. (In current JavaScript syntax, `throw x` is a statement rather than expression). Apart from them, I’d also like to implement `prettyPrint` as a prelude functions so the generated code is more readable. Current prelude code are implemented ad-hoc. It is not extensible and introduces mutable states to `JSBackend` class. To solve this, we may need a helper class maintaining which prelude code are needed.

chengluyu:


LPTK:
Let me know when this is ready for review.

shared/src/test/diff/codegen/Polyfill.mls
```diff
@@ -1,3 +1,62 @@
...
+// Now the add function is not the default one.
+:js
+add 1 2
+//│ // Query 1
+//│ res = add(1)(2);
+//│ // End of generated code
+//│ res: int
+//│    = 6
+
+// `withConstruct` will be inserted only when `with` is used.
```
chengluyu, line 59: These lines could explain what this PR does.

LPTK (approved):
I don't understand why you removed `intToString`. It breaks some useful tests. Put it back?
Besides that, LGTM.

chengluyu:
> I don't understand why you removed `intToString`. It breaks some useful tests. Put it back?
> 
> Besides that, LGTM.

That function is defined in the test file but not defined in `Typer` and only used in that test file. So if we put it back, it seems that we should define un-implemented functions in other files. How do you think?

LPTK:
Okay. Well, one way or the other we really do want to have _something_ that's implemented so we can make sure theses tests work! Apparently in JS one writes `n.toString()`. Can we add a `toString: anything -> string` prelude function to call `.toString` on its argument? And use it in that test.

LPTK:
Cool, you can go ahead and squash-merge whenever you're ready.

PR #59, opened 2022-01-05
Refactor method typing

fo5for:
The code for method typing is refactored to remove the use of mutable maps for storing method signatures and minimize messy substitutions of signatures. Below is an overview of the major changes:

**Refactor TypeDef and TypeRef**:
- `TypeDef` now stores the typed type of its body instead of the syntactic type.
- In order to type the body of a type definition before the `TypeDef` is added to the `Ctx`, `typeType` now takes an additional optional parameter `newDefsInfo` with the kind and number of type paramerters of the new definitions.
- The auxilliary `typeType2` also returns the list of `TypeVariables` used to instantiate implicit `TypeVar`s in the body, which are instantiated by substitution when a `TypeRef` is expanded.
- `TypeRef` now takes a `TypeName` as the base instead of a `TypeDef`, and no longer stores a `Ctx`, instead requiring a `Ctx` argument to retrieve the definition for expansion.
- `TypeRef.expand` no longer takes a `Raise` argument as it only performs substitution instead of typing the body every time. This eliminated much of the error swallowing.
- Caveat: `SimpleType.<:<` also needs to take a `Ctx` argument, which is problematic for normal forms as they implement `<:<` with `toType` and `SimpleType.<:<`, but `toType` would never produce a `TypeRef`. An "empty" `Ctx` is pass to `SimpleType.<:<` for now.

**Refactor method typing**:
- Removed mutable maps in context used for storing method signatures.
- Moved method bindings to a separate environment. Beside the actual binding, the inferred type is also stored with key delimited by `#`. This is only used for printing in `DiffTests`.
- Instead of looking up the signatures of inherited methods and substituting type variables for the type arguments, `typeMethod` now recursively traverse the type definition body, and collect inherited method signatures by typing them with the proper arguments applied.

New subclasses of `PolymorphicType`:
- `MethodType`: stores the parents of the method in addition. `&` intersects the bodies of two inherited methods and concatenate their parents; `+` overwrites a binding with the new signature with the same parent, or concatenate the parents if the call is ambiguous.
- `AbstractConstructor`: stores the set of abstract methods for printing when a constructor call of an abstract type is attempted. This is needed since it is no longer possible to look up the list of all declared and defined methods after the `TypeDef` is processed from the mutable maps.

**Miscellaneous improvements**:
- `checkRegular` no longer expands `TypeRef`s with parameter tags, which eliminates duplicate errors.
- A polymorphism issue for methods was accidentally identified and fixed by the refactoring in `MethodAndMatches.mls`

### DIFF
--- a/shared/src/main/scala/mlscript/NormalForms.scala
+++ b/shared/src/main/scala/mlscript/NormalForms.scala
@@ -73,6 +73,7 @@ class NormalForms extends TyperDatatypes { self: Typer =>
       case (_, LhsTop) => true
       case (LhsTop, _) => false
       case (LhsRefined(b1, ts1, rt1), LhsRefined(b2, ts2, rt2)) =>
+        implicit val ctx: Ctx = Ctx.empty
         b2.forall(b2 => b1.exists(_ <:< b2)) && ts2.forall(ts1) && rt1 <:< rt2
     }
   }
@@ -159,7 +160,7 @@ class NormalForms extends TyperDatatypes { self: Typer =>
         S(RhsBases(p, S(R(RhsField(n1, t1 | that._2)))))
       case _: RhsField | _: RhsBases => N
     }
-    def <:< (that: RhsNf): Bool = this.toType() <:< that.toType() // TODO less inefficient! (uncached calls to toType)
+    def <:< (that: RhsNf): Bool = this.toType().<:<(that.toType())(Ctx.empty) // TODO less inefficient! (uncached calls to toType)
   }
   case class RhsField(name: Var, ty: SimpleType) extends RhsNf
     { def name_ty: Var -> ST = name -> ty }
@@ -202,7 +203,7 @@ class NormalForms extends TyperDatatypes { self: Typer =>
       // }(r => s"!! $r")
     def & (that: Conjunct): Opt[Conjunct] =
       // trace(s"?? $this & $that ${lnf & that.lnf} ${rnf | that.rnf}") {
-      if (lnf.toType() <:< that.rnf.toType()) N // TODO support <:< on any Nf? // TODO less inefficient! (uncached calls to toType)
+      if (lnf.toType().<:<(that.rnf.toType())(Ctx.empty)) N // TODO support <:< on any Nf? // TODO less inefficient! (uncached calls to toType)
       else S(Conjunct.mk(lnf & that.lnf getOrElse (return N), vars | that.vars
         , rnf | that.rnf getOrElse (return N)
         , nvars | that.nvars))
@@ -317,7 +318,7 @@ class NormalForms extends TyperDatatypes { self: Typer =>
     def of(tvs: Set[TypeVariable]): DNF = DNF(Conjunct.of(tvs) :: Nil)
     def extr(pol: Bool): DNF = if (pol) of(LhsTop) else DNF(Nil)
     def merge(pol: Bool)(l: DNF, r: DNF): DNF = if (pol) l | r else l & r
-    def mk(ty: SimpleType, pol: Bool): DNF = (if (pol) ty.pushPosWithout(_ => ()) else ty) match {
+    def mk(ty: SimpleType, pol: Bool)(implicit ctx: Ctx): DNF = (if (pol) ty.pushPosWithout else ty) match {
       case bt: BaseType => of(bt)
       case bt: TraitTag => of(bt)
       case rt @ RecordType(fs) => of(rt)
@@ -326,7 +327,7 @@ class NormalForms extends TyperDatatypes { self: Typer =>
       case NegType(und) => DNF(CNF.mk(und, !pol).ds.map(_.neg))
       case tv: TypeVariable => of(Set.single(tv))
       case ProxyType(underlying) => mk(underlying, pol)
-      case tr @ TypeRef(defn, targs) => mk(tr.expand(_ => ()), pol) // TODO try to keep them?
+      case tr @ TypeRef(defn, targs) => mk(tr.expand, pol) // TODO try to keep them?
       case TypeBounds(lb, ub) => mk(if (pol) ub else lb, pol)
     }
   }
@@ -351,7 +352,7 @@ class NormalForms extends TyperDatatypes { self: Typer =>
       Disjunct(RhsField(f._1, f._2), Set.empty, LhsTop, Set.empty)).toList)
     def extr(pol: Bool): CNF = if (pol) CNF(Nil) else of(RhsBot)
     def merge(pol: Bool)(l: CNF, r: CNF): CNF = if (pol) l | r else l & r
-    def mk(ty: SimpleType, pol: Bool): CNF =
+    def mk(ty: SimpleType, pol: Bool)(implicit ctx: Ctx): CNF =
       // trace(s"?C $ty") {
       ty match {
         case bt: BaseType => of(bt)
@@ -362,7 +363,7 @@ class NormalForms extends TyperDatatypes { self: Typer =>
         case NegType(und) => CNF(DNF.mk(und, !pol).cs.map(_.neg))
         case tv: TypeVariable => of(Set.single(tv))
         case ProxyType(underlying) => mk(underlying, pol)
-        case tr @ TypeRef(defn, targs) => mk(tr.expand(_ => ()), pol) // TODO try to keep them?
+        case tr @ TypeRef(defn, targs) => mk(tr.expand, pol) // TODO try to keep them?
         case TypeBounds(lb, ub) => mk(if (pol) ub else lb, pol)
       }
       // }(r => s"!C $r")
--- a/shared/src/main/scala/mlscript/NormalForms.scala
+++ b/shared/src/main/scala/mlscript/NormalForms.scala
@@ -160,7 +160,7 @@ class NormalForms extends TyperDatatypes { self: Typer =>
         S(RhsBases(p, S(R(RhsField(n1, t1 | that._2)))))
       case _: RhsField | _: RhsBases => N
     }
-    def <:< (that: RhsNf): Bool = this.toType().<:<(that.toType())(Ctx.empty) // TODO less inefficient! (uncached calls to toType)
+    def <:< (that: RhsNf): Bool = (this.toType() <:< that.toType())(Ctx.empty) // TODO less inefficient! (uncached calls to toType)
   }
   case class RhsField(name: Var, ty: SimpleType) extends RhsNf
     { def name_ty: Var -> ST = name -> ty }
@@ -203,7 +203,7 @@ class NormalForms extends TyperDatatypes { self: Typer =>
       // }(r => s"!! $r")
     def & (that: Conjunct): Opt[Conjunct] =
       // trace(s"?? $this & $that ${lnf & that.lnf} ${rnf | that.rnf}") {
-      if (lnf.toType().<:<(that.rnf.toType())(Ctx.empty)) N // TODO support <:< on any Nf? // TODO less inefficient! (uncached calls to toType)
+      if ((lnf.toType() <:< that.rnf.toType())(Ctx.empty)) N // TODO support <:< on any Nf? // TODO less inefficient! (uncached calls to toType)
       else S(Conjunct.mk(lnf & that.lnf getOrElse (return N), vars | that.vars
         , rnf | that.rnf getOrElse (return N)
         , nvars | that.nvars))
--- a/shared/src/main/scala/mlscript/Typer.scala
+++ b/shared/src/main/scala/mlscript/Typer.scala
@@ -20,7 +20,7 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
   type Raise = Diagnostic => Unit
   type Binding = Str -> TypeScheme
   type Bindings = Map[Str, TypeScheme]
-
+  
   case class TypeDef(
     kind: TypeDefKind,
     nme: TypeName,
@@ -37,13 +37,14 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
         baseClasses.iterator.filterNot(traversed).flatMap(v =>
           ctx.tyDefs.get(v.name).fold(Set.empty[Var])(_.allBaseClasses(ctx)(traversed + v)))
     val (tparams: List[TypeName], targs: List[TypeVariable]) = tparamsargs.unzip
-    def thisTy(implicit prov: TypeProvenance): TypeRef = TypeRef(nme, targs)(prov)
-    def wrapMethod(pt: PolymorphicType)(implicit prov: TypeProvenance): MethodType =
+    def wrapMethod(pt: PolymorphicType)(implicit prov: TypeProvenance): MethodType = {
+      val thisTy = TypeRef(nme, targs)(prov)
       MethodType(pt.level, S(FunctionType(singleTup(thisTy), pt.body)(prov)), nme)
+    }
     def wrapMethod(mt: MethodType)(implicit prov: TypeProvenance): MethodType =
       if (mt.body.nonEmpty) wrapMethod(mt.toPT) else mt.copy(parents = nme :: Nil)
   }
-
+  
   private case class MethodDefs(
     tn: TypeName,
     parents: List[MethodDefs],
@@ -67,26 +68,18 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
             S(mn -> MethodType(defns.head.level, N, tn :: Nil))
         }
       )
-    private def ++(that: MethodDefs): MethodDefs = {
-      require(parents.isEmpty)
-      MethodDefs(
-        that.tn,
-        that.parents,
-        this.decls ++ that.decls,
-        this.defns ++ that.defns
-      )
-    }
     private def addParent(prt: TypeName): MethodDefs =
       copy(decls = decls.view.mapValues(mt => mt.copy(parents = prt :: mt.parents)).toMap)
     def propagate(top: Bool = true)(implicit raise: Raise): MethodDefs =
       parents.map(_.propagate(false)).reduceOption(_.&(_)(tn, defns.keySet)).map(_.addParent(tn))
-        .foldRight(if (top) copy(parents = Nil, decls = Map.empty, defns = Map.empty) else copy(parents = Nil))(_ ++ _)
+        .foldRight(if (top) MethodDefs(tn, Nil, Map.empty, Map.empty) else copy(parents = Nil))((mds1, mds2) =>
+          mds2.copy(decls = mds1.decls ++ mds2.decls, defns = mds1.defns ++ mds2.defns))
   }
   
   /** Keys of `mthEnv`:
    *  `L` represents the inferred types of method definitions. The first value is the parent name, and the second value is the method name.
    *  `R` represents the actual method types. The first optional value is the parent name, with `N` representing implicit calls,
-   *    and the second value is the method name.
+   *    and the second value is the method name. (See the case for `Sel` in `typeTerm` for documentation on explicit vs. implicit calls.)
    *  The public helper functions should be preferred for manipulating `mthEnv`
    */
   case class Ctx(parent: Opt[Ctx], env: MutMap[Str, TypeInfo], mthEnv: MutMap[(Str, Str) \/ (Opt[Str], Str), MethodType],
@@ -529,69 +522,93 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
         val targsMap = td.targs.lazyZip(rigidtargs).toMap[SimpleType, SimpleType]
         def ss(mt: MethodType, bmt: MethodType)(implicit prov: TypeProvenance) =
           constrain(subst(mt.toPT, targsMap).instantiate, subst(bmt.toPT, targsMap).rigidify)
-        def registerImplicit(nme: Str, mthTy: MethodType) = ctx.getMth(N, nme) match {
-          case S(MethodType(_, _, parents)) if parents.iterator.map(prt => Var(prt.name.decapitalize)).toSet
-              .subsetOf(ctx.allBaseClassesOf(tn.name)) =>
-          case S(MethodType(_, _, parents)) if parents.forall(prt =>
-              ctx.allBaseClassesOf(prt.name).contains(Var(tn.name.decapitalize))) =>
-            ctx.addMth(N, nme, mthTy)
-          case S(mt2) => ctx.addMth(N, nme, mt2 + mthTy)
-          case N => ctx.addMth(N, nme, mthTy)
+        def registerImplicit(mn: Str, mthTy: MethodType) = ctx.getMth(N, mn) match {
+          // If the currently registered method belongs to one of the base classes of this class,
+          // then we don't need to do anything.
+          // This is because implicit method calls always default to the parent methods.
+          case S(MethodType(_, _, parents)) if {
+            val bcs = ctx.allBaseClassesOf(tn.name)
+            parents.forall(prt => bcs(Var(prt.name.decapitalize)))
+          } =>
+          // If this class is one of the base classes of the parent(s) of the currently registered method,
+          // then we need to register the new method. Only happens when the class definitions are "out-of-order",
+          // and can disambiguate implicit calls previously marked as ambiguous.
+          // Example:
+            // class B: A
+            //   method F: 0
+            // class C: A
+            //   method F: 42
+            // class A
+            //   method F: int
+          case S(MethodType(_, _, parents)) if {
+            val v = Var(tn.name.decapitalize)
+            parents.forall(prt => ctx.allBaseClassesOf(prt.name).contains(v)) 
+          } => ctx.addMth(N, mn, mthTy)
+          // If this class is unrelated to the parent(s) of the currently registered method, then mark it as ambiguous
+          case S(mt2) => ctx.addMth(N, mn, mt2 + mthTy)
+          case N => ctx.addMth(N, mn, mthTy)
         }
-        def overrideError(nme: Str, mt: MethodType, mt2: MethodType) = {
+        def overrideError(mn: Str, mt: MethodType, mt2: MethodType) = {
           mt2.parents.foreach(parent => 
-            err(msg"Overriding method ${parent}.${nme} without explicit declaration is not allowed." -> mt.prov.loco ::
+            err(msg"Overriding method ${parent}.${mn} without explicit declaration is not allowed." -> mt.prov.loco ::
               msg"Note: method definition inherited from" -> mt2.prov.loco :: Nil)(raise, noProv))
-          println(s">> Checking subsumption (against inferred type) for inferred type of $nme : $mt")
+          println(s">> Checking subsumption (against inferred type) for inferred type of $mn : $mt")
         }
-        declsInherit.foreach { case nme -> mt =>
+        declsInherit.foreach { case mn -> mt =>
           implicit val prov: TypeProvenance = mt.prov
           val mthTy = td.wrapMethod(mt)
-          ctx.addMth(S(tn.name), nme, mthTy)
+          ctx.addMth(S(tn.name), mn, mthTy)
         }
-        defnsInherit.foreach { case nme -> mt => 
+        defnsInherit.foreach { case mn -> mt => 
           implicit val prov: TypeProvenance = mt.prov
-          println(s">> Checking subsumption for inferred type of $nme : $mt")
-          (if (decls.isDefinedAt(nme) && !defns.isDefinedAt(nme)) decls.get(nme) else N).orElse(declsInherit.get(nme))
+          println(s">> Checking subsumption for inferred type of $mn : $mt")
+          (if (decls.isDefinedAt(mn) && !defns.isDefinedAt(mn)) decls.get(mn) else N).orElse(declsInherit.get(mn))
             .foreach(ss(mt, _))
           val mthTy = td.wrapMethod(mt)
-          if (!declsInherit.get(nme).exists(decl => decl.single && decl.parents.last === mt.parents.head))
-            ctx.addMth(S(tn.name), nme, mthTy)
+          if (!declsInherit.get(mn).exists(decl => decl.single && decl.parents.last === mt.parents.last))
+            ctx.addMth(S(tn.name), mn, mthTy)
         }
-        decls.foreach { case nme -> mt =>
+        decls.foreach { case mn -> mt =>
           implicit val prov: TypeProvenance = mt.prov
-          println(s">> Checking subsumption for declared type of $nme : $mt")
-          ((declsInherit.get(nme), defnsInherit.get(nme)) match {
+          println(s">> Checking subsumption for declared type of $mn : $mt")
+          ((declsInherit.get(mn), defnsInherit.get(mn)) match {
             case (S(decl), S(defn)) =>
-              if (mt.body =/= decl.body && !defns.isDefinedAt(nme)) defn.parents.foreach(parent => 
-                err(msg"Overriding method ${parent}.${nme} without an overriding definition is not allowed." -> mt.prov.loco ::
+              if (mt.body =/= decl.body && !defns.isDefinedAt(mn)) defn.parents.foreach(parent => 
+                err(msg"Overriding method ${parent}.${mn} without an overriding definition is not allowed." -> mt.prov.loco ::
                   msg"Note: method definition inherited from" -> defn.prov.loco :: Nil)(raise, noProv))
               S(decl)
             case (S(decl), N) => S(decl)
             case (N, S(defn)) =>
-              overrideError(nme, mt, defn)
+              overrideError(mn, mt, defn)
               S(defn)
             case (N, N) => N
           }).foreach(ss(mt, _))
           val mthTy = td.wrapMethod(mt)
-          ctx.addMth(S(tn.name), nme, mthTy)
-          registerImplicit(nme, mthTy)
+          ctx.addMth(S(tn.name), mn, mthTy)
+          registerImplicit(mn, mthTy)
         }
-        defns.foreach { case nme -> mt => 
+        defns.foreach { case mn -> mt => 
           implicit val prov: TypeProvenance = mt.prov
-          println(s">> Checking subsumption for inferred type of $nme : $mt")
-          decls.get(nme).orElse((declsInherit.get(nme), defnsInherit.get(nme)) match {
+          println(s">> Checking subsumption for inferred type of $mn : $mt")
+          decls.get(mn).orElse((declsInherit.get(mn), defnsInherit.get(mn)) match {
             case (S(decl), S(defn)) if defn.parents.toSet.subsetOf(decl.parents.toSet) => S(decl)
             case (_, S(defn)) =>
-              overrideError(nme, mt, defn)
+              overrideError(mn, mt, defn)
               S(defn)
             case (S(decl), N) => S(decl)
             case (N, N) => N
           }).foreach(ss(mt, _))
           val mthTy = td.wrapMethod(mt)
-          if (!ctx.containsMth(S(tn.name), nme) || !decls.isDefinedAt(nme)) ctx.addMth(S(tn.name), nme, mthTy)
-          ctx.addMthDefn(tn.name, nme, mthTy)
-          if (!decls.isDefinedAt(nme) && !declsInherit.isDefinedAt(nme)) registerImplicit(nme, mthTy)
+          if (!decls.isDefinedAt(mn)) {
+            // If the class declares that method explicitly,
+            //   the declared signature is used so we don't have to do anything
+            // If the class does not declare that method explicitly (it could be inherited),
+            //   we still want to make the method available using its inferred signature
+            //   both for implicit method calls and for explicit ones
+            ctx.addMth(S(tn.name), mn, mthTy)
+            registerImplicit(mn, mthTy)
+          }
+          ctx.addMthDefn(tn.name, mn, mthTy)
         }
       }
       newDefs.foreach(td => if (ctx.tyDefs.isDefinedAt(td.nme.name)) typeMethodsSingle(td))
@@ -901,6 +918,12 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
         val raw_fun_ty = fun_ty.unwrapProxies
         resTy
       case Sel(obj, fieldName) =>
+        // Explicit method calls have the form `x.(Class.Method)`
+        // Implicit method calls have the form `x.Method`
+        //   If two unrelated classes define methods of the same name,
+        //   implicit calls to this method are marked as ambiguous and are forbidden
+        // Explicit method retrievals have the form `Class.Method`
+        //   Returns a function expecting an additional argument of type `Class` before the method arguments
         def rcdSel(obj: Term, fieldName: Var) = {
           val o_ty = typeTerm(obj)
           val res = freshVar(prov)
@@ -909,15 +932,17 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
         }
         def mthCall(obj: Term, fieldName: Var) = 
           (fieldName.name match {
-            case s"$parent.$nme" => ctx.getMth(S(parent), nme)
-            case nme => ctx.getMth(N, nme)
+            case s"$parent.$nme" => ctx.getMth(S(parent), nme) // explicit calls
+            case nme => ctx.getMth(N, nme) // implicit calls
           }) match {
             case S(mth_ty) =>
-              if (mth_ty.parents.size > 1)
+              if (mth_ty.body.isEmpty)
                 err(msg"Implicit call to method ${fieldName.name} is forbidden because it is ambiguous." -> term.toLoc
                   :: msg"Unrelated methods named ${fieldName.name} are defined by:" -> N
-                  :: mth_ty.parents.iterator.map(prt => ctx.tyDefs(prt.name))
-                    .map(td => msg"• ${td.kind.str} ${td.nme}" -> td.nme.toLoc).toList)
+                  :: mth_ty.parents.map { prt =>
+                    val td = ctx.tyDefs(prt.name)
+                    msg"• ${td.kind.str} ${td.nme}" -> td.nme.toLoc
+                  })
               val o_ty = typeTerm(obj)
               val res = freshVar(prov)
               con(mth_ty.instantiate, FunctionType(singleTup(o_ty), res)(prov), res)
@@ -926,7 +951,7 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
               else rcdSel(obj, fieldName) // TODO: no else?
           }
         obj match {
-          case Var(name) if name.isCapitalized && ctx.tyDefs.isDefinedAt(name) =>
+          case Var(name) if name.isCapitalized && ctx.tyDefs.isDefinedAt(name) => // explicit retrieval
             ctx.getMth(S(name), fieldName.name) match {
               case S(mth_ty) => mth_ty.instantiate
               case N =>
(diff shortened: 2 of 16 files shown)


LPTK:
Awesome improvements! Can you please rebase your branch on top of the current `mlscript` main branch?
We can keep your commits as they seem fairly atomic. Each of them passes all the tests, right?

shared/src/main/scala/mlscript/NormalForms.scala
```diff
@@ -159,7 +160,7 @@ class NormalForms extends TyperDatatypes { self: Typer =>
         S(RhsBases(p, S(R(RhsField(n1, t1 | that._2)))))
       case _: RhsField | _: RhsBases => N
     }
-    def <:< (that: RhsNf): Bool = this.toType() <:< that.toType() // TODO less inefficient! (uncached calls to toType)
+    def <:< (that: RhsNf): Bool = this.toType().<:<(that.toType())(Ctx.empty) // TODO less inefficient! (uncached calls to toType)
```
LPTK:
```suggestion
    def <:< (that: RhsNf): Bool = (this.toType() <:< that.toType())(Ctx.empty) // TODO less inefficient! (uncached calls to toType)
```

shared/src/main/scala/mlscript/NormalForms.scala
```diff
@@ -202,7 +203,7 @@ class NormalForms extends TyperDatatypes { self: Typer =>
       // }(r => s"!! $r")
     def & (that: Conjunct): Opt[Conjunct] =
       // trace(s"?? $this & $that ${lnf & that.lnf} ${rnf | that.rnf}") {
-      if (lnf.toType() <:< that.rnf.toType()) N // TODO support <:< on any Nf? // TODO less inefficient! (uncached calls to toType)
+      if (lnf.toType().<:<(that.rnf.toType())(Ctx.empty)) N // TODO support <:< on any Nf? // TODO less inefficient! (uncached calls to toType)
```
LPTK:
```suggestion
      if (lnf.toType() <:< that.rnf.toType())(Ctx.empty)) N // TODO support <:< on any Nf? // TODO less inefficient! (uncached calls to toType)
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -20,60 +20,94 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
   type Raise = Diagnostic => Unit
   type Binding = Str -> TypeScheme
   type Bindings = Map[Str, TypeScheme]
-  
+
```
```suggestion
  
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -20,60 +20,94 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
   case class TypeDef(
     kind: TypeDefKind,
     nme: TypeName,
-    tparams: List[TypeName],
-    targs: List[TypeVariable],
-    body: Type,
+    tparamsargs: List[(TypeName, TypeVariable)],
+    tvars: Iterable[TypeVariable], // "implicit" type variables. instantiate every time a `TypeRef` is expanded
```
LPTK, line 28: Why `Iterable`? I try to avoid using such abstract collections as they can obscure the underlying data structure shape and complexity of operations we do on them.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -20,60 +20,94 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
     toLoc: Opt[Loc],
   ) {
     def allBaseClasses(ctx: Ctx)(implicit traversed: Set[Var]): Set[Var] =
-      baseClasses.map(v => Var(v.name.toList.mapHead(_.toLower).mkString)) ++
+      baseClasses.map(v => Var(v.name.decapitalize)) ++
         baseClasses.iterator.filterNot(traversed).flatMap(v =>
           ctx.tyDefs.get(v.name).fold(Set.empty[Var])(_.allBaseClasses(ctx)(traversed + v)))
-    val targsMap: Map[Str, TypeVariable] = tparams.map(_.name).zip(targs).toMap
+    val tparams: List[TypeName] = tparamsargs.map(_._1)
+    val targs: List[TypeVariable] = tparamsargs.map(_._2)
```
```suggestion
    val (tparams: List[TypeName], targs: List[TypeVariable]) = tparamsargs.unzip
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -20,60 +20,94 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+    def thisTy(implicit prov: TypeProvenance): TypeRef = TypeRef(nme, targs)(prov)
+    def wrapMethod(bodyTy: PolymorphicType)(implicit prov: TypeProvenance): MethodType =
+      MethodType(bodyTy.level, FunctionType(singleTup(thisTy), bodyTy.body)(prov), nme)
+  }
+
```
```suggestion
  
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -20,60 +20,94 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
   ) {
     def allBaseClasses(ctx: Ctx)(implicit traversed: Set[Var]): Set[Var] =
-      baseClasses.map(v => Var(v.name.toList.mapHead(_.toLower).mkString)) ++
+      baseClasses.map(v => Var(v.name.decapitalize)) ++
         baseClasses.iterator.filterNot(traversed).flatMap(v =>
           ctx.tyDefs.get(v.name).fold(Set.empty[Var])(_.allBaseClasses(ctx)(traversed + v)))
-    val targsMap: Map[Str, TypeVariable] = tparams.map(_.name).zip(targs).toMap
+    val tparams: List[TypeName] = tparamsargs.map(_._1)
+    val targs: List[TypeVariable] = tparamsargs.map(_._2)
+    def thisTy(implicit prov: TypeProvenance): TypeRef = TypeRef(nme, targs)(prov)
```
LPTK: Seems to be only used in `wrapMethod`, so would probably be better as a local `val`.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -20,60 +20,94 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+    private def ++(that: MethodDefs): MethodDefs = {
+      require(prts.isEmpty)
+      MethodDefs(
+        that.tn,
+        that.prts,
+        this.decls ++ that.decls,
+        this.defns ++ that.defns
+      )
+    }
+    private def addprt(prt: TypeName): MethodDefs =
```
LPTK: Method naming consistency:
```suggestion
    private def addParent(prt: TypeName): MethodDefs =
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -20,60 +20,94 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+    val tparams: List[TypeName] = tparamsargs.map(_._1)
+    val targs: List[TypeVariable] = tparamsargs.map(_._2)
+    def thisTy(implicit prov: TypeProvenance): TypeRef = TypeRef(nme, targs)(prov)
+    def wrapMethod(bodyTy: PolymorphicType)(implicit prov: TypeProvenance): MethodType =
+      MethodType(bodyTy.level, FunctionType(singleTup(thisTy), bodyTy.body)(prov), nme)
+  }
+
+  private case class MethodDefs(
+    tn: TypeName,
+    prts: List[MethodDefs],
```
LPTK: I'd rather call this `parents`. Took me a while to get what `prts` meant 😄

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -20,60 +20,94 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+        this.prts ::: that.prts,
+        mergeMap(this.decls, that.decls)(_ & _),
+        (this.defns.iterator ++ that.defns.iterator).toSeq.groupMap(_._1)(_._2).flatMap {
+          case _ -> Nil => die
+          case mn -> (defn :: Nil) => S(mn -> defn)
+          case mn -> defns if newdefns(mn) => N
+          case mn -> defns =>
+            implicit val prov: TypeProvenance = TypeProvenance(tn.toLoc, "inherited method declaration")
+            S(mn -> MethodType(defns.head.level,
+              err(msg"A method definition must be given when inheriting multiple method definitions" -> tn.toLoc
```
```suggestion
              err(msg"An overriding method definition must be given when inheriting from multiple method definitions" -> tn.toLoc
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -186,14 +220,17 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
   
-  def clsNameToNomTag(td: TypeDef)(prov: TypeProvenance, ctx: Ctx): SimpleType = {
-    require(td.kind is Cls)
-    ClassTag(Var(td.nme.name.head.toLower.toString + td.nme.name.tail),
-      td.allBaseClasses(ctx)(Set.single(Var(td.nme.name))))(prov)
+  def clsNameToNomTag(tn: Str)(prov: TypeProvenance, ctx: Ctx): SimpleType = ctx.tyDefs.get(tn) match {
+    case S(td) =>
+      require(td.kind is Cls)
+      ClassTag(Var(td.nme.name.decapitalize), ctx.allBaseClassesOf(td.nme.name))(prov)
+    case N => errType(prov)
```
LPTK: This change seems unnecessary and may silence genuine bugs.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -250,6 +283,8 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
   def processTypeDefs(newDefs0: List[mlscript.TypeDef])(implicit ctx: Ctx, raise: Raise): Ctx = {
     var allDefs = ctx.tyDefs
     var allEnv = ctx.env.clone
+    var allMthEnv = ctx.mthenv.clone
```
```suggestion
    val allEnv = ctx.env.clone
    val allMthEnv = ctx.mthenv.clone
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -20,60 +20,94 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
   type Raise = Diagnostic => Unit
   type Binding = Str -> TypeScheme
   type Bindings = Map[Str, TypeScheme]
-  
+
   case class TypeDef(
     kind: TypeDefKind,
     nme: TypeName,
```
LPTK, line 26: This should probably be a `Var` instead.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -365,201 +393,232 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
-                ss(mt, bmt)
-              }
-            })
-          }
+        decls.foreach { case nme -> mt =>
+          implicit val prov: TypeProvenance = mt.prov
+          val fullName = s"${tn.name}.${nme}"
+          println(s">> Checking subsumption for declared type of $nme : $mt")
+          declsInherit.get(nme).orElse(defnsInherit.get(nme).tap(_.foreach { mt2 =>
+            err(msg"Overriding method ${mt2.prts.headOption.fold("")(_.name)}.${nme} without explicit declaration is not allowed."
```
LPTK: As discussed in the meeting, this `mt2.prts.headOption.fold("")(_.name)` is a bit fishy. Maybe better done as `mt2.prts.headOption.foreach { prt => ... }`.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -365,201 +393,232 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+          registerImplicit(nme, mthTy)
+        }
+        defns.foreach { case nme -> mt => 
+          implicit val prov: TypeProvenance = mt.prov
+          val fullName = s"${tn.name}.${nme}"
+          println(s">> Checking subsumption for inferred type of $nme : $mt")
+          decls.get(nme).orElse((declsInherit.get(nme), defnsInherit.get(nme)) match {
+            case (S(decl), S(defn)) if defn.prts.toSet.subsetOf(decl.prts.toSet) => S(decl)
+            case (_, S(defn)) =>
+              err(msg"Overriding method ${defn.prts.headOption.fold("")(_.name)}.${nme} without explicit declaration is not allowed."
```
LPTK: As discussed in the meeting, we could do some deduplication with the above here, at least for this specific error-reporting part.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -365,201 +393,232 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+                msg"Note: method definition inherited from" -> defn.prov.loco :: Nil)(raise, noProv)
+              println(s">> Checking subsumption (against inferred type) for inferred type of $nme : $mt")
+              S(defn)
+            case (S(decl), N) => S(decl)
+            case (N, N) => N
+          }).foreach(ss(mt, _))
+          val mthTy = td.wrapMethod(mt)
+          if (!ctx.containsmth(fullName) || !decls.isDefinedAt(nme)) ctx.mthenv += fullName -> mthTy
+          ctx.mthenv += s"${tn.name}#${nme}" -> mthTy
+          if (!decls.isDefinedAt(nme) && !declsInherit.isDefinedAt(nme)) registerImplicit(nme, mthTy)
```
LPTK: It would be good to add some (very brief) documentation for this logic and in the two lines above. When I read it first I had trouble understanding it.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -365,201 +393,232 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+              msg"Note: method definition inherited from" -> mt2.prov.loco :: Nil)(raise, noProv)
+            println(s">> Checking subsumption (against inferred type) for inferred type of $nme : $mt")
+          })).foreach(ss(mt, _))
+          val mthTy = td.wrapMethod(mt)
+          ctx.mthenv += fullName -> mthTy
+          registerImplicit(nme, mthTy)
+        }
+        defns.foreach { case nme -> mt => 
+          implicit val prov: TypeProvenance = mt.prov
+          val fullName = s"${tn.name}.${nme}"
```
LPTK: Let's use proper structured keys instead of raw strings here, as discussed in the meeting. With a doc comment on the meaning of the shapes of the keys.

shared/src/main/scala/mlscript/TyperDatatypes.scala
```diff
@@ -31,6 +31,36 @@ abstract class TyperDatatypes extends TyperHelpers { self: Typer =>
...
+  }
+  object MethodType {
+    def apply(level: Int, body: SimpleType, prt: TypeName): MethodType = new MethodType(level, body, prt :: Nil, true)
+    def apply(level: Int, body: SimpleType, prts: List[TypeName]): MethodType = new MethodType(level, body, prts, true)
+    private def apply(level: Int, body: SimpleType, prts: List[TypeName], single: Bool): MethodType =
+      new MethodType(level, body, prts, single)
+    def unapply(mt: MethodType): S[(Int, SimpleType, List[TypeName])] = S((mt.level, mt.body, mt.prts))
+  }
+
+  class AbstractConstructor(val absMths: Set[TypeName])(body: SimpleType) extends PolymorphicType(0, body) {
```
LPTK: Not a big fan of this subtype. Its semantics relies on downstream code to _remember_ to check for it, and if they forget, it will silently do the wrong thing. It's unclear why it needs to be a subtype of `PolymorphicType` at all. For instance, does the `TyperHelpers` method below make sense for it?

```scala
  def subst(ts: PolymorphicType, map: Map[SimpleType, SimpleType]): PolymorphicType = 
    PolymorphicType(ts.level, subst(ts.body, map))
```

Probably not.

Better make this and `TypeScheme` both subclasses of a new `TypeInfo` base class, and make typing environments map variable names to `TypeInfo`s, to be checked in the `Var` typing case later on.

shared/src/main/scala/mlscript/TyperDatatypes.scala
```diff
@@ -31,6 +31,36 @@ abstract class TyperDatatypes extends TyperHelpers { self: Typer =>
     def rigidify(implicit lvl: Int): SimpleType = freshenAbove(level, body, rigidify = true)
   }
   
+  class MethodType(level: Int, body: SimpleType, val prts: List[TypeName], val single: Bool)
```
LPTK: I'm also uneasy about this being a subclass of `PolymorphicType`. Typing environments never map to this, right? Maybe a different base trait/class can be used for the purposes of method type handling.

LPTK (changes requested):
It looks pretty good overall, except for the few comments I added.

Another thing: I think the following should be disallowed:
```scala
class C
    method F: int
    method F = 1
//│ Defined class C
//│ Declared C.F: C -> int
//│ Defined C.F: C -> 1

class D: C
    method F: 1
//│ Defined class D
//│ Declared D.F: D -> 1
```
Indeed, the whole point of requiring a signature for overridden definitions is that the original base class may later change the method _definition_ (leaving the declaration unchanged) without breaking downstream code. But in this example, if class `C` changes `F` to return `2`, it will actually break downstream code!

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -36,62 +36,72 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+   *    and the second value is the method name.
+   *  The public helper functions should be preferred for manipulating `mthEnv`
+   */
+  case class Ctx(parent: Opt[Ctx], env: MutMap[Str, TypeInfo], mthEnv: MutMap[(Str, Str) \/ (Opt[Str], Str), MethodType],
       lvl: Int, inPattern: Bool, tyDefs: Map[Str, TypeDef]) {
-    def +=(b: Binding): Unit = env += b
-    def ++=(bs: IterableOnce[Binding]): Unit = bs.iterator.foreach(+=)
-    def get(name: Str): Opt[TypeScheme] = env.get(name) orElse parent.dlof(_.get(name))(N)
-    def getmth(name: Str): Opt[MethodType] = mthenv.get(name) orElse parent.dlof(_.getmth(name))(N)
+    def +=(b: Str -> TypeInfo): Unit = env += b
```
LPTK, line 88: Why expand the `Binding` alias everywhere?

shared/src/main/scala/mlscript/Typer.scala
LPTK: Actually `TypeName` is fine. This name serves both as a type name and a term name (for the constructor), but it's primarily a type one.

shared/src/main/scala/mlscript/TyperDatatypes.scala
```diff
@@ -31,34 +35,30 @@ abstract class TyperDatatypes extends TyperHelpers { self: Typer =>
     def rigidify(implicit lvl: Int): SimpleType = freshenAbove(level, body, rigidify = true)
   }
   
-  class MethodType(level: Int, body: SimpleType, val prts: List[TypeName], val single: Bool)
-      extends PolymorphicType(level, body) {
+  class MethodType(val level: Int, val body: Opt[SimpleType], val parents: List[TypeName], val single: Bool)
```
LPTK, line 39: You could make this `case class MethodType(level: Int, body: Opt[SimpleType], parents: List[TypeName], single: Bool)` and avoid having to implement `copy`, `apply`, and `unapply`. Also, could you document what `single` is for?

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -20,60 +20,104 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+        that.parents,
+        this.decls ++ that.decls,
+        this.defns ++ that.defns
+      )
+    }
+    private def addParent(prt: TypeName): MethodDefs =
+      copy(decls = decls.view.mapValues(mt => mt.copy(parents = prt :: mt.parents)).toMap)
+    def propagate(top: Bool = true)(implicit raise: Raise): MethodDefs =
+      parents.map(_.propagate(false)).reduceOption(_.&(_)(tn, defns.keySet)).map(_.addParent(tn))
+        .foldRight(if (top) copy(parents = Nil, decls = Map.empty, defns = Map.empty) else copy(parents = Nil))(_ ++ _)
```
LPTK: This `copy` call may be cleaner as:
```suggestion
        .foldRight(if (top) MethodDefs(tn, Nil, Map.empty, Map.empty) else copy(parents = Nil))(_ ++ _)
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -20,60 +20,104 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+          case mn -> defns if newdefns(mn) => N
+          case mn -> defns =>
+            implicit val prov: TypeProvenance = TypeProvenance(tn.toLoc, "inherited method declaration")
+            err(msg"An overriding method definition must be given when inheriting from multiple method definitions" -> tn.toLoc
+              :: msg"Definitions of method $mn inherited from:" -> N
+              :: defns.iterator.map(mt => msg"• ${mt.parents.head}" -> mt.prov.loco).toList)
+            S(mn -> MethodType(defns.head.level, N, tn :: Nil))
+        }
+      )
+    private def ++(that: MethodDefs): MethodDefs = {
```
LPTK: Not sure why this `private def` is needed. It's only used once in `propagate` and does not seem to make a lot of sense on its own.

LPTK:
(Sorry I keep on piling comments on this PR while you're trying to merge! 😛)

shared/src/main/scala/mlscript/NormalForms.scala
fo5for: Addressed in f48c96fcda9dfbf2f7633ccfec76d4387d9aa354

shared/src/main/scala/mlscript/NormalForms.scala
fo5for: Addressed in f48c96fcda9dfbf2f7633ccfec76d4387d9aa354

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in f48c96fcda9dfbf2f7633ccfec76d4387d9aa354

shared/src/main/scala/mlscript/Typer.scala
fo5for: The `TypeRef` carries the same prov as the `MethodType`, which is different for every method. Making this a `val` would require changing it to a `noProv`, which loses some information for error reporting.

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in f48c96fcda9dfbf2f7633ccfec76d4387d9aa354

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in f48c96fcda9dfbf2f7633ccfec76d4387d9aa354

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in f48c96fcda9dfbf2f7633ccfec76d4387d9aa354

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in f48c96fcda9dfbf2f7633ccfec76d4387d9aa354

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in f48c96fcda9dfbf2f7633ccfec76d4387d9aa354

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in f48c96fcda9dfbf2f7633ccfec76d4387d9aa354

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in f48c96fcda9dfbf2f7633ccfec76d4387d9aa354

shared/src/main/scala/mlscript/Typer.scala
fo5for: You probably meant the one in `MethodDef`, which is addressed in 0fcee02d805237ba9580d4eb0154da5b1bb0c69a

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in 78d1a5ef3b58faaaa85466519f26dad19f5524c0

shared/src/main/scala/mlscript/TyperDatatypes.scala
fo5for: Addressed in 822bcd4ac635eb10660981cdb4d39cb501d3d8cb

shared/src/main/scala/mlscript/TyperDatatypes.scala
fo5for: Addressed in d7e4fecb249181bed9b09864e3ca35b645bbd7d8

shared/src/main/scala/mlscript/Typer.scala
fo5for: This was needed when the `TypeInfo` base class was introduced, and `env` was changed to map to `TypeInfo`s. We could either expand it here, or change the definition of the `Binding` alias and expand the reference in the return type of `typeStatement`, which only returns and is required to return a binding to a `TypeScheme`. I chose the former based on my interpretation of the word "binding", as a `TypeInfo` is not strictly a "binding". I can change it to the other way if you think that's more appropriate.

shared/src/main/scala/mlscript/TyperDatatypes.scala
fo5for: I wrote it this way to avoid explicitly dealing with `single` outside the class, which can only be turned into `false` by `&`. The documentation is added in dbf9e69d1f0b3293e36d97c6cf378bdfdb422fa8

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in dbf9e69d1f0b3293e36d97c6cf378bdfdb422fa8. (I wrote it this way to make the two branches more symmetric.)

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in dbf9e69d1f0b3293e36d97c6cf378bdfdb422fa8

shared/src/main/scala/mlscript/Typer.scala
fo5for: You were right to have trouble! Some of the conditions are redundant, which have been fixed in dbf9e69d1f0b3293e36d97c6cf378bdfdb422fa8. The `#` line should now be more readable after structured keys are used for `mthEnv`.

shared/src/main/scala/mlscript/Typer.scala
fo5for: I used the return type of `Map.values` (in `typeType2`) to avoid creating unnecessary collections. This would be removed when we fix the soundness issue with type variables.

shared/src/main/scala/mlscript/Typer.scala
LPTK: I'm not suggesting to make it a `val` _field_. I meant a _local_ `val` in `wrapMethod`.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -365,201 +376,239 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
-          case st: SimpleType => constrain(subst(mt, rigidSubsMap).instantiate,
-            subst(bno.fold(st)(bn => subst(st, subsMap(bn))), rigidSubsMap))(raise, prov, ctx)
+        val MethodDefs(_, _, decls, defns) = mds
+        val MethodDefs(_, _, declsInherit, defnsInherit) = mds.propagate()
+        val rigidtargs = td.targs.map(freshenAbove(ctx.lvl, _, true))
+        val targsMap = td.targs.lazyZip(rigidtargs).toMap[SimpleType, SimpleType]
+        def ss(mt: MethodType, bmt: MethodType)(implicit prov: TypeProvenance) =
+          constrain(subst(mt.toPT, targsMap).instantiate, subst(bmt.toPT, targsMap).rigidify)
+        def registerImplicit(nme: Str, mthTy: MethodType) = ctx.getMth(N, nme) match {
+          case S(MethodType(_, _, parents)) if parents.iterator.map(prt => Var(prt.name.decapitalize)).toSet
```
LPTK: This could be written much as the more efficient `if { val bcs = ctx.allBaseClassesOf(tn.name); parents.forall(...) }`.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -365,201 +376,239 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+              .subsetOf(ctx.allBaseClassesOf(tn.name)) =>
```
LPTK: This non-trivial condition really needs some doc. Be empathetic to the people who will be trying to understand your intention later on!! 😉
Despite the fact I _know_ the semantics of methods very well, I have no idea what this is checking...
```suggestion
              .subsetOf(ctx.allBaseClassesOf(tn.name)) =>
                // If [...] then we don't need to do anything
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -365,201 +376,239 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
-            subst(bno.fold(pt)(bn => subst(pt, subsMap(bn))), rigidSubsMap).rigidify)(raise, prov, ctx)
-          case st: SimpleType => constrain(subst(mt, rigidSubsMap).instantiate,
-            subst(bno.fold(st)(bn => subst(st, subsMap(bn))), rigidSubsMap))(raise, prov, ctx)
+        val MethodDefs(_, _, decls, defns) = mds
+        val MethodDefs(_, _, declsInherit, defnsInherit) = mds.propagate()
+        val rigidtargs = td.targs.map(freshenAbove(ctx.lvl, _, true))
+        val targsMap = td.targs.lazyZip(rigidtargs).toMap[SimpleType, SimpleType]
+        def ss(mt: MethodType, bmt: MethodType)(implicit prov: TypeProvenance) =
+          constrain(subst(mt.toPT, targsMap).instantiate, subst(bmt.toPT, targsMap).rigidify)
+        def registerImplicit(nme: Str, mthTy: MethodType) = ctx.getMth(N, nme) match {
```
LPTK: `nme` is the name of what?

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -365,201 +376,239 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+          case S(MethodType(_, _, parents)) if parents.iterator.map(prt => Var(prt.name.decapitalize)).toSet
+              .subsetOf(ctx.allBaseClassesOf(tn.name)) =>
+          case S(MethodType(_, _, parents)) if parents.forall(prt =>
+              ctx.allBaseClassesOf(prt.name).contains(Var(tn.name.decapitalize))) =>
+            ctx.addMth(N, nme, mthTy)
```
```suggestion
            // If [...], then we need to register the new method
            ctx.addMth(N, nme, mthTy)
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -365,201 +376,239 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+          decls.get(nme).orElse((declsInherit.get(nme), defnsInherit.get(nme)) match {
+            case (S(decl), S(defn)) if defn.parents.toSet.subsetOf(decl.parents.toSet) => S(decl)
+            case (_, S(defn)) =>
+              overrideError(nme, mt, defn)
+              S(defn)
+            case (S(decl), N) => S(decl)
+            case (N, N) => N
+          }).foreach(ss(mt, _))
+          val mthTy = td.wrapMethod(mt)
+          if (!decls.isDefinedAt(nme)) {
```
```suggestion
          if (!decls.isDefinedAt(nme)) {
            // If the class does not declare that method explicitly (it is inherited),
            // we still want to make the method available
            // both for implicit method calls (of the form `x.Foo`)
            // and for explicit ones (`x.(TheClass.Foo)`)
```

shared/src/main/scala/mlscript/Typer.scala
LPTK: Cool, but I'd still like some doc!! The semantics of methods may be crystal clear to you, but imagine how much trouble someone who is not sure will have, reading this code.

(I'm not saying the code _I_ write isn't similarly hard to understand BTW – it also needs to be better documented.)

shared/src/main/scala/mlscript/TyperDatatypes.scala
LPTK: Ok sure. Not sure it's worth it, but it's a detail.

shared/src/main/scala/mlscript/Typer.scala
fo5for: Oh... 😅 Addressed in c5f6b0817ec9905325a697a9281652782b2d2e30

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in c5f6b0817ec9905325a697a9281652782b2d2e30

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in c5f6b0817ec9905325a697a9281652782b2d2e30

shared/src/main/scala/mlscript/Typer.scala
fo5for: I have changed all uses of `nme` in `checkSubsume` to `mn` to clarify that it's the method name in c5f6b0817ec9905325a697a9281652782b2d2e30

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in c5f6b0817ec9905325a697a9281652782b2d2e30

shared/src/main/scala/mlscript/Typer.scala
fo5for: Addressed in c5f6b0817ec9905325a697a9281652782b2d2e30. The examples for explicit vs. implicit calls are removed as they were documented elsewhere.

shared/src/main/scala/mlscript/Typer.scala
fo5for: Actually, it was not crystal clear to me either. I realized that at some point during this refactoring, and I drew a graph for what to do for the 16 cases where a subset of inherited declaration, inherited definition, new declaration and new definition are present. 😅 Documenting them in text would be a huge mess but I'm not sure if including graphics in the source for documentation is a good idea 😅

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -365,201 +377,260 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
-            subst(bno.fold(st)(bn => subst(st, subsMap(bn))), rigidSubsMap))(raise, prov, ctx)
+        val MethodDefs(_, _, decls, defns) = mds
+        val MethodDefs(_, _, declsInherit, defnsInherit) = mds.propagate()
+        val rigidtargs = td.targs.map(freshenAbove(ctx.lvl, _, true))
+        val targsMap = td.targs.lazyZip(rigidtargs).toMap[SimpleType, SimpleType]
+        def ss(mt: MethodType, bmt: MethodType)(implicit prov: TypeProvenance) =
+          constrain(subst(mt.toPT, targsMap).instantiate, subst(bmt.toPT, targsMap).rigidify)
+        def registerImplicit(mn: Str, mthTy: MethodType) = ctx.getMth(N, mn) match {
+          // If the currently registered method belongs to one of the base classes of this class,
+          // then we don't need to do anything
```
```suggestion
          // then we don't need to do anything.
          // This is because implicit method calls always default to the parent methods.
```

shared/src/main/scala/mlscript/Typer.scala
LPTK: Great, the comments make things infinitely clearer now, especially the example.

shared/src/main/scala/mlscript/Typer.scala
LPTK: Okay, I find it much better now 👍

shared/src/main/scala/mlscript/Typer.scala
LPTK: Hmm okay, I guess that's fine.

LPTK (approved):
Okay, great work! I think it's almost ready to merge now. I think it's best to rebase-merge it. But be sure to squash the last two commits ("Cleanup and documentation" and "More cleanup and documentation") into a single commit first.

fo5for:
I noticed that `codegen/Classes.mls` is failing after rebasing, with the following error:
```scala
class MyBox: Box[int] & { info: string }
  method Map f = Box (f this.inner) with { info = this.info }
  method Inc = MyBox { inner = this.inner + 1; info = this.info }
def MyBox inner info = MyBox { inner; info }
//│ ╔══[ERROR] Type mismatch in inherited method definition:
//│ ║  l.10: 	  method Map f = Box { inner = f this.inner }
//│ ║        	         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//│ ╟── expression of type `{inner: ?a}` does not match type `'a`
//│ ║  l.10: 	  method Map f = Box { inner = f this.inner }
//│ ║        	                     ^^^^^^^^^^^^^^^^^^^^^^^^
//│ ╟── but it flows into application with expected type `Box['a]`
//│ ║  l.10: 	  method Map f = Box { inner = f this.inner }
//│ ║        	                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//│ ╟── Note: constraint arises from type variable:
//│ ║  l.9: 	  method Map: (T -> 'a) -> Box['a]
//│ ║       	                               ^^
//│ ╟── from record type:
//│ ║  l.8: 	class Box[T]: { inner: T }
//│ ╙──     	              ^^^^^^^^^^^^
//| ...
```
It took me a while to understand what was going on. In the old version, methods are typed once, then substituted when derived classes inherit them. So after the shorthand `Box` constructor is defined, the body of `Box.Map` still refers to the original constructor. But now that methods are typed every time a derived class inherits them, the body of `Box.Map` refers to the shorthand constructor instead when the inherited methods of `MyBox` are typed...

Should I push it anyways so you can have a look at it?

LPTK:
Oh! Method _bodies_ should definitely not be retyped every time.

Now, I understand that this probably means it'd be simpler to just go back to the previous approach where things were typed once and then we used type substitution. But I wouldn't want you to spend much more time on this, so I think it's fine if we capture and store the `ctx` in each type definition, and retype method bodies based on that context instead. Does that seem feasible to you?

fo5for:
Yeah I was thinking of the same thing. Storing a bunch of `MutMap`s with mostly the same content seems like a waste of space, which can be avoided by returning a nested `Ctx` instead when `TypeDef`s are processed. But a long list would potentially need to be traversed when looking up the environment. Basically this would compromise either the space complexity of `TypeDef`s or the time complexity of lookups...

LPTK:
But the maximum depth of any inheritance hierarchy in real programs is very small, and does not grow linearly as the program grows at all. So traversing a list the size the of inheritance hierarchy depth upon lookup seems totally fine to me, and in fact it'll probably be faster than copying mutable maps around _or_ going full immutable.

On the other hand, storing plain context copies seems much easier so I'd encourage you to do that. Eventually we'll change it and do the right thing anyway. (If someone is reading this 20 years later and this still has not changed, and I'm  not dead, ping me.)

fo5for:
I got something kinda working, but it's very hacky, and ultimately storing a context isn't nice. I came up with another solution: how about we change `env` to map to a list (or vector) instead, and add an optional index value to `Var`, indicating the index it refers to in the list. Then we traverse the bodies of method definitions after registering the constructors to the context, and register the index of the last element in the list to all `Var`s. For `Var`s outside of method definitions, they would point to index `None`, and would just be typed to the last element in the list. This sounds relatively clean and doesn't require too much changes to the existing code.

LPTK:
I don't think I'd change the entire context and `Var` representations just to avoid doing something that should and will be done eventually.

If what you have now is too hacky, how much work would it be to just do it right? As in, only type methods once, and then do substitutions.

fo5for:
I started experimenting with changing `Var` and context, got too invested and completely missed your reply until I got it working. 😅 

Actually probably not too much, since the definitions are already stored in `mthEnv` for printing in `DiffTests`.

fo5for:
Turns out doing it the right way is the easiest! 😅

fo5for:
Yikes I realized I forgot to do the substitution. No idea why it worked. Fixing now 😅

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -24,56 +24,97 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
   
-  /** targsMaps: stores the map from type parameters to type variables for polymorphic classes and methods
-   *    first layer: class name;
-   *    second layer: method name or None for top-level class type parameters;
-   *    third layer: type parameter.
-   *    For "normal" contexts, all innermost values should be `TypeVariable`s.
-   *    For the nested context during method typing, the innermost values can also be `TraitTag`s.
+  /** Keys of `mthEnv`:
+   *  `L` represents the inferred types of method definitions. The first value is the parent name, and the second value is the method name.
+   *    The method definitions are *not* wrapped with the implicit `this` parameter, but is instead wrapped at `getMthDefn`.
```
```suggestion
   *    The method definitions are *not* wrapped with the implicit `this` parameter, but are instead wrapped at `getMthDefn`.
```

LPTK:
> No idea why it worked.

Whoops! The best thing to do would be to add a test that fails because of this, before fixing it.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -491,8 +495,12 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
             val reverseRigid2 = reverseRigid ++ dummyTargs2.map(t =>
               t -> freshVar(t.prov, S(t.id.idStr))(thisCtx.lvl + 1))
             val bodyTy = subst(rhs.fold(
-              term =>
-                typeLetRhs(rec, nme.name, term)(thisCtx, raise, targsMap2),
+              term => ctx.getMthDefn(prt.name, nme.name, false)
+                .fold(typeLetRhs(rec, nme.name, term)(thisCtx, raise, targsMap2))(_.toPT match {
+                  case PolymorphicType(level, ProvType(underlying)) =>
```
LPTK: This look pretty weird and ad-hoc. I'd like a comment explaining why it's necessary.

LPTK (commented):
Note about the merging strategy: since you rebased on top of the main branch, your intermediate commits now fail because of the new `Classes` test. This is not ideal as intermediate commits should ideally all pass tests. 
Since your branch has meaningful commits over a long period, it would be fine to make a proper merge commit instead (rebase back to where you were, then make a merge commit after the fix). That, or we'll just squash the branch.

shared/src/main/scala/mlscript/Typer.scala
fo5for: It's not *that* necessary, but it makes the error reporting nicer. I have added a comment to explain that.

fo5for:
I think it's fine to squash the PR. There were some changes that I made and reverted afterwards. The weird class hierarchy of the two new classes also didn't make much sense.

> > No idea why it worked.
> 
> Whoops! The best thing to do would be to add a test that fails because of this, before fixing it.

Took me some time to find something that should fail but didn't!

PR #61, opened 2022-01-08
Add reformat rules and couple of sample files

### LINKED ISSUE #60: Tooling based formatter for consistent style and better readability
Being new to the project and Scala, I found the dense code difficult to navigate and understand. By having a tooling based formatter the styling will stay consistent with good settings for readability.

I found that scala metals vs code extension natively supports formatting by creating a `.scalafmt.conf`. The bare minimum file with default settings contains this -

```
version = "3.0.5"
```

More settings are [here](https://scalameta.org/scalafmt/docs/configuration.html).

twitu:
Closes #60

### DIFF
--- a/shared/src/main/scala/mlscript/Typer.scala
+++ b/shared/src/main/scala/mlscript/Typer.scala
@@ -1,69 +1,110 @@
 package mlscript
 
 import scala.collection.mutable
-import scala.collection.mutable.{Map => MutMap, Set => MutSet}
-import scala.collection.immutable.{SortedSet, SortedMap}
+import scala.collection.mutable.{Map => MutMap}
+import scala.collection.mutable.{Set => MutSet}
+import scala.collection.immutable.SortedSet
+import scala.collection.immutable.SortedMap
 import scala.util.chaining._
 import scala.annotation.tailrec
-import mlscript.utils._, shorthands._
+import mlscript.utils._
+import shorthands._
 import mlscript.Message._
 
-/** A class encapsulating type inference state.
- *  It uses its own internal representation of types and type variables, using mutable data structures.
- *  Inferred SimpleType values are then turned into CompactType values for simplification.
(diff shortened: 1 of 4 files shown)


shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -73,213 +114,411 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+        TypeName("unit"),
+        Nil,
+        Nil,
+        Top,
+        List.empty,
+        List.empty,
+        Set.empty,
+        N
+      ) ::
+      Nil
```
LPTK, line 287: Oh my, I'd rather not...!

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -363,55 +619,85 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+                      else
+                        f._1 ->
+                          freshVar(
+                            noProv,
+                            S(
+                              f._1.name.drop(f._1.name.indexOf('#') + 1)
+                            ) // strip any "...#" prefix
+                          )(1).tap(_.upperBounds ::= f._2)
+                    )
+                    .toList
```
LPTK, line 649: The original was fine. This just looks like obscene formatting to me 😝

LPTK (commented):
Yeah, as explained in the issue I don't like it, as is. If we were a team of 20 developers I'd consider it, but right now it seems overkill and to have too big drawbacks.

Besides, now is not the time to do this as it would create lots of tricky conflicts with pending PRs, most notably https://github.com/hkust-taco/mlscript/pull/59.

shared/src/main/scala/mlscript/Typer.scala
twitu: I agree this one is pretty bad.

twitu:
Closing for further discussion on #60

PR #67, opened 2022-01-13
Convert mlscript terms to typescript declarations

### LINKED ISSUE #57: Generate TypeScript type definition files from MLscript modules
This would allow reusing MLscript definitions from TypeScript.

I don't foresee any major difficulty to implement this, although the handling of recursive types will need the introduction of auxiliary type alias definitions on the TS side, as TS does not support "inline" recursive types such as `(int -> ‘a) as 'a`.

twitu:
Closes #57


shared/src/main/scala/mlscript/codegen/Typegen.scala
```diff
@@ -0,0 +1,74 @@
...
+        case Literal(IntLit(n)) => SourceCode(n.toString)
+        case Literal(DecLit(n)) => SourceCode(n.toString)
+        case Literal(StrLit(s)) => SourceCode("\"" + s + "\"")
+        case TypeVar(identifier, nameHint) => SourceCode("type") ++ SourceCode.space ++ SourceCode(this.toString)
+    }
+
+    // function changes how bound variables are processed so
+    // they have to be handled in a look ahead style
+    // function input is positive polarity and takes upper bound
+    // function output is negative polarity and takes lower bound
```
LPTK: The polarities are the other way around. Input is negative.

shared/src/main/scala/mlscript/codegen/Typegen.scala
```diff
@@ -0,0 +1,74 @@
+package mlscript
+
+/**
+  * Convert types to typescript declarations
+  */
+sealed trait TypeSourceCode { self: Type => 
+
+    def toSourceCode: SourceCode = this match {
```
LPTK:
```suggestion
  def toSourceCode: SourceCode = this match {
```
The standard indentation width in Scala is 2 spaces.

shared/src/main/scala/mlscript/codegen/Typegen.scala
```diff
@@ -0,0 +1,74 @@
...
+        case Rem(base, names) => SourceCode("Omit") ++ SourceCode.openAngleBracket ++ base.toSourceCode ++ SourceCode.commaSpace ++ SourceCode.record(names.map(name => SourceCode(name.name))) ++ SourceCode.closeAngleBracket
+        // TODO: handle conversion since typescript does not support lower bounds
+        case Bounds(lb, ub) => SourceCode("TODO") ++ lb.toSourceCode ++ ub.toSourceCode
+        case WithExtension(base, rcd) => Inter(Rem(base, rcd.fields.map(tup => tup._1)), rcd).toSourceCode
+        case Top => SourceCode("unknown")
+        case Bot => SourceCode("never")
+        case TypeName(name) => SourceCode(name)
+        case Literal(IntLit(n)) => SourceCode(n.toString)
+        case Literal(DecLit(n)) => SourceCode(n.toString)
+        case Literal(StrLit(s)) => SourceCode("\"" + s + "\"")
```
LPTK: This would gain to be factored with:
https://github.com/hkust-taco/mlscript/blob/7d04016c9c702ccc8a896728b28e39e638b5839f/shared/src/main/scala/mlscript/JSBackend.scala#L178-L180

Note that BigInt literals with `n` are valid in TS: https://www.typescriptlang.org/play?target=99#code/MYewdgzgLgBAHgLhgRgAztWGBeFHMBQQA

shared/src/main/scala/mlscript/codegen/Typegen.scala
```diff
@@ -0,0 +1,74 @@
...
+
+    // function changes how bound variables are processed so
+    // they have to be handled in a look ahead style
+    // function input is positive polarity and takes upper bound
+    // function output is negative polarity and takes lower bound
+    // otherwise handle normally
+    def funcTypeHelper(lhs: Type, rhs: Type): SourceCode = {
+        (lhs, rhs) match {
+            case (Bounds(llb, lub), Bounds(rlb, rub)) => SourceCode.concat(List(
+                                                (SourceCode("arg0") ++ SourceCode.colon ++ lub.toSourceCode).parenthesized,
```
LPTK: Why the huge indentation? Are you making a point about automatic formatters here? 😛 
Personally I'm against any kind of vertical alignment, as that tends to lead to useless diffs upon trivial changes.

shared/src/main/scala/mlscript/syntax.scala
```diff
@@ -76,7 +76,7 @@ sealed trait Terms extends DesugaredStatement
 
 // Types
 
-sealed abstract class Type extends TypeImpl
+sealed abstract class Type extends TypeImpl with TypeSourceCode
```
LPTK: Some points:
1. You marked `TypeSourceCode` as `sealed` so it can only be extended in the same file

2. I wouldn't make this type-gen stuff a method in `Type`. See: separation of concerns. Later type-gen may even live in a different project.

3. If I did want to make this a method in `Type`, I would have probably made `TypeImpl` extends `TypeSourceCode`.

shared/src/main/scala/mlscript/syntax.scala
```diff
@@ -94,8 +94,13 @@ sealed abstract class NullaryType                        extends Type
 
 case object Top                                          extends NullaryType
 case object Bot                                          extends NullaryType
+
+// Literal type or singleton type i.e.
+// `type t: 0` is a type with only one possible value
+// which is the literal integer 0 itself
 final case class Literal(lit: Lit)                       extends NullaryType
 
+// declare a new type with a given name
```
LPTK: This is not about declaring types; it's about referencing them.

LPTK (commented):
This will need some tests.

Similarly to the `:js` option, I think it'd be really good to have a `:ts` option in the diff-tests, showing the generated TS signature for selected test cases. See how it's implemented [here](https://github.com/hkust-taco/mlscript/blob/7d04016c9c702ccc8a896728b28e39e638b5839f/shared/src/test/scala/mlscript/DiffTests.scala#L122).

Pro-tip: Running the diff-tests continuously on source change (which makes development nicer) is done with SBT command `~mlscriptJVM/testOnly mlscript.DiffTests`.

shared/src/main/scala/mlscript/codegen/Typegen.scala
twitu: :grimacing: fixing this.

LPTK:
By the way, the right thing to do here is to keep track of polarities as an `Option[Boolean]` parameter, not special-case function types. For example a function type in negative position has its polarities reversed. And bounds types may appear nested deep in other types, as in `S -> List[A..B]` where `A..B` is the bounds type.
You start at positive polarity (`pol = Some(true)`). When you traverse contravariant type positions, such as function parameters, you switch it with `pol.map(!_)`, and when you traverse an invariant one (such as the type arguments of a random `TypeRef`), you make it `None`.

twitu:
> By the way, the right thing to do here is to keep track of polarities as an `Option[Boolean]` parameter, not special-case function types. For example a function type in negative position has its polarities reversed. And bounds types may appear nested deep in other types, as in `S -> List[A..B]` where `A..B` is the bounds type. You start at positive polarity (`pol = Some(true)`). When you traverse contravariant type positions, such as function parameters, you switch it with `pol.map(!_)`, and when you traverse an invariant one (such as the type arguments of a random `TypeRef`), you make it `None`.

In this case it's best that to have a context instance parameter threaded through the type gen function just like in code gen. That will help handle all the places where some kind of state is required. As you pointed out some of them are -

* Handle type position polarity to resolve bounded types
* Generate unique names

shared/src/main/scala/mlscript/codegen/Typegen.scala
twitu: Fixed

shared/src/main/scala/mlscript/codegen/Typegen.scala
twitu: This happens sometimes when copy pasting I've fixed this in the latest commit.

shared/src/main/scala/mlscript/syntax.scala
twitu: Yes I've learnt a little more about how `sealed` is meant to be used, and this was not the right place to use it.

I've moved the ts type gen to `TypeImpl` in a separate method called `toTsType` which converts a type to `SourceCode` that can then be printed or exported to a file.

shared/src/main/scala/mlscript/codegen/Typegen.scala
twitu: Fixed.

shared/src/main/scala/mlscript/syntax.scala
twitu: Fixed

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -77,11 +77,97 @@ abstract class TypeImpl extends Located { self: Type =>
...
+    * To give unique names arguments in nested types we increment an argument
+    * counter
+    * @param ctx
+    * @return
+    */
+  def toTsType(implicit ctx: Option[Boolean] = None, argCounter: Int = 0): SourceCode =
+    this match {
+    // these types can be inferred from expression that do not need to be
+    // assigned to a variable
+    case Union(TypeName("true"), TypeName("false")) | Union(TypeName("false"), TypeName("true")) => SourceCode("bool")
```
```suggestion
    case Union(TypeName("true"), TypeName("false")) | Union(TypeName("false"), TypeName("true")) => SourceCode("boolean")
```

twitu:
Since quite a few examples have single letter names for functions or variables it might be confusing to have a single letter type alias as well.

If `t` style prefix does not look good I can remove it.

twitu:
Fixed. I've replaced `t` with empty string.

LPTK:
> it might be confusing to have a single letter type alias as well

You mean type _variables_, right? Please make sure to use the correct terminology, to be sure we're talking about the same things.

shared/src/main/scala/mlscript/codegen/Codegen.scala
```diff
@@ -122,15 +131,32 @@ object SourceCode {
...
+  val fatArrow: SourceCode = SourceCode(" => ")
+  val equalSign: SourceCode = SourceCode(" = ")
 
   def concat(codes: Ls[SourceCode]): SourceCode =
     codes.foldLeft(SourceCode.empty) { _ + _ }
 
+  /**
+    * Source entries with curly braces
+    * 
+    * e.g. [1, 1] = {1, 1}
```
LPTK: I don't understand what `[1, 1] = {1, 1}` means. Are you using Haskell/JS syntax to document Scala parameters?

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -6,6 +6,55 @@ import scala.collection.immutable.SortedSet
...
 import mlscript.utils._, shorthands._
 
+/**
+  * A class to hold to the context while converting mlscript types to
+  * typescript representation.
+  * 
+  * polarity tracks polarity of the current type.
+  * * a type has no polarity by default
+  * * a type has negative polarity at function input position
+  * * a type has positive polarity at function output position
```
LPTK: That's actually not how polarities work. For a given definition's type, we start at positive polarity. Function parameters don't turn the polarity negative, but instead invert it. Results leave the polarity unchanged.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -6,6 +6,55 @@ import scala.collection.immutable.SortedSet
...
+  * @param polarity
+  * @param argCounter
+  * @param existingTypeVars
+  * @param newTypeAlias
+  * @param typeParams
+  */
+class ToTsTypegenContext(
+  var polarity: Option[Boolean] = None,
+  var argCounter: Int = 0,
+  var existingTypeVars: Map[TypeVar, String] = Map.empty,
```
LPTK: Why use a `var` for this? Mutable variables should be avoided if possible.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -6,6 +6,55 @@ import scala.collection.immutable.SortedSet
...
+  * `TypeVar` to their SourceCode representation.
+  *
+  * @param polarity
+  * @param argCounter
+  * @param existingTypeVars
+  * @param newTypeAlias
+  * @param typeParams
+  */
+class ToTsTypegenContext(
+  var polarity: Option[Boolean] = None,
```
LPTK: `polarity` should probably be immutable, and you should copy the context and pass it locally when going into a type subexpression. OTOH `argCounter` can remain a `var` but would be better as a local variable in `toTsTypeSourceCode`, as I tried to explain in https://github.com/hkust-taco/mlscript/issues/57#issuecomment-1027636268 (did you see the comment?)

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -78,10 +127,180 @@ abstract class TypeImpl extends Located { self: Type =>
...
+      case Literal(DecLit(n)) => SourceCode(n.toString)
+      case Literal(StrLit(s)) => SourceCode(JSLit.makeStringLiteral(s))
+      
+      // the below types are inferred from expressions that need to be assigned to variables or definitions
+      // or require information from previous declarations to be completed
+      
+      // typescript function types must have a named parameter
+      // e.g. (val: number) => string
+      case Function(lhs, rhs) =>
+        {
```
LPTK: Curly braces here do nothing. Similar remark in other places of the code.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -78,10 +127,180 @@ abstract class TypeImpl extends Located { self: Type =>
...
+    * @return
+    */
+  def toTsTypeSourceCode(termName: Option[String] = None): SourceCode = {
+    val ctx = ToTsTypegenContext.empty;
+    
+    // Create a mapping from type var to their friendly name for lookup
+    // Also add them as type parameters to the current type because typescript
+    // uses parametric polymorphism
+    ctx.existingTypeVars = ShowCtx.mk(this :: Nil, "").vs;
+    val tsType = this.toTsType(ctx);
```
LPTK: `this.` here does nothing.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -6,6 +6,55 @@ import scala.collection.immutable.SortedSet
...
+  * @param typeParams
+  */
+class ToTsTypegenContext(
+  var polarity: Option[Boolean] = None,
+  var argCounter: Int = 0,
+  var existingTypeVars: Map[TypeVar, String] = Map.empty,
+  // adhoc type vars introduced during conversion
+  var newTypeAlias: List[SourceCode] = List.empty,
+  // adhoc type parameters introduced during conversion
+  var newTypeParams: List[SourceCode] = List.empty
```
LPTK: `newTypeAlias`, `newTypeParams`: same remark as for `argCounter`.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -78,10 +127,180 @@ abstract class TypeImpl extends Located { self: Type =>
...
+    val ctx = ToTsTypegenContext.empty;
+    
+    // Create a mapping from type var to their friendly name for lookup
+    // Also add them as type parameters to the current type because typescript
+    // uses parametric polymorphism
+    ctx.existingTypeVars = ShowCtx.mk(this :: Nil, "").vs;
+    val tsType = this.toTsType(ctx);
+    
+    ctx.newTypeParams = ctx.newTypeParams ++ ctx.existingTypeVars.map(tup => SourceCode(tup._2));
+    val completeTypegenStatement = termName match {
```
LPTK: This intermediate `val` seems redundant.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -78,10 +127,180 @@ abstract class TypeImpl extends Located { self: Type =>
...
+    * keep tracks of extra type variables and type parameters defined through the context.
+    *
+    * Takes a context to maintain state.
+    * @param ctx
+    * @return
+    */
+  def toTsType(implicit ctx: ToTsTypegenContext = ToTsTypegenContext.empty): SourceCode = {
+    this match {
+      // these types can be inferred from expression that do not need to be
+      // assigned to a variable
```
LPTK: Not clear what this comment means. Reformulate? (And punctuate.)
Why do you talk about "inference" here?

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -78,10 +127,180 @@ abstract class TypeImpl extends Located { self: Type =>
...
+      case Union(TypeName("true"), TypeName("false")) | Union(TypeName("false"), TypeName("true")) => SourceCode("boolean")
+      case Union(lhs, rhs) => SourceCode.sepBy(List(lhs.toTsType, rhs.toTsType), SourceCode.separator)
+      case Inter(lhs, rhs) => SourceCode.sepBy(List(lhs.toTsType, rhs.toTsType), SourceCode.ampersand)
+      case Record(fields) => SourceCode.recordWithEntries(fields.map(entry => (SourceCode(entry._1.name), entry._2.toTsType)))
+      // unwrap extra parenthesis when tuple has single element
+      case Tuple(fields) => if (fields.length === 1) {
```
LPTK: This does not seem right. `[T]` is not the same as `T`.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -78,10 +127,180 @@ abstract class TypeImpl extends Located { self: Type =>
...
+      // typescript function types must have a named parameter
+      // e.g. (val: number) => string
+      case Function(lhs, rhs) =>
+        {
+          val currArgCount = ctx.argCounter;
+          val currPolarity = ctx.polarity;
+          
+          ctx.argCounter += 1;
+          ctx.multiplyPolarity(false)
+          val lhsTypeSource = lhs.toTsType;
```
LPTK: Here you'd make it `lhs.toTsType(ctx.copy(polarity = ...))` instead.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -78,10 +127,180 @@ abstract class TypeImpl extends Located { self: Type =>
...
+      }
+      // this occurs when polymorphic types (usually classes) are applied
+      case AppliedType(base, targs) => if (targs.length =/= 0) {
+          SourceCode(base.name) ++ SourceCode.openAngleBracket ++ SourceCode.sepBy(targs.map(_.toTsType)) ++ SourceCode.closeAngleBracket
+        } else {
+          // no type arguments required then print without brackets
+          SourceCode(base.name)
+        }
+      // Neg requires creating a parameterized type alias to hold the negated definition
+      case Neg(base) => {
```
LPTK: We should probably fail with an error here unless we're in negative position, as we've discussed.

Also, why not introduce a single `Neg<T>` definition used by all negations?

LPTK (changes requested):
I don't really like these definitions being added to `TypeImpl`. Why not put all the ts-gen things in its own file, to more cleanly separate concerns?

LPTK:
Also notice that the CI is failing: https://github.com/hkust-taco/mlscript/runs/5152982111?check_suite_focus=true

Make sure to run the tests on a clean repo to catch any changes.

shared/src/main/scala/mlscript/codegen/Codegen.scala
twitu: Removed the confusing example with better explanation.

shared/src/main/scala/mlscript/helpers.scala
twitu: Removed unneeded curly braces from this and other parts of code.

shared/src/main/scala/mlscript/helpers.scala
twitu: This was not well phrased earlier. 

The main different between typegen for say Union and Function is that Union does not affect the typegen context while Function will affect argCounter in typegen context. I've updated the comment to make this different clear.

shared/src/main/scala/mlscript/helpers.scala
twitu:
```
:d
def f x y = x + y
//│ 1. Typing term ((x,) => ((y,) => ((+ (x,)) (y,)))) for class class mlscript.Lam
```

The arguments to Lambda get converted to Tuples with single elements like `(x,)`, converting it directly would make the type `(arg0: [number]) => ..` in ts.

I unwrapped it because I couldn't see the difference between single element and single element tuple. What is the difference?

shared/src/main/scala/mlscript/helpers.scala
twitu: Updated the logic and the documentation.

I really need to understand polarity better though.

shared/src/main/scala/mlscript/helpers.scala
twitu: I've added the check and returning something like `SourceCode("ERROR unable to ....")

For a single common `Neg<T>`, this what I can think of.

```
type Neg<NegateType, FromType> = FromType extends NegateType ? never : FromType
```

twitu (commented):
One key change requested is to make typegen context immutable and use copy semantics to pass it around.

The challenge is that apart from polarity all the other fields - argCounter, existingTypeVars, newTypeAlias, and newTypeParams are modified by the recursive `toTsType` function. And then these changed values need to be passed backup because they are needed for generating final output. For e.g. `Recursive(uv, body)` adds a new type alias and removes a type variable.

This can be done using variable capture as mentioned in a previous comment. But internally this still works by making a function object with some mutable variable fields. By creating a separate class for it, I found it easier to read and document.

twitu:
> You mean type variables, right? Please make sure to use the correct terminology, to be sure we're talking about the same things.

```
//│ type a = (arg0: int) => a
//│ export declare const l: a
```

As an example. I was saying that `a` is a type alias and that identifier `l` is ascribed with. I thought it would be a confusing, but it looks fine.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -80,10 +128,166 @@ abstract class TypeImpl extends Located { self: Type =>
...
+          SourceCode(base.name) ++ SourceCode.openAngleBracket ++ SourceCode.sepBy(targs.map(_.toTsType)) ++ SourceCode.closeAngleBracket
+        } else {
+          // no type arguments required then print without brackets
+          SourceCode(base.name)
+        }
+      // Neg requires creating a parameterized type alias to hold the negated definition
+      case Neg(base) =>
+        // Negative type only works in positions of negative polarity
+        if (ctx.polarity =/= Some(false)) {
+          SourceCode("ERROR: cannot convert negative type in non-negative polarity position")
```
LPTK: This expression does not do anything! You should be throwing an exception here.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -80,10 +128,166 @@ abstract class TypeImpl extends Located { self: Type =>
...
+        SourceCode.openAngleBracket ++ base.toTsType ++ SourceCode.commaSpace ++
+        SourceCode.record(names.map(name => SourceCode(name.name))) ++ SourceCode.closeAngleBracket
+      case Bounds(lb, ub) =>
+        ctx.polarity match {
+          // positive polarity takes upper bound
+          case Some(true) => ub.toTsType
+          // negative polarity takes lower bound
+          case Some(false) => lb.toTsType
+          // TODO: Yet to handle invariant types
+          case None => SourceCode("TODO") ++ lb.toTsType ++ ub.toTsType
```
LPTK: Exception here please.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -80,10 +128,166 @@ abstract class TypeImpl extends Located { self: Type =>
...
+      case Recursive(uv, body) =>
+        val uvTsType = uv.toTsType
+        
+        // use modified context for converting body type. This will replace
+        // all use of uv type var with it's wrapped type.
+        ctx.newTypeAlias = (SourceCode("type ") ++ uvTsType ++
+          SourceCode.equalSign ++ body.toTsType) +: ctx.newTypeAlias
+          
+        // recursive type is no longer a type variable it's an alias instead
+        ctx.existingTypeVars = ctx.existingTypeVars.removed(uv)
```
LPTK: What does this line achieve? What if there happens to be the same recursive variable used twice, as in `(... as 'a) -> (... as 'a)`? (This is valid; `'a` happens to be bound twice.) This logic is very fishy.

shared/src/main/scala/mlscript/helpers.scala
LPTK: Each function type should have a tuple type as its parameter type, which corresponds to a TS parameter list. It's never okay to turn tuples of one field into their element type. You do see that this cannot be the right thing to do, right?

LPTK: To be clearer: think of the function type `(int,) -> (int,)`. It should NOT be converted to `(arg0: int) -> int`, but rather to `(arg0: int) -> [int]`. The function returns an array (of one integer element), not an integer! In other words, you should handle function parameter types specially.

shared/src/main/scala/mlscript/helpers.scala
LPTK: It's a really bad habit to return an incorrect value instead of cleanly failing early. The former approach makes debugging much harder and introduces silent failures that can go unnoticed and screw things up in production. It's the "JavaScript" way.

>     type Neg<NegateType, FromType> = FromType extends NegateType ? never : FromType

Yeah that seems right. Maybe use `NegatedType` as the name instead of `NegateType`?

shared/src/main/scala/mlscript/helpers.scala
LPTK: Also, try `def f (x, y) = x + y`. You have to get into the mindset of thinking to handle all possible situations that could arise, instead of making an over-specialized fix for one concrete example you're currently looking at.

LPTK:
> As an example. I was saying that `a` is a type alias and that identifier `l` is ascribed with. I thought it would be a confusing, but it looks fine.

For type variables that we end up treating as type aliases (namely, recursive ones), it may indeed probably be better to give them another name. But that should not be at the expense of mangling the names of _all_ type variables.

LPTK:
> One key change requested is to make typegen context immutable and use copy semantics to pass it around.
> 
> The challenge is that apart from polarity all the other fields - argCounter, existingTypeVars, newTypeAlias, and newTypeParams are modified by the recursive `toTsType` function. And then these changed values need to be passed backup because they are needed for generating final output. For e.g. `Recursive(uv, body)` adds a new type alias and removes a type variable.
> 
> This can be done using variable capture as mentioned in a previous comment. But internally this still works by making a function object with some mutable variable fields. By creating a separate class for it, I found it easier to read and document.

So, here is what I think we should do.

First, we should remember the big picture – the goal here is to generate a `.d.ts` file from a given MLscript typing unit. Crucially, the type aliases we introduce as helpers alongside the main exported definitions should not have name clashes between each other. Therefore, what you need to do here really is a form of "fresh name" management. This is why I originally suggested you make use of @chengluyu's `Scope` abstraction – you'd use it for generating type alias names. This abstraction already handles unique name generation in a nice way.

Second, polarity should simply be an immutable function parameter.

Third, `newTypeAlias` and `newTypeParams` should be local mutable variables (or local mutable collections like `ListBuffer`). They have no business being exported as visible mutable state outside of the function.

Finally, I don't quite understand why `existingTypeVars` needs to be mutable. Chances are it's better as an immutable `val` simply storing the current ShowCtx's variables.

twitu:
> I don't really like these definitions being added to TypeImpl. Why not put all the ts-gen things in its own file, to more cleanly separate concerns?

I've moved the current ts type generation logic into a separate class called `TsTypeGen` it's own package - `mlscript.codegen.typescript`. This is considering that the typedef type gen related changes will live in this package, while the common codegen functionality between js and ts may be shared through the `mlscript.codegen` package.

shared/src/main/scala/mlscript/codegen/Codegen.scala
```diff
@@ -149,6 +149,10 @@ object SourceCode {
   def concat(codes: Ls[SourceCode]): SourceCode =
     codes.foldLeft(SourceCode.empty) { _ + _ }
 
+  // concatenate source codes without intermediate allocations
+  def bulkConcat(codes: Iterable[SourceCode]): SourceCode =
+    new SourceCode(codes.map(_.lines).foldRight(List.empty[SourceLine])((lines, accum) => lines ::: accum))
```
LPTK: If you really want to avoid too many intermediate allocations:
```suggestion
    new SourceCode(codes.iterator.map(_.lines).foldRight(List.empty[SourceLine])((lines, accum) => lines ::: accum))
```

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
```diff
@@ -0,0 +1,240 @@
...
+    * information is collected to arrange the source code in proper order.
+    *
+    * Existing type vars is used to maintain a mapping from originally created type variables to their names. This is
+    * then used to resolve `TypeVar` to their SourceCode representation.
+    *
+    * @param existingTypeVars
+    * @param newTypeAlias
+    * @param typeParams
+    */
+  class ToTsTypegenContext(
```
LPTK: Remove defaults; just construct with the right arguments directly.

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
```diff
@@ -0,0 +1,240 @@
...
+          )
+        }
+
+        // try to allocate common Negate type alias
+        // skip if it already exists
+        try {
+          typegenScope.allocateJavaScriptName("Neg")
+          typegenCode += SourceCode("type Neg<NegatedType, FromType> = FromType extends NegatedType ? never: FromType")
+        } catch {
+          case _: Throwable => ()
```
LPTK: Code smell! Just check before the call.

LPTK (changes requested):
Few comments besides what we already discussed in the meeting.

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
twitu: Fixed

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
twitu: Fixed

shared/src/main/scala/mlscript/codegen/Codegen.scala
twitu: Oh ok using an iterator does not allocate the intermediate map.

shared/src/main/scala/mlscript/helpers.scala
twitu: In that sense each argument is actually a parameter list much like Scala. Typegen now does shows the correct behaviour.

twitu:
> What does this line achieve? What if there happens to be the same recursive variable used twice, as in (... as 'a) -> (... as 'a)? (This is valid; 'a happens to be bound twice.) This logic is very fishy.

Currently an example like this will converted like this.

```
(f -> a as a) -> (g -> a as a)
type newF = f -> newF
type newG = g -> newG
export const res: newF -> newG
```

Is this is an acceptable conversion? The two recursive functions will be aliased with different names. I am not able to think up an example where two different recursive types will produce the same type.

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
```diff
@@ -11,37 +11,85 @@ import mlscript.TypeDef
...
+      *   Right[None, String] is from existing type variable names
+      */
+    def findTypeVarName(t: TypeVar): Either[String, String] = {
+      this.typeVarMapping.get(t).map(Left(_)).getOrElse({
+        this.existingTypeVars
+          .get(t)
+          .map(Right(_))
+          .getOrElse(
+            throw IllFormedTsTypeError(s"Did not find mapping for type variable $t. Unable to generated ts type.")
+          )
```
LPTK: Minor improvement:
```suggestion
          .toRight(
            throw IllFormedTsTypeError(
              s"Did not find mapping for type variable $t. Unable to generated ts type."))
```

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
```diff
@@ -11,37 +11,85 @@ import mlscript.TypeDef
...
+
+  object TypegenContext {
+    def apply(mlType: Type): TypegenContext = {
+      val existingTypeVars = ShowCtx.mk(mlType :: Nil, "").vs;
+      // initialize local term scope with global term scope as parent
+      // this will reduce cases where global names are similar to function
+      // argument names. For e.g. the below case will be avoided
+      // export const arg0: int
+      // export const f: (arg0: int) => int
+      TypegenContext(existingTypeVars, MutMap.empty, Scope(Seq.empty, termScope), Scope())
```
LPTK: Shouldn't the type scope also initialize its parent to the outer type scope?

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
```diff
@@ -224,16 +242,18 @@ final case class TsTypegenCodeBuilder() {
...
-          .map(s => SourceCode(s))
-          .getOrElse(
-            throw IllFormedTsTypeError(s"Did not find mapping for type variable $t. Unable to generated ts type.")
-          )
+        typegenCtx.findTypeVarName(t).fold(
+          name => SourceCode(name),
+          // type variable has been accessed for the first time
+          // allocate a clash free name in type variable mapping
+          friendlyName => {
+            val newName = typegenCtx.typeScope.allocateJavaScriptName(s"t$friendlyName")
```
LPTK: Why did we revert to using `t` prefixes?

LPTK (commented):
This is starting to look better!

But you need a lot more tests. For example, please add this test:

```scala

:ts
def f: ('c -> 'a as 'a) -> 'c -> int
//│ f: ('c -> 'a as 'a) -> 'c -> int
//│ // start ts
//│ export type ta = (arg1: [tc]) => ta
//│ export declare const f<tc>: (arg: [ta]) => (arg2: [tc]) => int
//│ // end ts
```

Here you can see several things are wrong:

* You still don't handle function parameter lists correctly. Remember that every function type has a tuple type as its left-hand side. It's a sort of invariant (in a sense we could refactor FunctionType to be declared as `case class FunctionType(lhs: TupleType, rhs: SimpleType)` but we don't currently do it for various reasons). These tuple types should be interpreted as TS multi-parameter lists.

* Things you bind to type aliases may capture local type variables, like `'c` above. You need to add these type variables as parameters to the type aliases.

LPTK:
@twitu 
> Is this is an acceptable conversion? 

Yes. I was just pointing out that the previous logic was broken.

> I am not able to think up an example where two different recursive types will produce the same type.

I don't actually know what you mean by this. A typical example is `int -> 'a as 'a` and `int -> int -> 'b as 'b`, which are different recursive type expressions which refer to equivalent types. But this is out of topic, so I don't think that's what you meant.

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
twitu: I saw cases like this pop-up.

```
//│ export declare const g<a>: (arg: [{a: a, b: int} & a]) => (arg1: [unknown]) => a
```

Record names having same identifiers as type parameters and becoming confusing. But now that I think of this is a contrived case, mostly because I'm writing toy tests with short names. This should not be a problem if records are defined well.

Still would it be good to have some convention for type parameters? Maybe using capital letters like, how mlscript type aliases need to start with captial letters.

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
twitu: Hmmm yes type aliases could clash with type parameter names. I'll add the parent scope.

twitu:
> Things you bind to type aliases may capture local type variables, like 'c above. You need to add these type variables as parameters to the type aliases.

```
:ts
def f: ('c -> 'a as 'a) -> 'c -> int
//│ f: ('c -> 'a as 'a) -> 'c -> int
//│ // start ts
//│ export type ta = (arg1: [tc]) => ta
//│ export declare const f<tc>: (arg: [ta]) => (arg2: [tc]) => int
//│ // end ts
```

I tried the fixing the above example you showed in a6a4e00. The current output is.

```
//│ // start ts
//│ export type a<c> = (arg: [c]) => a
//│ export declare const f<c>: (arg: [a<c>]) => (arg1: [c]) => int
//|  // end ts
```

It's still incorrect because it's slightly more challenging than I expected because of this -

1. inside the type alias `a<c>` a is now an applied type but inside the `Type` tree it's still being considered as a type variable.

We could pre-compute all the type variable names using `ShowCtx` and then use type scope to give each type variable a friendly name and in the type variable mapping map `a` to -> `a<c_friendly_name, d_friendly_name.....>`. So that when type var a was found it's friendly name would be it's applied version. But this approach will not work if there are nested recursive types. For e.g.

```
def f: (('b -> 'a as 'b) -> 'c as 'c) -> 'b -> 'c -> int
```

When we begin typegen for c, we map `c -> c<b>`. But once we visit b and figure out it's recursive function it becomes a separate type alias and is not longer a type variable. The above approach, where we we pre-compute all the type variables as type parameters before completely visiting type c, fails to consider that a type variable might be removed.

The approach I can think of might require a combination of both
* add another field to typegen context like `ignoreTypeVar` this keeps track of any type variables that have been removed.
* modify the type tree of a recursive body and map the type nodes to applied type nodes with appropriate parameters.

2. There's another case I wanted to confirm.

The case where a recursive type might be used as a type variable outside it's body. As far as I understood `as a`, `a` will only be used inside the body of the recursive type it cannot exist outside, because it's ascribing the whole function to it's own output type.

```
def f: ('c -> 'a as 'a) -> 'a -> 'c -> int
('c -> 'a as 'a) -> anything -> 'c -> int
```
Again 'a will be considered a type variable even though it maybe an applied type. Although in this particular case it's being inferred as anything.

twitu:
> You still don't handle function parameter lists correctly. Remember that every function type has a tuple type as its left-hand side. It's a sort of invariant (in a sense we could refactor FunctionType to be declared as case class FunctionType(lhs: TupleType, rhs: SimpleType) but we don't currently do it for various reasons). These tuple types should be interpreted as TS multi-parameter lists.

I wanted to check I fully understand the issue well, here's the current state.
```scala
def g (x, y) = x + y
def h (x, (y, z)) = y z + x
def l m n = m + n
//│ // start ts
//│ g: (int, int,) -> int
//│ h: (int, ('a -> int, 'a,),) -> int
//│ l: int -> int -> int
//│ export declare const g: (arg: [int, int]) => int
//│ export declare const h<a1>: (arg: [int, [(arg1: [a1]) => int, a1]]) => int
//│ export declare const l: (arg: [int]) => (arg1: [int]) => int
//│ // end ts
```

Each separate argument and `()` tuple/parameter list is getting converted into it's own fixed length array. Nested tuples are becoming fixed length nested arrays. Have I misunderstood something in this part?

twitu:
> Things you bind to type aliases may capture local type variables, like 'c above. You need to add these type variables as parameters to the type aliases.

I realized that my previous comment was proposing a complex solution because I was constrained to a single pass approach. Instead I think this can be solved in a better way by doing two passes on the type.

1. Find all the type variables that will no longer by type variables after the type gen
2. Do the type gen

Two passes might also help with some complexities with classes and methods type gen that I'm not aware of right now.

shared/src/main/scala/mlscript/TyperDatatypes.scala
```diff
@@ -21,6 +21,8 @@ abstract class TyperDatatypes extends TyperHelpers { self: Typer =>
 
   sealed abstract class TypeInfo
 
+  /** A type for abstract classes and is used to check and throw
```
LPTK:
```suggestion
  /** A type for abstract classes that is used to check and throw
```

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
```diff
@@ -0,0 +1,288 @@
...
+  Top, Bot, Literal, Function, Recursive, AppliedType, Neg, Rem, Bounds, WithExtension,
+  IntLit, DecLit, StrLit, Arr }
+import mlscript.{JSBackend, JSLit}
+import mlscript.ShowCtx
+import mlscript.Typer
+import mlscript.TypeDef
+import mlscript.Terms
+import mlscript.SourceLine
+import scala.collection.mutable.{ListBuffer, Map => MutMap}
+import mlscript.codegen.Scope
```
LPTK:
```suggestion
package mlscript
package codegen.typescript

import scala.collection.mutable.{ListBuffer, Map => MutMap}

import mlscript.utils._
import mlscript.codegen.Scope
```

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
```diff
@@ -0,0 +1,288 @@
...
+
+final case class IllFormedTsTypeError(message: String) extends Exception(message);
```
LPTK: I don't like the name of this thing. In fact, let's just remove it and use the existing `mlscript.codegen.CodeGenError` instead.

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
```diff
@@ -0,0 +1,288 @@
...
+
+/** Typescript typegen code builder for an mlscript typing unit
+  */
+final case class TsTypegenCodeBuilder() {
+  // store converted mlscript type definitions and terms created in mlscript
```
LPTK: Does this slightly cryptic comment refer to `private val typegenCode`? If so, should be moved.

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
```diff
@@ -0,0 +1,288 @@
...
+  object TypegenContext {
+    def apply(mlType: Type): TypegenContext = {
+      val existingTypeVars = ShowCtx.mk(mlType :: Nil, "").vs
+      val (recVarSet, nonRecVarSet) = mlType.partitionRecTypeVarSet
+      val typegenTypeScope = Scope("localTypeScope", List.empty, typeScope)
+      val typegenTermScope = Scope("localTermScope", List.empty, termScope)
+      val typeVarMapping = MutMap.empty[TypeVar, SourceCode]
+      existingTypeVars.iterator.foreach(kv => {
+        val name = typegenTypeScope.declareRuntimeSymbol(kv._2)
+        typeVarMapping += ((kv._1, SourceCode(kv._2)))
```
LPTK: More idiomatic:
```suggestion
        typeVarMapping += kv._1 -> SourceCode(kv._2)
```

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
```diff
@@ -0,0 +1,288 @@
...
+  )
+
+  object TypegenContext {
+    def apply(mlType: Type): TypegenContext = {
+      val existingTypeVars = ShowCtx.mk(mlType :: Nil, "").vs
+      val (recVarSet, nonRecVarSet) = mlType.partitionRecTypeVarSet
+      val typegenTypeScope = Scope("localTypeScope", List.empty, typeScope)
+      val typegenTermScope = Scope("localTermScope", List.empty, termScope)
+      val typeVarMapping = MutMap.empty[TypeVar, SourceCode]
+      existingTypeVars.iterator.foreach(kv => {
```
LPTK: More idiomatic:
```suggestion
      existingTypeVars.iterator.foreach { case (key, value) =>
```

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
```diff
@@ -0,0 +1,288 @@
...
+
+        // try to allocate common Negate type alias
+        if (!typeScope.existsRuntimeSymbol("Neg")) {
+          typeScope.declareRuntimeSymbol("Neg")
+          typegenCode += SourceCode("type Neg<NegatedType, FromType> = FromType extends NegatedType ? never: FromType")
+        }
+
+        // introduce a new type parameter for the `FromType`
+        val typeParam = typegenCtx.typeScope.declareRuntimeSymbol()
+        typegenCtx.typeVarMapping += ((TypeVar(Right(typeParam), None), SourceCode(typeParam)))
```
LPTK, line 241: Use `->`.

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
```diff
@@ -0,0 +1,288 @@
...
+        }
+        // recursive type has other type variables
+        // so now it is an applied type
+        else {
+          val uvTypeParams = typeVarMapping.iterator
+            .filter(tup => nestedTypegenCtx.nonRecTypeVarSet.contains(tup._1))
+            .map(_._2)
+            .toList
+          val uvAppliedName = uvName ++ SourceCode.paramList(uvTypeParams)
+          typeVarMapping += ((uv, uvAppliedName))
```
LPTK: Use `->`.

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
```diff
@@ -0,0 +1,288 @@
...
+    * @param recTypeVarSet
+    *   Set of non-recursive type variables for the type where context was intialized
+    * @param termScope
+    *   scope for argument names
+    * @param typeScope
+    *   scope for type parameter names
+    */
+  protected case class TypegenContext(
+      val typeVarMapping: MutMap[TypeVar, SourceCode],
+      val recTypeVarSet: Set[TypeVar],
```
LPTK: This is unused. Please remove.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -16,6 +16,30 @@ abstract class TypeImpl extends Located { self: Type =>
     case Recursive(n, b) => n :: b.typeVarsList
     case _ => children.flatMap(_.typeVarsList)
   }
+
+  /** Return two sets of type variables - recursive and non-recursive
+    *
+    * @return
+    */
+  lazy val partitionRecTypeVarSet: (Set[TypeVar], Set[TypeVar]) = {
```
LPTK: Since the first set is not used, let's make this simply compute the set free type variables, as I had originally suggested.

Also, the point of making this a `lazy val` is that we can store all intermediate computations in the nodes of the type tree itself, so that it can be accessed there later on. But for this to happen, the definition of the `lazy val` needs to use itself recursively! (As opposed to recursing through a local `def`.)

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
```diff
@@ -0,0 +1,288 @@
...
+      val nonRecTypeVarSet: Set[TypeVar],
```
LPTK: Since this will be a `lazy val` in `Type`, I don't think there's a need to store it here in the context.

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
```diff
@@ -0,0 +1,288 @@
...
+    * @param pol
+    *   polarity of type
+    * @param funcArg
+    *   true if Tuple type is the lhs of a function. Used to translate tuples as multi-parameter functions
+    * @return
+    */
+  private def toTsType(mlType: Type)(implicit
+      typegenCtx: TypegenContext,
+      pol: Option[Boolean],
+      funcArg: Boolean
```
LPTK: This probably shouldn't be an implicit parameter, as it's too easy to forget to set it back to `false` in most recursive calls. You can make it an explicit param with a default value of `false`, but make sure to consider all recursive calls carefully.

In fact, it seems you forget to do just that in most places!! The fact this is not caught in a test is concerning, and shows we probably don't have enough tests.

Please add this test and make sure the output is corrected:

```scala
:ts
def f: ((1,2),(3,4)) -> 5
//│ f: ((1, 2,), (3, 4,),) -> 5
//│  = <missing implementation>
//│ // start ts
//│ export declare const f: (arg1: (arg2: 1, arg3: 2), arg4: (arg5: 3, arg6: 4)) => 5
//│ // end ts
```

shared/src/main/scala/mlscript/syntax.scala
```diff
@@ -96,8 +108,13 @@ sealed abstract class NullaryType                        extends Type
 
 case object Top                                          extends NullaryType
 case object Bot                                          extends NullaryType
+
+// Literal type or singleton type i.e.
+// `type t: 0` is a type with only one possible value
```
LPTK: The syntax `type t: 0` means nothing.
```suggestion
// type `0` is a type with only one possible value
```

shared/src/test/diff/typegen/TypegenTerms.mls
```diff
@@ -0,0 +1,156 @@
+:ts
+:ne
```
LPTK: If you never want to execute the code, you could just add `:NoJS` at the beginning of the file.

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -53,9 +57,9 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
       fixme: Bool, showParse: Bool, verbose: Bool, noSimplification: Bool,
       explainErrors: Bool, dbg: Bool, fullExceptionStack: Bool, stats: Bool,
       stdout: Bool, noExecution: Bool, noGeneration: Bool, showGeneratedJS: Bool,
-      expectRuntimeErrors: Bool, showRepl: Bool, allowEscape: Bool)
+      showDeclarationTS: Bool, expectRuntimeErrors: Bool, showRepl: Bool, allowEscape: Bool)
```
```suggestion
      generateTsDeclarations: Bool, expectRuntimeErrors: Bool, showRepl: Bool, allowEscape: Bool)
```

LPTK (changes requested):
Almost good, but please address my comments and make the final refactoring I suggested therein.

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
twitu: I've fixed this and I'm adding more tests.

twitu:
I'm observing a tricky case where the definition will come after the declaration.
```
def convert: string -> int
def convert d = 4
```
Currently this will throw an error because typegen has already allocated the name `convert` in the term scope. This can be fixed by adding extra checks or by creating additional `Scope` helper methods.

Q. In such cases the type of the declaration should be used or the type inferred from the definition? I remember you had mentioned that in such situations for methods, mlscript only shows the declared type and not the type inferred from the definition.

Then there is the case where there is only a declaration. In such cases, `export "declare" const convert` will not be correct. Because the function won't be defined in the runtime.
```
def convert: string -> int
```

This grid summarizes it.

| declaration | definition | translate |
| -- | -- | -- |
| :heavy_check_mark: | :heavy_check_mark: |  which type to use? |
| :heavy_check_mark: | :red_circle: | export but don't declare |
| :red_circle: | :heavy_check_mark: | export declare and use inferred type |
| :red_circle: | :red_circle: | do nothing |

LPTK:
You should always review the diffs that your changes effect on the test files. It seems you did not notice  several ones like this:

```diff
//│ res: int
//│ // start ts
//│ export declare const cont: (arg1: int) => int
-//│ export declare const app<a>: (arg1: int) => (arg3: (arg5: int) => a) => a
+//│ export declare const app: (arg1: int) => (arg3: (arg5: int) => a) => a
//│ export declare const res: int
//│ // end ts
```

This is because you forgot the base case for `TypeVar` in `nonRecTypeVarSet`.

LPTK:
> Q. In such cases the type of the declaration should be used or the type inferred from the definition? I remember you had mentioned that in such situations for methods, mlscript only shows the declared type and not the type inferred from the definition.

You should use the declared type indeed. Note that in the diff-tests, things can also be re-declared. I don't think you should crash here. Just re-emit the same name like you do for `res`.

> Then there is the case where there is only a declaration. In such cases, `export "declare" const convert` will not be correct. Because the function won't be defined in the runtime.

Ideally, an MLscript file could be used to provide type definitions without implementations, just like TS declaration files. So the declaration may well be correct even if it's not accompanied by an MLscript definition. Just generate it normally.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -16,6 +16,16 @@ abstract class TypeImpl extends Located { self: Type =>
...
     case _ => children.flatMap(_.typeVarsList)
   }
+
+  /**
+    * @return
+    *  set of non-recursive type variables in type
+    */
+  lazy val nonRecTypeVarSet: Set[TypeVar] = this match {
+    case Recursive(uv, body) => body.nonRecTypeVarSet - uv
+    case t@TypeVar(_, _) => Set(t)
```
LPTK: More efficient (avoids allocating a vararg):
```suggestion
    case t: TypeVar => Set.single(t)
```

twitu:
I've removed the check for already declared terms so shadowing is now possible.

However there is a challenge with translating a term definition to show it's type declaration (if previously declared). Currently `TsTypeGen` does not store such a mapping from a term name to it's declaration. So there is no way to access the previous declaration.

It is possible to re-use `Scope`'s `lexicalTypeSymbols` field to store the declaration as a TypeSymbol, although it's not really a type symbol. Then for any new definition the type from the declaration can be looked up and reused.

A draft implementation is in #85

LPTK:
> It is possible to re-use `Scope`'s `lexicalTypeSymbols` field to store the declaration as a TypeSymbol, although it's not really a type symbol.

That sounds unholy. I don't know why you'd need that. Can't you just check if the thing was declared before, and if it was, simply do not re-emit a TS type?

Anyway, better to keep things as simple as possible here, rather than find ugly workarounds. In the end this logic will likely change anyway!

twitu:
TS requires exported values to have an explicit type so just having `export declare const <name>` will not be valid TS. In the latest commit I've added a `// skip <name>` in place of an already defined/declared term.

LPTK:
By "do not re-emit a TS type" I of course meant to skip the const declaration, not to literally leave the declaration without a type 🙄

LPTK:
Okay, so after looking at the latest change, I think it's better to redeclare the type of all declared/defined identifiers even if they have already been declared before. 

I prefer to see this:

```
//│ export type b<a> = a | (arg1: (arg3: b<a>) => [int, int] & a) => [int, int]
//│ export declare const weird<a>: b<a>
//│ export type a = [int, a]
//│ export declare const weird: (arg1: a) => int
//│ export declare const weird<a>: (arg1: a | [int, bool]) => a
```

Rather than this:


```
//│ // skip weird
//│ // skip weird
//│ // skip weird
```
Remember that this is just the logic for the tests, and in this context we don't want to skip any possibly useful output.  Again, the actual logic for generating valid `.d.ts` files will adapted adequately. Sorry for the back and forth!

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
```diff
@@ -0,0 +1,262 @@
...
+package codegen.typescript
+
+import scala.collection.mutable.{ListBuffer, Map => MutMap}
+
+import mlscript.utils._
+import mlscript.codegen.{Scope, CodeGenError}
+
+/** Typescript typegen code builder for an mlscript typing unit
+  */
+final case class TsTypegenCodeBuilder() {
```
LPTK:
```suggestion
final class TsTypegenCodeBuilder {
```

shared/src/main/scala/mlscript/codegen/typescript/TsTypegen.scala
```diff
@@ -0,0 +1,262 @@
...
+    *   Set of non-recursive type variables for the type where context was intialized
+    * @param termScope
+    *   scope for argument names
+    * @param typeScope
+    *   scope for type parameter names
+    */
+  protected case class TypegenContext(
+      val typeVarMapping: MutMap[TypeVar, SourceCode],
+      val termScope: Scope,
+      val typeScope: Scope
```
LPTK:
```suggestion
      typeVarMapping: MutMap[TypeVar, SourceCode],
      termScope: Scope,
      typeScope: Scope
```

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -16,6 +16,16 @@ abstract class TypeImpl extends Located { self: Type =>
     case Recursive(n, b) => n :: b.typeVarsList
     case _ => children.flatMap(_.typeVarsList)
   }
+
+  /**
+    * @return
+    *  set of non-recursive type variables in type
+    */
+  lazy val nonRecTypeVarSet: Set[TypeVar] = this match {
```
```suggestion
  lazy val freeTypeVariables: Set[TypeVar] = this match {
```

shared/src/test/diff/typegen/TypegenTerms.mls
```diff
@@ -0,0 +1,171 @@
...
+//│ res: int
+//│ res: {b: 'a} -> 'a | 1
+//│ x: 1
+//│ y: 2
+//│ z: int
+//│ res: (1, 3, {a: {b: {c: "d"}}}, "hello hello",)
+//│ // start ts
+//│ export declare const res: {a: "hello", b: "world"}
+//│ export declare const res: int
+//│ export declare const res<a>: (arg1: {b: a}) => a | 1
```
LPTK: TS parses `|` as having precedence over `=>` so some parentheses should be used here.

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -37,6 +38,9 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
     val strw = new java.io.StringWriter
     val out = new java.io.PrintWriter(strw)
     def output(str: String) = out.println(outputMarker + str)
+    def outputSourceCode(code: SourceCode) = code.lines.foreach(line => {
```
```suggestion
    def outputSourceCode(code: SourceCode) = code.lines.foreach { line =>
```

LPTK (approved):
LGTM!

PR #71, opened 2022-01-16
Run tests in parallel

fo5for:
Self-explanatory. The individual tests seem to be taking a lot longer, but the actual resource usage did not increase by much.


LPTK (approved):
That's useful, thanks! It makes the running time of tests much shorter.

The way it works is pretty weird though: it instantiates the test class many times (close to 90 times on my machine) and runs through all the file names every time, calling the `test` command needlessly (at least it doesn't re-run tests). That may in part explain the observed overhead. And that's also why I moved the filename filtering to the companion object.

Issue #66, opened 2022-01-12
Implement `originProv`


LPTK:
As discussed in the meeting.

LPTK:
Fixed in https://github.com/hkust-taco/mlscript/pull/72.

PR #72, opened 2022-01-17
Handle "origin" type provenances

Elrouille:
Add a boolean isOrigin to type provenances to reflect whether a provenance is an origin prov or not.
When using the toString method of a type provenance we prepend "o:" in front of an origin prov.
When we report an error, we add a note at the end of the error to explain where the type is defined when there is an origin prov.
We also modified the selection of lhs prov so that the one printed for the error is not an origin prov.

#66

### DIFF
--- a/shared/src/main/scala/mlscript/ConstraintSolver.scala
+++ b/shared/src/main/scala/mlscript/ConstraintSolver.scala
@@ -360,7 +360,7 @@ class ConstraintSolver extends NormalForms { self: Typer =>
           }}.toList
         else Nil
       
-      val lhsProv = cctx.head.find(_._1.prov.loco.isDefined).map(_._1.prov).getOrElse(lhs.prov)
+      val lhsProv = cctx.head.find(l => l._1.prov.loco.isDefined && !l._1.prov.isOrigin).map(_._1.prov).getOrElse(lhs.prov)
       
       // TODO re-enable
       // assert(lhsProv.loco.isDefined) // TODO use soft assert
@@ -404,6 +404,21 @@ class ConstraintSolver extends NormalForms { self: Typer =>
           msg"${msgHead}from ${p.desc}:" -> p.loco
       }
       
+      val originProvList = cctx.flatMap{ subCtx =>
+        subCtx.iterator.flatMap { case (l,r) => 
+          //we only one origin prov for each subCtx
+          val considered = List(l optionIf (_.prov.isOrigin), r optionIf (_.prov.isOrigin)).flatten
+          considered
+        }.nextOption()
+      }
+
+      first = true
+      val originProvHints = originProvList.map { l => 
+        val msgHead = if (first) msg"Note: " else msg""
+          first = false
+          msg"${msgHead}Type ${l.prov.desc} is defined at: " -> l.prov.loco 
+      }
+
       val msgs: Ls[Message -> Opt[Loc]] = List(
         msg"Type mismatch in ${prov.desc}:" -> prov.loco :: Nil,
         msg"expression of type `${lhs.expPos}` $error" ->
@@ -436,6 +451,7 @@ class ConstraintSolver extends NormalForms { self: Typer =>
           (if (isSameType) Nil else msg"which $fail" -> N :: Nil)
         }.toList.flatten,
         constraintProvenanceHints,
+        originProvHints,
         detailedContext,
       ).flatten
       raise(TypeError(msgs))
--- a/shared/src/main/scala/mlscript/ConstraintSolver.scala
+++ b/shared/src/main/scala/mlscript/ConstraintSolver.scala
@@ -406,7 +406,7 @@ class ConstraintSolver extends NormalForms { self: Typer =>
       
       val originProvList = cctx.flatMap{ subCtx =>
         subCtx.iterator.flatMap { case (l,r) => 
-          //we only one origin prov for each subCtx
+          // We only keep one origin prov for each subCtx.
           val considered = List(l optionIf (_.prov.isOrigin), r optionIf (_.prov.isOrigin)).flatten
           considered
         }.nextOption()
--- a/shared/src/main/scala/mlscript/ConstraintSolver.scala
+++ b/shared/src/main/scala/mlscript/ConstraintSolver.scala
@@ -382,7 +382,7 @@ class ConstraintSolver extends NormalForms { self: Typer =>
       
       val tighestLocatedRHS = cctx.flatMap { subCtx =>
         subCtx.flatMap { case (l, r) =>
-          val considered =  (r, r.prov) :: Nil
+          val considered = (r, r.prov) :: Nil
           considered.filter { case (_, p) =>
             p.loco =/= prov.loco && (p.loco match {
               case Some(loco) =>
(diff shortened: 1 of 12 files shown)


shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -121,13 +121,13 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
     TypeProvenance(trm.toLoc, if (desc === "") trm.describe else desc)
   def originProv(loco: Opt[Loc], desc: Str): TypeProvenance = {
     // TODO make a new sort of provenance for where types and type varianles are defined
-    // tp(loco, desc)
+    tp(loco, desc,true)
```
```suggestion
    tp(loco, desc, true)
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -121,13 +121,13 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
     // ^ This yields unnatural errors like:
       //│ ╟── expression of type `B` is not a function
       //│ ║  l.6: 	    method Map[B]: B -> A
       //│ ║       	               ^
     // So we should keep the info but not shadow the more relevant later provenances
-    noProv
+    //noProv
```
```suggestion
```

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -398,18 +398,30 @@ class ConstraintSolver extends NormalForms { self: Typer =>
...
           val msgHead = if (first) msg"Note: constraint arises " else msg""
           first = false
           msg"${msgHead}from ${p.desc}:" -> p.loco
-        }
-        else msg"in the context of ${p.desc}" -> p.loco
       }
       
+      val originProvList = cctx.flatMap{ subCtx =>
+        subCtx.iterator.flatMap { case (l,r) => 
+          //we only one origin prov for each subCtx
```
LPTK: I don't understand what this comment means.

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -382,12 +382,12 @@ class ConstraintSolver extends NormalForms { self: Typer =>
       
       val tighestLocatedRHS = cctx.flatMap { subCtx =>
         subCtx.flatMap { case (l, r) =>
-          val considered = (true, r, r.prov) :: Nil
-          considered.filter { case (isMainProv, _, p) =>
+          val considered =  (r, r.prov) :: Nil
```
LPTK:
```suggestion
          val considered = (r, r.prov) :: Nil
```

shared/src/main/scala/mlscript/TyperDatatypes.scala
```diff
@@ -12,9 +12,9 @@ abstract class TyperDatatypes extends TyperHelpers { self: Typer =>
   
   // The data types used for type inference:
   
-  case class TypeProvenance(loco: Opt[Loc], desc: Str) {
+  case class TypeProvenance(loco: Opt[Loc], desc: Str, isOrigin: Bool = false) {
     def & (that: TypeProvenance): TypeProvenance = this // arbitrary; maybe should do better
-    override def toString: Str = "‹"+loco.fold(desc)(desc+":"+_)+"›"
+    override def toString: Str = (if(isOrigin) "o: " else "") + "‹"+loco.fold(desc)(desc+":"+_)+"›"
```
LPTK:
```suggestion
    override def toString: Str = (if (isOrigin) "o: " else "") + "‹"+loco.fold(desc)(desc+":"+_)+"›"
```

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -398,18 +398,30 @@ class ConstraintSolver extends NormalForms { self: Typer =>
...
+          val considered = List(l optionIf (_.prov.isOrigin), r optionIf (_.prov.isOrigin)).flatten
+          considered
+        }.nextOption()
+      }
+
+      first = true
+      val originProvHints = originProvList.map { l => 
+        val msgHead = if (first) msg"Note: " else msg""
+          first = false
+          msg"${msgHead}Type ${l.prov.desc} is defined at: " -> l.prov.loco
```
LPTK: This results in messages like "`Type class A is defined at ...`". But a "type class" is something else. Perhaps use this instead:
```suggestion
          msg"${msgHead}${l.prov.desc.capitalize} is defined at: " -> l.prov.loco 
```

shared/src/test/diff/mlscript/BadMethods.mls
```diff
@@ -9,9 +9,14 @@ class Bar[A]: Foo[A]
 //│ ║  l.7: 	    rec method Map f = f Map (fun x -> Map)
 //│ ║       	               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 //│ ╟── expression of type `B` is not a function
+//│ ║  l.6: 	    method Map[B]: B -> A
```
LPTK, line 12: It looks like this is an origin-prov being printed. It should not.

LPTK (changes requested):
LGTM, besides the various proposed changes. Let's address them and revisit.

LPTK:
Also, you should run the tests again. As shown in the CI run, some line number changes were not committed: https://github.com/hkust-taco/mlscript/runs/4837434285?check_suite_focus=true

shared/src/main/scala/mlscript/ConstraintSolver.scala
Elrouille:
```suggestion
          //we only keep one origin prov for each subCtx
```

shared/src/main/scala/mlscript/ConstraintSolver.scala
Elrouille: Yes, it makes sense

shared/src/test/diff/mlscript/BadMethods.mls
Elrouille: Yes, but otherwise nothing is printed because the list of provenances does not have any information that we can print except for the origin prov. I figured origin prov should not be printed when we had more precise provenances, but in this case, they might be of interest?

Below is the look of the error before implementing origin provenances :

```
//CONSTRAINT FAILURE: B <: ([([α18])] -> α20)
//CTX: List(List(B <: ([([α18])] -> α20) [‹expression›] [‹application:Loc(126,131,BadMethods:+3)›], B <: [([([α18])] -> α20)] [‹expression›] [‹applied expression:Loc(126,127,BadMethods:+3)›], B <: α19 [‹expression›] [‹variable›], (B) <: (α19) [‹expression›] [‹expression›]))
//│ ╔══[ERROR] Type mismatch in method definition:
//│ ║  l.7: 	    rec method Map f = f Map (fun x -> Map)
//│ ║       	               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
//│ ╟── expression of type `B` is not a function
//│ ╟── Note: constraint arises from application:
//│ ║  l.7: 	    rec method Map f = f Map (fun x -> Map)
//│ ╙──  
```

shared/src/main/scala/mlscript/ConstraintSolver.scala
LPTK:
```suggestion
          // We only keep one origin prov for each subCtx.
```

shared/src/test/diff/mlscript/BadMethods.mls
LPTK: Hmm sure, why not. But then it shouldn't be printed in the note below, because it's redundant. Anyway, this part of the code will change soon.

LPTK (approved):
Okay, you can go for the squash-merge (after making the small comment change). Don't forget to clean up the commit message (don't leave useless info in it such as the bullet-list of intermediate commit messages).

PR #73, opened 2022-01-18
Refactor type provenance tracking and error reporting

LPTK:
This PR deeply cleans up the way type provenances are recorded and used for error reporting. This took a lot more effort to accomplish than I originally thought.

Now the type provenance chain construction is more principled, simpler, and importantly no longer mixes the provenances of type flows at different levels. For instance, before this PR we used to have provenances chains like:

```ts
//│ ╟── expression of type `false` does not match type `int`
//│ ║  l.+1: 	f { prop: false }
//│ ║        	          ^^^^^
//│ ╟── but it flows into record with expected type `{prop: ?a}`
//│ ║  l.+1: 	f { prop: false }
//│ ║        	  ^^^^^^^^^^^^^^^
```

This approach turns out (from long experience, notably on another project) to be unmaintainable and inextricably buggy. Now we do not mix flows in this way – every part of a provenance chain is supposed to have the same type (or at most one actual and one expected type, in case of a mismatch).

The tracking is currently still only for _first-order_ provenances.
Investigating higher-order ones is left for future work (Cc: @Elrouille).


shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -325,139 +340,106 @@ class ConstraintSolver extends NormalForms { self: Typer =>
...
-              => doesntHaveField(n.name)
-            case _ => doesntMatch(r)
-          }
-          msg"but it flows into ${l.prov.desc}$expTyMsg" -> l.prov.loco ::
-          (if (isSameType) Nil else msg"which $fail" -> N :: Nil)
+        tighestRelevantFailure.map { l =>
+          val expTyMsg = msg" with expected type `${rhs.expNeg}`"
+          msg"but it flows into ${l.prov.desc}$expTyMsg" -> l.prov.loco :: Nil
         }.toList.flatten,
-        constraintProvenanceHints,
```
Elrouille, line 453: The variable constraintProvenanceHints containing the updated content can probably be kept to keep the logic used with originProvHints and detailedContext

LPTK: Sure, good idea. I've made the change in the follow-up PR though, to avoid conflicts: https://github.com/hkust-taco/mlscript/pull/74/commits/eb83e2475a6ebf7c2c000e146cf2fda05ab46af8.

Elrouille:
I've reviewed all the tests diffs. They seem correct to me. The only thing I'm not sure about is the result of the :stats option -I'm not saying it's incorrect, just that I could not check whether it was.

LPTK:
No worries, `:stats` simply shows the number of constraining calls just to make sure it doesn't go crazy. It's normal if it changes a little after making changes to constraint solving.

PR #74, opened 2022-01-18
Improve error messages

LPTK:


shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -656,6 +656,8 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
         .orElse(ctx.tyDefs.get(name).map(td => (td.kind, td.tparamsargs.size)))
         .toRight(() => err("type identifier not found: " + name, loc)(raise))
     val localVars = mutable.Map.empty[TypeVar, TypeVariable]
+    def tp(loco: Opt[Loc], desc: Str, originName: Opt[Str] = N) =
```
Elrouille: I find it weird to see tp being redefined here considering it's the name we give to TypeProvenance.apply l.119. I understand why it's necessary to define it like this (as we want the boolean isType to be true) but it's currently confusing.

shared/src/test/diff/mlscript/Annoying.mls
```diff
@@ -71,14 +71,14 @@ def test x = case x of { int -> x.f | _ -> x.x }
 //│ ╔══[ERROR] Type mismatch in def definition:
 //│ ║  l.57: 	def test x = case x of { int -> x.f | _ -> x.x }
```
Elrouille, line 72: It seems like the ^^^^ should be pointing toward the second part of the test.

Elrouille, line 72: It seems like the ^^^^^ should be highlighting part `_ -> x.x`

shared/src/test/diff/mlscript/Arrays.mls
```diff
@@ -137,7 +143,7 @@ ty2A = ty2B
 //│ ╔══[ERROR] Type mismatch in def definition:
 //│ ║  l.+1: 	ty2A = ty2B
```
Elrouille, line 144: line number should be l.139

shared/src/main/scala/mlscript/Typer.scala
LPTK: You're right that it was confusing! I've changed it to `tyTp`.

shared/src/test/diff/mlscript/Annoying.mls
LPTK: Good catch! The locations of `case` and `def` trees were off. Fixed in faaf465385da5a3b17f661a183274fd49c1e04ef.

shared/src/test/diff/mlscript/Arrays.mls
LPTK: This is just the way line numbers local to the current block are printed with option `:ShowRelativeLineNums`. It's used to minimize diffs due to line number changes in errors. We should really use that approach more often!

PR #77, opened 2022-01-24
Fix comma expression precedence in generated code

chengluyu:
Currently, the generated code duplicates results sometimes, causing wrong orders in the output.

telegram-cloud-photo-size-4-5999058118752974196-y

The bug occurred because when generating comma expressions, we didn’t wrap them in parentheses when in array expressions, and invocations. This PR fixes the problem.


shared/src/test/diff/codegen/Parentheses.mls
```diff
@@ -0,0 +1,101 @@
...
+//│ res = x;
+//│ // Query 2
+//│ res = x > 0 && x < 0;
+//│ // End of generated code
+//│ x: 0
+//│  = 0
+//│ res: bool
+//│    = false
+
+// Some complicated expressiones.
```
LPTK:
```suggestion
// Some complicated expressions.
```

LPTK (approved):
Thanks for the fix!

Please rebase on the new mlscript head **and recompile the JS file** (with `fullOptJS`), as `Main.scala` has changed in the meantime (see https://github.com/hkust-taco/mlscript/commit/704c43eaa9938aae019eafb7a01d4114ce45e991).

Then you can squash-merge.

Issue #69, opened 2022-01-14
Properly support tuple field accessors typing


LPTK:
There are currently at least two ways in which they are broken:

1. While they work on simple types thanks to the case in `ConstraintSolver`:

    ```ts
    t = (1, 2, 3)
    //│ t: (1, 2, 3,)
    //│  = [ 1, 2, 3 ]
    
    t._1
    //│ res: 1
    //│    = undefined
    ```

    they don't work for more complex cases, because the normal forms don't know about them:

    ```ts
    t = (1, 2, 3) with {x = 1}
    // t = (1, 2, 3)
    //│ t: (1, 2, 3,) & {x: 1}
    //│  = Array { '0': 1, '1': 2, '2': 3, x: 1 }
    
    t._1
    //│ ╔══[ERROR] Type mismatch in field selection:
    //│ ║  l.301: 	t._1
    //│ ║         	^^^^
    //│ ╟── expression of type `(1, 2, 3,) & {x: 1}` does not match type `{_1: ?a}`
    //│ ║  l.294: 	t = (1, 2, 3) with {x = 1}
    //│ ║         	     ^^^^^^^^^^^^^^^^^^^^^
    //│ ╟── but it flows into reference with expected type `{_1: ?a}`
    //│ ║  l.301: 	t._1
    //│ ╙──       	^
    ```

2. We can subvert them, causing a soundness problem:

    ```ts
    t = (1, 2, 3) with {_1 = "oops"}
    //│ t: (1, 2, 3,) & {_1: "oops"}
    //│  = Array { '0': 1, '1': 2, '2': 3, _1: 'oops' }
    
    (t: (int,int,int,))._1
    //│ res: int
    //│    = 'oops'
    ```

LPTK:
Related JS-gen issue: https://github.com/hkust-taco/mlscript/issues/65#issuecomment-1013010123

LPTK:
I realize the right way of doing this is to change the way field removal types act on tuples (and arrays).

Currently it leaves tuples untouched:

https://github.com/hkust-taco/mlscript/blob/b15b7765ed91217c31c3bdb44db603392fa11ba6/shared/src/main/scala/mlscript/TyperHelpers.scala#L318

https://github.com/hkust-taco/mlscript/blob/b15b7765ed91217c31c3bdb44db603392fa11ba6/shared/src/main/scala/mlscript/TyperHelpers.scala#L358

[...]

PR #79, opened 2022-02-04
Added `Array type`

### LINKED ISSUE #63: Add Array types
MLscript already supports tuples, which in JS/TS are represented using arrays.

We should add proper (immutable for now) array types which are supertypes of tuple types.

The easiest would be to have some `Array[T]` with subtyping relationships like `(1, 2, 3) <: Array[1 | 2 | 3]`.

Note for later: it would be nice if we could type [_spreads_](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#spread-expressions-with-tuple-types), as in types of shapes such as `(S, T, ...U, V, W, ...X)` with an arbitrary number of "spliced" types. These types would precisely type the corresponding value spreads, and they would reduce when more information becomes known about the involved types. Cc @chengluyu.

Meowcolm024:
Added type `Array[T]` as a super type of tuples. Following: `(a, b, c) <: Array[a | b | c]` as in #63


shared/src/main/scala/mlscript/MLParser.scala
```diff
@@ -51,11 +51,21 @@ class MLParser(origin: Origin, indent: Int = 0, recordLocations: Bool = true) {
...
-      case (prt, id) => Var(s"${prt}.${id}")
-    }))).rep ).map {
-      case (st, sels) => sels.foldLeft(st)(Sel)
+
+  def subterm[_: P]: P[Term] = P( subtermNoSel ~ (
+    // fields
+    ("." ~/ (variable | locate(("(" ~/ ident ~ "." ~ ident ~ ")")
+      .map {case (prt, id) => Var(s"${prt}.${id}")})))
+      .map {(t: Var) => Left(t)} |
+    // array subscription
```
LPTK:
```suggestion
    // array subscript
```

shared/src/main/scala/mlscript/NormalForms.scala
```diff
@@ -53,6 +68,9 @@ class NormalForms extends TyperDatatypes { self: Typer =>
           case (S(b), w: Without) => S(Without(b & w, SortedSet.empty)(noProv))
           case (S(w: Without), b) => S(Without(w & b, SortedSet.empty)(noProv))
           case (S(_), _) => N
+          case (N, tup: TupleType) =>
+            r1Final = RecordType(mergeSortedMap(r1Final.fields, tup.toRecord.fields)(_ & _).toList)(noProv)
+            S(that)
           case (N, _) => S(that)
         }) map { b => LhsRefined(S(b), ts, r1) }
```
LPTK: Wait... why is `r1` used here and not `r1Final`? It seems `r1Final` isn't even used!

That's probably a bug, but it seems to have no effect on the current test suite. It would be nice to come up with a test case that makes this incorrect version fail.

shared/src/main/scala/mlscript/NormalForms.scala
```diff
@@ -122,30 +140,35 @@ class NormalForms extends TyperDatatypes { self: Typer =>
       case (RhsBases(ps, bf), p: ClassTag) =>
         S(RhsBases(if (ps.contains(p)) ps else p :: ps , bf))
       case (RhsBases(ps, N), that: MiscBaseType) => S(RhsBases(ps, S(L(that))))
-      case (RhsBases(ps, S(L(TupleType(fs1)))), TupleType(fs2)) =>
-        if (fs1.size =/= fs2.size) N
+      case (RhsBases(ps, S(L(t1@TupleType(fs1)))), t2@TupleType(fs2)) =>
+        if (fs1.size =/= fs2.size) 
+          RhsBases(ps, S(L(t1.toArray))) | t2.toArray // upcast tuple to array
```
LPTK:
```suggestion
          RhsBases(ps, S(L(t1.toArray))) | t2.toArray // upcast tuples of different sizes to array
```

shared/src/main/scala/mlscript/TyperHelpers.scala
```diff
@@ -384,6 +404,7 @@ abstract class TyperHelpers { self: Typer =>
       case ComposedType(_, l, r) => l :: r :: Nil
       case RecordType(fs) => fs.map(_._2)
       case TupleType(fs) => fs.map(_._2)
+      case ArrayType(inner) => inner :: Nil // ? not sure
```
LPTK:
```suggestion
      case ArrayType(inner) => inner :: Nil
```

shared/src/main/scala/mlscript/TyperDatatypes.scala
```diff
@@ -110,8 +110,22 @@ abstract class TyperDatatypes extends TyperHelpers { self: Typer =>
...
+  case class ArrayType(val inner: SimpleType)(val prov: TypeProvenance) extends ArrayBase {
+    lazy val level: Int = inner.level
+    override def toString = s"Array[${inner}]"
+  }
+
+  case class TupleType(fields: List[Opt[Var] -> SimpleType])(val prov: TypeProvenance) extends ArrayBase {
+    lazy val inner: SimpleType = fields.map(_._2).fold(ExtrType(true)(noProv))(_ | _)
     lazy val level: Int = fields.iterator.map(_._2.level).maxOption.getOrElse(0)
+    lazy val toArray: ArrayType = ArrayType(inner)(prov)  // upcast to array
+    // still keep this?
```
LPTK: Still keep what?

shared/src/main/scala/mlscript/TyperHelpers.scala
```diff
@@ -101,6 +102,10 @@ abstract class TyperHelpers { self: Typer =>
       case ((no1, t1), (no2, t2)) => (N, t1 | t2)
     }
   }
+
+  def tupleMerge(fs1: Ls[Opt[Var] -> SimpleType], fs2: Ls[Opt[Var] -> SimpleType]): SimpleType = {
```
LPTK: It seems this is only used once, and it does not make that much sense on its own.

shared/src/test/diff/mlscript/Repro.mls
```diff
@@ -0,0 +1,318 @@
+
```
LPTK: This file is to help conveniently reproduce issues. It should be left empty. Please move these test cases to a new file with a different name.

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -203,12 +203,18 @@ class ConstraintSolver extends NormalForms { self: Typer =>
...
-              if b.fields.size === ty.fields.size
-              => (b.fields.unzip._2 lazyZip ty.fields.unzip._2).foreach(rec(_, _, false))
-            case (LhsRefined(S(b: TupleType), ts, r), _) => reportError
+            case (LhsRefined(N, ts, r), RhsBases(pts, N | S(L(_: FunctionType | _: ArrayBase)))) => reportError
+            case (LhsRefined(S(b: TupleType), ts, r), RhsBases(pts, S(L(ty: TupleType)))) =>
+              if (b.fields.size === ty.fields.size)
+                (b.fields.unzip._2 lazyZip ty.fields.unzip._2).foreach(rec(_, _, false))
+              else
+                rec(b.toArray, ty.toArray, false)  // ?
+                // fail
```
LPTK: This is wrong. Something like `(1, 2)` should not be constrainable to `(2, 1, 2)`.
```suggestion
            case (LhsRefined(S(b: TupleType), ts, r), RhsBases(pts, S(L(ty: TupleType)))) 
              if b.fields.size === ty.fields.size
              => (b.fields.unzip._2 lazyZip ty.fields.unzip._2).foreach(rec(_, _, false))```

File: shared/src/main/scala/mlscript/ConstraintSolver.scala [hunk 9]
```diff
@@ -203,12 +203,18 @@ class ConstraintSolver extends NormalForms { self: Typer =>
...
+            case (LhsRefined(S(b: TupleType), ts, r), RhsBases(pts, S(L(ar: ArrayType)))) =>
+              rec(b.toArray, ar, false)
+            case (LhsRefined(S(ArrayType(i1)), ts, r), RhsBases(pts, S(L(ArrayType(i2))))) =>
+              rec(i1, i2, false)
```
LPTK: If you had a `toArray` method declared in `ArrayBase`, you could make this just one case:

```suggestion
            case (LhsRefined(S(b: ArrayBase), ts, r), RhsBases(pts, S(L(ar: ArrayType)))) =>
              rec(b.toArray, ar, false)
```

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -297,6 +303,8 @@ class ConstraintSolver extends NormalForms { self: Typer =>
                   msg"Wrong tuple field name: found '${ln.name}' instead of '${rn.name}'", lhs.prov.loco) } } // TODO better loco
               rec(l, r, false)
             }
+          case (t: TupleType, _: ArrayType) => rec(t.toArray, rhs, false)
+          case (ArrayType(ar1), ArrayType(ar2)) => rec(ar1, ar2, false)
```
LPTK: Same remark as above about `toArray`.

shared/src/main/scala/mlscript/NormalForms.scala
```diff
@@ -122,30 +140,35 @@ class NormalForms extends TyperDatatypes { self: Typer =>
...
         else S(RhsBases(ps, S(L(TupleType(fs1.lazyZip(fs2).map {
           case ((S(n1), ty1), (S(n2), ty2)) => (if (n1 === n2) S(n1) else N, ty1 | ty2)
           case ((n1o, ty1), (n2o, ty2)) => (n1o orElse n2o, ty1 | ty2)
         })(noProv)))))
+      case (RhsBases(ps, S(L(ArrayType(in)))), TupleType(_)) => N
```
LPTK: Shouldn't this upcast to the common array type? It seems to break the symmetry with other cases.

shared/src/main/scala/mlscript/NormalForms.scala
```diff
@@ -229,11 +252,21 @@ class NormalForms extends TyperDatatypes { self: Typer =>
             S(Conjunct(
               LhsRefined(S(FunctionType(l1 & l2, r1 | r2)(noProv)), ts, rcdU), vs1, RhsBot, nvs1))
           case (S(TupleType(fs1)), S(TupleType(fs2))) => // TODO Q: records ok here?!
-            if (fs1.size =/= fs2.size) S(Conjunct(LhsRefined(N, ts, rcdU), vs1, RhsBot, nvs1))
+            if (fs1.size =/= fs2.size) S(Conjunct(
+              LhsRefined(S(ArrayType(tupleMerge(fs1, fs2))(noProv)), ts, rcdU), vs1, RhsBot, nvs1))
             else S(Conjunct(
               LhsRefined(S(TupleType(tupleUnion(fs1, fs2))(noProv)), ts, rcdU), vs1, RhsBot, nvs1))
+          // not sure
```
LPTK: This looks correct.

```suggestion
```

shared/src/main/scala/mlscript/NormalForms.scala
```diff
@@ -229,11 +252,21 @@ class NormalForms extends TyperDatatypes { self: Typer =>
...
+          case (S(TupleType(fs)), S(ArrayType(ar))) =>
+            S(Conjunct(
+              LhsRefined(S(ArrayType(fs.map(_._2).fold(ar)(_|_))(noProv)), ts, rcdU), vs1, RhsBot, nvs1))
```
LPTK: Why not use `ArrayType(tup.toArray.inner | ar)`?

shared/src/main/scala/mlscript/NormalForms.scala
```diff
@@ -309,7 +342,7 @@ class NormalForms extends TyperDatatypes { self: Typer =>
   
   object DNF {
     def of(lnf: LhsNf): DNF = DNF(Conjunct(lnf, SortedSet.empty, RhsBot, SortedSet.empty) :: Nil)
-    def of(bt: BaseType): DNF = DNF.of(LhsRefined(S(bt), SortedSet.empty, RecordType.empty))
+    def of(bt: BaseType): DNF = DNF.of(LhsTop & bt getOrElse (???))
```
LPTK: Let's make this cleaner:
```suggestion
    def of(bt: BaseType): DNF = DNF.of(LhsRefined(S(bt), SortedSet.empty, bt.toRecord)
```
where we could declare `toRecord` in `BaseType` and make it simply return `Top` for everything except tuples...

shared/src/main/scala/mlscript/TypeSimplifier.scala
```diff
@@ -361,14 +366,24 @@ trait TypeSimplifier { self: Typer =>
...
-                    rcd.fields.filterNot(_._1.name.isCapitalized).mapValues(go(_, pol)))(rcd.prov).sorted
-                  LhsRefined(bo.map {
-                    case ct: ClassTag => ct
-                    case ft @ FunctionType(l, r) => FunctionType(go(l, !pol), go(r, pol))(ft.prov)
-                    case tt @ TupleType(fs) => TupleType(fs.mapValues(go(_, pol)))(tt.prov)
-                    case wt @ Without(b, ns) => Without(go(b, pol), ns)(wt.prov)
-                  }, tts, newRcd).toType(sort = true)
+                  val nFields = rcd.fields.filterNot(_._1.name.isCapitalized).mapValues(go(_, pol))
+                  val (res, nfs) = bo match {
+                    case S(tt @ TupleType(fs)) =>
```
LPTK, line 371: The implementation here is too crude. Consider this:

```scala
def test: (string, 1) & { _1: "hello" }
//│ test: (string, 1,) & {_1: "hello", _2: 1}
```

It should instead result in `("hello", 1,)`.

You need to merge all tuple fields into the tuple, and to only leave non-tuple fields in the refinement!

It's okay though. I suggest to open a new issue to improve this later, and merge this PR without the improvement.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -263,7 +267,7 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
       case p: ProxyType => fieldsOf(p.underlying, paramTags)
       case Without(base, ns) => fieldsOf(base, paramTags).filter(ns contains _._1)
       case TypeBounds(lb, ub) => fieldsOf(ub, paramTags)
-      case _: ObjectTag | _: FunctionType | _: TupleType | _: TypeVariable
+      case _: ObjectTag | _: FunctionType | _: ArrayBase | _: TypeVariable
```
LPTK, line 271: Add in the doc of the method: `Note: this does not retrieve the positional fields implicitly defined by tuples`

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1191,6 +1206,7 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
             case ComposedType(false, l, r) => Inter(go(l, polarity), go(r, polarity))
             case RecordType(fs) => Record(fs.map(nt => nt._1 -> go(nt._2, polarity)))
             case TupleType(fs) => Tuple(fs.map(nt => nt._1 -> go(nt._2, polarity)))
+            case ArrayType(inner) => Arr(go(inner, polarity)) // ? not sure
```
LPTK: Yeah that's fine because our array types are covariant. (We may later introduce an invariant mutable variation.)
```suggestion
            case ArrayType(inner) => Arr(go(inner, polarity))
```

shared/src/main/scala/mlscript/TyperDatatypes.scala
```diff
@@ -110,8 +110,22 @@ abstract class TyperDatatypes extends TyperHelpers { self: Typer =>
       if (fields.isEmpty) ExtrType(false)(prov) else RecordType(fields)(prov)
   }
   
-  case class TupleType(fields: List[Opt[Var] -> SimpleType])(val prov: TypeProvenance) extends MiscBaseType {
+  // currently use MiscBaseType
```
LPTK: Why this comment?

shared/src/main/scala/mlscript/TyperDatatypes.scala
```diff
@@ -110,8 +110,22 @@ abstract class TyperDatatypes extends TyperHelpers { self: Typer =>
...
+  sealed abstract class ArrayBase extends MiscBaseType {
+    def inner: SimpleType
+  }
+
+  // "normal" array type to be added in the future
```
LPTK: What is `"normal" array type`?

shared/src/main/scala/mlscript/TyperHelpers.scala
```diff
@@ -68,6 +68,7 @@ abstract class TyperHelpers { self: Typer =>
     case FunctionType(lhs, rhs) => FunctionType(subst(lhs, map), subst(rhs, map))(ts.prov)
     case RecordType(fields) => RecordType(fields.map { case fn -> ft => fn -> subst(ft, map) })(ts.prov)
     case TupleType(fields) => TupleType(fields.map { case fn -> ft => fn -> subst(ft, map) })(ts.prov)
+    case ArrayType(inner) => ArrayType(subst(inner, map))(ts.prov)  // ? also not sure
```
LPTK:
```suggestion
    case ArrayType(inner) => ArrayType(subst(inner, map))(ts.prov)
```

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -196,6 +198,7 @@ trait TermImpl extends StatementImpl { self: Term =>
     case Test(l, r) => "'is' test"
     case With(t, fs) =>  "`with` extension"
     case CaseOf(scrut, cases) =>  "`case` expression" 
+    case Subs(arr, idx) => "array subscription"
```
```suggestion
    case Subs(arr, idx) => "array subscript"
```

LPTK (changes requested):
Very impressive work, thanks!

I have many lower-level comments and change requests, but overall it looks good. Please go through all my comments and solve them.

shared/src/main/scala/mlscript/TypeSimplifier.scala
LPTK: Actually I just added a comment in https://github.com/hkust-taco/mlscript/issues/69#issuecomment-1031281164.

shared/src/main/scala/mlscript/NormalForms.scala
Meowcolm024: I didn't wrote this part, so I'm not sure @_@
But there is a 
``` scala
var r1Final = r1
```
above

shared/src/main/scala/mlscript/NormalForms.scala
Meowcolm024: I'm not quite sure about this part now... @_@

shared/src/main/scala/mlscript/NormalForms.scala
LPTK: Yeah, I wrote it when we were meeting. Obviously I forgot to use `r1Final` down there, otherwise it makes no sense to have this `var r1Final` (that we update in several places) at all since it does not change anything!

```suggestion
        }) map { b => LhsRefined(S(b), ts, r1Final) }
```

shared/src/main/scala/mlscript/NormalForms.scala
```diff
@@ -229,11 +252,21 @@ class NormalForms extends TyperDatatypes { self: Typer =>
             S(Conjunct(
               LhsRefined(S(FunctionType(l1 & l2, r1 | r2)(noProv)), ts, rcdU), vs1, RhsBot, nvs1))
           case (S(TupleType(fs1)), S(TupleType(fs2))) => // TODO Q: records ok here?!
-            if (fs1.size =/= fs2.size) S(Conjunct(LhsRefined(N, ts, rcdU), vs1, RhsBot, nvs1))
+            if (fs1.size =/= fs2.size) S(Conjunct(
+              LhsRefined(S(ArrayType(
+                (fs1 ++ fs2).map(_._2).fold(ExtrType(true)(noProv))(_ | _))(noProv)), ts, rcdU), vs1, RhsBot, nvs1))
```
LPTK:
```suggestion
              LhsRefined(S(ArrayType(tup1.toArray.inner | tup2.toArray.inner)(noProv)), ts, rcdU), vs1, RhsBot, nvs1))
```

shared/src/main/scala/mlscript/NormalForms.scala
```diff
@@ -229,11 +252,21 @@ class NormalForms extends TyperDatatypes { self: Typer =>
...
             else S(Conjunct(
               LhsRefined(S(TupleType(tupleUnion(fs1, fs2))(noProv)), ts, rcdU), vs1, RhsBot, nvs1))
+          case (S(tup @ TupleType(fs)), S(ArrayType(ar))) =>
+            S(Conjunct(
+              LhsRefined(S(ArrayType(tup.toArray | ar)(noProv)), ts, rcdU), vs1, RhsBot, nvs1))
```
LPTK: You forgot `.inner`!! Same below.

shared/src/test/diff/mlscript/TupleArray.mls
```diff
@@ -319,14 +319,14 @@ hg1 ((5,5,5))
 
 def ta1: Array[int] | (int, bool)
 def test: (string, 1) & { _1: "hello" }
-//│ ta1: Array[bool | int]
+//│ ta1: Array[Array[bool | int] | int]
```
LPTK: You should always have a quick look over what diff-test changes your modifications cause. Here you can clearly see that something went wrong.

shared/src/main/scala/mlscript/TypeSimplifier.scala
LPTK: Please feel free to close this conversation when you've read it.

shared/src/main/scala/mlscript/NormalForms.scala
LPTK: This case computes `(tag1 | tag2 | tag3 | ... | Array[in]) | (T1, T2, ...)`, which is the same as `tag1 | tag2 | tag3 | ... | (Array[in] | (T1, T2, ...))` and `(Array[in] | (T1, T2, ...)` should reduce to `Array[in | T1 | T2 | ...]`, no?

```suggestion
      case (RhsBases(ps, S(L(ArrayType(in)))), tup @ TupleType(_)) => this | tup.toArray
```

shared/src/main/scala/mlscript/TyperDatatypes.scala
```diff
@@ -110,8 +110,19 @@ abstract class TyperDatatypes extends TyperHelpers { self: Typer =>
       if (fields.isEmpty) ExtrType(false)(prov) else RecordType(fields)(prov)
   }
   
-  case class TupleType(fields: List[Opt[Var] -> SimpleType])(val prov: TypeProvenance) extends MiscBaseType {
+  sealed abstract class ArrayBase extends MiscBaseType {
+    def inner: SimpleType
+  }
+
+  case class ArrayType(val inner: SimpleType)(val prov: TypeProvenance) extends ArrayBase {
+    lazy val level: Int = inner.level
```
LPTK: This does not need to be a `lazy val` as the computation is trivial and needs not be stored.

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -203,12 +203,13 @@ class ConstraintSolver extends NormalForms { self: Typer =>
                     case _ => reportError
                   }
               }
-            case (LhsRefined(N, ts, r), RhsBases(pts, N | S(L(_: FunctionType | _: TupleType)))) =>
-              reportError
+            case (LhsRefined(N, ts, r), RhsBases(pts, N | S(L(_: FunctionType | _: ArrayBase)))) => reportError
```
LPTK: We try to keep lines below 100 columns.

```suggestion
            case (LhsRefined(N, ts, r), RhsBases(pts, N | S(L(_: FunctionType | _: ArrayBase)))) =>
              reportError
```

LPTK:
@Meowcolm024 Okay, we're almost there! But there are a bunch of old and new unresolved conversations. Please don't forget to mark these as resolved when you're sure you've resolved them. It'll helps us a lot.

shared/src/main/scala/mlscript/Typer.scala
Meowcolm024: I think I would just resolve this conversion :P

shared/src/main/scala/mlscript/NormalForms.scala
Meowcolm024: This seems not possible, as `TopType` is not a subtype of `RecordType`?
I'm trying to do this:
``` scala
def of(bt: BaseType): DNF = DNF.of(LhsRefined(S(bt), SortedSet.empty, bt.toRecord match {
      case r @ RecordType(_) => r
      case _ => RecordType.empty
    }))
```
with:
``` scala
  sealed abstract class BaseType extends BaseTypeOrTag {
    def toRecord: SimpleType = TopType
  }
```
But it still looks strange...

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -203,12 +203,14 @@ class ConstraintSolver extends NormalForms { self: Typer =>
                     case _ => reportError
                   }
               }
-            case (LhsRefined(N, ts, r), RhsBases(pts, N | S(L(_: FunctionType | _: TupleType)))) =>
+            case (LhsRefined(N, ts, r), RhsBases(pts, N | S(L(_: FunctionType | _: ArrayBase)))) =>
```
LPTK:
```suggestion
            case (LhsRefined(N, ts, r), RhsBases(pts, N | S(L(_: FunctionType | _: ArrayBase)))) =>
```

shared/src/main/scala/mlscript/TyperDatatypes.scala
```diff
@@ -110,8 +110,19 @@ abstract class TyperDatatypes extends TyperHelpers { self: Typer =>
       if (fields.isEmpty) ExtrType(false)(prov) else RecordType(fields)(prov)
   }
   
-  case class TupleType(fields: List[Opt[Var] -> SimpleType])(val prov: TypeProvenance) extends MiscBaseType {
+  sealed abstract class ArrayBase extends MiscBaseType {
+    def inner: SimpleType
+  }
+
+  case class ArrayType(val inner: SimpleType)(val prov: TypeProvenance) extends ArrayBase {
+    val level: Int = inner.level
```
LPTK: This should be a `def`. There is no need to store this value.

shared/src/main/scala/mlscript/Typer.scala
LPTK: Did you not understand that change I was proposing? You should try to apply my suggestions and if you're not sure how ask for clarification.

Meowcolm024: Is it adding this comment to the code? @_@ (not sure

shared/src/main/scala/mlscript/NormalForms.scala
LPTK: Right, it's more logical if `toRecord` always returns a `RecordType`, and when the thing is not a record it just returns `RecordType(Nil)`.

shared/src/main/scala/mlscript/Typer.scala
LPTK: I said "in the doc of the method", which means here, after this line: https://github.com/hkust-taco/mlscript/blob/11a2ba2505f05aabfeefff558cd01557c0dc8810/shared/src/main/scala/mlscript/Typer.scala#L258

Meowcolm024: I see

LPTK (approved):
Okay great. I'll squash-merge this now!

PR #80, opened 2022-02-04
Support `this` types (close #20)

### LINKED ISSUE #20: Support "this types"
Would actually be quite straightforward to support. Example:

```scala
class A
  method Foo: Int -> this

class B: A
  method Foo n = console.log n; this

a = B{}
a.Foo 1 // : B

b: A = a
b.Foo 1 // : A
```

chengluyu:
This PR is not done yet. Fix this problem.

```
class Box[T]: { inner: T }
  method Get = this.inner
//│ Defined class Box
//│ Defined Box.Get: (Box['T] & {inner: 'T & 'a}) -> 'a
```

`'a` could be merged with `'T`.


LPTK:
> `'a` could be merged with `'T`.

Actually this is normal. It happens because type simplification does not yet reason about the variance of type parameters. Maybe variance analysis could be your next mini project 😛

LPTK:
As explained in the meeting, there is no additional simplification to do here.

In particular, consider something like:

```scala
def b: Box[string] & { inner: "hello" }
b = Box { inner = "hello" }

b.Get
```

which should be typed as `"hello"`. If we unified `'a` with `T` here, we'd get type `string`.

LPTK:
As discussed in https://github.com/hkust-taco/mlscript/issues/20#issuecomment-1060196682, we need to make `this` rigid. I pushed some changes to start doing that. However, we're still missing a substitution, after typing the method, to remove these rigid `this` types (we can replace them with `anything`). Without it, we're currently failing to type the _uses_ of the method, and we also pretty-print ugly `& this` noise in the signatures.

@chengluyu can you try to implement this missing substitution?

chengluyu:
@LPTK This PR is ready! I added some comments which I will remove after review.

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -492,26 +538,40 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
+            val mthTy = MethodType(bodyTy.level, S((thisTv & td.thisTy(noProv), bodyTy.body)), td2.nme :: Nil, false)(prov)
+            println(s">>> ${td2.nme.name}.${md.nme.name} : ${mthTy.toPT}")
             if (rhs.isRight || !declared.isDefinedAt(nme.name)) {
               if (top) thisCtx.addMth(S(td.nme.name), nme.name, mthTy)
               thisCtx.addMth(N, nme.name, mthTy)
             }
             nme.name -> mthTy
           }
+          // restore `this` in type context
+          thisCtx += "this" -> tr
```
LPTK: What is this for? Why would `this` need to be restored here?

shared/src/test/diff/mlscript/BadClasses.mls
```diff
@@ -0,0 +1,41 @@
...
+ //│ ╔══[ERROR] Type mismatch in field selection:
+ //│ ║  l.6: 	  method Get2 = this.x
+ //│ ║       	                ^^^^^^
+ //│ ╟── reference of type `Box[?T] & this` does not have field 'x'
+ //│ ║  l.6: 	  method Get2 = this.x
+ //│ ╙──     	                ^^^^
+ //│ Defined class Box
+ //│ Defined Box.Get: (Box['T] & this) -> 'T
+ //│ Defined Box.Get2: (Box['T] & this) -> error
+//│ ╔══[ERROR] Type mismatch in field selection:
```
LPTK, line 7: Why is this error duplicated?

shared/src/test/diff/mlscript/Repro.mls
```diff
@@ -0,0 +1,145 @@
+// Preamble
+def plus1 x = x + 1
+//│ plus1: int -> int
+//│      = [Function: plus1]
+
+// Let's try something with type parameters.
```
LPTK: There are no type parameters here...

chengluyu:
I copied the old tests file in 42bc0b6.

LPTK:
As we found in the group meeting, the `this` type variable from inherited definitions is not the same as for the base class, which causes unsoundness. You should do substitution accordingly.

LPTK (changes requested):

LPTK:
Also, make sure to remove the extra comments and debugging info before requesting a review when you are done.

shared/src/main/scala/mlscript/TyperHelpers.scala
```diff
@@ -79,7 +79,9 @@ abstract class TyperHelpers { self: Typer =>
       cache += tv -> tv
       tv
     case tv: TypeVariable => cache.getOrElse(tv, {
-      val v = freshVar(tv.prov)(tv.level)
+      println(s">>> subst($tv) where cache is $cache")
+      // Inherit name hints so I can identify this variables
```
LPTK: I suppose this comment is no longer valid? You should most definitely not use name hints to base the typing logic on.

LPTK (commented):
Please make a clean up pass, removing all extra/outdated comments, then request a review.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -547,16 +548,20 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool)
...
-          case Var(name) if name.isCapitalized && ctx.tyDefs.isDefinedAt(name) => // explicit retrieval
-            ctx.getMth(S(name), fieldName.name) match {
-              case S(mth_ty) => mth_ty.toPT.instantiate
-              case N =>
-                err(msg"Class ${name} has no method ${fieldName.name}", term.toLoc)
-                mthCallOrSel(obj, fieldName)
-            }
-          case _ => mthCallOrSel(obj, fieldName)
-        }
+        trace(s"$lvl. Typing Sel") {
```
LPTK: I don't think this is useful, for example. Remove?

chengluyu: Removed.

shared/src/main/scala/mlscript/TyperHelpers.scala
chengluyu: Removed. But I keep if name hint is "this" for clarity. How do you think?

shared/src/test/diff/mlscript/Repro.mls
chengluyu: Removed.

shared/src/test/diff/mlscript/BadClasses.mls
chengluyu: Caused by ill-indentation. Removed.

shared/src/main/scala/mlscript/TypeDefs.scala
chengluyu: Removed.

chengluyu:
OK. I just merged with main branch and inspected the changed tests. This PR is done!

shared/src/main/scala/mlscript/TyperHelpers.scala
LPTK: Wait, what... we should _always_ keep the name hint! It was a bug that we did not keep it here!

LPTK: So... after this I ended up propagating and introducing name hint in many places! See https://github.com/hkust-taco/mlscript/commit/305e5bcf95898e9b14a4b65b6675a9798cb0abd5.

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -41,9 +41,8 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
         baseClasses.iterator.filterNot(traversed).flatMap(v =>
           ctx.tyDefs.get(v.name).fold(Set.empty[Var])(_.allBaseClasses(ctx)(traversed + v)))
     val (tparams: List[TypeName], targs: List[TypeVariable]) = tparamsargs.unzip
-    def thisTy(prov: TypeProvenance): TypeRef = TypeRef(nme, targs)(prov)
-    def wrapMethod(pt: PolymorphicType, prov: TypeProvenance): MethodType =
-      MethodType(pt.level, S((thisTy(prov), pt.body)), nme :: Nil, isInherited = false)(prov)
+    def thisTy(prov: TypeProvenance): ST = thisTv
+    val thisTv: TypeVariable = freshVar(noProv, S("this"), Nil, TypeRef(nme, targs)(noProv) :: Nil)(1)
```
LPTK: This `thisTy` definition doesn't make any sense...

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -33,7 +33,7 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool)
     * The public helper functions should be preferred for manipulating `mthEnv`
    */
   case class Ctx(parent: Opt[Ctx], env: MutMap[Str, TypeInfo], mthEnv: MutMap[(Str, Str) \/ (Opt[Str], Str), MethodType],
-      lvl: Int, inPattern: Bool, tyDefs: Map[Str, TypeDef]) {
+      lvl: Int, inPattern: Bool, tyDefs: Map[Str, TypeDef]) { // <-- last tyDefs
```
LPTK: ...

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -49,6 +49,7 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool)
     def nest: Ctx = copy(Some(this), MutMap.empty, MutMap.empty)
     def nextLevel: Ctx = copy(lvl = lvl + 1)
     private val abcCache: MutMap[Str, Set[Var]] = MutMap.empty
+    // USE THIS!!!
```
LPTK: You should not make me have to clean after yourself... Please properly review your diff before asking for a review! 😠

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -253,6 +254,11 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool)
               tyTp(App(n, Var("").withLocOf(f)).toCoveringLoc, "extension field")) }
           )(tyTp(r.toLoc, "extension record")))(tyTp(ty.toLoc, "extension type"))
       case Literal(lit) => ClassTag(lit, lit.baseClasses)(tyTp(ty.toLoc, "literal type"))
+      case TypeName("this") =>
+        ctx.env.getOrElse("this", err(msg"undeclared this" -> ty.toLoc :: Nil)) match {
+          case AbstractConstructor(_, _) => ??? // how to handle this case?
```
LPTK: Again... We've even discussed this case in a meeting.

shared/src/main/scala/mlscript/TyperDatatypes.scala
```diff
@@ -58,7 +58,7 @@ abstract class TyperDatatypes extends TyperHelpers { self: Typer =>
   case class MethodType(
     level: Int,
     body: Opt[(SimpleType, SimpleType)],
-    parents: List[TypeName],
+    parents: List[TypeName], // -> multiple this TV
```
LPTK: Come on now...

shared/src/test/diff/mlscript/This.mls
```diff
@@ -0,0 +1,329 @@
...
+//│ Defined Animal.Eat: (Animal & 'this) -> (Animal & 'this)
+
+class Cat: Animal
+//│ Defined class Cat
+
+Animal.Eat
+//│ res: (Animal & 'this) -> (Animal & 'this)
+
+Cat.Eat
+//│ res: (Cat & 'this) -> (Animal & 'this)
```
LPTK: This test file is full of garbage, like stuff copied from other tests that doesn't have anything to do with the `this`-type feature and your old personal comments... I had to go through it and clean it up myself.

LPTK (approved):
Ok, please go ahead and squash-merge this when it's ready.

In the future, please carefully review every line of your diff yourself to avoid wasting my time.

PR #81, opened 2022-02-09
Enhance JavaScript code generation scopes with symbols

chengluyu:
This PR enhances scopes with symbols. The old code base lacks the clarification between lexical names (names used in MLscript) and runtime names (names used in generated JavaScript code), making translating declarations error-prune. Also, there is a need to memorize where symbols are declared and their kinds. For example, we want to know if `error` refers to the built-in `error` or a recent shadowed symbol.

Here is a brief change log.

- We added `ValueSymbol`, `TypeSymbol`, `BuiltinSymbol`, `StubValueSymbol`, `ClassSymbol`, `TraitSymbol` and `ParamSymbol` and related declaration methods in `Scope`.
- We refactored the result of `JSTestBackend`, making it more meaningful.
- Refering to a unimplemented symbol will not cause runtime errors. Instead, the test will output a short message. Also, the unimplemented status is cascadable. See dea229e.
- Provide an escape hatch to access built-in variables provided by Node.js. See a5ba631.


shared/src/test/diff/mlscript/MLTuples.mls
```diff
@@ -34,24 +44,24 @@ trait Hey: { x: int }
...
+//│ ║  l.45: 	Hey t
 //│ ║        	    ^
 //│ ╟── Note: constraint arises from record type:
-//│ ║  l.31: 	trait Hey: { x: int }
+//│ ║  l.41: 	trait Hey: { x: int }
 //│ ╙──      	           ^^^^^^^^^^
 //│ res: (1, 2, 3,) & hey | error
 //│    = [ 1, 2, 3 ]
 
 // TODO: fix `withConstruct` on arrays
```
LPTK: Seems like this is now obsolete.
```suggestion
```

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -5,51 +5,45 @@ import scala.util.matching.Regex
...
+import mlscript.codegen.{CodeGenError, UnimplementedError, Scope, ValueSymbol}
+import mlscript.codegen.ClassSymbol
+import mlscript.codegen.BuiltinSymbol
+import mlscript.codegen.StubValueSymbol
+import mlscript.codegen.TraitSymbol
+import mlscript.codegen.LexicalSymbol
+import mlscript.codegen.FreeSymbol
+import mlscript.codegen.ParamSymbol
+import mlscript.codegen.TypeSymbol
+import scala.collection.mutable.ArrayBuffer
```
```suggestion
import mlscript.codegen._
```

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -69,46 +63,46 @@ class JSBackend {
...
-        JSIdent(jsName)
-      }
-    // TODO: need scope to track variables so that we can rename reserved words
+    case Var(name) => translateVar(scope.resolve(name), false)
     case Lam(params, body) =>
       val patterns = translateParams(params)
       val lamScope = Scope("Lam", patterns flatMap { _.bindings }, scope)
-      JSArrowFn(patterns, lamScope withTempVarDecls translateTerm(body)(lamScope))
+      JSArrowFn(patterns, lamScope.tempVars `with` translateTerm(body)(lamScope))
     // TODO: when scope symbols are ready, rewrite this
```
LPTK: > TODO: when scope symbols are ready, rewrite this

So should this be rewritten now?

shared/src/main/scala/mlscript/codegen/Symbol.scala
```diff
@@ -0,0 +1,208 @@
...
+import mlscript.JSClassDecl
+
+trait RuntimeSymbol {
+  def runtimeName: Str
+}
+
+// trait TypeSymbol {
+//   def baseType: Ls[Str] -> Type
+//   def isParametric: Bool
+// }
```
LPTK:
```suggestion
```

shared/src/main/scala/mlscript/codegen/Symbol.scala
```diff
@@ -0,0 +1,208 @@
...
+    * The lexical name of the symbol. This is different from the runtime name,
+    * the name of the symbol in the generated code. We allow duplicates lexical
+    * names in the same scope.
+    */
+  def lexicalName: Str
+
+  /**
+    * A short name for the declaration used in debugging and error messages.
+    */
+  def shortName: Str
```
LPTK: Seems unnecessary.

shared/src/main/scala/mlscript/codegen/Symbol.scala
```diff
@@ -0,0 +1,208 @@
...
+trait RuntimeSymbol {
+  def runtimeName: Str
+}
+
+// trait TypeSymbol {
+//   def baseType: Ls[Str] -> Type
+//   def isParametric: Bool
+// }
+
+abstract class LexicalSymbol extends RuntimeSymbol {
```
LPTK: This and other base types should probably be `sealed`.

shared/src/main/scala/mlscript/codegen/Symbol.scala
```diff
@@ -0,0 +1,208 @@
...
+final case class TraitSymbol(
+    override val lexicalName: Str,
+    override val runtimeName: Str,
+    override val params: Ls[Str],
+    base: Type
+) extends TypeSymbol(lexicalName, runtimeName, params, base) {
+  override def shortName: Str = s"trait $lexicalName"
+}
+
+final case class ParamSymbol(override val lexicalName: Str, override val runtimeName: Str)
```
LPTK: Seems unnecessary.

shared/src/main/scala/mlscript/codegen/Symbol.scala
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+  * The not-found symbol.
+  */
+final case class FreeSymbol(val lexicalName: Str) extends LexicalSymbol {
+  override def runtimeName: Str = throw new CodeGenError("free symbol has no runtime name")
+  def shortName: Str = s"free $lexicalName"
+}
+
+final case class TemporarySymbol(val runtimeName: Str) extends RuntimeSymbol
```
LPTK: Seems unnecessary.

LPTK (changes requested):
Here's a summary of the change needs we identified during the meeting:

1. Fix eta expansion of classes shadowing a name

    ```scala
    Test = 1
    //│ Test: 1
    //│     = 1

    class Test
    //│ Defined class Test

    :js
    x = Test
    //│ // Query 1
    //│ globalThis.x = Test;
    //│ res = x;
    //│ // End of generated code
    //│ x: anything -> Test
    //│  = 1

    x 1
    //│ res: Test
    //│ Runtime error:
    //│   TypeError: x is not a function
    ```

2. Don't register type symbols as runtime names

3. What's the `location` variable for? Can it be removed?

I'm also going to commit and push the changes we made then.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -1,18 +1,16 @@
 package mlscript
 
-import mlscript.utils._, shorthands._
+import mlscript.codegen.Helpers._
+import mlscript.codegen._
+import mlscript.utils._
```
LPTK: The `import mlscript.utils._, shorthands._` pattern is everywhere in the code base. No reason to make it different here. It sort if implicitly emphasizes that the two imports are related.

chengluyu: Fixed in eb9dd5f.

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: It is a tiny optimization which turns built-in functions like `add 1 2` into `1 + 2`. I think it is not necessary now. So I just removed the TODO here.

chengluyu:
OK. I think I finished the cleanup.

shared/src/main/scala/mlscript/codegen/Helpers.scala
```diff
@@ -0,0 +1,67 @@
...
+        case NoCases        => "NoCases"
+      }
+      s"CaseOf(${inspect(trm)}, ${inspectCaseBranches(cases)}"
+    case IntLit(value)  => s"IntLit($value)"
+    case DecLit(value)  => s"DecLit($value)"
+    case StrLit(value)  => s"StrLit($value)"
+    case Subs(arr, idx) => s"Subs(${inspect(arr)}, ${inspect(idx)})"
+  }
+
+  def topologicallySort[A: Ordering](relations: Ls[A -> A]): Iterable[A] = {
```
LPTK:
```suggestion
  def topologicalSort[A: Ordering](relations: Ls[A -> A]): Iterable[A] = {
```

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -243,287 +218,213 @@ class JSBackend {
...
-      case TypeDef(Als, TypeName(name), tparams, body, _, _) =>
-        val tnames = tparams map { case TypeName(nme) => nme }
-        typeAliasMap(name) = tnames -> body
-      case TypeDef(Trt, TypeName(name), _, _, _, _) =>
-        traitNames += name
-      case _ => ()
+    val classSymbols = declareTypeDefs(typeDefs)
+    val (isolatedClasses, relations) = resolveInheritance(classSymbols)
+    val sortedClasses = try topologicallySort(relations) catch {
+      case e: RuntimeException => throw CodeGenError("cyclic inheritance involving")
```
LPTK: The error message is incomplete! Just `"cyclic inheritance"` would actually be fine. We'll get a better error through the type anyway.

shared/src/main/scala/mlscript/codegen/Helpers.scala
LPTK: Also, this general-purpose function should be moved to `shared/src/main/scala/mlscript/utils`, probably in a new file.

shared/src/main/scala/mlscript/JSBackend.scala
LPTK: > `case e: RuntimeException `

This is disgusting. It will swallow _any_ error and make it pass for a cyclic inheritance, even if it's another source or error. The better thing to do would be to throw a proper catchable exception in the topological sort function.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -243,287 +218,213 @@ class JSBackend {
...
+      case e: UnimplementedError => JSTestBackend.Unimplemented(e.getMessage())
+      case e: Throwable => JSTestBackend.UnexpectedCrash(e.getMessage())
+    }
 
-  // Generate code for test.
-  private def generate(pgrm: Pgrm): TestCode = {
+  /**
+    * Generate JavaScript code which targets MLscript test from the given block.
+    */
+  private def generate(pgrm: Pgrm)(implicit scope: Scope, allowEscape: Bool): JSTestBackend.TestCode = {
```
LPTK, line 391: What does `allowEscape` do?! Document it?

shared/src/main/scala/mlscript/codegen/Symbol.scala
```diff
@@ -0,0 +1,168 @@
...
+
+  /**
+    * Fill up this field after the class translation.
+    */
+  var body: Opt[JSClassDecl] = N
+
+  /**
+    * Fill up this field after resolving classes.
+    */
+  var baseClass: Opt[ClassSymbol] = N
```
LPTK: Can you explain why `baseClass` and `body` need to be mutable variables? Why not have them as immutable parameters to the class?

shared/src/test/diff/codegen/TrickyShadowing.mls
```diff
@@ -0,0 +1,240 @@
...
+:js
+class add
+//│ ╔══[ERROR] Type names must start with a capital letter
+//│ ║  l.27: 	class add
+//│ ╙──      	      ^^^
+//│ // Prelude
+//│ class add1 {}
+//│ // End of generated code
+
+// FIXME
```
LPTK:
```suggestion
```

shared/src/test/diff/codegen/TrickyShadowing.mls
```diff
@@ -0,0 +1,240 @@
...
+//│ // Query 1
+//│ globalThis.x = function x(x) {
+//│   return new Test1(x);
+//│ };
+//│ res = x;
+//│ // End of generated code
+//│ x: anything -> Test
+//│  = [Function: x]
+
+// FIXME
```
LPTK:
```suggestion
```

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -204,8 +177,10 @@ class JSBackend {
     case Bra(_, trm) => translateTerm(trm)
     case Tup(terms) =>
       JSArray(terms map { case (_, term) => translateTerm(term) })
-    case _: Bind | _: Test =>
-      throw CodeGenError(s"cannot generate code for term ${JSBackend.inspectTerm(term)}")
+    case Subs(arr, idx) =>
+      JSMember(translateTerm(arr), translateTerm(idx))
+    case App(_, _) | _: Bind | _: Test =>
+      throw CodeGenError(s"cannot generate code for term ${inspect(term)}")
     case _ => die // FIXME "Exhaustivity analysis reached max recursion depth, not all missing cases are reported."
```
LPTK: Let's try to fix this in this PR. Try with deeper max depth, and if it does not work maybe try to refactor a bit? Such as moving complicated patterns like `case App(App(App(Var("if"), tst), con), alt)` into the `App` case.

LPTK (commented):
Otherwise LGTM! But before merging I'd like to make sure we really can't avoid that last bit of symbol mutability.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -415,14 +415,19 @@ class JSTestBackend extends JSBackend {
 
   /**
     * Generate JavaScript code which targets MLscript test from the given block.
+    *
+    * @param pgrm the program to translate
+    * @param scope the top-level scope
+    * @param allowEscape whether allow unimplemented values can be accessible
```
```suggestion
    * @param allowEscape whether to try executing code even if it refers to unimplemented definitions
```

shared/src/main/scala/mlscript/codegen/Symbol.scala
chengluyu: They were removed!

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: Now we throw and catch `CyclicGraphError`.

chengluyu: Fixed in f499c7d.

shared/src/main/scala/mlscript/codegen/Helpers.scala
chengluyu: Fixed in f499c7d.

shared/src/main/scala/mlscript/utils/algorithms.scala
```diff
@@ -17,8 +24,8 @@ object algorithms {
         sort(SortedMap.from(hasPreds.view.mapValues(_ -- found)), done ++ found)
       }
     }
-    val toPred = relations.foldLeft(SortedMap[A, Set[A]]()) { (acc, e) => 
-      acc + (e._1 -> acc.getOrElse(e._1, Set())) + (e._2 -> (acc.getOrElse(e._2, Set()) + e._1))
+    val toPred = edges.foldLeft(SortedMap.from(nodes.map { _ -> Set[A]() })) { (acc, e) =>
```
LPTK: Faster:
```suggestion
    val toPred = edges.foldLeft(SortedMap.from(nodes.map { _ -> Set.empty[A] })) { (acc, e) => 
```

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: I extract all cases of `App` into a new method. So exhaustivity check in `translateTerm` becomes easier.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -407,123 +244,157 @@ class JSBackend {
...
+    // Cache base classes for class symbols.
+    // Note: it may include previously defined class symbols.
+    val baseClasses = Map.from(classSymbols.iterator.flatMap { derivedClass =>
+      resolveBaseClass(derivedClass.actualType).map(derivedClass -> _)
+    })
+    val sorted = try topologicalSort(baseClasses, classSymbols) catch {
+      case e: CyclicGraphError => throw CodeGenError("cyclic inheritance detected")
+    }
+    // We only need class symbols declared in current translation unit.
+    sorted.flatMap(sym => if (classSymbols.contains(sym)) S(sym -> baseClasses.get(sym)) else N)
```
chengluyu: Here we resolve the base class for each newly-defined classes and put them in a `Map`. The information is used immediately in `topologicalSort` (this is **the first consumer**). Note that `sortClassSymbols` returns all newly-defined classes with their base classes (`Iterable[ClassSymbol, Opt[ClassSymbol]`).

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -407,123 +244,157 @@ class JSBackend {
...
-            nameClsMap += name -> cls
-            S(cls)
-          case Trt => N
-          case Als => N
-        }
-      }
+    val classSymbols = declareTypeDefs(typeDefs)
+    val defStmts = sortClassSymbols(classSymbols).map { case (derived, base) =>
+      translateClassDeclaration(derived, base)(topLevelScope)
+    }.toList
```
chengluyu, line 397: After we got all sorted newly-defined classes (together with their base classes!), we pass them to `translateClassDeclaration`, which is **the second consumer** of base classes.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -407,123 +244,157 @@ class JSBackend {
...
-  // Name of the helper function for `with` construction.
-  private val withConstructFnName = topLevelScope allocateJavaScriptName "withConstruct"
+  /**
+    * Resolve inheritance of all declared classes.
+    * 
+    * @return sorted class symbols with their base classes
+    */
+  protected def sortClassSymbols(classSymbols: Ls[ClassSymbol]): Iterable[(ClassSymbol, Opt[ClassSymbol])] = {
+    // Cache base classes for class symbols.
+    // Note: it may include previously defined class symbols.
```
LPTK: What is `it`? What is this comment trying to say?

shared/src/main/scala/mlscript/codegen/Symbol.scala
```diff
@@ -0,0 +1,158 @@
...
+  def lexicalName: Str
+}
+
+sealed trait RuntimeSymbol extends LexicalSymbol {
+  def runtimeName: Str
+}
+
+sealed trait TypeSymbol extends LexicalSymbol {
+  val params: Ls[Str]
+  val actualType: Type
```
LPTK: These two params seem unused. Remove them? Or justify why we'll need them in the future.

shared/src/main/scala/mlscript/codegen/Symbol.scala
```diff
@@ -0,0 +1,158 @@
...
+)(implicit val accessible: Bool)
+    extends RuntimeSymbol {
+  override def toString: Str = s"value $lexicalName"
+}
+
+final case class ClassSymbol(
+    lexicalName: Str,
+    runtimeName: Str,
+    params: Ls[Str],
+    actualType: Type,
```
LPTK: Why is the body type called the "actual" type? Very strange and nonstandard terminology. Please change.

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -382,17 +434,22 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
...
             results match {
-              case L((message, crashed)) =>
-                if (crashed)
-                  output("Code generation crashed:")
-                else
-                  output("Code generation met an error:")
-                output(s"  $message")
+              case IllFormedCode(message) =>
+                totalRuntimeErrors += 1
+                output("Code generation met an error:")
```
LPTK: `met` is a little strange here
```suggestion
                output("Code generation encountered an error:")
```

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -382,17 +434,22 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
...
+                output(s"  ${message}")
+              case Unimplemented(message) =>
+                output("Unable to execute the code:")
+                output(s"  ${message}")
+              case UnexpectedCrash(message) =>
```
LPTK: This message should include the name of the error. For example try making it crash with a `MatchError` — it won't even say it's a match error.

LPTK (approved):
Okay I think this is finally in a mergeable state! Apply my remaining comments and then feel free to squash-merge it.

shared/src/main/scala/mlscript/JSBackend.scala
LPTK: Extracting into a different method seems unnecessary, but okay I guess.

shared/src/main/scala/mlscript/codegen/Symbol.scala
chengluyu: It was because there had been another member named `body`. Changed in `52324df`.

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: Fixed in 918dee7.

shared/src/test/scala/mlscript/DiffTests.scala
chengluyu: Fixed in 22ab793.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -407,123 +244,157 @@ class JSBackend {
...
+  protected def sortClassSymbols(classSymbols: Ls[ClassSymbol]): Iterable[(ClassSymbol, Opt[ClassSymbol])] = {
+    // Cache base classes for class symbols.
+    val baseClasses = Map.from(classSymbols.iterator.flatMap { derivedClass =>
+      resolveBaseClass(derivedClass.body).map(derivedClass -> _)
+    })
+    val sorted = try topologicalSort(baseClasses, classSymbols) catch {
+      case e: CyclicGraphError => throw CodeGenError("cyclic inheritance detected")
+    }
+    // Their base classes might be class symbols defined in previous translation
+    // units. So we filter them here.
```
```suggestion
    // units. So we filter them out here.
```

shared/src/main/scala/mlscript/codegen/Symbol.scala
chengluyu: I noticed that @twitu is using two scopes (one for types and another for terms) in TypeScript code generation. So preserving `TypeSymbol` and related information could simplify the workflow there.

LPTK: Ok, though he's not using symbols yet because they weren't really available. We'll look into using that info.

chengluyu: OK. So I'm keeping it. We can remove them if it turns out to be useless.

PR #82, opened 2022-02-12
Create TypingUnit class for helper methods that find and return type information

twitu:


LPTK:
As discussed in the meeting, this seems like a useful change, but it should probably leave `JSBackend` alone as that file has been changed completely by @chengluyu. And let's rename this new class from `ClassInfo` to `TypingUnit` and put it in its own file.

shared/src/main/scala/mlscript/TypingUnit.scala
```diff
@@ -0,0 +1,155 @@
...
+    case Rem(_, _) | TypeVar(_, _) | Literal(_) | Recursive(_, _) | Bot | Top |
+        Tuple(_) | Neg(_) | Bounds(_, _) | WithExtension(_, _) |
+        Function(_, _) | Union(_, _) | Arr(_) =>
+      throw TypingUnitError(s"unable to derive from type $ty")
+  }
+
+  // This function normalizes a type, removing all `AppliedType`s.
+  private def substitute(
+      body: Type,
+      subs: collection.immutable.HashMap[Str, Type] =
```
LPTK, line 96: `collection.immutable.HashMap` can just be `Map`

shared/src/main/scala/mlscript/TypingUnit.scala
```diff
@@ -0,0 +1,155 @@
...
+      if (traitNames contains base)
+        Nil -> N
+      else
+        getBaseClassAndFields(applyTypeAlias(base, targs map { substitute(_) }))
+    // There is some other possibilities such as `class Fun[A]: A -> A`.
+    // But it is not achievable in JavaScript.
+    case Rem(_, _) | TypeVar(_, _) | Literal(_) | Recursive(_, _) | Bot | Top |
+        Tuple(_) | Neg(_) | Bounds(_, _) | WithExtension(_, _) |
+        Function(_, _) | Union(_, _) | Arr(_) =>
+      throw TypingUnitError(s"unable to derive from type $ty")
```
LPTK, line 90: I know this is the error message @chengluyu wrote, but it's very poor and should be improved. What "cannot derive"?

shared/src/main/scala/mlscript/TypingUnit.scala
```diff
@@ -0,0 +1,155 @@
+package mlscript
+
+import mlscript.utils._, shorthands._
+import scala.collection.immutable.{Map, Set, HashSet}
+
+final case class TypingUnitError(message: String) extends Exception(message)
```
LPTK, line 6: I'd rather call that "`IllFormedCodeError`" (and reuse it in more places, though for now it's okay if it's only used here).

LPTK (commented):
I don't actually agree with the logic. Why would `traitNames` be handled separately? Why are trait fields not aggregated? And where is the info on class and trait _bodies_? These are not type aliases. So shouldn't `typeAliasMap` be `typeDefBodies` instead?

LPTK:
Also, can we make these changes as part of the TS code-gen for classes? It feels weird to make a PR for some definitions that are never used nor tested at all.

You can just change the PR's name and make the TS code-gen for classes changes here I guess.

twitu:
I'll open a separate PR for this because a github PR is tied to a branch name and this name would not match the typegen functionality added to it.

I'll keep the above points in mind when making the changes.

PR #83, opened 2022-02-15
Add mutable fields

### LINKED ISSUE #8: Add mutable assignment
The scheme is to use an invariant `Ref` type (or call it `Mut`) everywhere, where we have `Ref[T] <: T`, and with a `Ref: 'a -> Ref['a]` constructor. For example we have `{ x = Ref 1 } : { x: Ref[0 | 1] } <: { x: int }`

The advantage over explicitly making record fields optionally mutable is to make the implementation in MLscript easy and modular. But we need to put a restriction on where these `Ref` types can appear. The restriction could be checked in a pass after type checking:
**_One can only call the `Ref` constructor as part of a field initialization, as in `{ x = Ref 1 }` or as part of a local variable initialization, as in `let x = Ref 1 in ...`._**

The code generator will simply ignore these Ref constructors, as in JS all fields are mutable by default.

LPTK:
Closes #8.


shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -414,6 +414,7 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
                           val fv = freshVar(noProv,
                             S(f._1.name.drop(f._1.name.indexOf('#') + 1)) // strip any "...#" prefix
                           )(1).tap(_.upperBounds ::= f._2.ub)
+                          // ? not sure if we can do this
```
LPTK: This is a bit funky. It may work in most cases but that's not really the logic we want. Here, we should decide based on whether the field was declared mutable or not. I think you should do what I suggested on Discord:

> it may be better to make the LB of FieldType an Option, and if it's None it means it's immutable

Then things will be a lot cleaner, including here.

This can also give better errors.

Meowcolm024: So it means that we're using `None` for place of `BotType` originally(for immutable ones), and use `Some(T)` for the mutable ones? @_@

LPTK: Yes, exactly!

Meowcolm024: But it seems that the lower bound would always equals to the upper bound for mutable fields, so what is the difference between this way and simply using a flag @_@
Or is there scenario where the lower bound is different?

LPTK: The problem is that if we used a single bound, it would be invariant, which would make other parts of the compiler awkward because they currently assume that everything is either covariant or contravariant. For example, merging two mutable tuples in type simplification would require making the two types equal by constraining, but currently no constraining occurs in simplification. It's more of a technical detail, really.

shared/src/main/scala/mlscript/TypeSimplifier.scala
```diff
@@ -385,13 +388,14 @@ trait TypeSimplifier { self: Typer =>
...
                       val componentFieldsMap = componentFields.toMap
                       val tupleComponents = fs.iterator.zipWithIndex.map { case ((nme, ty), i) =>
-                        nme -> go(ty & componentFieldsMap.getOrElse(i + 1, TopType), pol)
+                        nme -> (ty && componentFieldsMap.getOrElse(i + 1, TopType.toUpper(noProv))).update(go(_, !pol), go(_, pol))
                       }.toList
-                      S(TupleType(tupleComponents)(tt.prov)) -> rcdFields.mapValues(go(_, pol))
+                      S(TupleType(tupleComponents)(tt.prov)) -> rcdFields.mapValues(_.update(go(_, !pol), go(_, pol)))
                     case S(ct: ClassTag) => S(ct) -> nFields
-                    case S(ft @ FunctionType(l, r)) =>
+                    case S(ft @ FunctionType(l, r)) =>
```
LPTK:
```suggestion
                    case S(ft @ FunctionType(l, r)) =>
```

shared/src/main/scala/mlscript/TyperDatatypes.scala
```diff
@@ -253,12 +254,38 @@ abstract class TyperDatatypes extends TyperHelpers { self: Typer =>
   case class TypeBounds(lb: SimpleType, ub: SimpleType)(val prov: TypeProvenance) extends SimpleType {
     def level: Int = lb.level max ub.level
     override def toString = s"$lb..$ub"
+    def && (that: TypeBounds, prov: TypeProvenance = noProv): TypeBounds =
```
LPTK: Question: is this actually still used?

shared/src/main/scala/mlscript/TypeSimplifier.scala
LPTK: IS this resolved? If so, please mark it so.

Meowcolm024: yep

LPTK: > If so, please mark it so. (resolved)

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -203,6 +204,7 @@ class JSBackend {
       JSArray(terms map { case (_, (term, _)) => translateTerm(term) })
     case Subs(arr, idx) =>
       JSMember(translateTerm(arr), translateTerm(idx))
+    case Assign(field, value) => ???  // TODO
```
LPTK: I think it should be easy to implement this case while we're at it. You just need to generate a JS assignment expression.

shared/src/main/scala/mlscript/TyperDatatypes.scala
Meowcolm024: Seems not used from code lens, both`&&` and `||`

LPTK: Okay, I think we can remove it then.

shared/src/main/scala/mlscript/JSBackend.scala
Meowcolm024: I find that I also need to return a `unit`, but I don't know how @_@

Meowcolm024: Okay, I think I've find a way

LPTK: You can return an empty array as unit, which is the current interpretation:

```scala
()
//│ res: ()
//│    = []
```

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -51,7 +51,7 @@ class JSBackend {
     case Bra(_, trm) => translatePattern(trm)
     case Tup(fields) => JSArrayPattern(fields map { case (_, (t, _)) => translatePattern(t) })
     // Others are not supported yet.
-    case _: Lam | _: App | _: Sel | _: Let | _: Blk | _: Bind
+    case _: Lam  | _: App  | _: Sel    | _: Let  | _: Blk    | _: Bind
```
LPTK: Please don't... Horizontal alignment is a sin except in a few exceptional cases (such as aligning `case ... =>`).

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -204,11 +204,8 @@ class JSBackend {
       JSArray(terms map { case (_, (term, _)) => translateTerm(term) })
     case Subs(arr, idx) =>
       JSMember(translateTerm(arr), translateTerm(idx))
-    case Assign(field, value) => JSImmEvalFn(N, Nil, Left(
-      JSCommaExpr(
-        JSAssignExpr(translateTerm(field), translateTerm(value)) :: 
-        JSArray(Nil) :: Nil   // return unit
-      )), Nil)
+    case Assign(field, value) => JSImmEvalFn(N, JSNamePattern("_") :: Nil, Left(JSArray(Nil)), 
+      JSAssignExpr(translateTerm(field), translateTerm(value)) :: Nil)
```
LPTK: So why not use a comma expression here? As in `(x = a, [])`.

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -133,7 +133,7 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
   
   
   def processTypeDefs(newDefs0: List[mlscript.TypeDef])(implicit ctx: Ctx, raise: Raise): Ctx = {
-    
+
```
LPTK: Can you please _not_ remove the indentation space? Please revert all these indentation deletions, they're unnecessary changes.

LPTK (commented):
Nice, LGTM. I thin it's almost ready to merge except for my three comments below.

shared/src/main/scala/mlscript/JSBackend.scala
Meowcolm024: I'm not sure if using a comma is suitable @_@, I wonder if there is a semicolon that I can use

LPTK: AFAIK that's how JS does sequencing of effectful expressions in an expression. JS does not have first-class `;` nor expression blocks like Scala.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comma_Operator

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -194,15 +194,18 @@ class JSBackend {
...
           name -> translateTerm(value)
         }) :: Nil
       )
     case Bra(_, trm) => translateTerm(trm)
     case Tup(terms) =>
-      JSArray(terms map { case (_, term) => translateTerm(term) })
+      JSArray(terms map { case (_, (term, _)) => translateTerm(term) })
     case Subs(arr, idx) =>
       JSMember(translateTerm(arr), translateTerm(idx))
+    case Assign(field, value) => JSImmEvalFn(N, Nil,
```
LPTK: But what is the point of wrapping thins in a function?

Meowcolm024: It seems that there's no need @_@

LPTK:
Okay, merging now!

PR #84, opened 2022-02-22
JavaScript code generation for traits

chengluyu:
This PR is WIP. I'll write a summary when it's ready.


shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -223,6 +226,82 @@ class JSBackend {
...
+          )
+      ) :: Nil
+    )
+    const(
+      traitSymbol.lexicalName,
+      JSFuncExpr(
+        N,
+        Nil,
+        Ls(
+          const("tag", id("Object")()),
```
chengluyu:
```suggestion
          const("tag", id("Symbol")()),
```

LPTK: Didn't you want to use the `Symbol` as a key?

chengluyu: Yes! I fixed it!

chengluyu:
We probably need to clean up `JSBackend.translateTraitDeclaration`.

LPTK (approved):
LGTM!

Do you have a test case that shows matching on a trait constructed on a primitive value?

As in:

```
t1 = T1 1
//│ t1: 1 & t1
//│   = [Number: 1]

case t1 of { T1 -> () }
//│ res: ()
//│    = []
```

You can go ahead and squash-merge.

PR #88, opened 2022-03-02
Refactor `MethodType`

fo5for:
In the past, method body types were wrapped in a `FunctionType` with the implicit `this` parameter after type checking is completed. This resulted in some ambiguity as the `MethodType`s within `Typer.processTypeDefs.typeMethods` are not wrapped, while the ones registered to the `Ctx` are wrapped, but with no way to distinguish them.

This PR moves the wrapping logic to `MethodType`, requiring a pair of the `this` parameter and and the actual body type, and wrapping them in `toPT`. A new `bodyPT` val stores the unwrapped body type.

~~Idk why but somehow saved 4 constrain calls~~

### DIFF
--- a/shared/src/main/scala/mlscript/Typer.scala
+++ b/shared/src/main/scala/mlscript/Typer.scala
@@ -37,12 +37,9 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
         baseClasses.iterator.filterNot(traversed).flatMap(v =>
           ctx.tyDefs.get(v.name).fold(Set.empty[Var])(_.allBaseClasses(ctx)(traversed + v)))
     val (tparams: List[TypeName], targs: List[TypeVariable]) = tparamsargs.unzip
-    def wrapMethod(pt: PolymorphicType, prov: TypeProvenance): MethodType = {
-      val thisTy = TypeRef(nme, targs)(prov)
-      MethodType(pt.level, S(FunctionType(singleTup(thisTy), pt.body)(prov)), nme)(prov)
-    }
-    def wrapMethod(mt: MethodType): MethodType =
-      if (mt.body.nonEmpty) wrapMethod(mt.toPT, mt.prov) else mt.copy(parents = nme :: Nil)
+    def thisTy(implicit prov: TypeProvenance): TypeRef = TypeRef(nme, targs)(prov)
+    def wrapMethod(pt: PolymorphicType, prov: TypeProvenance): MethodType =
+      MethodType(pt.level, S((thisTy(prov), pt.body)), nme)(prov)
   }
   
   private case class MethodDefs(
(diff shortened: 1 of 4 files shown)


shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -37,12 +37,9 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
         baseClasses.iterator.filterNot(traversed).flatMap(v =>
           ctx.tyDefs.get(v.name).fold(Set.empty[Var])(_.allBaseClasses(ctx)(traversed + v)))
     val (tparams: List[TypeName], targs: List[TypeVariable]) = tparamsargs.unzip
-    def wrapMethod(pt: PolymorphicType, prov: TypeProvenance): MethodType = {
-      val thisTy = TypeRef(nme, targs)(prov)
-      MethodType(pt.level, S(FunctionType(singleTup(thisTy), pt.body)(prov)), nme)(prov)
-    }
-    def wrapMethod(mt: MethodType): MethodType =
-      if (mt.body.nonEmpty) wrapMethod(mt.toPT, mt.prov) else mt.copy(parents = nme :: Nil)
+    def thisTy(implicit prov: TypeProvenance): TypeRef = TypeRef(nme, targs)(prov)
```
LPTK, line 40: The parameter should not be implicit – it's only passed explicitly! Try to refrain from marking parameters as implicit unless there's a tangible benefit to it. It causes more potential harm and confusion.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -68,9 +65,13 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
-    private def addParent(prt: TypeName): MethodDefs =
-      copy(decls = decls.view.mapValues(mt => mt.copy(parents = prt :: mt.parents)).toMap)
-    def propagate(top: Bool = true)(implicit raise: Raise): MethodDefs =
+    private def addParent(prt: TypeName)(implicit ctx: Ctx): MethodDefs = {
+      val td = ctx.tyDefs(prt.name)
+      def addThis(mt: MethodType): MethodType = mt.copy(body = mt.body.map(b => b.copy(_1 = td.thisTy(mt.prov))))
+      def add(mt: MethodType): MethodType = mt.copy(parents = prt :: mt.parents)
+      copy(decls = decls.view.mapValues(addThis).mapValues(add).toMap, defns = defns.view.mapValues(addThis).toMap)
+    }
+    def propagate(top: Bool = true)(implicit ctx: Ctx, raise: Raise): MethodDefs =
```
LPTK, line 74: The method is only used twice. Once with `top = true` and once with `top = false`. Using a default argument value make little sense here and just makes the logic unnecessarily error-prone.

shared/src/main/scala/mlscript/TyperDatatypes.scala
```diff
@@ -39,32 +39,39 @@ abstract class TyperDatatypes extends TyperHelpers { self: Typer =>
     def rigidify(implicit lvl: Int): SimpleType = freshenAbove(level, body, rigidify = true)
   }
   
-  // single: whether the method declaration comes from a single class, and not the intersection of multiple inherited declarations
-  class MethodType(val level: Int, val body: Opt[SimpleType], val parents: List[TypeName], val single: Bool)
+  /** body._1: implicit `this` parameter
+   *  body._2: actual body of the method
+   *  single: whether the method declaration comes from a single class, and not the intersection of multiple inherited declarations
```
LPTK, line 44: Why not name this `inherited` (and `s/single/!inherited`)? Good names are important.

LPTK (approved):
Okay, I realize there's _a lot_ of awkward and confusing logic in there, which I had missed as I reviewed this part of the code too superficially.

Let's merge this as is, though. You can just mark my comments as resolved after reading them. I have started making the changes myself locally, and will make another PR later today.

PR #89, opened 2022-03-04
Do a serious cleaning/revamp of the logic, naming, and documentation of method typing

LPTK:
@fo5for please take a look at this to validate the changes, and in particular go through the new documentation. This is the minimum standard of documentation I would expect for such nontrivial (borderline wicked) logic, otherwise there's no way for someone who hasn't recently worked on it to understand what's going on (including the original author 😛).

### DIFF
--- a/shared/src/main/scala/mlscript/Typer.scala
+++ b/shared/src/main/scala/mlscript/Typer.scala
@@ -37,52 +37,66 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
         baseClasses.iterator.filterNot(traversed).flatMap(v =>
           ctx.tyDefs.get(v.name).fold(Set.empty[Var])(_.allBaseClasses(ctx)(traversed + v)))
     val (tparams: List[TypeName], targs: List[TypeVariable]) = tparamsargs.unzip
-    def thisTy(implicit prov: TypeProvenance): TypeRef = TypeRef(nme, targs)(prov)
+    def thisTy(prov: TypeProvenance): TypeRef = TypeRef(nme, targs)(prov)
     def wrapMethod(pt: PolymorphicType, prov: TypeProvenance): MethodType =
-      MethodType(pt.level, S((thisTy(prov), pt.body)), nme)(prov)
+      MethodType(pt.level, S((thisTy(prov), pt.body)), nme :: Nil, isInherited = false)(prov)
   }
   
-  private case class MethodDefs(
-    tn: TypeName,
-    parents: List[MethodDefs],
+  /** Represent a set of methods belonging to some owner type.
+    * This includes explicitly declared/defined as well as inherited methods. */
+  private case class MethodSet(
+    ownerType: TypeName,
+    parents: List[MethodSet],
     decls: Map[Str, MethodType],
     defns: Map[Str, MethodType],
   ) {
-    private def &(that: MethodDefs)(tn: TypeName, newdefns: Set[Str])(implicit raise: Raise): MethodDefs =
-      MethodDefs(
+    private def &(that: MethodSet)(tn: TypeName, overridden: Set[Str])(implicit raise: Raise): MethodSet =
+      MethodSet(
         tn,
         this.parents ::: that.parents,
         mergeMap(this.decls, that.decls)(_ & _),
         (this.defns.iterator ++ that.defns.iterator).toSeq.groupMap(_._1)(_._2).flatMap {
           case _ -> Nil => die
           case mn -> (defn :: Nil) => S(mn -> defn)
-          case mn -> defns if newdefns(mn) => N
+          case mn -> defns if overridden(mn) => N  // ignore a previously-defined method if it's overridden
           case mn -> defns =>
             err(msg"An overriding method definition must be given when inheriting from multiple method definitions" -> tn.toLoc
               :: msg"Definitions of method $mn inherited from:" -> N
               :: defns.iterator.map(mt => msg"• ${mt.parents.head}" -> mt.prov.loco).toList)
-            S(mn -> MethodType(defns.head.level, N, tn :: Nil)(
+            S(mn -> MethodType(defns.head.level, N, tn :: Nil, isInherited = false)(
               TypeProvenance(tn.toLoc, "inherited method declaration")))
         }
       )
-    private def addParent(prt: TypeName)(implicit ctx: Ctx): MethodDefs = {
-      val td = ctx.tyDefs(prt.name)
-      def addThis(mt: MethodType): MethodType = mt.copy(body = mt.body.map(b => b.copy(_1 = td.thisTy(mt.prov))))
-      def add(mt: MethodType): MethodType = mt.copy(parents = prt :: mt.parents)
-      copy(decls = decls.view.mapValues(addThis).mapValues(add).toMap, defns = defns.view.mapValues(addThis).toMap)
+    /** Returns a `MethodSet` of the _inherited_ methods only,
+      *   disregarding the current MethodSet's methods... */
+    def processInheritedMethods(top: Bool)(implicit ctx: Ctx, raise: Raise): MethodSet = {
+      def addParent(mt: MethodSet, prt: TypeName): MethodSet = {
+        val td = ctx.tyDefs(prt.name)
+        def addThis(mt: MethodType): MethodType =
+          mt.copy(body = mt.body.map(b => b.copy(_1 = td.thisTy(mt.prov))))(mt.prov)
+        def add(mt: MethodType): MethodType =
+          mt.copy(parents = prt :: mt.parents)(mt.prov)
+        mt.copy(decls = mt.decls.view.mapValues(addThis).mapValues(add).toMap, defns = mt.defns.view.mapValues(addThis).toMap)
+      }
+      parents.map(_.processInheritedMethods(false))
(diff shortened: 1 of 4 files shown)


shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -37,52 +37,66 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+    private def &(that: MethodSet)(tn: TypeName, overridden: Set[Str])(implicit raise: Raise): MethodSet =
+      MethodSet(
         tn,
         this.parents ::: that.parents,
         mergeMap(this.decls, that.decls)(_ & _),
         (this.defns.iterator ++ that.defns.iterator).toSeq.groupMap(_._1)(_._2).flatMap {
           case _ -> Nil => die
           case mn -> (defn :: Nil) => S(mn -> defn)
-          case mn -> defns if newdefns(mn) => N
+          case mn -> defns if overridden(mn) => N  // ignore a previously-defined method if it's overridden
```
fo5for: This is actually the inherited methods definitions.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -37,52 +37,66 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
               TypeProvenance(tn.toLoc, "inherited method declaration")))
         }
       )
-    private def addParent(prt: TypeName)(implicit ctx: Ctx): MethodDefs = {
-      val td = ctx.tyDefs(prt.name)
-      def addThis(mt: MethodType): MethodType = mt.copy(body = mt.body.map(b => b.copy(_1 = td.thisTy(mt.prov))))
-      def add(mt: MethodType): MethodType = mt.copy(parents = prt :: mt.parents)
-      copy(decls = decls.view.mapValues(addThis).mapValues(add).toMap, defns = defns.view.mapValues(addThis).toMap)
+    /** Returns a `MethodSet` of the _inherited_ methods only,
+      *   disregarding the current MethodSet's methods... */
```
fo5for: You can also call with `top = false` to get the full set of methods (i.e. where the current methods override and inherited ones). It's just never used anywhere since we need to consider them separately for subsumption checking.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -37,52 +37,66 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
   /** Keys of `mthEnv`:
-   *  `L` represents the inferred types of method definitions. The first value is the parent name, and the second value is the method name.
-   *    The method definitions are *not* wrapped with the implicit `this` parameter, but are instead wrapped at `getMthDefn`.
-   *  `R` represents the actual method types. The first optional value is the parent name, with `N` representing implicit calls,
-   *    and the second value is the method name. (See the case for `Sel` in `typeTerm` for documentation on explicit vs. implicit calls.)
-   *  The public helper functions should be preferred for manipulating `mthEnv`
+    * `L` represents the inferred types of method definitions. The first value is the parent name,
+    *   and the second value is the method name.
+    *   The method definitions are *not* wrapped with the implicit `this` parameter,
+    *   but are instead wrapped at `getMthDefn`.
```
fo5for: This should have been removed after the previous `MethodType` refactoring.

shared/src/test/diff/mlscript/BadMethods.mls
```diff
@@ -753,6 +753,9 @@ class Test7C: Test7A & Test7B
 //│ ╙──       	         ^^^^^^^^^^
 //│ Defined class Test7C
 
+Test7C.Mth7A
+//│ res: error
```
fo5for, line 757: While I agree that the logic has room for improvement, I don't think an error should be reported here. It would just be identical to the one reported at the definition. There's nothing more to report here until the definition is fixed.

fo5for (approved):
There are a few places where the documentation is inaccurate. I will make some changes and push them. Otherwise it looks good to me.

> This is the minimum standard of documentation I would expect for such nontrivial (borderline wicked) logic

I will do my best 😅

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -37,52 +37,66 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
+        val td = ctx.tyDefs(ownerType.name)
+        def addThis(mt: MethodType): MethodType =
+          mt.copy(body = mt.body.map(b => b.copy(_1 = td.thisTy(mt.prov))))(mt.prov)
+        def add(mt: MethodType): MethodType =
+          mt.copy(parents = ownerType :: mt.parents)(mt.prov)
+        mt.copy(decls = mt.decls.view.mapValues(addThis).mapValues(add).toMap, defns = mt.defns.view.mapValues(addThis).toMap)
+      }
+      parents.map(_.processInheritedMethods(false))
+        .reduceOption(_.&(_)(ownerType, defns.keySet)).map(addParent)
+        .foldRight(if (top) MethodSet(ownerType, Nil, Map.empty, Map.empty) else copy(parents = Nil)) {
```
LPTK: Actually, why do we set `parents = Nil` here? (Document it?)

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -37,52 +37,66 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
-    private def addParent(prt: TypeName)(implicit ctx: Ctx): MethodDefs = {
-      val td = ctx.tyDefs(prt.name)
-      def addThis(mt: MethodType): MethodType = mt.copy(body = mt.body.map(b => b.copy(_1 = td.thisTy(mt.prov))))
-      def add(mt: MethodType): MethodType = mt.copy(parents = prt :: mt.parents)
-      copy(decls = decls.view.mapValues(addThis).mapValues(add).toMap, defns = defns.view.mapValues(addThis).toMap)
+    /** When called with top = true, returns a `MethodSet` of the _inherited_ methods only,
+      *   disregarding the current MethodSet's methods... 
+      * Useful for subsumption checking
+      *   as inherited and current methods should be considered separately */
+    def processInheritedMethods(top: Bool)(implicit ctx: Ctx, raise: Raise): MethodSet = {
```
LPTK: Actually the doc of this method is still incomplete. It also performs some checks, possibly raising errors. That's why I renamed the method to include the verb "process". But the doc should mention this more specifically.

shared/src/main/scala/mlscript/Typer.scala
LPTK: Yeah, to me this `top` parameter is an implementation detail that is only exposed for coding convenience. I'm not sure it will do the right thing when called with `true` though (seems it would raise wrong error messages if a method is defined several times in a class).

Also, why name this `top` without documenting what it does? Why not name it `includeCurrentClassMethods`?

shared/src/test/diff/mlscript/BadMethods.mls
LPTK: There are no errors reported here. And it behaves the same as before my changes AFAIK. I just wanted to exercise this specific case in the tests.

LPTK (commented):
Otherwise LGTM. Please make the requested changes by amending your last commit, then rebase-merge the PR.

shared/src/main/scala/mlscript/Typer.scala
fo5for: I think we can hide the `top` parameter by making this a private helper method, and only expose a method that calls this with `top = true`.

> seems it would raise wrong error messages if a method is defined several times in a class

It actually wouldn't see that a method is defined multiple times since the method declaration and definitions are stored as `Map`s. The errors are raised in `typeMethods` instead.

I named `top` `top` because something special needs to be done if this is the "top level" set of methods, and I needed a boolean parameter to represent that. I will rename it to `includeCurrentMethods` (since the current `TypeDef` can also be a trait).

shared/src/test/diff/mlscript/BadMethods.mls
fo5for: Oh I thought it was related to the note above `MethodType`, which says that the error should be checked at the call site.

shared/src/main/scala/mlscript/Typer.scala
fo5for: It actually does nothing. I guess I put it there because I had the mental picture of folding up a tree into a single node recursively. I will remove that.

shared/src/main/scala/mlscript/Typer.scala
LPTK: > I think we can hide the top parameter by making this a private helper method, and only expose a method that calls this with top = true.

Yeah I know but I don't think it's worth it.

shared/src/test/diff/mlscript/BadMethods.mls
LPTK: No, in that note I meant the call sites of the `toPT`/`bodyPT` methods.

shared/src/main/scala/mlscript/Typer.scala
LPTK: Okay, good. It was very puzzling!

shared/src/main/scala/mlscript/Typer.scala
fo5for: Oops I didn't see this in time 😅

PR #90, opened 2022-03-08
Properly simplify intersections of tuple and record types

### LINKED ISSUE #69: Properly support tuple field accessors typing

LPTK:
Fixes https://github.com/hkust-taco/mlscript/issues/69.


shared/src/test/diff/mlscript/TupleArray.mls
```diff
@@ -239,7 +257,7 @@ k1._3
 //│    = [ 1, 2, 3, _4: true ]
 
 (1, 2, true) with {_0 = "hello"}
-//│ res: (1, 2, true,) & {_0: "hello"}
+//│ res: (1, 2, true,)
```
Meowcolm024: I think the  `_0` shall be kept in the type, otherwise the info is lost @_@

Meowcolm024 (approved):

PR #91, opened 2022-03-09
Support recursive `let` definition (close #64)

### LINKED ISSUE #64: Support recursive definitions in JS
Currently, recursive definitions don't work:

```ts
:js
rec def foo =
  let _ = console.log "hey" in foo
//│ // Query 1
//│ let foo1;
//│ foo1 = (function (_) {
//│   return foo;
//│ })(console.log("hey"));
//│ res = foo1;
//│ // End of generated code
//│ foo: nothing
//│    = [Function: foo]

:js
rec def foo() =
  let _ = console.log "hey" in foo()
//│ // Query 1
//│ let foo2;
//│ foo2 = () => (function (_) {
//│   return foo1();
//│ })(console.log("hey"));
//│ res = foo2;
//│ // End of generated code
//│ foo: () -> nothing
//│    = [Function: foo2]
```

Weirdly, sometimes they do seem to work:

```ts

// TODO define concat
:e
def concat: string -> string -> string
concat s t = s + t
//│ concat: string -> string -> string
//│ int -> int -> int
//│   <:  concat:
//│ string -> string -> string
//│ ╔══[ERROR] Type mismatch in def definition:
//│ ║  l.56: 	concat s t = s + t
//│ ║        	^^^^^^^^^^^^^^^^^^
//│ ╟── expression of type `string` does not match type `int`
//│ ║  l.55: 	def concat: string -> string -> string
//│ ║        	            ^^^^^^
//│ ╟── Note: constraint arises from reference:
//│ ║  l.56: 	concat s t = s + t
//│ ╙──      	             ^
//│ ╔══[ERROR] Type mismatch in def definition:
//│ ║  l.56: 	concat s t = s + t
//│ ║        	^^^^^^^^^^^^^^^^^^
//│ ╟── expression of type `int` does not match type `string`
//│ ║  l.56: 	concat s t = s + t
//│ ║        	             ^^^^^
//│ ╟── but it flows into function of type `?a -> ?b`
//│ ║  l.56: 	concat s t = s + t
//│ ║        	         ^^^^^^^^^
//│ ╟── which does not match type `string -> string`
//│ ╟── Note: constraint arises from type reference:
//│ ║  l.55: 	def concat: string -> string -> string
//│ ╙──      	                                ^^^^^^
//│       = [Function: concat]

:js
rec def test n acc =
  if n == 0 then acc
  else concat (test (n - 1) (concat acc "+")) "!"
//│ // Query 1
//│ let test;
//│ test = (n) => (acc) => (n == 0 ? acc : concat(test(n - 1)(concat(acc)("+")))("!"));
//│ res = test;
//│ // End of generated code
//│ test: int -> string -> string
//│     = [Function: test]

test 1 "hello"
//│ res: string
//│    = 'hello+!'

test 10 "hello"
//│ res: string
//│    = 'hello++++++++++!!!!!!!!!!'
```
Is this inconsistency a bug in the handling of scopes/names?

Other examples:
```ts
:js
def pow x =
  let rec go n =
    if n > 0 then go (n - 1) * x
    else 1
  in go
//│ // Query 1
//│ let pow;
//│ pow = (x) => (function (go) {
//│   return go;
//│ })((n) => (n > 0 ? go(n - 1) * x : 1));
//│ res = pow;
//│ // End of generated code
//│ pow: int -> int -> int
//│    = [Function: pow]

pow 3 4
//│ res: int
//│ Unexpected runtime error
//│   ReferenceError: go is not defined

:js
rec def pow n = { call = fun x -> if n > 0 then (pow (n - 1)).call x * x else 1 }
//│ // Query 1
//│ let pow1;
//│ pow1 = (n) => { call: (x) => (n > 0 ? pow(n - 1).call(x) * x : 1) };
//│ res = pow1;
//│ // End of generated code
//│ pow: int -> {call: int -> int}
//│    = [Function: pow1]

(pow 3).call 4
//│ res: int
//│ Unexpected runtime error
//│   TypeError: Cannot read properties of undefined (reading 'call')
```

chengluyu:
This PR implements recursive `let` definition in JavaScript code generation.

### DIFF
--- a/shared/src/main/scala/mlscript/JSBackend.scala
+++ b/shared/src/main/scala/mlscript/JSBackend.scala
@@ -129,16 +129,33 @@ class JSBackend {
     case Sel(receiver, fieldName) =>
       JSField(translateTerm(receiver), fieldName.name)
     // Turn let into an IIFE.
-    case Let(isRec, Var(name), value, body) =>
+    case Let(true, Var(name), Lam(args, body), expr) =>
+      val letScope = scope.derive("Let", name :: Nil)
+      val fn = {
+        val params = translateParams(args)
+        val bindings = name :: params.flatMap { _.bindings }
+        val fnScope = scope.derive("Function", bindings)
+        val fnBody = fnScope.tempVars.`with`(translateTerm(body)(fnScope))
+        JSFuncExpr(S(name), params, fnBody.fold(_.`return` :: Nil, identity))
+      }
+      JSImmEvalFn(
+        N,
+        JSNamePattern(name) :: Nil,
+        letScope.tempVars.`with`(translateTerm(expr)(letScope)),
+        fn :: Nil
+      )
+    case Let(_, Var(name), value, body) =>
       val letScope = scope.derive("Let", name :: Nil)
       JSImmEvalFn(
-        name :: Nil,
+        N,
+        JSNamePattern(name) :: Nil,
         letScope.tempVars `with` translateTerm(body)(letScope),
         translateTerm(value)(letScope) :: Nil
       )
     case Blk(stmts) =>
       val blkScope = scope.derive("Blk")
       JSImmEvalFn(
+        N,
         Nil,
         R(blkScope.tempVars `with` (stmts flatMap (_.desugared._2) map {
           case t: Term             => JSExprStmt(translateTerm(t))
@@ -195,7 +212,7 @@ class JSBackend {
     case Case(pat, body, rest) =>
       translateCase(scrut, pat)(translateTerm(body), translateCaseBranch(scrut, rest))
     case Wildcard(body) => translateTerm(body)
-    case NoCases        => JSImmEvalFn(Nil, R(JSInvoke(
+    case NoCases        => JSImmEvalFn(N, Nil, R(JSInvoke(
       JSNew(JSIdent("Error")),
       JSExpr("non-exhaustive case expression") :: Nil
     ).`throw` :: Nil), Nil)
@@ -360,7 +377,7 @@ class JSWebBackend extends JSBackend {
             ).stmt :: Nil
         })
     val epilogue = resultsIdent.member("map")(JSIdent(prettyPrinterName)).`return` :: Nil
-    JSImmEvalFn(Nil, R(polyfill.emit() ::: stmts ::: epilogue), Nil).toSourceCode.toLines
+    JSImmEvalFn(N, Nil, R(polyfill.emit() ::: stmts ::: epilogue), Nil).toSourceCode.toLines
   }
 }
 
--- a/shared/src/main/scala/mlscript/JSBackend.scala
+++ b/shared/src/main/scala/mlscript/JSBackend.scala
@@ -145,7 +145,7 @@ class JSBackend {
         fn :: Nil
       )
     case Let(true, Var(name), _, _) =>
-      throw new CodeGenError(s"recursive value $name is not supported")
+      throw new CodeGenError(s"recursive non-function definition $name is not supported")
     case Let(_, Var(name), value, body) =>
       val letScope = scope.derive("Let", name :: Nil)
       JSImmEvalFn(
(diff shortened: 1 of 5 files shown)


shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -129,16 +129,33 @@ class JSBackend {
     case Sel(receiver, fieldName) =>
       JSField(translateTerm(receiver), fieldName.name)
     // Turn let into an IIFE.
-    case Let(isRec, Var(name), value, body) =>
+    case Let(true, Var(name), Lam(args, body), expr) =>
```
LPTK, line 132: You should probably raise a code-gen error if the thing is not a lambda, instead of falling through to the next case and generating something wrong.

shared/src/test/diff/codegen/Recursion.mls
```diff
@@ -37,8 +42,7 @@ def pow1 x =
 // FIXME support recursive let bindings
```
LPTK:
```suggestion
```

LPTK (approved):
Cool, thanks! You can go ahead and squash-merge after addressing my comments.

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: But `rhs` (in `Let(isRec: Bool, name: Var, rhs: Term, body: Term)`) might be non-`Lam`s, right? For example, the parse result of `let y = 0 in y` is `Let(false, y, IntLit(0), Var(y))`.

LPTK: Yes, that's the whole point of my comment. Non-`Lam` recursive bindings aren't code-generable.

LPTK: Here's a hint:
```ml
let rec x = { f = x } in x
//│ res: {f: 'a} as 'a
//│ Runtime error:
//│   ReferenceError: x is not defined
```

chengluyu: Just fixed that!

chengluyu: There is also a scoping issue in following code. It should report: unresolved symbol x. I just fixed it.

```
let x = { f = x } in x
//│ res: {f: 'a} as 'a
//│ Runtime error:
//│   ReferenceError: x is not defined
```

shared/src/test/diff/codegen/Recursion.mls
```diff
@@ -54,3 +54,15 @@ rec def pow2 n = { call = fun x -> if n > 0 then (pow2 (n - 1)).call x * x else
 (pow2 3).call 4
 //│ res: int
 //│    = 64
+
+:js
+let rec x = { f = x } in x
+//│ res: {f: 'a} as 'a
+//│ Code generation encountered an error:
+//│   recursive value x is not supported
```
LPTK:
```suggestion
//│   recursive non-function definition x is not supported
```

shared/src/main/scala/mlscript/JSBackend.scala
LPTK: Great!

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -129,16 +129,35 @@ class JSBackend {
...
+        JSFuncExpr(S(name), params, fnBody.fold(_.`return` :: Nil, identity))
+      }
+      JSImmEvalFn(
+        N,
+        JSNamePattern(name) :: Nil,
+        letScope.tempVars.`with`(translateTerm(expr)(letScope)),
+        fn :: Nil
+      )
+    case Let(true, Var(name), _, _) =>
+      throw new CodeGenError(s"recursive value $name is not supported")
```
chengluyu:
```suggestion
      throw new CodeGenError(s"recursive non-function definition $name is not supported")
```

PR #92, opened 2022-03-10
Fix discrepancy in diff test logic

LPTK:
I noticed some files do not properly test for code-gen errors, whose testing logic was missing.

@chengluyu can you please update these tests?


LPTK:
Thanks!!

PR #96, opened 2022-03-13
Fix locations, type provenances, and error messages for mutable fields

LPTK:
@Meowcolm024 Here, I made many fixes to your branch regarding the rather tricky error reporting mechanism.

Most of the type provenance information you were passing along was wrong. But I don't blame you, because this system is a bit like black magic if you don't know all the details 😛 

Make sure to review the diff to validate the changes.

Also, you should really merge the changes from the main `mlscript` branch more regularly. The longer you wait, the more it risks diverging, and the more painful the merge conflict resolution will be!

### DIFF
--- a/shared/src/main/scala/mlscript/Typer.scala
+++ b/shared/src/main/scala/mlscript/Typer.scala
@@ -969,16 +969,17 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
         val vl = typeTerm(rhs)
         con(vl, fieldType, UnitType.withProv(prov))
       case Assign(s @ Subs(a, i), rhs) => 
-        val a_ty = typeTerm(a)
         val i_ty = typeTerm(i)
         con(i_ty, IntType, TopType)
+        val a_ty = typeTerm(a)
         val sprov = tp(s.toLoc, "assigned array element")
         val elemType = freshVar(sprov)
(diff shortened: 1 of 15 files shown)


LPTK:
Ah, and there was a bug in the typing of array updates! The new value was typed at `int` and the index never typed!

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -918,56 +928,65 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool) extend
...
         val vl = typeTerm(rhs)
-        val arr_ty = mkProxy(a_ty, tp(a.toCoveringLoc, "receiver"))
-        con(arr_ty, ArrayType(FieldType(Some(res), res)(prov))(prov), res)
-        con(vl, IntType, vl)
-        UnitType
-      case Assign(_, rhs) => 
-        err(msg"Cannot not assign to non-field", term.toLoc)(raise)
+        val arr_ty =
+            // Note: this proxy does not seem to make any difference:
+            mkProxy(a_ty, tp(a.toCoveringLoc, "receiver"))
```
Meowcolm024, line 980: I actually don't know what this is proxy is doing, as I just copied that from `rcdSel` @_@, since the typing looks similar

PR #107, opened 2022-04-25
Add variance analysis for type variables

### LINKED ISSUE #101: Implement variance analysis
This will be important to simplify inferred types further.

twitu:
Closes #101

### DIFF
--- a/shared/src/main/scala/mlscript/TypeDefs.scala
+++ b/shared/src/main/scala/mlscript/TypeDefs.scala
@@ -554,6 +554,7 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
     * @param ctx
     */
   def computeVariances(tyDefs: List[TypeDef], ctx: Ctx): Unit = {
+    println(s"VARIANCE ANALYSIS")
     var varianceUpdated: Bool = false;
 
     /** Update variance information for all type variables belonging
@@ -577,7 +578,7 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
           updateVariance(fieldTy.ub, curVariance)
       }
 
-      trace(s"updateVariance(ty: $ty of ${ty.getClass()}, curVariance: $curVariance)(${tyDef.nme.name}.${tyDef.tvarVariances})") {
+      trace(s"upd[$curVariance] $ty") {
         ty match {
           case ProxyType(underlying) => updateVariance(underlying, curVariance)
           case TraitTag(_) | ClassTag(_, _) => ()
@@ -588,6 +589,7 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
             val newVariance = oldVariance && curVariance
             if (newVariance =/= oldVariance) {
               tyDef.tvarVariances(t) = newVariance
+              println(s"UPDATE ${tyDef.nme.name}.$t from $oldVariance to $newVariance")
               varianceUpdated = true
             }
             val (visitLB, visitUB) = (
@@ -634,7 +636,7 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
             updateVariance(rhs, curVariance)
           case Without(base, names) => updateVariance(base, curVariance.flip)
         }
-      } (_ => if (varianceUpdated) s"updated - ${tyDef.nme.name}.${tyDef.tvarVariances})" else "no update")
+      }()
     }
 
     // set default value for all type variables as bivariant
@@ -643,32 +645,26 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
     // and hence are not set in the variance info map
     tyDefs.foreach(t => t.tparamsargs.foreach{case (_, tvar) => t.tvarVariances.put(tvar, VarianceInfo.bi)})
 
-    do {
+    var i = 1
+    do trace(s"⬤ ITERATION $i") {
       val visitedSet: MutSet[Bool -> TypeVariable] = MutSet()
       varianceUpdated = false;
-      tyDefs.foreach{
-        case t@TypeDef(Cls, nme, _, _, body, mthDecls, mthDefs, _, _) =>
-          updateVariance(body, VarianceInfo.co)(t, visitedSet)
-          val stores = mthDecls.foreach { mthDef => 
-            val mthBody = ctx.mthEnv.getOrElse(
-              Right(Some(nme.name), mthDef.nme.name),
-              throw new Exception(s"Method ${mthDef.nme.name} does not exist in the context")
-            ).body
-            mthBody.foreach { case (_, body) => updateVariance(body, VarianceInfo.co)(t, visitedSet) }
-          }
-        case t@TypeDef(Trt, nme, _, _, body, mthDecls, mthDefs, _, _) =>
-          updateVariance(body, VarianceInfo.co)(t, visitedSet)
-          val stores = mthDecls.foreach { mthDef => 
-            val mthBody = ctx.mthEnv.getOrElse(
-              Right(Some(nme.name), mthDef.nme.name),
-              throw new Exception(s"Method ${mthDef.nme.name} does not exist in the context")
-            ).body
-            mthBody.foreach { case (_, body) => updateVariance(body, VarianceInfo.co)(t, visitedSet) }
-          }
-        case t@TypeDef(Als, nme, _, _, body, mthDecls, mthDefs, _, _) =>
-          updateVariance(body, VarianceInfo.co)(t, visitedSet)
+      tyDefs.foreach {
+        case t@TypeDef(k, nme, _, _, body, mthDecls, mthDefs, _, _) =>
+          trace(s"${k.str} ${nme.name}  ${t.tvarVariances.iterator.map(kv => s"${kv._2} ${kv._1}").mkString("  ")}") {
+            updateVariance(body, VarianceInfo.co)(t, visitedSet)
+            val stores = mthDecls.foreach { mthDef => 
+              val mthBody = ctx.mthEnv.getOrElse(
+                Right(Some(nme.name), mthDef.nme.name),
+                throw new Exception(s"Method ${mthDef.nme.name} does not exist in the context")
+              ).body
+              mthBody.foreach { case (_, body) => updateVariance(body, VarianceInfo.co)(t, visitedSet) }
+            }
+          }()
       }
-    } while (varianceUpdated)
+      i += 1
+    }() while (varianceUpdated)
+    println(s"DONE")
   }
 
   case class VarianceInfo(isCovariant: Bool, isContravariant: Bool) {
--- a/shared/src/main/scala/mlscript/TypeDefs.scala
+++ b/shared/src/main/scala/mlscript/TypeDefs.scala
@@ -44,6 +44,7 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
     def thisTy(prov: TypeProvenance): TypeRef = TypeRef(nme, targs)(prov)
     def wrapMethod(pt: PolymorphicType, prov: TypeProvenance): MethodType =
       MethodType(pt.level, S((thisTy(prov), pt.body)), nme :: Nil, isInherited = false)(prov)
+    val tvarVariances: VarianceStore = MutMap()
   }
   
   /** Represent a set of methods belonging to some owner type.
@@ -541,5 +542,160 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
     typeMethods(typeTypeDefs(ctx.copy(env = allEnv, mthEnv = allMthEnv, tyDefs = allDefs)))
   }
   
-  
+  /**
+    * Finds the variances of all type variables in the given type definitions with the given
+    * context using a fixed point computation. The algorithm starts with each type variable
+    * as bivariant by default and each type defintion position as covariant and
+    * then keeps updating the position variance based on the types it encounters.
+    * 
+    * It uses the results to update variance info in the type defintions
+    *
+    * @param tyDefs
+    * @param ctx
+    */
+  def computeVariances(tyDefs: List[TypeDef], ctx: Ctx): Unit = {
+    var varianceUpdated: Bool = false;
+
+    /** Update variance information for all type variables belonging
+      * to a type definition.
+      *
+      * @param ty
+      *   type tree to check variance for
+      * @param curVariance
+      *   variance of current position where the type tree has been found
+      * @param tyDef
+      *   type definition which is currently being processed
+      * @param visited
+      *   set of type variables visited along with the variance
+      *   true polarity if covariant position visit
+      *   false polarity if contravariant position visit
+      *   both if invariant position visit
+      */
+    def updateVariance(ty: SimpleType, curVariance: VarianceInfo)(implicit tyDef: TypeDef, visited: MutSet[Bool -> TypeVariable]): Unit = {
+      def fieldVarianceHelper(fieldTy: FieldType): Unit = {
+          fieldTy.lb.foreach(lb => updateVariance(lb, curVariance.flip))
+          updateVariance(fieldTy.ub, curVariance)
+      }
+
+      trace(s"updateVariance(ty: $ty of ${ty.getClass()}, curVariance: $curVariance)(${tyDef.nme.name}.${tyDef.tvarVariances})") {
+        ty match {
+          case ProxyType(underlying) => updateVariance(underlying, curVariance)
+          case TraitTag(_) | ClassTag(_, _) => ()
+          case ExtrType(pol) => ()
+          case t: TypeVariable =>
+            // update the variance information for the type variable
+            val oldVariance = tyDef.tvarVariances.getOrElseUpdate(t, VarianceInfo.bi)
+            val newVariance = oldVariance && curVariance
+            if (newVariance =/= oldVariance) {
+              tyDef.tvarVariances(t) = newVariance
+              varianceUpdated = true
+            }
+            val (visitLB, visitUB) = (
+              !curVariance.isContravariant && !visited(true -> t),
+              !curVariance.isCovariant && !visited(false -> t),
+            )
+            if (visitLB) visited += true -> t
+            if (visitUB) visited += false -> t
+            if (visitLB) t.lowerBounds.foreach(lb => updateVariance(lb, VarianceInfo.co))
+            if (visitUB) t.upperBounds.foreach(ub => updateVariance(ub, VarianceInfo.contra))
+          case RecordType(fields) => fields.foreach {
+            case (_ , fieldTy) => fieldVarianceHelper(fieldTy)
+          }
+          case NegType(negated) =>
+            updateVariance(negated, curVariance.flip)
+          case TypeRef(defn, targs) =>
+            // it's possible that the type definition may not exist in the
+            // context because it is malformed or incorrect. Do nothing in
+            // such cases
+            ctx.tyDefs.get(defn.name).foreach(typeRefDef => {
+              // variance for all type parameters of type definitions has been preset
+              // do nothing if variance for the parameter does not exist
+              targs.zip(typeRefDef.tparamsargs).foreach { case (targ, (_, tvar)) =>
+                typeRefDef.tvarVariances.get(tvar).foreach {
+                  case in @ VarianceInfo(false, false) => updateVariance(targ, in)
+                  case VarianceInfo(true, false) => updateVariance(targ, curVariance)
+                  case VarianceInfo(false, true) => updateVariance(targ, curVariance.flip)
+                  case VarianceInfo(true, true) => ()
+                }
+              }
+            })
+          case ComposedType(pol, lhs, rhs) =>
+            updateVariance(lhs, curVariance)
+            updateVariance(rhs, curVariance)
+          case TypeBounds(lb, ub) =>
+            updateVariance(lb, VarianceInfo.contra)
+            updateVariance(ub, VarianceInfo.co)
+          case ArrayType(inner) => fieldVarianceHelper(inner)
+          case TupleType(fields) => fields.foreach {
+              case (_ , fieldTy) => fieldVarianceHelper(fieldTy)
+            }
+          case FunctionType(lhs, rhs) =>
+            updateVariance(lhs, curVariance.flip)
+            updateVariance(rhs, curVariance)
+          case Without(base, names) => updateVariance(base, curVariance.flip)
+        }
+      } (_ => if (varianceUpdated) s"updated - ${tyDef.nme.name}.${tyDef.tvarVariances})" else "no update")
+    }
+
+    // set default value for all type variables as bivariant
+    // this prevents errors when printing type defintions in
+    // DiffTests for type variables that are not used at all
+    // and hence are not set in the variance info map
+    tyDefs.foreach(t => t.tparamsargs.foreach{case (_, tvar) => t.tvarVariances.put(tvar, VarianceInfo.bi)})
+
+    do {
+      val visitedSet: MutSet[Bool -> TypeVariable] = MutSet()
+      varianceUpdated = false;
+      tyDefs.foreach{
+        case t@TypeDef(Cls, nme, _, _, body, mthDecls, mthDefs, _, _) =>
+          updateVariance(body, VarianceInfo.co)(t, visitedSet)
+          val stores = mthDecls.foreach { mthDef => 
+            val mthBody = ctx.mthEnv.getOrElse(
+              Right(Some(nme.name), mthDef.nme.name),
+              throw new Exception(s"Method ${mthDef.nme.name} does not exist in the context")
+            ).body
+            mthBody.foreach { case (_, body) => updateVariance(body, VarianceInfo.co)(t, visitedSet) }
+          }
+        case t@TypeDef(Trt, nme, _, _, body, mthDecls, mthDefs, _, _) =>
+          updateVariance(body, VarianceInfo.co)(t, visitedSet)
+          val stores = mthDecls.foreach { mthDef => 
+            val mthBody = ctx.mthEnv.getOrElse(
+              Right(Some(nme.name), mthDef.nme.name),
+              throw new Exception(s"Method ${mthDef.nme.name} does not exist in the context")
+            ).body
+            mthBody.foreach { case (_, body) => updateVariance(body, VarianceInfo.co)(t, visitedSet) }
+          }
+        case t@TypeDef(Als, nme, _, _, body, mthDecls, mthDefs, _, _) =>
+          updateVariance(body, VarianceInfo.co)(t, visitedSet)
+      }
+    } while (varianceUpdated)
+  }
+
+  case class VarianceInfo(isCovariant: Bool, isContravariant: Bool) {
+
+    /** Combine two pieces of variance information together
+     */
+    def &&(that: VarianceInfo): VarianceInfo =
+      VarianceInfo(isCovariant && that.isCovariant, isContravariant && that.isContravariant)
+
+    /*  Flip the current variance if it encounters a contravariant position
+     */
+    def flip: VarianceInfo = VarianceInfo(isContravariant, isCovariant)
+
+    override def toString(): String = this match {
+      case (VarianceInfo(true, true)) => "±"
+      case (VarianceInfo(false, true)) => "-"
+      case (VarianceInfo(true, false)) => "+"
+      case (VarianceInfo(false, false)) => "="
+    }
+  }
+
+  object VarianceInfo {
+    val bi: VarianceInfo = VarianceInfo(true, true)
+    val co: VarianceInfo = VarianceInfo(true, false)
+    val contra: VarianceInfo = VarianceInfo(false, true)
+    val in: VarianceInfo = VarianceInfo(false, false)
+  }
+
+  type VarianceStore = MutMap[TypeVariable, VarianceInfo]
 }
--- a/shared/src/main/scala/mlscript/Typer.scala
+++ b/shared/src/main/scala/mlscript/Typer.scala
@@ -22,7 +22,8 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool)
   type Binding = Str -> TypeScheme
   type Bindings = Map[Str, TypeScheme]
   
-  /** Keys of `mthEnv`:
+  /**  `env`: maps the names of all global and local bindings to their types
+    *  Keys of `mthEnv`:
     * `L` represents the inferred types of method definitions. The first value is the parent name,
     *   and the second value is the method name.
     * `R` represents the actual method types.
@@ -115,13 +116,17 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool)
     TypeDef(Cls, TypeName("unit"), Nil, Nil, TopType, Nil, Nil, Set.empty, N) ::
     {
       val tv = freshVar(noTyProv)(1)
-      TypeDef(Als, TypeName("Array"), List(TypeName("A") -> tv), Nil,
+      val tyDef = TypeDef(Als, TypeName("Array"), List(TypeName("A") -> tv), Nil,
         ArrayType(FieldType(None, tv)(noTyProv))(noTyProv), Nil, Nil, Set.empty, N)
+      tyDef.tvarVariances += tv -> VarianceInfo.co
+      tyDef
     } ::
     {
       val tv = freshVar(noTyProv)(1)
-      TypeDef(Als, TypeName("MutArray"), List(TypeName("A") -> tv), Nil,
+      val tyDef = TypeDef(Als, TypeName("MutArray"), List(TypeName("A") -> tv), Nil,
         ArrayType(FieldType(Some(tv), tv)(noTyProv))(noTyProv), Nil, Nil, Set.empty, N)
+      tyDef.tvarVariances += tv -> VarianceInfo.in
+      tyDef
     } ::
     Nil
   val primitiveTypes: Set[Str] =
--- a/shared/src/test/diff/analysis/Variance.mls
+++ b/shared/src/test/diff/analysis/Variance.mls
@@ -1,4 +1,4 @@
-:NoJS
+
 type UseT[T] = T -> int
 class C[X]: { y: X }
   method Create: int -> C[X]
--- a/shared/src/test/diff/analysis/Variance.mls
+++ b/shared/src/test/diff/analysis/Variance.mls
@@ -1,0 +1,39 @@
+:NoJS
+type UseT[T] = T -> int
+class C[X]: { y: X }
+  method Create: int -> C[X]
+  method MapWrap: (int -> int) -> int -> D[C[X]]
+class D[R]: { z: R }
+  method Produce: string -> D[R]
+  method Producer: int -> C[R] -> D[R]
+  method Unwrap: D[R] -> R
+//│ Defined type alias UseT[-T]
+//│ Defined class C[=X]
+//│ Declared C.Create: C['X] -> int -> C['X]
+//│ Declared C.MapWrap: C['X] -> (int -> int) -> int -> D[c & {C#X = 'X, y: 'X}]
+//│ Defined class D[=R]
+//│ Declared D.Produce: D['R] -> string -> D['R]
+//│ Declared D.Producer: D['R] -> int -> C['R] -> D['R]
+//│ Declared D.Unwrap: D['R] -> D['R] -> 'R
+
+:w
+class F[S, T]: E[T]
+  method Create: S -> T
+class E[G]
+  method Id: G -> G
+class Z[X, R]
+  method Phantom: C[X]
+//│ Defined class F[-S, =T]
+//│ Declared F.Create: F['S, 'T] -> 'S -> 'T
+//│ Defined class E[=G]
+//│ Declared E.Id: E['G] -> 'G -> 'G
+//│ Defined class Z[=X, ±R]
+//│ Declared Z.Phantom: Z['X, 'R] -> C['X]
+//│ ╔══[WARNING] Type definition Z has bivariant type parameters:
+//│ ║  l.24: 	class Z[X, R]
+//│ ║        	      ^
+//│ ╟── R is irrelevant and may be removed
+//│ ║  l.24: 	class Z[X, R]
+//│ ╙──      	           ^
+
+
--- a/shared/src/test/scala/mlscript/DiffTests.scala
+++ b/shared/src/test/scala/mlscript/DiffTests.scala
@@ -5,6 +5,7 @@ import fastparse.Parsed.Failure
 import fastparse.Parsed.Success
 import sourcecode.Line
 import scala.collection.mutable
+import scala.collection.mutable.{Map => MutMap}
 import scala.collection.immutable
 import mlscript.utils._, shorthands._
 import mlscript.JSTestBackend.IllFormedCode
@@ -67,6 +68,7 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
       noGeneration: Bool = false,
       showGeneratedJS: Bool = false,
       generateTsDeclarations: Bool = false,
+      debugVariance: Bool = false,
       expectRuntimeErrors: Bool = false,
       expectCodeGenErrors: Bool = false,
       showRepl: Bool = false,
@@ -109,6 +111,7 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
           case "ng" => mode.copy(noGeneration = true)
           case "js" => mode.copy(showGeneratedJS = true)
           case "ts" => mode.copy(generateTsDeclarations = true)
+          case "dv" => mode.copy(debugVariance = true)
           case "ge" => mode.copy(expectCodeGenErrors = true)
           case "re" => mode.copy(expectRuntimeErrors = true)
           case "ShowRepl" => mode.copy(showRepl = true)
@@ -289,16 +292,44 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
                 ctx.tyDefs.contains(td.nme.name) && !oldCtx.tyDefs.contains(td.nme.name)
               ).foreach(td => tsTypegenCodeBuilder.declareTypeDef(td))
             }
-            
+
+            val curBlockTypeDefs = typeDefs
+              // add check from process type def block below
+              .flatMap(typeDef => if (!oldCtx.tyDefs.contains(typeDef.nme.name)) ctx.tyDefs.get(typeDef.nme.name) else None)
+            // activate typer tracing if variance debugging is on and then set it back
+            // this makes it possible to debug variance in isolation
+            var temp = typer.dbg
+            typer.dbg = mode.debugVariance
+            typer.computeVariances(curBlockTypeDefs, ctx)
+            typer.dbg = temp
+            val varianceWarnings: MutMap[TypeName, Ls[TypeName]] = MutMap()
+
             // process type definitions first
             typeDefs.foreach(td =>
               // check if type def is not previously defined
               if (ctx.tyDefs.contains(td.nme.name)
                   && !oldCtx.tyDefs.contains(td.nme.name)) {
                   // ^ it may not end up being defined if there's an error
+
                 val tn = td.nme.name
-                output(s"Defined " + td.kind.str + " " + tn)
                 val ttd = ctx.tyDefs(tn)
+
+                // generate warnings for bivariant type variables
+                val bivariantTypeVars = ttd.tparamsargs.iterator.filter{ case (tname, tvar) =>
+                  ttd.tvarVariances.get(tvar).contains(typer.VarianceInfo.bi)
+                }.map(_._1).toList
+                if (!bivariantTypeVars.isEmpty) {
+                  varianceWarnings.put(td.nme, bivariantTypeVars)
+                }
+                
+                val params = if (!ttd.tparamsargs.isEmpty)
+                    SourceCode.horizontalArray(ttd.tparamsargs.map{ case (tname, tvar) =>
+                      val tvarVariance = ttd.tvarVariances.getOrElse(tvar, throw new Exception(s"Type variable $tvar not found in variance store ${ttd.tvarVariances} for $ttd"))
+                      SourceCode(s"$tvarVariance${tname.name}")
+                    }.toList).toString()
+                  else
+                    SourceCode("")
+                output(s"Defined " + td.kind.str + " " + tn + params)
 
                 // calculate types for all method definitions and declarations
                 // only once and reuse for pretty printing and type generation
@@ -335,6 +366,15 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
               }
             )
             
+            if (!varianceWarnings.isEmpty) {
+              import Message._
+              val diags = varianceWarnings.map{ case (tname, biVars) =>
+                val warnings = biVars.map( tname => msg"${tname.name} is irrelevant and may be removed" -> tname.toLoc).toList
+                Warning(msg"Type definition ${tname.name} has bivariant type parameters:" -> tname.toLoc :: warnings)
+              }.toList
+              report(diags)
+            }
+
             final case class ExecutedResult(var replies: Ls[ReplHost.Reply]) extends JSTestBackend.Result {
               def showFirst(prefixLength: Int): Unit = replies match {
                 case ReplHost.Error(err) :: rest =>
--- a/shared/src/test/diff/basics/Data.fun
+++ b/shared/src/test/diff/basics/Data.fun
@@ -4,7 +4,7 @@ data Test a b
 //│ Parsed: data ((Test a) b);
 //│ Desugared: class Test[a, b]: {a: a, b: b}
 //│ Desugared: def Test: [a, b] -> a -> b -> Test[a, b]
-//│ Defined class Test
+//│ Defined class Test[+a, +b]
 //│ Test: 'a -> 'b -> Test['a, 'b]
 
 :p
--- a/shared/src/test/diff/basics/Datatypes.fun
+++ b/shared/src/test/diff/basics/Datatypes.fun
@@ -43,7 +43,7 @@ data type Bool2 of True2 & False2
 //│ ║  l.25: 	data type Bool2 of True2 & False2
 //│ ╙──      	                   ^^^^^
 //│ Defined type alias Bool2
-//│ Defined class &
+//│ Defined class &[+True2, +False2]
 //│ &: 'a -> 'b -> &['a, 'b]
 
 data type Bool3 of
@@ -106,6 +106,7 @@ Tru : Boolean
 // data type List of Nil { T: Nothing }, Cons head tail { T: head | tail.T }
 
 :p
+:w
 data type List a of
   Nil
   Cons (head: a) (tail: List a)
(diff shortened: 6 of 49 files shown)


LPTK:
What happened to the `var` we added to `TypeDef`? Why did you revert to an external store?

And hat's with the strange `curStoreHash` trick? All you need is to make sure to test when something changed and flip a boolean flag when it does. A hash code is going to have collisions (even if unlikely) and thus be unreliable.

twitu:
> What happened to the `var` we added to `TypeDef`? Why did you revert to an external store?
> 
> And hat's with the strange `curStoreHash` trick? All you need is to make sure to test when something changed and flip a boolean flag when it does. A hash code is going to have collisions (even if unlikely) and thus be unreliable.

I'll make the change to use typedef store and maintain an update variable to check for changes.

twitu:
Also currently the variance analysis is not giving correct results. I'll fix that and more tests.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -951,6 +962,13 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool)
       case (VarianceInfo(false, false)) => this
     }
 
+    override def toString(): String = this match {
```
LPTK: How about using
```
±
-
+
=
```

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -537,6 +539,34 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
...
+              // in current block that have at least one
+              // type variable
+              val curBlockTypeDefs = typeDefs
+                .flatMap(typeDef => ctx.tyDefs.get(typeDef.nme.name))
+                .filter(tyDef => !tyDef.tparamsargs.isEmpty)
+              val varianceStore = ctx.computerVariance(curBlockTypeDefs)
+
+              val allSource = curBlockTypeDefs
+                .map(tyDef => {
+                  val nameSource = tyDef.kind match {
```
LPTK: There's already `tyDef.kind.str` for that.

LPTK:
Also please move the logic to `TypeDefs.scala`, and in particular don't put `updateVariance` in the `Ctx` – it has nothing to do there. This recursive traversal function should be a local `def` nested _inside_ `computerVariance`, so it has access to the boolean flag variable.

Also, `computerVariance` should be called `computerVariances`.

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -541,5 +541,136 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
     typeMethods(typeTypeDefs(ctx.copy(env = allEnv, mthEnv = allMthEnv, tyDefs = allDefs)))
   }
   
-  
+  /**
+    * Finds variances of all type variables in the given type definitions with the given
+    * context. It starts each type defintion as a bivariant and then keeps changing
```
LPTK:
```suggestion
    * context using a fixed point computation.
    * The algorithm starts with each type parameter as bivariant and then keeps updating it
```

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -541,5 +541,136 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
+    *
+    * @param tyDefs
+    * @param ctx
+    * @return Mapping from type variable to variance info
+    */
+  def computeVariances(tyDefs: List[TypeDef], ctx: Ctx): VarianceStore = {
+    var varianceUpdated: Bool = false;
+    def updateVariance(ty: SimpleType, curVariance: VarianceInfo)(implicit store: VarianceStore, ctx: Ctx): Unit = ty match {
+      case NegTrait(tt) => ???
+      case NegVar(tv) => ???
```
LPTK: These are just proxies for negated tags and vars. You can remove the cases and match on `ProxyType(underlying)` instead, recursing into `underlying`.

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -541,5 +541,136 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
+      case TraitTag(id) => ???
+      case ClassTag(_, _) => ???
```
LPTK: Nothing to do for tags.

twitu:
I think the current example `type UseT[T] = T -> int` is giving bivariant for T. It's going wrong because the starting bivariance is not becoming contra when flipping. Maybe we'll need one more parameter on flip, something to indicate the variance of the position, something like this -

```
def flip(pol: bool) = this match {
  bi => if (pol) co else contra
  co => if (pol) co else contra
  contra => if (pol) contra else co
  inv => inv
}
```

But I tried this out and now it's not terminating. :thinking:

LPTK:
When we implemented it together, we _did_ give to flip a `VarianceInfo` parameter which was supposed to be the variance of the traversed position... Please try to be more careful so we don't have to go over the same things again and again. Similarly for the `var` in `TypeDef` – I don't understand why you removed it. Is there something you're not sure about it?

Please make use of the tracing methods to help you understand what's going on. Have a look at how `trace` is used and do the same in the traversal method.

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -44,6 +44,7 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
     def thisTy(prov: TypeProvenance): TypeRef = TypeRef(nme, targs)(prov)
     def wrapMethod(pt: PolymorphicType, prov: TypeProvenance): MethodType =
       MethodType(pt.level, S((thisTy(prov), pt.body)), nme :: Nil, isInherited = false)(prov)
+    val tvarVariance: VarianceStore = MutMap()
```
LPTK:
```suggestion
    val tvarVariances: VarianceStore = MutMap()
```

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -541,5 +542,117 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
+    override def toString(): String = this match {
+      case (VarianceInfo(true, true)) => "±"
+      case (VarianceInfo(false, true)) => "-"
+      case (VarianceInfo(true, false)) => "+"
+      case (VarianceInfo(false, false)) => "="
+    }
+  }
+
+  type VarianceStore = MutMap[TypeVariable, VarianceInfo]
+  object VarianceInfo {
```
LPTK, line 689: Why not put this right below its companion class definition?

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -541,5 +542,117 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
+    def apply(): VarianceInfo = VarianceInfo(true, true)
```
LPTK: I don't see a need for this when we already have `bi`.

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -541,5 +542,117 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
+    def bi(): VarianceInfo = VarianceInfo()
+    def co(): VarianceInfo = VarianceInfo(true, false)
+    def contra(): VarianceInfo = VarianceInfo(false, true)
+    def inv(): VarianceInfo = VarianceInfo(false, false)
```
LPTK: These should all be values (`val`). No need to allocate new `VarianceInfo` all the time.

LPTK: To be consistent with the other names, this should be named `in`.

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -541,5 +542,117 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
+    * @param tyDefs
+    * @param ctx
+    * @return Mapping from type variable to variance info
+    */
+  def computeVariances(tyDefs: List[TypeDef], ctx: Ctx): Unit = {
+    var varianceUpdated: Bool = false;
+
+    def fieldVarianceHelper(fieldTy: FieldType, curVariance: VarianceInfo)(implicit ctx: Ctx, tyDef: TypeDef): Unit = {
+        updateVariance(fieldTy.ub, curVariance.flip(VarianceInfo.co()))
+        fieldTy.lb.foreach(lb => updateVariance(lb, curVariance.flip(VarianceInfo.contra())))
```
LPTK: For consistency, better process the LB before the UB.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -22,7 +22,8 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool)
   type Binding = Str -> TypeScheme
   type Bindings = Map[Str, TypeScheme]
   
-  /** Keys of `mthEnv`:
+  /**  env: maps the names of all global terms, and type constructors to their types
```
LPTK: `env` is also used for local bindings!

shared/src/test/diff/analysis/variance.mls
```diff
@@ -0,0 +1,23 @@
...
+//│ Declared C.MapWrap: C['X] -> (int -> int) -> int -> D[c & {C#X = 'X, y: 'X}]
+//│ Defined class D
+//│ Declared D.Produce: D['R] -> string -> D['R]
+//│ Declared D.Producer: D['R] -> int -> C['R] -> D['R]
+//│ Declared D.Unwrap: D['R] -> D['R] -> 'R
+//│ // start variance
+//│ type alias UseT[-T7']
+//│ class C[+X8']
+//│ class D[+R9']
+//│ // end variance
```
LPTK: You should extend this test with at least one other block of definitions which refer to previously-defined blocks

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -537,6 +539,30 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
...
             if (mode.generateTsDeclarations) outputSourceCode(tsTypegenCodeBuilder.toSourceCode())
+
+            if (mode.showVariance) {
+              // process typer typedefs for type defintions
+              // in current block that have at least one
+              // type variable
+              val curBlockTypeDefs = typeDefs
+                .flatMap(typeDef => ctx.tyDefs.get(typeDef.nme.name))
+                .filter(tyDef => !tyDef.tparamsargs.isEmpty)
+              typer.computeVariances(curBlockTypeDefs, ctx)
```
LPTK: This should always be done, not just when `mode.showVariance` is true.

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -537,6 +539,30 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
...
+
+              val allSource = curBlockTypeDefs
+                .map(tyDef => {
+                  val nameSource = SourceCode(s"${tyDef.kind.str} ${tyDef.nme.name}")
+                  val typeDefSource = SourceCode.horizontalArray(tyDef.tparamsargs.map{case (t, tvar) =>
+                      SourceCode(s"${tyDef.tvarVariance.getOrElse(tvar, throw new Exception(s"Type variable $tvar not found in variance store"))}$tvar")
+                    }.toList)
+                  nameSource ++ typeDefSource
+                })
+              outputSourceCode(
```
LPTK: Actually, it would be better to always show the inferred variance alongside the "`Defined <kind> <name>`" messages – updating these messages to the format "`Defined <kind> <name>[<variance><type param name>*]`".

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -110,6 +111,7 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
           case "ng" => mode.copy(noGeneration = true)
           case "js" => mode.copy(showGeneratedJS = true)
           case "ts" => mode.copy(generateTsDeclarations = true)
+          case "variance" => mode.copy(showVariance = true)
```
LPTK: I think it would be more useful to have a `dv` flag (for "debug variance") which shows the fixed analysis point progressing through the use of `trace`.

shared/src/test/diff/analysis/variance.mls
LPTK: Also, why not add the very good example we devised when discussing this algorithm?

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -541,5 +542,117 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
+      }
+    } while (varianceUpdated)
+  }
+
+  case class VarianceInfo(isCovariant: Bool, isContravariant: Bool) {
+
+    def &&(that: VarianceInfo): VarianceInfo =
+      VarianceInfo(isCovariant && that.isCovariant, isContravariant && that.isContravariant)
+
+    def flip(that: VarianceInfo): VarianceInfo = that match {
```
LPTK: So, looking at the use you make of it, this definition is pretty redundant. Let's make it a simple `def flip: VarianceInfo` method that simply flips the `isCovariant` and `isContravariant` fields, as I originally had in mind when I suggested this method.

Then when you go through a covariant position, you just use `curVariance`. When you go through a contravariant one, you use `curVariance.flip`. And when you go through an invariant one, you use `VarianceInfo .in`.

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -541,5 +542,117 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
+    def fieldVarianceHelper(fieldTy: FieldType, curVariance: VarianceInfo)(implicit ctx: Ctx, tyDef: TypeDef): Unit = {
+        updateVariance(fieldTy.ub, curVariance.flip(VarianceInfo.co()))
+        fieldTy.lb.foreach(lb => updateVariance(lb, curVariance.flip(VarianceInfo.contra())))
+    }
+
+    def updateVariance(ty: SimpleType, curVariance: VarianceInfo)(implicit ctx: Ctx, tyDef: TypeDef): Unit = ty match {
+      case ProxyType(underlying) => updateVariance(underlying, curVariance)
+      case TraitTag(_) | ClassTag(_, _) => ()
+      case ExtrType(pol) => ()
+      case t: TypeVariable => val _ = tyDef.tvarVariance.updateWith(t)(tVariance => tVariance match {
```
LPTK:
```suggestion
      case t: TypeVariable => tyDef.tvarVariance.updateWith(t) {
```

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -541,5 +542,117 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
     typeMethods(typeTypeDefs(ctx.copy(env = allEnv, mthEnv = allMthEnv, tyDefs = allDefs)))
   }
   
-  
+  /**
+    * Finds variances of all type variables in the given type definitions with the given
```
LPTK:
```suggestion
    * Finds the variances of all type variables in the given type definitions with the given
```

LPTK (changes requested):
Cool! But the code needs to be polished.

shared/src/test/diff/analysis/variance.mls
LPTK: Also `variance.mls` should be renamed `Variance.mls`.

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -537,6 +539,30 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
...
             // generate typescript typegen block
             if (mode.generateTsDeclarations) outputSourceCode(tsTypegenCodeBuilder.toSourceCode())
+
+            if (mode.showVariance) {
+              // process typer typedefs for type defintions
+              // in current block that have at least one
+              // type variable
+              val curBlockTypeDefs = typeDefs
+                .flatMap(typeDef => ctx.tyDefs.get(typeDef.nme.name))
+                .filter(tyDef => !tyDef.tparamsargs.isEmpty)
```
LPTK: I don't think this is a good idea. You analysis also finds the variances of method type variables. It would be inconsistent not to also compute the method type variable variances of type definitions that happen not to have type parameters.

LPTK:
I just realized something. Since your analysis is on internal type representations, it should definitely also visit type variable bounds when encountering a type variable.

You should recurse into `tv.lowerBounds` when `tv` is found in positive position and `tv.upperBounds` when found in negative position, and both when in invariant position.

Type variable bounds may be cyclic, so you'll need a `mutable.Set[(Bool, TV)]` to avoid re-traversing previously-traversed type variable bounds.

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -541,5 +542,117 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
+      case NegType(negated) =>
+        updateVariance(negated, curVariance.flip(VarianceInfo.contra()))
+      case TypeRef(defn, targs) =>
+        val typeRefDef = ctx.tyDefs.getOrElse(defn.name, throw new Exception(s"Cannot find $defn type definition in context"))
+        targs.zip(typeRefDef.tvars).foreach { case (targ, tvar) =>
+          val tvarVariance = if (curVariance === VarianceInfo.contra())
+              typeRefDef.tvarVariance.getOrElse(tvar, VarianceInfo.bi()).flip(VarianceInfo.contra())
+            else
+              typeRefDef.tvarVariance.getOrElse(tvar, VarianceInfo.bi()).flip(VarianceInfo.co())
+          updateVariance(targ, tvarVariance)
```
LPTK: Actually this case should read:
```suggestion
          val tvarVariance = typeRefDef.tvarVariances.getOrElseUpdate(tvar, VarianceInfo.bi)
          if (tvarVariance.isInvariant) updateVariance(targ, tvarVariance)
          else if (!tvarVariance.isCovariant) updateVariance(targ, curVariance.flip)
          else if (!tvarVariance.isContravariant) updateVariance(targ, curVariance)
```
or equivalently (probably cleaner):
```suggestion
          typeRefDef.tvarVariances.getOrElseUpdate(tvar, VarianceInfo.bi) match {
            case in @ VarianceInfo(false, false) => updateVariance(targ, in)
            case VarianceInfo(true, false) => updateVariance(targ, curVariance)
            case VarianceInfo(false, true) => updateVariance(targ, curVariance.flip)
            case VarianceInfo(true, true) =>
          }
```

shared/src/test/scala/mlscript/DiffTests.scala
twitu: Currently `MethodType` does not maintain a list of type variables, so a mapping is not maintained for it's variance information, even if it is calculated. So do we still want to process type defs that don't have type parameters.

LPTK: There's no reason for `MethodType` to maintain a list of type variables anyway.

> So do we still want to process type defs that don't have type parameters.

So you agree with me, then? It's not clear to me if your message is just to confirm my suggestion or make another point.

twitu: I've already removed this check. I'm trying to understand how method type parameters variance information be used since it's not stored anywhere.

LPTK: It is, you do seem to store it in `case t: TypeVariable` where you handle _all_ type variables, not just the type def's type params.

twitu: Aaah you're right it's storing all the `TypeVariables` it's finding. So for a method we simply have to check the `tvarVariances` of the top level TypeDef.

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -555,89 +555,92 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
-      case ComposedType(pol, lhs, rhs) =>
-        updateVariance(lhs, curVariance)
-        updateVariance(rhs, curVariance)
-      case TypeBounds(lb, ub) =>
-        updateVariance(lb, curVariance.flip(VarianceInfo.contra()))
-        updateVariance(ub, curVariance.flip(VarianceInfo.co()))
-      case ArrayType(inner) => fieldVarianceHelper(inner, curVariance)
-      case TupleType(fields) => fields.foreach {
-          case (_ , fieldTy) => fieldVarianceHelper(fieldTy, curVariance)
+    def updateVariance(ty: SimpleType, curVariance: VarianceInfo)(implicit ctx: Ctx, tyDef: TypeDef): Unit =
```
LPTK: No need to pass `ctx`. It can already be captured from the outer scope.
```suggestion
    def updateVariance(ty: SimpleType, curVariance: VarianceInfo)(implicit tyDef: TypeDef): Unit =
```

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -555,89 +555,92 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
+      trace(s"updateVariance(ty: $ty of ${ty.getClass()}, curVariance: $curVariance)(${tyDef.nme.name}.${tyDef.tvarVariances})") {
+        ty match {
+          case ProxyType(underlying) => updateVariance(underlying, curVariance)
+          case TraitTag(_) | ClassTag(_, _) => ()
+          case ExtrType(pol) => ()
+          case t: TypeVariable => val _ = tyDef.tvarVariances.updateWith(t) {
```
LPTK: _Again:_
```suggestion
          case t: TypeVariable => tyDef.tvarVariances.updateWith(t) {
```

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -555,89 +555,92 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
+            case None => Some(curVariance)
+            case Some(value) =>
+              val newVariance = value && curVariance
+              if (newVariance =/= curVariance) varianceUpdated = true;
```
LPTK:
```suggestion
              if (newVariance =/= curVariance) varianceUpdated = true
```

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -555,89 +555,92 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
+          }
+          case RecordType(fields) => fields.foreach {
+            case (_ , fieldTy) => fieldVarianceHelper(fieldTy, curVariance)
+          }
+          case NegType(negated) =>
+            updateVariance(negated, curVariance.flip)
+          case TypeRef(defn, targs) =>
+            val typeRefDef = ctx.tyDefs.getOrElse(defn.name, throw new Exception(s"Cannot find $defn type definition in context"))
+            targs.zip(typeRefDef.tvars).foreach { case (targ, tvar) =>
+              typeRefDef.tvarVariances.getOrElse(tvar, VarianceInfo.bi) match {
```
LPTK:
```suggestion
              typeRefDef.tvarVariances.getOrElseUpdate(tvar, VarianceInfo.bi) match {
```

Please be more careful when reviewing my comments and suggestions. This is the second one you partially disregarded here, which I take as a lack of respect for my time spent reviewing your changes.

(You can ignore this if you plan to but simply haven't gotten to making the change yet – but then the comment should not have been marked resolved, because you may forget to do it and I will not necessarily find out!)

shared/src/main/scala/mlscript/TypeDefs.scala
LPTK: I'm adding a TODO comment here so we don't forget: https://github.com/hkust-taco/mlscript/pull/107#issuecomment-1111738808

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -555,89 +555,92 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
     def &&(that: VarianceInfo): VarianceInfo =
       VarianceInfo(isCovariant && that.isCovariant, isContravariant && that.isContravariant)
 
-    def flip(that: VarianceInfo): VarianceInfo = that match {
-      case (VarianceInfo(true, true)) => this
-      case (VarianceInfo(true, false)) => if (this === VarianceInfo.bi()) that else this
-      case (VarianceInfo(false, true)) => if (this === VarianceInfo.bi()) that else VarianceInfo(isContravariant, isCovariant)
-      case (VarianceInfo(false, false)) => this
-    }
+    /*  Flips the current variance if it encounters a contravariant position
```
LPTK: For consistency with the doc for `&&`:
```suggestion
    /*  Flip the current variance when going through a contravariant position
```

shared/src/test/diff/analysis/Variance.mls
```diff
@@ -1,23 +1,17 @@
...
-//│ Defined class D
+//│ Defined class D[+R]
 //│ Declared D.Produce: D['R] -> string -> D['R]
 //│ Declared D.Producer: D['R] -> int -> C['R] -> D['R]
 //│ Declared D.Unwrap: D['R] -> D['R] -> 'R
-//│ // start variance
-//│ type alias UseT[-T7']
-//│ class C[+X8']
-//│ class D[+R9']
-//│ // end variance
```
LPTK: TODO: https://github.com/hkust-taco/mlscript/pull/107#discussion_r860458238

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -555,89 +555,92 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
+                case VarianceInfo(false, true) => updateVariance(targ, curVariance.flip)
+                case VarianceInfo(true, true) => ()
+              }
+            }
+          case ComposedType(pol, lhs, rhs) =>
+            updateVariance(lhs, curVariance)
+            updateVariance(rhs, curVariance)
+          case TypeBounds(lb, ub) =>
+            updateVariance(lb, curVariance.flip)
+            updateVariance(ub, curVariance)
```
LPTK: So the funny thing with `TypeBounds` is that the LB and UB are always at the same polarities:
```suggestion
            updateVariance(lb, VarianceInfo.contra)
            updateVariance(ub, VarianceInfo.co)
```

shared/src/main/scala/mlscript/TypeDefs.scala
twitu: bloops marks this whole block yellow saying non-Unit value is discarded if I don't add the `val _`

LPTK: Ok, that's because `updateWith` returns a value. In Scala `val _ =` really isn't a common idiom, as it binds an identifier named `_`, which is more of a quirk. You could instead add a `()` afterward this call to explicitly return Unit.

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -541,5 +542,120 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
+          case ProxyType(underlying) => updateVariance(underlying, curVariance)
+          case TraitTag(_) | ClassTag(_, _) => ()
+          case ExtrType(pol) => ()
+          case t: TypeVariable => val _ = tyDef.tvarVariances.updateWith(t) {
+            case None => Some(curVariance)
+            case Some(value) =>
+              val newVariance = value && curVariance
+              if (newVariance =/= curVariance) varianceUpdated = true;
+              Some(newVariance)
+          }
```
LPTK: Actually I think this logic isn't correct because in the `None` case we should definitely set `varianceUpdated = true`. Also shouldn't the test in the `Some` case be `newVariance =/= value`?!

I'd rewrite the code like this instead:
```suggestion
          case t: TypeVariable =>
            val oldVariance = tyDef.tvarVariances.getOrElseUpdate(t, VarianceInfo.co)
            val newVariance = oldVariance && curVariance
            if (newVariance =/= oldVariance) {
              tyDef.tvarVariances(t) = newVariance
              varianceUpdated = true
            }
```

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -302,18 +303,30 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
             typer.dbg = mode.debugVariance
             typer.computeVariances(curBlockTypeDefs, ctx)
             typer.dbg = temp
-            
+            val varianceWarnings: MutSortedMap[Str, Ls[Str]] = MutSortedMap()
```
LPTK: Instead of a `Ls[Str]`, better keep a `Ls[TypeName]` here, as `TypeName` also carries location info.

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -302,18 +303,30 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
...
               if (ctx.tyDefs.contains(td.nme.name)
                   && !oldCtx.tyDefs.contains(td.nme.name)) {
                   // ^ it may not end up being defined if there's an error
+
                 val tn = td.nme.name
                 val ttd = ctx.tyDefs(tn)
+
+                // generate warnings for bivariant type variables
+                val bivariantTypeVars = ttd.tparamsargs.filter{ case (tname, tvar) =>
+                  ttd.tvarVariances.get(tvar).map(v => v.equals(typer.VarianceInfo.bi)).getOrElse(false)
```
LPTK: Can it happen in a well-typed program that the variance not be found in `ttd.tvarVariances`? If yes, under what circumstances? => a comment explaining this and justifying the `.getOrElse(false)` is needed.

Also, don't use Java's `.equals`. Just use `===` (it's shadowed by ScalaTest here, so use the `=:=` synonym).

Also, the current logic could be written:
```suggestion
                  ttd.tvarVariances.get(tvar).contains(typer.VarianceInfo.bi)
```

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -354,6 +367,16 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
               }
             )
             
+            if (!varianceWarnings.isEmpty) {
+              import Message._
+              var warnings: List[Message -> Opt[Loc]] = Nil
+              varianceWarnings.foreach{ case (tname, biVars) =>
+                warnings ::= (msg"${tname} does not use ${biVars.mkString(",")}" -> None)
```
LPTK: You'll need to put the location of the type parameter here instead of `None`.

LPTK: > ${tname} does not use ${biVars.mkString(",")}

It may actually be using it these. For example in `class Foo[A] method F: Foo[A]`. So I'd rephrase this as I suggested in the issue, "type parameter T is irrelevant and could be removed".

Also, better generate one warning line per type parameter, this way each one will get its own location.

```suggestion
                warnings :::= biVars.map(v => msg"type parameter $v is irrelevant and could be removed" -> ...)
```

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -354,6 +367,16 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
...
+              }
+              warnings ::= (msg"Unused type parameters" -> None)
```
```suggestion
              warnings ::= msg"${tname} has unused type parameters:" -> None
```

shared/src/test/scala/mlscript/DiffTests.scala
twitu: There really is no situation when variance will not be found. Inside computeVariances we are initializing the variance for all type parameters to bi. The refactor you've shared looks like the best way to express this.

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -369,12 +369,11 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
...
             if (!varianceWarnings.isEmpty) {
               import Message._
-              var warnings: List[Message -> Opt[Loc]] = Nil
-              varianceWarnings.foreach{ case (tname, biVars) =>
-                warnings ::= (msg"${tname} does not use ${biVars.mkString(",")}" -> None)
-              }
-              warnings ::= (msg"Unused type parameters" -> None)
-              report(Warning(warnings) :: Nil)
+              val diags = varianceWarnings.map{ case (tname, biVars) =>
+                val warnings = biVars.map( tname => msg"${tname.name} is irrelevant and maybe removed" -> tname.toLoc).toList
```
```suggestion
                val warnings = biVars.map( tname => msg"${tname.name} is irrelevant and may be removed" -> tname.toLoc).toList
```

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -369,12 +369,11 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
...
+                Warning(msg"type definition ${tname.name} has bivariant type parameters" -> tname.toLoc :: warnings)
```
```suggestion
                Warning(msg"Type definition ${tname.name} has bivariant type parameters:" -> tname.toLoc :: warnings)
```

shared/src/main/scala/mlscript/TypeDefs.scala
twitu: A few points regarding the implementation. `computeVariances` maintains a mutable set of visited type variables. For any unvisited type variable it adds it to the visited set and recurses into the lb or ub types.

A question here is why are the lb and ub for `TypeVariable` expressed a `List[SimpleType]`. Why is a list needed here?

During recursion when it sees the typevariable again with it present in the visited set it updates it's variance information.

However there is one other case that needs to be handled i.e. for type variables that do not recurse. In such case their lower and upper bounds are empty and this is their first an only visit in the current branch of the type tree. So their variance information is updated in that visit itself.

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -556,24 +556,51 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
   def computeVariances(tyDefs: List[TypeDef], ctx: Ctx): Unit = {
     var varianceUpdated: Bool = false;
 
-    def fieldVarianceHelper(fieldTy: FieldType, curVariance: VarianceInfo)(implicit tyDef: TypeDef): Unit = {
+    def fieldVarianceHelper(fieldTy: FieldType, curVariance: VarianceInfo)(implicit tyDef: TypeDef, visited: MutSet[TypeVariable]): Unit = {
```
LPTK: You can move this function to `updateVariance`'s body and remove all parameters except for `fieldTy`.

shared/src/main/scala/mlscript/TypeDefs.scala
LPTK: The list is not really needed but is here for historical reasons. In the future we may replace it by single bounds with intersections and unions.

> for type variables that do not recurse. In such case their lower and upper bounds are empty

No, that's not true at all. Bounds may well be non-recursive. You don't see many cases like these in simplified `Type`s (though this will change soon), but the internal representations often have non-recursive bounds.

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -556,24 +556,51 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
         fieldTy.lb.foreach(lb => updateVariance(lb, curVariance.flip))
         updateVariance(fieldTy.ub, curVariance)
     }
 
-    def updateVariance(ty: SimpleType, curVariance: VarianceInfo)(implicit tyDef: TypeDef): Unit =
+    def updateVariance(ty: SimpleType, curVariance: VarianceInfo)(implicit tyDef: TypeDef, visited: MutSet[TypeVariable]): Unit =
```
LPTK: As I mentioned in my original comment, this should be a `MutSet[(Bool, TV)]`. Can please you make sure to follow my suggestions so I don't have to repeat myself?
Finding a variable `tv` in positive position should set `visited += true -> tv`. Negative position `visited += false -> tv`. Invariant position, both.

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -618,25 +645,25 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
       varianceUpdated = false;
       tyDefs.foreach{
         case t@TypeDef(Cls, nme, _, _, body, mthDecls, mthDefs, _, _) =>
-          updateVariance(body, VarianceInfo.co)(t)
+          updateVariance(body, VarianceInfo.co)(t, MutSet())
```
LPTK: Use a single new `MutSet` everywhere, created in each iteration of the loop.

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -556,24 +556,51 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
+                  else
+                    t.upperBounds.foreach(ub => updateVariance(ub, curVariance))
+                case VarianceInfo(true, false) =>
+                  if (t.lowerBounds.isEmpty)
+                    updateVariance(t, curVariance)
+                  else
+                    t.lowerBounds.foreach(lb => updateVariance(lb, curVariance))
+                case VarianceInfo(false, false) => updateVariance(t, curVariance)
+              }
+            }
```
LPTK: This logic is very wrong.

There should be no special-casing when the bounds are empty.

You should test for `visited.contains(pol -> t)` where `pol` depends on `curVariance`, as explained before.

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -556,24 +556,51 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
+            // update it's variance information
+            else {
+              // a type variable is bivariant when we know nothing about it
+              val oldVariance = tyDef.tvarVariances.getOrElseUpdate(t, VarianceInfo.bi)
+              val newVariance = oldVariance && curVariance
+              if (newVariance =/= oldVariance) {
+                tyDef.tvarVariances(t) = newVariance
+                varianceUpdated = true
+              }
             }
```
LPTK: This has nothing to do in such an `else` branch. 

Note that it could be done solely in the "not yet visited" branches, but it's simpler to just do it always:
```suggestion
            // a type variable is bivariant when we know nothing about it
            val oldVariance = tyDef.tvarVariances.getOrElseUpdate(t, VarianceInfo.bi)
            val newVariance = oldVariance && curVariance
            if (newVariance =/= oldVariance) {
              tyDef.tvarVariances(t) = newVariance
              varianceUpdated = true
            }
```
Also let's put it before the logic to visit the bounds.

LPTK (changes requested):

twitu:
Running the tests I can see that `Array` is a corner case. It's a type reference that's not defined in the user defined block. So it's never visited and it's default type parameter variance gets taken as bivariant thus generating warnings.

This'll be the same case with `MutArray`.

Do you suggest adding a specific checks to handle this corner case? Or solving this more generically?

One option is to set the variance information for the Array and MutArray to covariant and invariant, because we already know this.

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -556,54 +556,76 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
+      * @param tyDef
+      *   type definition which is currently being processed
+      * @param visited
+      *   set of type variables visited along with the variance
+      *   true polarity if covariant position visit
+      *   false polarity if contravariant position visit
+      *   both if invariant position visit
+      */
+    def updateVariance(ty: SimpleType, curVariance: VarianceInfo)(implicit tyDef: TypeDef, visited: MutSet[Bool -> TypeVariable]): Unit = {
+      // helper method for calling updateVariance on fields
```
LPTK: I think this comment is totally useless. You can get all this information immediately by just looking at the thing 😛

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -541,5 +542,180 @@ class TypeDefs extends ConstraintSolver { self: Typer =>
...
+                case VarianceInfo(true, true) => ()
+                case VarianceInfo(false, true) =>
+                    t.upperBounds.foreach(ub => updateVariance(ub, curVariance))
+                case VarianceInfo(true, false) =>
+                    t.lowerBounds.foreach(lb => updateVariance(lb, curVariance))
+                case VarianceInfo(false, false) =>
+                    t.upperBounds.foreach(ub => updateVariance(ub, curVariance))
+                    t.lowerBounds.foreach(lb => updateVariance(lb, curVariance))
+              }
+            }
```
LPTK: Type variable lower bounds have positive polarity and their upper bounds have negative polarity. So you should recurse in them accordingly (_not_ passing `curVariance`).

Also, all this code can be rewritten more concisely as follows:
```suggestion
            val (visitLB, visitUB) = (
              !curVariance.isContravariant && !visited(true -> t),
              !curVariance.isCovariant && !visited(false -> t),
            )
            if (visitLB) visited += true -> t
            if (visitUB) visited += false -> t
            if (visitLB) t.lowerBounds.foreach(lb => updateVariance(lb, VarianceInfo.co))
            if (visitUB) t.upperBounds.foreach(ub => updateVariance(ub, VarianceInfo.contra))
```

LPTK (commented):
Also, why are there many tests that are not updated and many that fail? Do't you run the tests locally before pushing? You need to fix them for this PR to be merged.

LPTK:
> One option is to set the variance information for the Array and MutArray to covariant and invariant, because we already know this.

Yes, I think this is the right thing to do!

LPTK:
> [Suppress badalias error with fixme](https://github.com/hkust-taco/mlscript/pull/107/commits/d849010857cf2679c2c3d00cff6141f9633dddc8)

I don't get it, why don't you just fix this? It's a bug. A type error in a type definition leading it to not be registered should _not_ crash the compiler in variance analysis!!

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -294,6 +294,10 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
               ).foreach(td => tsTypegenCodeBuilder.declareTypeDef(td))
             }
 
+            // intialize pre-defined type defs with their variance
+            ctx.tyDefs.get("Array").foreach(tydef => tydef.tvarVariances += tydef.tparamsargs(0)._2 -> typer.VarianceInfo.co)
+            ctx.tyDefs.get("MutArray").foreach(tydef => tydef.tvarVariances += tydef.tparamsargs(0)._2 -> typer.VarianceInfo.in)
+
```
LPTK: This is really not the place to do it. It should be done in `Typer.scala`, in the same place we define the actual type defs:

https://github.com/hkust-taco/mlscript/blob/21ed4a4c886e5ce55d29849778bec509abe7a639/shared/src/main/scala/mlscript/Typer.scala#L119-L126

LPTK (changes requested):

shared/src/main/scala/mlscript/TypeDefs.scala
LPTK: What's the reasoning for reverting this back yet again?

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -22,7 +22,8 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool)
   type Binding = Str -> TypeScheme
   type Bindings = Map[Str, TypeScheme]
   
-  /** Keys of `mthEnv`:
+  /**  env: maps the names of all global and local bindings to their types
```
LPTK: Consistency:
```suggestion
  /**  `env`: maps the names of all global and local bindings to their types
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -50,6 +51,7 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool)
     private val abcCache: MutMap[Str, Set[Var]] = MutMap.empty
     def allBaseClassesOf(name: Str): Set[Var] = abcCache.getOrElseUpdate(name,
       tyDefs.get(name).fold(Set.empty[Var])(_.allBaseClasses(this)(Set.empty)))
+
```
LPTK: Better avoid whitespace changes. No need to remove the indentation.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -867,5 +869,4 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool)
     
     go(st, polarity)(Set.empty)
   }
-
```
LPTK: Why?

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -536,7 +580,7 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
             }
             // generate typescript typegen block
             if (mode.generateTsDeclarations) outputSourceCode(tsTypegenCodeBuilder.toSourceCode())
-
```
LPTK: Idem.

twitu:
> I don't get it, why don't you just fix this? It's a bug. A type error in a type definition leading it to not be registered should not crash the compiler in variance analysis!!

This particular case is happening because the a type definition A depends on a type definition B which is malformed. Because B is malformed it is not added to the context.

How do you want me to handle such cases? Where the type definitions body refers to a type definition that is not part of the context. One option is to just skip any analysis and continue, this would leave the variance of any type variables used in the malformed type definition unchanged.

LPTK:
Yes for example, that seems reasonable. When there's an error in the program, generally, it's okay if the downstream code does not make perfect sense, because the program is wrong anyway.

LPTK:
```
[info] - node version *** FAILED ***
[info]   "v16.15.0" did not start with "v16.14", and "v16.15.0" did not start with "v17" (NodeTest.scala:11)
```

The Node version test could be adapted to accept your version.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -50,6 +51,7 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool)
     private val abcCache: MutMap[Str, Set[Var]] = MutMap.empty
     def allBaseClassesOf(name: Str): Set[Var] = abcCache.getOrElseUpdate(name,
       tyDefs.get(name).fold(Set.empty[Var])(_.allBaseClasses(this)(Set.empty)))
+
```
LPTK: Ishaaaaaaaaan

LPTK: Sorry for being such a pain, but I really dislike pointless spacing changes in unrelated parts of the code. You should see them when you review your own diff.

shared/src/main/scala/mlscript/TypeDefs.scala
twitu: Actually I realize I should do the computation inside the option monad using `foreach`. I'm pre-setting all type variables of a type definition to bi-variance, so the invariant logic here is that the variance for the type variable for the will always exist, otherwise just ignore it.

twitu: Feels like the GDP technique could be useful here :grin:. We know that values for these keys are set, but still have to handle the else case or use for each.

shared/src/main/scala/mlscript/Typer.scala
twitu: :lotus_position: let the white space flow through you :wind_chime: :leaves:

shared/src/main/scala/mlscript/TypeDefs.scala
twitu: I'll resolve this now, using the `foreach`. Using `foreach` in the above line also solves the error that was showing up in `BadAlias`.

LPTK:
@twitu I've pushed a commit to remove code that was **plain duplicated**. When you're done, you should clean up your code, making sure you avoid repeating yourself.

The commit also cleans up the variance analysis tracing info, which was completely unreadable. This is what it looks like now:

```ts
:dv
:w
class F[S, T]: E[T]
  method Create: S -> T
class E[G]
  method Id: G -> G
class Z[X, R]
  method Phantom: C[X]
//│ VARIANCE ANALYSIS
//│ ⬤ ITERATION 1
//│ | class F  ± S20'  ± T21'
//│ | | upd[+] E[T21']
//│ | | upd[+] ((None..S20',) -> T21')
//│ | | | upd[-] (None..S20',)
//│ | | | | upd[-] S20'
//│ | | | | | UPDATE F.S20' from ± to -
//│ | | | upd[+] T21'
//│ | | | | UPDATE F.T21' from ± to +
//│ | class E  ± G22'
//│ | | upd[+] ⊤
//│ | | upd[+] ((None..G22',) -> G22')
//│ | | | upd[-] (None..G22',)
//│ | | | | upd[-] G22'
//│ | | | | | UPDATE E.G22' from ± to -
//│ | | | upd[+] G22'
//│ | | | | UPDATE E.G22' from - to =
//│ | class Z  ± X23'  ± R24'
//│ | | upd[+] ⊤
//│ | | upd[+] C[X23']
//│ | | | upd[=] X23'
//│ | | | | UPDATE Z.X23' from ± to =
//│ ⬤ ITERATION 2
//│ | class F  - S20'  + T21'
//│ | | upd[+] E[T21']
//│ | | | upd[=] T21'
//│ | | | | UPDATE F.T21' from + to =
//│ | | upd[+] ((None..S20',) -> T21')
//│ | | | upd[-] (None..S20',)
//│ | | | | upd[-] S20'
//│ | | | upd[+] T21'
//│ | class E  = G22'
//│ | | upd[+] ⊤
//│ | | upd[+] ((None..G22',) -> G22')
//│ | | | upd[-] (None..G22',)
//│ | | | | upd[-] G22'
//│ | | | upd[+] G22'
//│ | class Z  = X23'  ± R24'
//│ | | upd[+] ⊤
//│ | | upd[+] C[X23']
//│ | | | upd[=] X23'
//│ ⬤ ITERATION 3
//│ | class F  - S20'  = T21'
//│ | | upd[+] E[T21']
//│ | | | upd[=] T21'
//│ | | upd[+] ((None..S20',) -> T21')
//│ | | | upd[-] (None..S20',)
//│ | | | | upd[-] S20'
//│ | | | upd[+] T21'
//│ | class E  = G22'
//│ | | upd[+] ⊤
//│ | | upd[+] ((None..G22',) -> G22')
//│ | | | upd[-] (None..G22',)
//│ | | | | upd[-] G22'
//│ | | | upd[+] G22'
//│ | class Z  = X23'  ± R24'
//│ | | upd[+] ⊤
//│ | | upd[+] C[X23']
//│ | | | upd[=] X23'
//│ DONE
//│ Defined class F[-S, =T]
//│ Declared F.Create: F['S, 'T] -> 'S -> 'T
//│ Defined class E[=G]
//│ Declared E.Id: E['G] -> 'G -> 'G
//│ Defined class Z[=X, ±R]
//│ Declared Z.Phantom: Z['X, 'R] -> C['X]
//│ ╔══[WARNING] Type definition Z has bivariant type parameters:
//│ ║  l.25: 	class Z[X, R]
//│ ║        	      ^
//│ ╟── R is irrelevant and may be removed
//│ ║  l.25: 	class Z[X, R]
//│ ╙──      	           ^
```

It's good to invest 5min to make debugging info easy to read. It quickly pays off by letting you be more productive down the line!

LPTK (approved):
Good job! Thanks for making variance analysis finally happen! 😄

PR #111, opened 2022-06-12
Support nested type provenance information

twitu:
The current approach does the following.
* A new class called NestedTypeProvenance that extends TypeProvenance
* Where nested provenance is required use a `ProvType` to store the extra nested prov info.

The initial approach of converting `ConCtx` from `Ls[SimpleType] -> Ls[SimpleType]` to `Ls[TypeProvenance] -> Ls[TypeProvenance]` was not working out well. Because `errorReport` heavily depends on the `SimpleType` information to create appropriate error messages.


twitu:
Current implementation does not store context chains across level changes.

```scala
    type ConCtx = Ls[SimpleType] -> Ls[SimpleType]

      recImpl(lhs, rhs)(raise,
        if (sameLevel)
          (if (cctx._1.headOption.exists(_ is lhs)) cctx._1 else lhs :: cctx._1)
          ->
          (if (cctx._2.headOption.exists(_ is rhs)) cctx._2 else rhs :: cctx._2)
        else (lhs :: Nil) -> (rhs :: Nil)
```

The new implementation tries to maintain previous context information by adding to the rhs chain (or second field of the context tuple).

```scala
    def rec(lhs: SimpleType, rhs: SimpleType, sameLevel: Bool, nestedProv: Opt[NestedTypeProvenance] = N)
          (implicit raise: Raise, cctx: ConCtx): Unit = {
        if (sameLevel)
        <snip>
        else (lhs :: Nil) -> nestedProv.map(prov => rhs.withProv(prov) :: Nil).getOrElse(rhs :: Nil)
```

This allows us to maintain the previous chain of information while also bundling it into a single SimpleType object, while also making minimum changes to type signatures. If no nested provenance is passed the existing rhs is used to construct the cctx.

---

Currently nested provenance is only being passed where lhs and rhs match FunctionTypes. However there are other cases where sameLevel is set to False but the nested type provenance is not being passed, thus losing out on the information. These are -

```scala
def recImpl
            case (LhsRefined(bo, ts, r, _), RhsBases(ots, S(R(RhsField(n, t2))), trs)) =>
            case (LhsRefined(S(b: ArrayBase), ts, r, _), RhsBases(pts, S(L(ar: ArrayType)), _)) =>
            case (TupleType(fs0), TupleType(fs1)) if fs0.size === fs1.size => // TODO generalize (coerce compatible tuples) =>
            case (t: ArrayBase, a: ArrayType) =>
            case (err @ ClassTag(ErrTypeId, _), FunctionType(l1, r1)) =>
            case (FunctionType(l0, r0), err @ ClassTag(ErrTypeId, _)) =>
            case (RecordType(fs0), RecordType(fs1)) =>
            case (err @ ClassTag(ErrTypeId, _), RecordType(fs1)) =>
            case (RecordType(fs1), err @ ClassTag(ErrTypeId, _)) =>
            case (tr1: TypeRef, tr2: TypeRef) if tr1.defn.name =/= "Array" =>

    def recLb(lhs: FieldType, rhs: FieldType)
```

Many of these might not be relevant. However for the current example we're solving from `ProvFlows.mls` we encounter the following flow.

1. (l0 -> r0) FunctionType <: (l1 -> r1) FunctionType (store nested prov)
2. (l1) Tuple <: (l0) Tuple (prov info lost because nested prov is not being passed)

This I can fix. However we'll need to figure which of the above cases need to pass nested prov.

---

Another behaviour I'm seeing is that cases where ProvTypes are matched adds a lot of elements to the context making long chains that has to be handled in error reporting. But I'm not sure if it adds extra information. Here's a debug trace snippet from `ProvFlows.mls`.

```scala
          case (p @ ProvType(und), _) => rec(und, rhs, true)
          case (_, p @ ProvType(und)) => rec(lhs, und, true)

// debug trace
//│ | C [[(Int -> Int)]] <! ([([false<TypeName(bool)>],)] -> α28)    (0) where lhs is ProvType and rhs is FunctionType and cctx is (List([[(Int -> Int)]]),List(([([false<TypeName(bool)>],)] -> α28)))}
//│ | | C [(Int -> Int)] <! ([([false<TypeName(bool)>],)] -> α28)    (0) where lhs is ProvType and rhs is FunctionType and cctx is (List([(Int -> Int)], [[(Int -> Int)]]),List(([([false<TypeName(bool)>],)] -> α28), ([([false<TypeName(bool)>],)] -> α28)))}
//│ | | | C (Int -> Int) <! ([([false<TypeName(bool)>],)] -> α28)    (0) where lhs is FunctionType and rhs is FunctionType and cctx is (List((Int -> Int), [(Int -> Int)], [[(Int -> Int)]]),List(([([false<TypeName(bool)>],)] -> α28), ([([false<TypeName(bool)>],)] -> α28), ([([false<TypeName(bool)>],)] -> α28)))}
```

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -280,18 +280,25 @@ class ConstraintSolver extends NormalForms { self: Typer =>
...
           (implicit raise: Raise, cctx: ConCtx): Unit =
     // trace(s"C $lhs <! $rhs") {
-    trace(s"C $lhs <! $rhs    (${cache.size})") {
+    trace(s"C $lhs <! $rhs    (${cache.size}) where lhs is ${lhs.getClass.getSimpleName} and rhs is ${rhs.getClass.getSimpleName} and cctx is $cctx}") {
     // trace(s"C $lhs <! $rhs  ${lhs.getClass.getSimpleName}  ${rhs.getClass.getSimpleName}") {
       // println(s"[[ ${cctx._1.map(_.prov).mkString(", ")}  <<  ${cctx._2.map(_.prov).mkString(", ")} ]]")
       // println(s"{{ ${cache.mkString(", ")} }}")
       
+      lazy val provChain = Some(NestedTypeProvenance(cctx._1 reverse_::: cctx._2))
+      lazy val revProvChain = Some(NestedTypeProvenance((cctx._1 reverse_::: cctx._2).reverse))
```
LPTK: I think this is equivalent (but more efficient!):
```suggestion
      lazy val revProvChain = Some(NestedTypeProvenance((cctx._2 reverse_::: cctx._1)))
```

shared/src/main/scala/mlscript/TyperDatatypes.scala
```diff
@@ -195,7 +197,7 @@ abstract class TyperDatatypes extends TyperHelpers { self: Typer =>
   
   /** The sole purpose of ProvType is to store additional type provenance info. */
   case class ProvType(underlying: SimpleType)(val prov: TypeProvenance) extends ProxyType {
-    override def toString = s"[$underlying]"
+    override def toString = if (prov is NestedTypeProvenance) s"[$underlying] prov: $prov" else s"[$underlying]"
```
LPTK, line 221: `is` is a shorthand for testing reference equality. `NestedTypeProvenance` here is the companion object of a case class. This check will never be true.

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -271,7 +271,7 @@ class ConstraintSolver extends NormalForms { self: Typer =>
         }
       }
     
-    def rec(lhs: SimpleType, rhs: SimpleType, sameLevel: Bool, nestedProv: Opt[TypeProvenance] = N)
+    def rec(lhs: SimpleType, rhs: SimpleType, sameLevel: Bool, nestedProv: Opt[NestedTypeProvenance] = N)
```
LPTK: It seems the `sameLevel` boolean has become redundant. Eventually it would have the same info as `nestedProv.isEmpty`, rigth?

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -538,7 +546,6 @@ class ConstraintSolver extends NormalForms { self: Typer =>
       
       val detailedContext =
         if (explainErrors)
-          // TODO show nested provs
```
LPTK: You removed this but you're still not showing them...

LPTK (commented):
Note: you apparently forgot to run the tests after the merge.

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -279,26 +286,108 @@ class ConstraintSolver extends NormalForms { self: Typer =>
...
+      
+      // check if these tuples are implicit
+      // don't add them to chain
+      (lhs, rhs) match {
+        case (lt: TupleType, rt: TupleType) =>
+          if (lt.implicitTuple || rt.implicitTuple)
+            sameLhs = true
+            sameRhs = true
+        case _ => ()
+      }
```
LPTK, line 315: Instead of this hack, wouldn't it be better to simply not associate any provenance to tuples types derived from implicit tuples?

twitu: Yes provenance is no longer associated with implicit tuples. I missed removing this old check.

LPTK: Ok, cool.

LPTK:
Ok so please re-open this PR from your fork's branch when it's ready.

PR #112, opened 2022-06-14
Half-completed SpliceType

Meowcolm024:
Support for Splice Types. Not finished, but at least some part works O_O

Should be working:
- Parser
- Type Simplifier

Not finished yet:
- Constraint Solver
- Normal Form
- Being used as a pattern in function arguments ?
- Codegen


LPTK (approved):
Ok let's merge this now to prevent it from diverging further from the main branch.

Issue #65, opened 2022-01-11
JavaScript code generation backlog


chengluyu:
When working on #53, I met a lot of things that we could improved in the future. Here are them.

- [ ] Find a neat way to store query results. Requirements:
  - If the initialization fails, the variable should not be declared.
  - Declaration should not pollute global context.
  
  [Source](https://github.com/hkust-taco/mlscript/pull/53#discussion_r780986469)
- [x] ~~Implement `Class.Member instance` calling convention. [Source](https://github.com/hkust-taco/mlscript/pull/53#issuecomment-1009170669)~~ This syntax might be changed in the new parser.
- [x] Add timeout to `stream.read()` otherwise we will occasionally suffer from infinite loops. [Source](https://github.com/hkust-taco/mlscript/pull/53#issuecomment-1004609618) (Fixed in #126)
- [x] Automatically check references to unimplemented variables. [Source](https://github.com/hkust-taco/mlscript/pull/53#issuecomment-1004609618)
- [x] ~~Support merging arrays in `withConstruct`. Source TBD.~~ The semantics of `with` might be changed.
- [ ] Some `// TODO` introduced in commits from #53. Source TBD.
    
  **As for 1 Aug, 2022. The `// TODO`s are as follows.** They are not urgent at this moment.
  - Translation of `Blk`. https://github.com/hkust-taco/mlscript/blob/ba99965401408470e4072d95081199c504c004a8/shared/src/main/scala/mlscript/JSBackend.scala#L177-L178
  - Consider traits when sorting classes. https://github.com/hkust-taco/mlscript/blob/ba99965401408470e4072d95081199c504c004a8/shared/src/main/scala/mlscript/JSBackend.scala#L482-L485
  - Improve parentheses in code generation. https://github.com/hkust-taco/mlscript/blob/ba99965401408470e4072d95081199c504c004a8/shared/src/main/scala/mlscript/codegen/Codegen.scala#L415-L416
  - Load built-in symbols from `Typer`. https://github.com/hkust-taco/mlscript/blob/ba99965401408470e4072d95081199c504c004a8/shared/src/main/scala/mlscript/codegen/Scope.scala#L23-L26 https://github.com/hkust-taco/mlscript/blob/ba99965401408470e4072d95081199c504c004a8/shared/src/main/scala/mlscript/codegen/Scope.scala#L45-L51
[...]

PR #113, opened 2022-06-21
Fix the `res` running variable is not consistent with DiffTests

NeilKleistGao:
Relative Issue: #65 

Since every statement will change the `res` variable, I tried to add the type information into the context so that the correct type could be found. I also used a flag variable to indicate self-define `res`. I'm not sure expected results and it may need further discussion.


shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -52,6 +52,7 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
       override def emitDbg(str: String): Unit = output(str)
     }
     var ctx: typer.Ctx = typer.Ctx.init
+    var selfDefineRes: Bool = false
```
LPTK: Better avoid mutable variables whenever possible. Here it should be easy to do away with it!

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -510,6 +511,8 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
                   typer.typeType(rhs)(ctx.nextLevel, raise,
                     vars = tps.map(tp => tp.name -> typer.freshVar(typer.noProv/*FIXME*/)(1)).toMap))
                 ctx += nme.name -> ty_sch
+                if (nme.name.contentEquals("res")) selfDefineRes = true
```
```suggestion
                val selfDefineRes = nme.name === "res"
```

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -521,13 +524,15 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
                 typer.dbg = mode.dbg
                 val ty_sch = typer.typeLetRhs(isrec, nme.name, rhs)(ctx, raise)
                 val exp = getType(ty_sch)
+                if (nme.name.contentEquals("res")) selfDefineRes = true
```
LPTK: Same.

LPTK (changes requested):
Cool, thanks! I'd just a minor change to be done, as explained below.

shared/src/test/scala/mlscript/DiffTests.scala
NeilKleistGao: I think `selfDefineRes` maybe not be a very good name for this variable. Actually, I use it to indicate whether there is a user-defined variable, whose name is also `res`, in the context. If we only check the current variable, this would happen:
```typescript
:js
res = 5
//│ // Query 1
//│ globalThis.res1 = 5;
//│ res = res1;
//│ // End of generated code
//│ res: 5
//│    = 5

:js
foo
//│ // Query 1
//│ res = foo;
//│ // End of generated code
//│ res: "oops"
//│    = 'oops'

:js
res
//│ // Query 1
//│ res = res1;
//│ // End of generated code
//│ res: "oops"
//│    = 5
```

Since users define their own `res` variable, I think our query actions should not modify it. :smile:

LPTK: Hmmm... I'm not convinced we should do something different if a `res` was defined somewhere else in the file. Imagine someone adds a `res = 123` at the beginning of a 1000-line fine, and this ends up changing a totally unrelated part of the test file that relied on auto-`res` variable. Better just be consistent here and override any previous `res`.

NeilKleistGao: I think it is acceptable to override the previous `res` when the auto-`res` is updated, and the result of the test above should be like this:
```typescript
:js
res = 5
//│ // Query 1
//│ res = 5;
//│ // End of generated code
//│ res: 5
//│    = 5

:js
foo
//│ // Query 1
//│ res = foo;
//│ // End of generated code
//│ res: "oops"
//│    = 'oops'

:js
res
//│ // Query 1
//│ res = res;
//│ // End of generated code
//│ res: "oops"
//│    = 'oops'
```

I firstly found this problem in `mlscript/shared/src/test/diff/mlscript/Yicong.mls`:
```typescript
res = if true then (1, 2, 3, 'hello') else (true, 4, false)
hhh = if false then (45,345, 'bye', true) else (false, 3)
t = if false then res else hhh
```

If I understood your suggestion correctly, this test file may need to be changed, and the user-defined `res` should not be renamed as `res1`.

LPTK: The Yicong test is fine. Note that in principle no `res` is introduced when the results of the lines are already bound.

> I think it is acceptable to override the previous res when the auto-res is updated, and the result of the test above should be like this:

Agreed! Do you see how to do that?

NeilKleistGao: Hmmm... I think I got some clues. I will try tomorrow. Thank you for you help. :smile:

NeilKleistGao: I've found several new problems today. The first one is:
```typescript
:js
42
//│ // Prelude
//│ let res;
//│ // Query 1
//│ res = 42;
//│ // End of generated code
//│ res: 42
//│    = 42

:js
def foo = "oops"
//│ // Query 1
//│ globalThis.foo = "oops";
//│ res = foo;
//│ // End of generated code
//│ foo: "oops"
//│    = 'oops'

:js
res
//│ // Query 1
//│ res = res;
//│ // End of generated code
//│ res: 42
//│    = 42
```
I'm not sure whether `res = 42` is correct. Because `"oops"` is bound to the variable `foo`, and `res` should not be introduced. But there is `res = foo;` here. Does it still mean that `res` should be updated?

The second problem is the results of the test below:
```typescript
:js
def res = 0
//│ // Query 1
//│ globalThis.res1 = 0;
//│ res = res1;
//│ // End of generated code
//│ res: 0
//│    = 0

// However, we can shadow `res`!

:js
res + 1
//│ // Query 1
//│ res = res1 + 1;
//│ // End of generated code
//│ res: int
//│    = 1

:js
res + 1
//│ // Query 1
//│ res = res1 + 1;
//│ // End of generated code
//│ res: int
//│    = 1
```
Since `res + 1` is not bound to any variable, so I think this should override the `res1`, and the final result of `res + 1` should be 2.

Hope to get your confirmation so that I can make sure my codes are correct. :smile:

LPTK: Yeah I also noticed `def`s currently seems to modify `res`. But it looks like it may just be a `:js` printing error as `res` is visibly not modified.

> Since res + 1 is not bound to any variable, so I think this should override the res1, and the final result of res + 1 should be 2.

Seems so, yes.

Another thing that might cause problems:

```scala
def res: bool

42

res
```

NeilKleistGao: I think `res` has been modified by `res = foo;` and that is why the inconsistency exists because the display of `foo` depends on this variable:
```scala
def query(prelude: Str, code: Str): ReplHost.Reply = {
  val wrapped = s"$prelude try { $code } catch (e) { console.log('\\u200B' + e + '\\u200B'); }"
  send(wrapped)
  consumeUntilPrompt() match {
    case _: ReplHost.Result =>
      send("res") // here
      consumeUntilPrompt()
    case t => t
  }
}
```
In order to change test files less, I just back up the value of `res` and recover it after the execution. Feel free to contact me if `res = foo;` could be removed.

LPTK: @chengluyu any better ideas?

chengluyu: I was mimicking the behavior of Node.js actually. In Node.js REPL, `_` represents the latest result. But if you override it with something like `const _ = 8`, you will lose this shortcut to the latest result forever.

LPTK: Let's not use JS as a reputable source to take inspiration from :P

NeilKleistGao: So I think this version is plausible, and I still wonder whether we should remove `res = foo;` or not. 😄

LPTK: Yes you should.

NeilKleistGao: Since I've removed all `res = someVariableName`, there are some changes in `BadInherit2.mls`, `BadInherit2Co.mls` and `Under.mls` that need further discussions. Some data are also outputted even though an error occurred because we used `res` as an intermediate variable. I'm not sure whether the results in those files mentioned above are correct or not.

LPTK: Thanks I'll take a look when I can (currently tight with time because of a deadline on July 7th).

shared/src/test/diff/mlscript/Issue65.mls
```diff
@@ -0,0 +1,65 @@
+:js
```
LPTK: Let's rename this file to `i65.mls`.

shared/src/test/diff/mlscript/Under.mls
```diff
@@ -56,12 +56,12 @@ def foo { _ ; _ } = 1
 //│ ║  l.48: 	def foo { _ ; _ } = 1
 //│ ╙──      	              ^
 //│ foo: {_: anything, _: anything} -> 1
-//│    = 1
+//│    = undefined
```
LPTK, line 61: Why is `res` overridden here?

shared/src/main/scala/mlscript/codegen/Scope.scala
```diff
@@ -301,6 +302,15 @@ class Scope(name: Str, enclosing: Opt[Scope]) {
     * Shorthands for deriving function scopes.
     */
   def derive(name: Str, params: Ls[Str]): Scope = Scope(name, params, this)
+  
+  def refreshRes(): Unit = {
```
LPTK, line 306: Why do we need this?

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -606,7 +605,14 @@ class JSTestBackend extends JSBackend {
       case term: Term =>
         try {
           val body = translateTerm(term)(scope)
-          JSTestBackend.CodeQuery(scope.tempVars.emit(), (resultIdent := body) :: Nil)
+          val code = body match {
+            case JSIdent(name) if (name.indexOf('\'') > -1) => (resultIdent := JSIdent(s"globalThis[\"${name}\"]"))
```
LPTK: This looks like a _huge_ hack. Making this logic rely on the fresh name encoding of the `Scope` class is totally brittle (what if we want to change it later?). You should add the correct logic earlier, nearer to the place where this thing is decided.

LPTK (commented):
The outputs look correct. But I'm having trouble figuring out if this implementation is doing the right thing. It looks like it's adding hacks here and there instead of cleanly solving the issue. @chengluyu can you please review and suggest a better way?

shared/src/test/diff/mlscript/Under.mls
NeilKleistGao: Since I removed all `res = VariableName`, we are not able to get the value by querying `res` any longer. Instead, now I search `VariableName` in `globalThis`  directly (here `VariableName` is `foo`).  However, it generates no JavaScript code and that's why it is `undefined`.

shared/src/main/scala/mlscript/codegen/Scope.scala
NeilKleistGao: I use this function to remove all self-defined `res` variables when they should be overwritten by the auto-`res`.

shared/src/test/diff/mlscript/Under.mls
LPTK: Do you know how come it generates no code? If the generation fails, it should at least print a code-gen error. Is am error swallowed somewhere in there?

shared/src/main/scala/mlscript/codegen/Scope.scala
LPTK: Why not just remove `res`? Is it needed to remove all `res$i` variables?

shared/src/test/diff/mlscript/Under.mls
NeilKleistGao: I think it does generate code, but there are some runtime errors:
```js
globalThis.foo5 = function foo5({
  _,
  _
}) {
  return 1;
};
```
It leads to `Error: Duplicate parameter name not allowed in this context`, and that's why `globalThis.foo5` is `undefined`.

In `globalThis.foo6`:
```js
globalThis.foo6 = function foo6({
  _,
  __: _
}) {
  return 1;
};
```
Should it be `__= _` ?

shared/src/main/scala/mlscript/codegen/Scope.scala
NeilKleistGao: I think you are right. Just resetting the `res` is enough.

shared/src/test/diff/mlscript/Under.mls
LPTK: > Should it be __= _ ?

You mean a default parameter? Why?

Generating `{ __ : _ }` is the right thing: a record field named `__` locally being given name `_` (which is already defined).

LPTK: > It leads to Error: Duplicate parameter name not allowed in this context, and that's why globalThis.foo5 is undefined.

Ok, so this is a case of the usual problem that we don't currently catch syntax errors as there is no easy way of doing it. It's fine, but can you please add a comment that explains this is the reason we get `undefined`?

NeilKleistGao: OK. I will do it later.

shared/src/test/diff/mlscript/Under.mls
```diff
@@ -44,24 +44,25 @@ def foo { _ } = { _ }
 //│ res: 1
 //│    = 1
 
+// The code leads to an error: "Duplicate parameter name not allowed in this context", so we got an undefined result here.
```
LPTK:
```suggestion
// * This code leads to a syntax error in NodeJS: "Duplicate parameter name not allowed in this context",
// *  so we got an undefined result here.
```

shared/src/main/scala/mlscript/JSBackend.scala
LPTK: Wait so this logic was not needed at all?

NeilKleistGao: I found something wrong in `shared\src\test\diff\mlscript\BadInherit2.mls`, where:
```
f'
//│ res: 1
//│    = A1 { Foo1: [Function (anonymous)] }
```

Initially, I thought it might be a mistake because `f'` should be `1`, so I added this check to avoid it. However, I found it wrong in the original file:
```
f'
//│ res: 1
//│    = 0
```

So I think maybe it should not print `1` at all, and this check is redundant (or at least should be fixed in another pr).

LPTK: Ah I see! This is actually an instance of https://github.com/hkust-taco/mlscript/issues/65#issuecomment-1168257197

It works without the tick:

```scala
def f_ = 1
//│ f_: 1
//│   = 1

f_
//│ res: 1
//│    = 1
```

We need to replace ticks in JS with some other symbol. This should be doable with a small modification to the scope class. Could you do it in this PR – or in another one if you prefer?

NeilKleistGao: I can fix it in another PR later.  😄

LPTK: Ok cool, please do!

LPTK (approved):
LGTM now!

LPTK:
Thanks @NeilKleistGao

PR #116, opened 2022-07-11
Fix DiffTests line ending and console output on Windows

fo5for:
- DiffTests used to updates all test input files to CRLF on the first run, causing all subsequent to fail to parse. Solved by overriding `java.io.PrintWriter.println()`.
- The console output of DiffTests used to contain random duplicated lines. Solved by using `Predef.println` instead of `Predef.print`.


LPTK:
The random duplicated lines is also a Windows-specific thing, right? I did not notice something like that on MacOS.

fo5for:
Probably, but I can't say for sure since I have no idea why it happened and how this fixed it...

LPTK:
Ok your fix in https://github.com/fo5for/mlscript-core-language works on my computer now! Can you cherry-pick and force-push here, erasing our previosu commits in this branch?

LPTK (approved):
LGTM

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -38,7 +38,9 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
     val fileContents = os.read(file)
     val allLines = fileContents.splitSane('\n').toList
     val strw = new java.io.StringWriter
-    val out = new java.io.PrintWriter(strw)
+    val out = new java.io.PrintWriter(strw) {
+      override def println(): Unit = print('\n')
```
LPTK: Though I'd add a comment to this effect:
```suggestion
      override def println(): Unit = print('\n') // to avoid inserting CRLF on Windows
```

LPTK:
Please squash-merge after addressing the comment.

PR #117, opened 2022-07-13
Replace ticks in JS with another symbol

NeilKleistGao:


LPTK:
I don't think the names of record fields should be affected. In fact, how come they are affected? Are they also handled by the `Scope` abstraction? That seems wrong.

NeilKleistGao:
> I don't thing the names of record fields should be affected. In fact, how come they are affected? Are they also handled by the `Scope` abstraction? That seems wrong.

Did you mean the test file should be like this?
```
:js
a' = {b' = 42}
//│ // Query 1
//│ globalThis["a$"] = { "b'": 42 };
//│ // End of generated code
//│ a': {b': 42}
//│   = { 'b'': 42 }
```

Sorry for not noticing the PR and now I could not recall what the code was like 3 hours ago. :cry:

LPTK:
> Did you mean the test file should be like this?

Yes. Field names are constants whose value is meaningful, and ticks are allowed in them. We can't just change them silently like this.

NeilKleistGao:
> > Did you mean the test file should be like this?
> 
> Yes. Field names are constants whose value is meaningful, and ticks are allowed in them. We can't just change them silently like this.

And is it ok if I directly modify the code in the file `codegen.scala` (just like the current commit)?

LPTK (changes requested):
This looks almost good now, but still not completely right.

What you're doing in `updatePatternName` is still partially wrong, because it changes field names as well as pattern variables.

Consider this:

```fsharp
:js
foo { x' } = x'
//│ // Query 1
//│ globalThis.foo = function foo({ x$ }) {
//│   return x$;
//│ };
//│ // End of generated code
//│ foo: {x': 'a} -> 'a
//│    = [Function: foo]

:js
foo { x' = y' } = y'
//│ // Query 1
//│ globalThis.foo1 = function foo1({ x$: y$ }) {
//│   return y$;
//│ };
//│ // End of generated code
//│ foo: {x': 'a} -> 'a
//│    = [Function: foo1]
```

The JS I actually expect here is:

```js
function foo({ "x'": x$ }) {
  return x$;
}

function foo({ "x'": y$ }) {
  return y$;
}
```

NeilKleistGao:
> Could you document what this does and why it's needed? In particular I have no idea what logic the `!sym.shadowing` and `!sym.accessed` are meant to implement.

Hmmm... I think I just reused the code from the function `declareValue`. Since the original code merely used the lexical name for code generation when translating parameters, it is necessary to replace them with their runtime name. That's why I added the function `getRuntimeName`. But maybe it is also ok if I simply replace ticks in them.

LPTK:
That's extremely brittle!! `accessed` is mutable, so `sym.accessed` may change, and the names you'll produce will be inconsistent!

Please be more careful to understand what's the intended semantics of the code you're modifying, and that what you write is robust and will work in _all_ situations.

What needs to be done here is to properly make use of the scope infrastructure to allocate names. Previously we assumed all names were fine as is, so we used things like `val letScope = scope.derive("Let", name :: Nil)`. This is no longer true because of ticks. So we need to instead make  this `val letScope = scope.derive("Let", Nil); ... letScope. declareValue(name)`, and the tick-sanitization should be happening in `declareValue` before passing it to `lexicalValueSymbols.get `!

NeilKleistGao:
> That's extremely brittle!! `accessed` is mutable, so `sym.accessed` may change, and the names you'll produce will be inconsistent!
> 
> Please be more careful to understand what's the intended semantics of the code you're modifying, and that what you write is robust and will work in _all_ situations.
> 
> What needs to be done here is to properly make use of the scope infrastructure to allocate names. Previously we assumed all names were fine as is, so we used things like `val letScope = scope.derive("Let", name :: Nil)`. This is no longer true because of ticks. So we need to instead make this `val letScope = scope.derive("Let", Nil); ... letScope. declareValue(name)`, and the tick-sanitization should be happening in `declareValue` before passing it to `lexicalValueSymbols.get `!

Sorry for the brittle code. However, I think `val letScope = scope.derive("Let", name :: Nil)` is acceptable because I've processed ticks when the `derive` is called:
```scala
def this(name: Str, params: Ls[Str], enclosing: Scope) = {
  this(name, Opt(enclosing))
  params foreach { param =>
    // TODO: avoid reserved keywords.
    val symbol = ValueSymbol(param, Scope.replaceTicks(param))
    register(symbol)
  }
}
```

The main problem is that I need to **get** the runtime name and use it for code generation. I think my code should only deal with `ValueSymbol` in `lexicalValueSymbols`.

LPTK:
I see. In that case, shouldn't `getRuntimeName` just do `lexicalValueSymbols.getOrElse(lexicalName, lastWords(s"unexpected absence of name ${lexicalName} from scope")).runtimeName`?

NeilKleistGao:
> I see. In that case, shouldn't `getRuntimeName` just do `lexicalValueSymbols.getOrElse(lexicalName, lastWords(s"unexpected absence of name ${lexicalName} from scope")).runtimeName`?

agree.

NeilKleistGao:
> I see. In that case, shouldn't `getRuntimeName` just do `lexicalValueSymbols.getOrElse(lexicalName, lastWords(s"unexpected absence of name ${lexicalName} from scope")).runtimeName`?

agree.

LPTK:
I had another, deeper look. I still don't like how the logic for updating the pattern names is spread across lots of code. There is no reason for this, as it's needlessly error-prone – if we add binding structures, one can easily forget to update the name to remove ticks. Instead, as I said, it would be better to move this logic to `declareValue`. To make this work, simply add a `(implicit scope: Scope)` parameter to `translatePattern` and bind the pattern names there using `scope.declareValue`. Then the `bindings` def in `JSPattern` should become useless and can be removed.

Also, while looking whether all names were properly handling ticks, I found that the current `translateTraitDeclaration` completely forgets to translate method parameters!

```js
:js
trait T : {x': int}
  method N' = this.x'
  method P' y' = this.x' + y'
  method Q ((y', {z'})) = this.x' + y' + z'
//│ Defined trait T
//│ Defined T.N': T -> int
//│ Defined T.P': T -> int -> int
//│ Defined T.Q: T -> (int, {z': int},) -> int
//│ // Prelude
//│ const T = function () {
//│   const tag = Symbol();
//│   return ({
//│     implement: function implement(instance) {
//│       if (tag in instance) {
//│         return;
//│       }
//│       Object.defineProperty(instance, tag, { value: {} });
//│       if (!("N'" in instance)) {
//│         Object.defineProperty(instance, "N'", {
//│           get: function () {
//│             return this["x'"];
//│           }
//│         });
//│       }
//│       if (!("P'" in instance)) {
//│         instance["P'"] = function () {
//│           return this["x'"] + y$;
//│         };
//│       }
//│       if (!("Q" in instance)) {
//│         instance.Q = function () {
//│           return this["x'"] + y$ + z$;
//│         };
//│       }
//│     },
//│     build: function build(instance) {
//│       if (typeof instance !== "object") {
//│         instance = Object.assign(instance, {});
//│       }
//│       this.implement(instance);
//│       return instance;
//│     },
//│     is: function is(x) {
//│       return typeof x === "object" && x !== null && tag in x;
//│     }
//│   });
//│ }();
//│ // End of generated code
```

NeilKleistGao:
> I had another, deeper look. I still don't like how the logic for updating the pattern names is spread across lots of code. There is no reason for this, as it's needlessly error-prone – if we add binding structures, one can easily forget to update the name to remove ticks. Instead, as I said, it would be better to move this logic to `declareValue`. To make this work, simply add a `(implicit scope: Scope)` parameter to `translatePattern` and bind the pattern names there using `scope.declareValue`. Then the `bindings` def in `JSPattern` should become useless and can be removed.
> 
> Also, while looking whether all names were properly handling ticks, I found that the current `translateTraitDeclaration` completely forgets to translate method parameters!
> 
> ```js
> :js
> trait T : {x': int}
>   method N' = this.x'
>   method P' y' = this.x' + y'
>   method Q ((y', {z'})) = this.x' + y' + z'
> //│ Defined trait T
> //│ Defined T.N': T -> int
> //│ Defined T.P': T -> int -> int
> //│ Defined T.Q: T -> (int, {z': int},) -> int
> //│ // Prelude
> //│ const T = function () {
> //│   const tag = Symbol();
> //│   return ({
> //│     implement: function implement(instance) {
> //│       if (tag in instance) {
> //│         return;
> //│       }
> //│       Object.defineProperty(instance, tag, { value: {} });
> //│       if (!("N'" in instance)) {
> //│         Object.defineProperty(instance, "N'", {
> //│           get: function () {
> //│             return this["x'"];
> //│           }
> //│         });
> //│       }
> //│       if (!("P'" in instance)) {
> //│         instance["P'"] = function () {
> //│           return this["x'"] + y$;
> //│         };
> //│       }
> //│       if (!("Q" in instance)) {
> //│         instance.Q = function () {
> //│           return this["x'"] + y$ + z$;
> //│         };
> //│       }
> //│     },
> //│     build: function build(instance) {
> //│       if (typeof instance !== "object") {
> //│         instance = Object.assign(instance, {});
> //│       }
> //│       this.implement(instance);
> //│       return instance;
> //│     },
> //│     is: function is(x) {
> //│       return typeof x === "object" && x !== null && tag in x;
> //│     }
> //│   });
> //│ }();
> //│ // End of generated code
> ```

Sounds great! I will give it a try later and fix the bug in `translateTraitDeclaration`.

LPTK:
Is this ready for review? Looks like the `bindings` defs are still there.

NeilKleistGao:
> Is this ready for review? Looks like the `bindings` defs are still there.

I will fix it as soon as possible.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -33,16 +33,19 @@ class JSBackend {
...
+      val runtimeName = scope.declareParameter(name)
+      JSNamePattern(runtimeName)
+    }
     // fun { x, y } -> ... ==> function ({ x, y }) { ... }
     // should returns ("{ x, y }", ["x", "y"])
     case Rcd(fields) =>
       JSObjectPattern(fields map {
-        case (Var(nme), (Var(als), _)) if nme === als => nme -> N
-        case (Var(nme), (subTrm, _))                  => nme -> S(translatePattern(subTrm))
+        case (Var(nme), (Var(als), _)) if nme === als => scope.declareParameter(nme);(); nme -> N
```
```suggestion
        case (Var(nme), (Var(als), _)) if nme === als => scope.declareParameter(nme); nme -> N
```

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -131,15 +134,14 @@ class JSBackend {
     case Let(true, Var(name), Lam(args, body), expr) =>
       val letScope = scope.derive("Let", name :: Nil)
       val fn = {
-        val params = translateParams(args)
-        val bindings = name :: params.flatMap { _.bindings }
-        val fnScope = scope.derive("Function", bindings)
+        val params = translateParams(args)(letScope)
```
LPTK: This seems incorrect. Why are you translating the parameters in the `letScope` instead of the `fnScope`?

shared/src/main/scala/mlscript/codegen/Scope.scala
```diff
@@ -57,8 +57,7 @@ class Scope(name: Str, enclosing: Opt[Scope]) {
     this(name, Opt(enclosing))
```
LPTK: Is this constructor still used? If it is, shouldn't the code that uses it be updated too? If yes to both, please do so and remove the constructor.

shared/src/main/scala/mlscript/codegen/Scope.scala
```diff
@@ -322,6 +332,8 @@ object Scope {
     new Scope(name, params, enclosing)
 
   private val nameAlphabet: Ls[Char] = Ls.from("abcdefghijklmnopqrstuvwxyz")
+
+  def replaceTicks(str: Str): Str = str.replace('\'', '$')
```
LPTK:
```suggestion
  private def replaceTicks(str: Str): Str = str.replace('\'', '$')
```

shared/src/main/scala/mlscript/JSBackend.scala
LPTK: Why not use the runtime name returned by `declareParameter` here?

shared/src/main/scala/mlscript/codegen/Scope.scala
```diff
@@ -291,8 +292,17 @@ class Scope(name: Str, enclosing: Opt[Scope]) {
     name
   }
 
+  def declareParameter(name: Str): Str = {
+    val symbol = ValueSymbol(name, Scope.replaceTicks(name))
```
LPTK: Should also check for `isValidIdentifier` here.

shared/src/main/scala/mlscript/codegen/Codegen.scala
```diff
@@ -269,42 +270,34 @@ abstract class JSCode {
...
     properties
       .map {
-        case name -> Some(JSWildcardPattern()) => SourceCode(name)
+        case name -> Some(JSWildcardPattern()) => SourceCode(if (JSField.isValidIdentifier(name)) name else JSLit.makeStringLiteral(name))
         case name -> Some(subPattern) =>
-          SourceCode(s"$name: ") ++ subPattern.toSourceCode
-        case name -> N => SourceCode(name)
+          SourceCode(s"${if (JSField.isValidIdentifier(name)) name else JSLit.makeStringLiteral(name)}: ") ++ subPattern.toSourceCode
+        case name -> N => if (JSField.isValidIdentifier(name)) SourceCode(name)
+          else SourceCode(s"${JSLit.makeStringLiteral(name)}: ${Scope.replaceTicks(name)}")
```
LPTK: This name sanitization logic doesn't belong here at all. Here we should assume the names are fine. The `isValidIdentifier` function should be called whenever names are translated in the first place (such as in `declareParameter` AND in other places too).

LPTK (changes requested):
This looks better, but still not quite there! Sorry for dragging this out, but we're making progress and the code is already better.

Also, since we're implementing name sanitization now, could you make sure we never generate JS keywords? For example:

```scala

:js
foo { return } = return
//│ // Prelude
//│ let res;
//│ // Query 1
//│ globalThis.foo = function foo({ return }) {
//│   return return;
//│ };
//│ // End of generated code
//│ foo: {return: 'a} -> 'a
//│    = undefined

foo { return = 1 }
//│ res: 1
//│ Runtime error:
//│   ReferenceError: foo is not defined
```

This can probably be done by simply registering JS keyword names in all scopes, so they're considered already taken.

BTW, please don't force-push your commits, which makes it hard for me to review what's changed since last time. Just push normal commits on top of the previous ones. We'll squash them all while merging anyway.

shared/src/main/scala/mlscript/JSBackend.scala
NeilKleistGao: These are names of record fields, so I just declared and kept using the original names.

NeilKleistGao:
> This can probably be done by simply registering JS keyword names in all scopes, so they're considered already taken.

I think we already have one in `Symbol.scala` and names will be checked in `allocateRuntimeName`:
```scala
if (!runtimeSymbols.contains(realPrefix) && !Symbol.isKeyword(realPrefix)) {
      return realPrefix
}
```

I guess this is because there is another check missing when we declare parameters (both in the new version and in the old version).

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -132,26 +132,28 @@ class JSBackend {
       JSField(translateTerm(receiver), fieldName.name)
     // Turn let into an IIFE.
     case Let(true, Var(name), Lam(args, body), expr) =>
-      val letScope = scope.derive("Let", name :: Nil)
+      val letScope = scope.derive("Let")
+      val runtimeParam = letScope.declareParameter(name)
```
LPTK: Why do you call this `Param`? It's not a parameter; it's the function's name!

LPTK: Oh ok it's because it's the parameter in the following `JSImmEvalFn`. Still, the name is quite confusing.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -47,12 +47,12 @@ class JSBackend {
       JSObjectPattern(fields map {
         case (Var(nme), (Var(als), _)) if nme === als => {
           val runtimeName = scope.declareParameter(nme);
-          if (JSField.isValidIdentifier(nme)) nme -> N
+          if (JSField.isValidIdentifier(nme) && !Symbol.isKeyword(nme)) nme -> N
```
LPTK: Why not move the `isKeyword` check inside `isValidIdentifier`, where it logically belongs? You should really think about how to minimize the possibility for mistakes and omissions by centralizing checking and sanitization...

shared/src/main/scala/mlscript/JSBackend.scala
NeilKleistGao: What about `runtimeName`?

shared/src/main/scala/mlscript/JSBackend.scala
NeilKleistGao: I also did some other refactors. I wonder if it is necessary to move all checks outside the codegen so that all names in codegen are fine.

NeilKleistGao:
Sorry for another force-push. It's just a wrong commits so I covered it. Please ignore it. :cry:

shared/src/main/scala/mlscript/codegen/Codegen.scala
```diff
@@ -607,6 +607,8 @@ object JSField {
   private val identifierPattern: Regex = "^[A-Za-z$][A-Za-z0-9$]*$".r
 
   def isValidIdentifier(s: Str): Bool = identifierPattern.matches(s) && !Symbol.isKeyword(s)
+
+  def makeIdentifierValid(s: Str): Str = if (isValidIdentifier(s)) s else JSLit.makeStringLiteral(s)
```
LPTK: The name of this function is wrong. It's not about making an identifier valid. It's about emitting a valid field name specification.

shared/src/main/scala/mlscript/JSBackend.scala
LPTK: Yes. We need to be consistent. Either we do all name sanitization in code-gen, or none at all. Since some checks are already done before, then better move everything there.

shared/src/main/scala/mlscript/JSBackend.scala
LPTK: Sure.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -395,7 +395,7 @@ class JSBackend {
   private def translateClassMember(
       method: MethodDef[Left[Term, Type]]
   )(implicit scope: Scope): JSClassMemberDecl = {
-    val name = JSField.makeIdentifierValid(method.nme.name)
+    val name = JSField.emitValidFieldName(method.nme.name)
```
LPTK: Why is this called here when the other is called in `Codegen`? This should be called in `Codegen` as well. The name of `method return: T` is `return`, not `"return"`...

shared/src/main/scala/mlscript/JSBackend.scala
NeilKleistGao: > Yes. We need to be consistent. Either we do all name sanitization in code-gen, or none at all. Since some checks are already done before, then better move everything there.

There are still some parts I am not sure about. Take this one for example:
```scala
if (JSField.isValidIdentifier(property.name)) {
  s".${property.name}"
} else {
  s"[${JSLit.makeStringLiteral(property.name)}]"
}
```

Should I move this? I think it is not a sanitization because it just assembles the code based on if the name is valid. However, it does do some checks.

NeilKleistGao: I think I got it. (from another conversation below)

LPTK: Oh sorry I think I was wrong in my previous comment (https://github.com/hkust-taco/mlscript/pull/117#discussion_r923989340) when I stated that `isValidIdentifier` should be moved to the `JSBackend`. I though it had to do with name sanitization, but in fact it doesn't! – it's a codegen detail.

shared/src/main/scala/mlscript/JSBackend.scala
LPTK: Please don't resolve the conversation before I had a chance to see your answer. I would have missed this if I did not manually go through looking at all resolved comments.

> I just declared and kept using the original names

Are you certain the original name will be the same as the one assigned by `declareParameter`? Even if you are, future changes may break this assumption. So why not do the safe thing here?

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -33,16 +34,27 @@ class JSBackend {
     * This function only translate name patterns and object patterns. I was thinking if we can
     * support literal parameter matching by merging multiple function `def`s into one.
     */
-  private def translatePattern(t: Term): JSPattern = t match {
+  private def translatePattern(t: Term)(implicit scope: Scope): JSPattern = t match {
     // fun x -> ... ==> function (x) { ... }
     // should returns ("x", ["x"])
-    case Var(name) => JSNamePattern(name)
+    case Var(name) => {
```
LPTK: Useless braces.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -33,16 +34,27 @@ class JSBackend {
...
+      val runtimeName = scope.declareParameter(name)
+      JSNamePattern(runtimeName)
+    }
     // fun { x, y } -> ... ==> function ({ x, y }) { ... }
     // should returns ("{ x, y }", ["x", "y"])
     case Rcd(fields) =>
       JSObjectPattern(fields map {
-        case (Var(nme), (Var(als), _)) if nme === als => nme -> N
-        case (Var(nme), (subTrm, _))                  => nme -> S(translatePattern(subTrm))
+        case (Var(nme), (Var(als), _)) if nme === als => {
```
LPTK: Useless braces.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -33,16 +34,27 @@ class JSBackend {
...
+          val runtimeName = scope.declareParameter(nme);
+          if (JSField.isValidIdentifier(nme)) nme -> N
+          else JSLit.makeStringLiteral(nme) -> S(JSNamePattern(runtimeName))
+        }
+        case (Var(nme), (subTrm, _)) => {
```
LPTK: Useless braces.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -33,16 +34,27 @@ class JSBackend {
...
+    }
     // fun { x, y } -> ... ==> function ({ x, y }) { ... }
     // should returns ("{ x, y }", ["x", "y"])
     case Rcd(fields) =>
       JSObjectPattern(fields map {
-        case (Var(nme), (Var(als), _)) if nme === als => nme -> N
-        case (Var(nme), (subTrm, _))                  => nme -> S(translatePattern(subTrm))
+        case (Var(nme), (Var(als), _)) if nme === als => {
+          val runtimeName = scope.declareParameter(nme);
+          if (JSField.isValidIdentifier(nme)) nme -> N
```
LPTK: Similar as my previous comment: this logic seems brittle. What if `runtimeName` ended up different than `nme` here?

What we should test here is whether `runtimeName === als`, not whether `nme === als`! And we should _always_ use `runtimeName`.

shared/src/main/scala/mlscript/codegen/Codegen.scala
```diff
@@ -748,12 +739,14 @@ final case class JSFuncDecl(name: Str, params: Ls[JSPattern], body: Ls[JSStmt])
 abstract class JSClassMemberDecl extends JSStmt;
 
 final case class JSClassGetter(name: Str, body: JSExpr \/ Ls[JSStmt]) extends JSClassMemberDecl {
-  def toSourceCode: SourceCode =
-    SourceCode(s"get $name() ") ++ (body match {
+  def toSourceCode: SourceCode = {
```
LPTK: Useless braces.

shared/src/main/scala/mlscript/codegen/Codegen.scala
```diff
@@ -790,8 +784,10 @@ final case class JSClassDecl(
       implements.foreach { name =>
         buffer += s"    $name.implement(this);"
       }
-      fields.foreach { name =>
-        buffer += s"    this.$name = fields.$name;"
+      fields.foreach { name => {
```
LPTK: Useless braces.

shared/src/main/scala/mlscript/codegen/Scope.scala
```diff
@@ -291,17 +282,21 @@ class Scope(name: Str, enclosing: Opt[Scope]) {
     name
   }
 
+  def declareParameter(name: Str): Str = {
+    val prefix = if (JSField.isValidIdentifier(name)) name
+     else if (Symbol.isKeyword(name)) name + "$" else Scope.replaceTicks(name)
```
LPTK: Indentation.
```suggestion
      else if (Symbol.isKeyword(name)) name + "$" else Scope.replaceTicks(name)
```

shared/src/main/scala/mlscript/JSBackend.scala
NeilKleistGao: > Please don't resolve the conversation before I had a chance to see your answer. I would have missed this if I did not manually go through looking at all resolved comments.
> 
> > I just declared and kept using the original names
> 
> Are you certain the original name will be the same as the one assigned by `declareParameter`? Even if you are, future changes may break this assumption. So why not do the safe thing here?

Sorry for using it as a TODO list. :cry: I will check it in detail.

NeilKleistGao: > These are names of record fields, so I just declared and kept using the original names.

And sorry for the unclear description.

Based on the previous discussion, we agreed that the name of the record field should not be changed:
```
:js
foo { x' } = x'
//│ // Query 1
//│ globalThis.foo = function foo({ "x'": x$ }) {
//│   return x$;
//│ };
//│ // End of generated code
//│ foo: {x': 'a} -> 'a
//│    = [Function: foo]
```

Here the `nme` is for `{"x'"}` and that's why I kept using the original one. If I used the `runtimeName`, the test file would be like this:
```
:js
foo { x' } = x'
//│ // Query 1
//│ globalThis.foo = function foo({ "x$": x$ }) {
//│   return x$;
//│ };
//│ // End of generated code
//│ foo: {x': 'a} -> 'a
//│    = [Function: foo]
```

chengluyu:
This PR contains #122.

shared/src/main/scala/mlscript/JSBackend.scala
LPTK: You didn't understand the problem.

I just noticed that @chengluyu independently reimplemented this same change (I don't know why), and he got the logic right. Please have a look:

https://github.com/hkust-taco/mlscript/pull/123/files/e20a99bf6815c0e6b12b9bf98957e13eb16e55b5..dc35036b3b2bef960e29c5eacec188c6c1979fc0#diff-47ab5806a2f49b005e96b51079228893fa2ee6c461349a441f99eb16bbcbd6dfL44-R56
```scala
        case (Var(fieldName), (Var(alias), _)) =>
          val paramSymbol = scope.declareValue(alias)
          if (paramSymbol.runtimeName === fieldName)
            fieldName -> N
          else
            fieldName -> S(JSNamePattern(paramSymbol.runtimeName))
```
(EDIT: I had pasted the wrong code)

NeilKleistGao: > You didn't understand the problem.
> 
> I just noticed that @chengluyu independently reimplemented this same change (I don't know why), and he got the logic right. Please have a look:
> 
> https://github.com/hkust-taco/mlscript/pull/123/files/e20a99bf6815c0e6b12b9bf98957e13eb16e55b5..dc35036b3b2bef960e29c5eacec188c6c1979fc0#diff-47ab5806a2f49b005e96b51079228893fa2ee6c461349a441f99eb16bbcbd6dfL44-R56
> 
> ```scala
>         case (Var(fieldName), (Var(alias), _)) =>
>           val paramSymbol = scope.declareValue(alias)
>           if (paramSymbol.runtimeName === fieldName)
>             fieldName -> N
>           else
>             fieldName -> S(JSNamePattern(paramSymbol.runtimeName))
> ```
> 
> (EDIT: I had pasted the wrong code)

Ah! I think I got it!

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -33,16 +34,24 @@ class JSBackend {
...
       JSObjectPattern(fields map {
-        case (Var(nme), (Var(als), _)) if nme === als => nme -> N
-        case (Var(nme), (subTrm, _))                  => nme -> S(translatePattern(subTrm))
+        case (Var(nme), (Var(als), _)) =>
+          val runtimeName = scope.declareParameter(als)
+          val fieldName = JSField.emitValidFieldName(nme)
+          if (runtimeName === fieldName) fieldName -> N
+          else fieldName -> S(JSNamePattern(runtimeName))
+        case (Var(nme), (subTrm, _)) => 
+          scope.declareParameter(nme)
```
LPTK: What's the purpose of this line?!

LPTK:
A couple of things:

1. Before you request a review, please take a detailed look at ALL your diff, and check all the logic you changed. There has been way too much back-and-forth in this PR. I feel like I've been wasting a lot of time. But I take it as a learning experience for you.

2. You _can_ resolve conversations where you made the suggested fix, as long as there's nothing for me to see or confirm. And if there is something I should see, you should comment on the comment thread to let me know.

shared/src/main/scala/mlscript/JSBackend.scala
NeilKleistGao: Removed.

NeilKleistGao:
Sorry for a series of back-and-forth. :cry: I'll pay more attention to these details in future work.

LPTK:
No problem. I have just clarified my expectations regarding contributions and PR here: https://github.com/hkust-taco/mlscript/blob/b6fc64e70a7833296686206a5c4517cc3608c269/CONTRIBUTING.md

LPTK (approved):
Okay, this finally LGTM, thanks!

NeilKleistGao:
Finally! I did learn a lot from this PR! Thank you for your patience and guidance.

PR #121, opened 2022-07-16
 Feat/timeout for stream read

mahzoun99:


mahzoun99:
@LPTK Please [check](https://github.com/hkust-taco/mlscript/runs/7371482686?check_suite_focus=true)
a Test failed with 10 seconds of timeout. (10sec was for each file)
And another one failed because of Unexpected diagnostics (or lack thereof) (idk why)

LPTK:
Does it work on your machine? 10 seconds seems like a REALLY long time for these tests to take... Maybe a quirk of the CI's environment?

Where is the "Unexpected diagnostics"? It doesn't show in the report.

mahzoun99:
> Does it work on your machine? 10 seconds seems like a REALLY long time for these tests to take... Maybe a quirk of the CI's environment?

Yes, it worked. But that warning we decided to ignore said ignore it if u don't care about the performance.


> Where is the "Unexpected diagnostics"? It doesn't show in the report.

Probably I saw another log by mistake for a moment. Both are timeout exceptions.

LPTK:
Weird, I re ran the test and now it says "Unexpected diagnostics". I'll have a look soon.

LPTK:
So I looked into this and ended up using a different design, relying on ScalaTest to implement a cleaner timeout and thread-killing mechanism. See the PR at https://github.com/hkust-taco/mlscript/pull/126.

It turns out the CI is sometimes extremely slow and some tests just take forever.

PR #122, opened 2022-07-19
Fix incorrect code generation due to single quotes in identifiers

chengluyu:
Consider single quotes in identifiers in `def allocateRuntimeName(rawPrefix: Str): Str`.


LPTK:
Argh... @NeilKleistGao has already been working on this specific thing in a PR which also now fixes other things: https://github.com/hkust-taco/mlscript/pull/117

Can you review that PR and make sure the changes are good and encompass what's done here?

PR #123, opened 2022-07-19
Fix `this` being captured by immediately invoked functions

chengluyu:
This is mentioned in #65 .

```
:js
class A: { x: int }
  method MA = let _ = 1 in this.x
//│ Defined class A
//│ Defined A.MA: A -> int
//│ // Prelude
//│ class A {
//│   constructor(fields) {
//│     this.x = fields.x;
//│   }
//│   get MA() {
//│     return ((function (_) {
//│       return this.x;
//│     })(1));
//│   }
//│ }
//│ // End of generated code

(A { x = 1 }).MA
//│ res: int
//│ Runtime error:
//│   TypeError: Cannot read properties of undefined (reading 'x')
```

### DIFF
--- a/shared/src/main/scala/mlscript/JSBackend.scala
+++ b/shared/src/main/scala/mlscript/JSBackend.scala
@@ -81,7 +81,9 @@ class JSBackend {
           JSIdent(sym.runtimeName)
         else
           throw new UnimplementedError(sym)
-      case S(sym: ValueSymbol) => JSIdent(sym.runtimeName)
+      case S(sym: ValueSymbol) =>
+        sym.accessed = true
+        JSIdent(sym.runtimeName)
       case S(sym: ClassSymbol) =>
         if (isCallee)
           JSNew(JSIdent(sym.runtimeName))
@@ -388,20 +390,37 @@ class JSBackend {
     JSClassDecl(classSymbol.runtimeName, fields, base, members, traits)
   }
 
+  /**
+   * Translate class methods and getters.
+   */
   private def translateClassMember(
       method: MethodDef[Left[Term, Type]]
   )(implicit scope: Scope): JSClassMemberDecl = {
     val name = method.nme.name
-    method.rhs.value match {
+    // Collect parameters and create the member scope.
+    val (memberParams, memberScope, body) = method.rhs.value match {
       case Lam(params, body) =>
-        val methodScope = scope.derive(s"Method $name")
+        val methodScope = scope.derive("method $name")
         val methodParams = translateParams(params)(methodScope)
-        methodScope.declareValue("this")
-        JSClassMethod(name, methodParams, L(translateTerm(body)(methodScope)))
+        (S(methodParams), methodScope, body)
       case term =>
-        val getterScope = scope.derive(s"Getter $name")
-        getterScope.declareValue("this")
-        JSClassGetter(name, L(translateTerm(term)(getterScope)))
+        (N, scope.derive(s"Getter $name"), term)
+    }
+    // Declare a value symbol for `this`.
+    val thisSymbol = memberScope.declareThis()
+    // Translate class member body.
+    val bodyResult = translateTerm(body)(memberScope).`return`
+    // If `this` is accessed, add `const self = this`.
+    val bodyStmts = if (thisSymbol.accessed) {
+      val thisDecl = JSConstDecl(thisSymbol.runtimeName, JSIdent("this"))
+      R(thisDecl :: bodyResult :: Nil)
+    } else {
+      R(bodyResult :: Nil)
+    }
+    // Returns members depending on what it is.
+    memberParams match {
+      case S(memberParams) => JSClassMethod(name, memberParams, bodyStmts)
+      case N => JSClassGetter(name, bodyStmts)
     }
   }
 
--- a/shared/src/main/scala/mlscript/JSBackend.scala
+++ b/shared/src/main/scala/mlscript/JSBackend.scala
@@ -411,7 +411,6 @@ class JSBackend {
       case _ => scope.derive(s"getter $name")
     }
     // Declare the alias for `this` before declaring parameters.
-    // TODO: Remove this after we don't use IIFEs.
     val selfSymbol = memberScope.declareThisAlias()
     // Declare parameters.
     val (memberParams, body) = method.rhs.value match {
--- a/shared/src/main/scala/mlscript/JSBackend.scala
+++ b/shared/src/main/scala/mlscript/JSBackend.scala
@@ -5,6 +5,7 @@ import mlscript.codegen.Helpers._
 import mlscript.codegen._
 import scala.collection.mutable.ListBuffer
 import mlscript.{JSField, JSLit}
+import scala.collection.mutable.{Set => MutSet}
 
 class JSBackend {
   /**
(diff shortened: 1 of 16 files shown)


LPTK:
Cool!

But does this really solve the problem in general? What if there is a captured `this` in a generated local `function`?

chengluyu:
Sadly this PR only works in this particular case.

The root of the problem is we implement `let` and pattern matches with IIFEs. IIFEs introduce extra `function` scopes in generated code, which shadow `this`. Although we can keep track of runtime scopes and add alias variable declarations but that would be cumbersome.

A better solution is to avoid IIFEs and generate a series of JavaScript statements.

LPTK:
Why not just add a `const self = this` at the beginning of methods and then refer to this `self` instead? It should be a trivial change that would avoid mistranslating the semantics of MLscript.

LPTK:
Of course we need to make sure that `self` itself is not shadowed. I think your `Scope` infrastructure should be able to deal with that.

Also, if the `self` symbol is never accessed in the method body, we can omit the binding in the generated code.

chengluyu:
It's done.

LPTK:
Please ad some tests making sure that this `self` cannot be shadowed.

chengluyu:
I added some test cases but found that shadowing `this` caused problems. The reason is that we didn't allocate runtime names for parameters (as well as parameter patterns) in `translatePattern`. The latest commit fixes this. I also added test cases for illegal parameters.

LPTK:
Ok, now that https://github.com/hkust-taco/mlscript/pull/117 is merged you can adapt your changes here and we can merge it.

chengluyu:
I rebased the branch. This PR is ready.

shared/src/main/scala/mlscript/codegen/Scope.scala
```diff
@@ -51,6 +51,16 @@ class Scope(name: Str, enclosing: Opt[Scope]) {
     }
   }
 
+  /**
+    * Shorthands for creating function scopes.
+    */
+  def this(name: Str, params: Ls[Str], enclosing: Scope) = {
```
LPTK: Is this used?

shared/src/test/diff/mlscript/TypeClasses.mls
```diff
@@ -1,4 +1,3 @@
-
```
LPTK: Let's not add useless space changes.

LPTK (approved):
Ok, LGTM.

But it actually occurred to me that we could also define `self` in the class scope, instead of in each method, which would produce leaner code. As in:

```js
class C { self = this; ... }
```

Would that be easy to change now?

shared/src/test/diff/mlscript/TypeClasses.mls
LPTK: (I actually dislike when tests start immediately on the first line. We separate blocks by empty lines, so this feels wrong.)

chengluyu:
> Ok, LGTM.
> 
> But it actually occurred to me that we could also define `self` in the class scope, instead of in each method, which would produce leaner code. As in:
> 
> ```js
> class C { self = this; ... }
> ```
> 
> Would that be easy to change now?

There had been some trouble implementing this. Now I think `self = this` is a temporary and ugly fix. It will be unnecessary after we remove IIFEs. So let’s not do it right now.

shared/src/main/scala/mlscript/codegen/Scope.scala
```diff
@@ -51,16 +51,6 @@ class Scope(name: Str, enclosing: Opt[Scope]) {
...
-  /**
-    * Shorthands for creating function scopes.
-    */
-  def this(name: Str, params: Ls[Str], enclosing: Scope) = {
-    this(name, Opt(enclosing))
-    params foreach { param =>
-      declareValue(param)
-    }
-  }
-
```
LPTK: See, this is exactly what I was telling you about being careful with merges... It is not enough to just click "accept both". Please be more mindful next time.

LPTK:
> Now I think self = this is a temporary and ugly fix. It will be unnecessary after we remove IIFEs.

I disagree. JS semantics is just stupid and we should NOT follow it, so the self-fix will be needed in general.

The following MLscript (using the future syntax):

```fsharp
class A {
  fun m() =
    fun f() = this.x
    f
}
```
should generate this JS:

```js
class A {
  m() {
    let self = this
    function f() { return self.x }
    return f
  }
}
```

or alternatively:

```js
class A {
  #self = this
  m() {
    function f() { return #self.x }
    return f
  }
}
```

But it's okay if we don't do the latter for now if it's too annoying to implement.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -388,20 +398,43 @@ class JSBackend {
...
   )(implicit scope: Scope): JSClassMemberDecl = {
     val name = method.nme.name
-    method.rhs.value match {
+    // Create the method/getter scope.
+    val memberScope = method.rhs.value match {
+      case _: Lam => scope.derive(s"method $name")
+      case _ => scope.derive(s"getter $name")
+    }
+    // Declare the alias for `this` before declaring parameters.
+    // TODO: Remove this after we don't use IIFEs.
```
```suggestion
```

shared/src/main/scala/mlscript/codegen/Scope.scala
chengluyu: 👀 Sorry, I didn't get. I deleted this because of [this comment](https://github.com/hkust-taco/mlscript/pull/123#discussion_r927398129). I didn't click any “accept both” button we deleting this.

chengluyu:
> > Now I think self = this is a temporary and ugly fix. It will be unnecessary after we remove IIFEs.
> 
> I disagree. JS semantics is just stupid and we should NOT follow it, so the self-fix will be needed in general.
> 
> The following MLscript (using the future syntax):
> 
> ```fsharp
> class A {
>   fun m() =
>     fun f() = this.x
>     f
> }
> ```
> 
> should generate this JS:
> 
> ```js
> class A {
>   m() {
>     let self = this
>     function f() { return self.x }
>     return f
>   }
> }
> ```
> 
> or alternatively:
> 
> ```js
> class A {
>   #self = this
>   m() {
>     function f() { return #self.x }
>     return f
>   }
> }
> ```
> 
> But it's okay if we don't do the latter for now if it's too annoying to implement.

This is another case. It’s not IIFE. In your case, we should create alias because of the nested function `f`. It should only be created when the body of `f` refers to `this`.

But now we add the alias anyway. That's why I say it’s ugly because the generated code is not easy to understand. In the future, we will deprecate IIFEs for pattern matching and `let` expressions. Then we can remove the fix in this PR. By the time when the nested function syntax arrives, we should add anther mechanism which prevents inner functions shadowing `this`.

shared/src/main/scala/mlscript/codegen/Scope.scala
LPTK: The point is that this constructor was deleted on MLScript's main branch. It was deleted in https://github.com/hkust-taco/mlscript/pull/117, on which you rebased your changes. You reintroduced it because you botched your merge. Please be more careful.

LPTK:
> By the time when the nested function syntax arrives, we should add anther mechanism which prevents inner functions shadowing this.

But why add another mechanism? It's the same problem. The fix could be _improved_ later, but there is no reason to just remove it.

PR #127, opened 2022-07-29
Display syntax errors and REPL history in `DiffTests`

chengluyu:
This PR makes several improvements on code generation error reporting.

- When `:ShowRepl` is on, shows history in test files rather than `stdout`.
- The reply to preludes (classes) is not shown previously because `consumeUntilPrompt` discarded the output. Now it shows.
- Now we extract syntax errors from the Node.js output and display them.

By the way, I need some examples that legally cause syntax errors.

### DIFF
--- a/shared/src/test/scala/mlscript/DiffTests.scala
+++ b/shared/src/test/scala/mlscript/DiffTests.scala
@@ -15,6 +15,9 @@ import mlscript.JSTestBackend.TestCode
 import mlscript.codegen.typescript.TsTypegenCodeBuilder
 import org.scalatest.time._
 import org.scalatest.concurrent.{TimeLimitedTests, Signaler}
+import mlscript.DiffTests.ReplHost.Result
+import mlscript.DiffTests.ReplHost.Empty
+import mlscript.DiffTests.ReplHost.Unexecuted
 
 
 class DiffTests
@@ -414,15 +417,15 @@ class DiffTests
 
             final case class ExecutedResult(var replies: Ls[ReplHost.Reply]) extends JSTestBackend.Result {
               def showFirst(prefixLength: Int): Unit = replies match {
-                case ReplHost.Error(err) :: rest =>
+                case ReplHost.Error(isSyntaxError, content) :: rest =>
                   if (!(mode.expectTypeErrors
                       || mode.expectRuntimeErrors
                       || allowRuntimeErrors
                       || mode.fixme
                   )) failures += blockLineNum
                   totalRuntimeErrors += 1
-                  output("Runtime error:")
-                  err.split('\n') foreach { s => output("  " + s) }
+                  output((if (isSyntaxError) "Syntax" else "Runtime") + " error:")
+                  content.linesIterator.foreach { s => output("  " + s) }
                   replies = rest
                 case ReplHost.Unexecuted(reason) :: rest =>
                   output(" " * prefixLength + "= <no result>")
@@ -488,8 +491,14 @@ class DiffTests
                             output(s"│ │   $line")
                           }
                           output(s"│ └── Reply")
-                          preludeReply.linesIterator.zipWithIndex.foreach { case (line, index) =>
-                            output(s"│     $line")
+                          // Display successful results in multiple lines.
+                          // Display other results in single line.
+                          preludeReply match {
+                            case Result(content) =>
+                              content.linesIterator.zipWithIndex.foreach {
+                                case (line, index) => output(s"│     $line")
+                              }
+                            case other => output(s"│     $other")
                           }
                         }
                       }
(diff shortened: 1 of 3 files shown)


chengluyu:
This PR is ready. @LPTK

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -446,41 +446,76 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
...
+                        val preludeReply = host.execute(lines mkString " ")
+                        if (mode.showRepl) {
+                          output(s"├─┬ Prelude")
+                          output(s"│ ├── Code")
+                          lines.iterator.foreach { line => output(s"│ │   $line") }
+                          // Display successful results in multiple lines.
+                          // Display other results in single line.
+                          preludeReply match {
+                            case ReplHost.Result(content, intermediate) =>
+                              intermediate.foreach { value =>
```
LPTK, line 492: `value` is not used here. Seems like a bug.

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -446,41 +446,76 @@ class DiffTests extends org.scalatest.funsuite.AnyFunSuite with org.scalatest.Pa
...
                         val prelude = preludeLines.mkString
                         val code = codeLines.mkString
+                        val p0 = if (i + 1 == queries.length) "  " else "│ "
                         if (mode.showRepl) {
-                          println(s"├── Query ${i + 1}/${queries.length}")
-                          println(s"│ ├── Prelude: ${if (preludeLines.isEmpty) "<empty>" else prelude}")
-                          println(s"│ └── Code: ${code}")
+                          val p1 = if (i + 1 == queries.length) "└─" else "├─"
+                          output(s"$p1┬ Query ${i + 1}/${queries.length}")
+                          output(s"$p0├── Prelude: ${if (preludeLines.isEmpty) "<empty>" else prelude}")
```
LPTK, line 515: It seems that this is always empty in practice (see tests). If it is, better remove this stuff. Otherwise, when is it not empty? Can we have a test for it?

shared/src/test/scala/mlscript/ReplHost.scala
```diff
@@ -0,0 +1,193 @@
...
+    override def toString(): Str = s"[unexecuted] $message"
+  }
+
+  /**
+    * If the `ReplHost` captured errors, it will response with `Error`.
+    * @param syntax if `true`, this is a syntax error; otherwise, this is a
+    *               runtime error
+    * @param message the error message
+    */
+  final case class Error(syntax: Bool, message: Str) extends Reply {
```
LPTK, line 185: Why the special-casing for syntax errors?

If you read my message again, I suggested testing for `buffer.endsWith("\nUncaught "`. It would be better to do that and then place the end of that line (after the `Uncaught `) inside an optional error string (instead of a boolean).

This way we'll always know if something goes wrong, syntax-wise or not.

LPTK (changes requested):
Nice improvement. But I have a couple small change requests.

shared/src/test/scala/mlscript/ReplHost.scala
chengluyu: Because we wrap code in `try`-`catch` statements, none-syntax errors will be caught by them. Code with syntax errors cannot be parsed, so `try`-`catch` statements cannot caught.

https://github.com/hkust-taco/mlscript/blob/6d5d382f7b314fecf7692a4aad2e63f416308597/shared/src/test/scala/mlscript/ReplHost.scala#L94

shared/src/test/scala/mlscript/DiffTests.scala
chengluyu: I added a test case indicating its usage in 5bf8d35.

shared/src/test/scala/mlscript/DiffTests.scala
chengluyu: Thank you so much! It’s a bug caused by copy'n'paste.

shared/src/test/scala/mlscript/ReplHost.scala
LPTK: Ok but are you 100% this try-catch wrapper is completely impervious to any sort of errors other than syntax errors? Does it really catch everything?

Also, is there a reason why we don't just rely on parsing `"Uncaught "` in the output to catch errors and exceptions instead of using the try/catch + `console.log('\\u200B' + e + '\\u200B')` hack?

LPTK: It's fine if there is no reason and this is just historical. Changing this is outside the scope of this PR.

chengluyu: I was thinking that, compared with `"Uncaught "`, invisible characters like [ZWSP](https://en.wikipedia.org/wiki/Zero-width_space) is rarer in the output. But as we detect syntax errors by searching for `"Uncaught "`, it is unnecessary to maintain this hack. We can remove this hack and detect errors in a unified and simple way in another PR.

chengluyu: Do you agree to merge this PR first and then remove this hack in the next PR?

LPTK: Yes that sounds good to me. In fact, removing the hack is low priority. Let's not do it right now – you have more urgent things to do now.

shared/src/test/scala/mlscript/ReplHost.scala
```diff
@@ -0,0 +1,193 @@
+package mlscript
+
+import mlscript.utils.shorthands._
+
+/**
+ * A helper class manipulate an interactive Node.js process.
```
LPTK: Grammar is wrong. Did you mean this?
```suggestion
 * A helper class to manipulate an interactive Node.js process.
```

LPTK (approved):
LGTM. The commits look clean, so please rebase-merge the PR after you've handled my last comment.

PR #132, opened 2022-08-31
Implement desugaring and code generation for ultimate conditional syntax

chengluyu:


LPTK:
You understand `typeTerm(body.desugar(otherwise))` is not the right thing to do, right? It doesn't fix anything we talked about.

shared/src/test/diff/nu/Match1.mls
```diff
@@ -0,0 +1,71 @@
...
+    Foo then 0
+    Bar then 1
+//│ f: (Bar | Foo) -> (0 | 1)
+//│  = [Function: f]
+
+fun f(x) =
+  if x
+    is Foo then 0
+    is Bar then 1
+//│ f: nothing -> (0 | 1 | undefined)
```
LPTK, line 64: What's the desugaring here? Seems it produces two cases instead of one, hence the `nothing`.

shared/src/test/diff/nu/Match2.mls
```diff
@@ -0,0 +1,120 @@
...
+    _ then 1
+//│ foo: ((Test with {value: 'value}) | ~Test) -> (1 | 'value)
+//│    = [Function: foo]
+
+let t = Test(0)
+//│ t: Test & {value: 0}
+//│  = Test { value: 0 }
+
+// FIXME: The evaluation result should not be `undefined`.
+foo of Test
```
LPTK, line 24: BTW, `Test` would be the class or its constructor here, not an instance.

PR #129, opened 2022-08-24
TypeScript to MLscript

NeilKleistGao:

PR #133, opened 2022-09-01
TypeScript to MLscript

NeilKleistGao:
A cleaner version of #129 without force-pushing.


ts2mls/js/src/main/scala/ts2mls/types/TSType.scala
```diff
@@ -0,0 +1,50 @@
...
+
+case class TSClassType(name: String,
+                       members: Map[String, TSMemberType],
+                       statics: Map[String, TSMemberType],
+                       typeVars: List[TSTypeParameter],
+                       parents: List[TSType]) extends TSType
+case class TSInterfaceType(name: String,
+                           members: Map[String, TSMemberType],
+                           typeVars: List[TSTypeParameter],
+                           parents: List[TSType]) extends TSType
```
LPTK: Horizontal alignment is an anti-pattern. Better do:

```suggestion
case class TSClassType(
    name: String,
    members: Map[String, TSMemberType],
    statics: Map[String, TSMemberType],
    typeVars: List[TSTypeParameter],
    parents: List[TSType],
  ) extends TSType

case class TSInterfaceType(
    name: String,
    members: Map[String, TSMemberType],
    typeVars: List[TSTypeParameter],
    parents: List[TSType],
  ) extends TSType
```

build.sbt
```diff
@@ -57,3 +57,27 @@ lazy val mlscript = crossProject(JSPlatform, JVMPlatform).in(file("."))
...
+  .dependsOn(mlscript % "compile->compile;test->test")
+
+lazy val ts2mlsJS = ts2mls.js
+lazy val ts2mlsJVM = ts2mls.jvm
+
+val tsTypeDiffTests = taskKey[Unit]("")
+
+tsTypeDiffTests := (Def.task{
+  (ts2mlsJVM / Test / test).value
+} triggeredBy (ts2mlsJS / Test / test)).value
```
LPTK:
```suggestion
} triggeredBy (ts2mlsJS / Test / test)).value

```

ts2mls/js/src/main/scala/ts2mls/JSWriter.scala
```diff
@@ -0,0 +1,40 @@
...
+  }
+}
+
+object JSWriter {
+  private val fs = g.require("fs") // must use fs module to manipulate files in JS
+
+  def apply(filename: String) = new JSWriter(filename)
+
+  private def createBuffer(length: Int) = g.Buffer.alloc(length)
+}
```
LPTK:
```suggestion
}

```

ts2mls/js/src/main/scala/ts2mls/JSWriter.scala
```diff
@@ -0,0 +1,40 @@
...
+  private var fileSize = 0 // how many bytes we've written in the file
+  private var needTruncate = false
+
+  def writeln(str: String) = {
+    val strln = str + "\n"
+    val buffer = createBuffer(strln.length)
+    fs.readSync(out, buffer, 0, strln.length)
+    
+    // override when the content is different
+    if (!strln.equals(buffer.toString())) {
```
LPTK: Style: use `===` after importing `mlscript.utils._` instead.

ts2mls/js/src/main/scala/ts2mls/TSCompilerInfo.scala
```diff
@@ -0,0 +1,170 @@
...
+
+  // the first declaration of this symbol
+  // if there is no overloading, there is only one declaration
+  lazy val declaration =
+    if (declarations.isUndefined) TSNodeObject(g.undefined)
+    else declarations.get(0)
+  lazy val escapedName: String = sym.escapedName.toString
+  lazy val declarations = TSNodeArray(sym.declarations)
+
+  lazy val symbolType: String = checker.getTypeStringOfSymbol(sym)
```
LPTK: Can you add some documentation? Why do we need the string of a type here?

ts2mls/js/src/main/scala/ts2mls/TSSourceFile.scala
```diff
@@ -0,0 +1,145 @@
...
+    else if (node.isClassDeclaration)
+      ns.put(name, parseMembers(ns.getFullPath(name), node, true))
+    else if (node.isInterfaceDeclaration)
+      ns.put(name, parseMembers(ns.getFullPath(name), node, false))
+    else if (node.isNamespace)
+      parseNamespace(node)
+
+  private def parseNamespace(node: TSNodeObject)(implicit ns: TSNamespace): Unit =
+    parseNamespaceLocals(node.locals)(ns.derive(node.symbol.escapedName))
+}
```
LPTK:
```suggestion
}

```

ts2mls/js/src/main/scala/ts2mls/TSProgram.scala
```diff
@@ -0,0 +1,23 @@
...
+  
+  implicit val checker = TSTypeChecker(program.getTypeChecker())
+  filenames.foreach(filename => TSSourceFile(program.getSourceFile(filename), globalNamespace))
+
+  def generate(writer: JSWriter): Unit = globalNamespace.generate(writer)
+}
+
+object TSProgram {
+  def apply(filenames: Seq[String]) = new TSProgram(filenames)
+}
```
LPTK:
```suggestion
}

```

ts2mls/js/src/main/scala/ts2mls/TSCompilerInfo.scala
```diff
@@ -0,0 +1,170 @@
...
+  def apply(obj: js.Dynamic)(implicit checker: TSTypeChecker) = new TSTypeObject(obj)
+}
+
+class TSIdentifierObject(id: js.Dynamic) extends TSAny(id) {
+  lazy val escapedText: String = id.escapedText.toString()
+}
+
+object TSIdentifierObject {
+  def apply(id: js.Dynamic) = new TSIdentifierObject(id)
+}
```
LPTK:
```suggestion
}

```

ts2mls/js/src/main/scala/ts2mls/TSCompilerInfo.scala
```diff
@@ -0,0 +1,170 @@
...
+  // `TypeScript.isModuleDeclaration` works on both namespaces and modules
+  // but namespaces are more recommended, so we merely use `isNamespace` here
+  lazy val isNamespace = TypeScript.isModuleDeclaration(node)
+  lazy val isArrayTypeNode = TypeScript.isArrayTypeNode(node)
+  lazy val isTupleTypeNode = TypeScript.isTupleTypeNode(node)
+  lazy val isImplementationOfOverload = checker.isImplementationOfOverload(node)
+
+  // if a node has an initializer or is marked by a question notation
+  // it is ok if we provide nothing
+  lazy val isOptional = !initializer.isUndefined || !IsUndefined(node.questionToken)
```
LPTK: What does "we provide nothing" mean here?

Also, why the `!initializer.isUndefined` check?

ts2mls/js/src/main/scala/ts2mls/TSData.scala
```diff
@@ -0,0 +1,75 @@
...
+
+class TSSymbolMap(map: js.Dynamic)(implicit checker: TSTypeChecker) extends TSAny(map) {
+  def foreach(f: TSSymbolObject => Unit): Unit =
+    if (!isUndefined)
+      map.forEach({(s: js.Dynamic) => f(TSSymbolObject(s))})
+}
+
+object TSSymbolMap {
+  def apply(map: js.Dynamic)(implicit checker: TSTypeChecker) = new TSSymbolMap(map)
+}
```
LPTK:
```suggestion
}

```

ts2mls/js/src/main/scala/ts2mls/types/TSType.scala
```diff
@@ -0,0 +1,50 @@
...
+        case i: TSIntersectionType => getOverloadTypeParameters(i)
+        case TSFunctionType(_, _, v) => v
+        case _ => List()
+      }) :::
+      (rhs match {
+        case i: TSIntersectionType => getOverloadTypeParameters(i)
+        case TSFunctionType(_, _, v) => v
+        case _ => List()
+      })).distinct
+    case _ => List()
```
LPTK: Isn't this case unreachable?

ts2mls/js/src/main/scala/ts2mls/TSNamespace.scala
```diff
@@ -0,0 +1,74 @@
...
+
+  def generate(writer: JSWriter): Unit =
+    order.toList.foreach((p) => p match {
+      case Left(name) => subSpace(name).generate(writer)
+      case Right(name) => {
+        val mem = members(name)
+        val fullName = getFullPath(name)
+        mem match {
+          case inter: TSIntersectionType => { // overloaded functions
+            val typeParams = TSIntersectionType.getOverloadTypeParameters(inter).map((t) => t.name)
```
LPTK: The logic of `getOverloadTypeParameters` seems very wrong. So if `foo: (<A>() -> A) & (<B>() -> B)` you'll get `List(A, B)` as the result...!

LPTK (changes requested):
I didn't review all the details but this seems like a significant piece of work. Nice!

However I think I found some issues.

Also, please make sure your files always end with an empty line.

ts2mls/js/src/main/scala/ts2mls/JSWriter.scala
NeilKleistGao: I've replaced all `something.equals` and `!something.equals` with `===` and `=/=`. Is it necessary to use abbreviation defined in `shorthands.scala` either?

ts2mls/js/src/main/scala/ts2mls/types/TSType.scala
NeilKleistGao: I think so

ts2mls/js/src/main/scala/ts2mls/TSCompilerInfo.scala
NeilKleistGao: This name seems ambiguous. I've changed it and added some explanations.

ts2mls/js/src/main/scala/ts2mls/TSCompilerInfo.scala
NeilKleistGao: ohh! some dumbass comments. :dizzy_face: I've updated it.

NeilKleistGao: e.g. `function f(x?: int) {}`, we can use it directly: `f()`.
in this case, there would be a question token.
e.g. `function f(x: int = 42) {}`, we can use it directly: `f()`.
in this case, the initializer would store the default value and be non-empty.

ts2mls/js/src/main/scala/ts2mls/TSNamespace.scala
NeilKleistGao: seems code like this is also acceptable:
```ts
function baz<T>(): T;
function baz<U, T>(): U;
function baz<U, T, P>(): P;
function baz() {
  // do something here
}
```

any better idea to deal with it? :worried:

ts2mls/js/src/main/scala/ts2mls/JSWriter.scala
LPTK: No it's not necessary, but you can do it at your own convenience.

ts2mls/js/src/main/scala/ts2mls/TSNamespace.scala
LPTK: > seems code like this is also acceptable

Of course; that's the point!

You should NEVER EVER return the wrong result out of convenience when you don't know what to do. I can't emphasize this enough. Correctness comes first, otherwise there's no way anyone can keep their sanity long term.

I mentioned that ts-to-mls should be robust and resilient, but not incorrect in doing so. For now, it's fine to reject this. Maybe we can raise a warning somewhere, and simply ignore the overloads except for the last one (which is usually the most general).

ts2mls/js/src/main/scala/ts2mls/types/TSType.scala
LPTK: It's bad style to add unreachable cases when they can be avoided.

ts2mls/js/src/main/scala/ts2mls/TSCompilerInfo.scala
LPTK: I see. This makes me wonder why we don't directly look at the _type_ of the definition, as determined by TS, instead of looking at the definition tree. TS will resolve all these details on its own already. Can't we just query the type of `f` here?

NeilKleistGao: Actually, I've tried to use `TypeObject` in the code as much as possible, and I think that is the information of type determined by TSC.

ts2mls/js/src/main/scala/ts2mls/TSNamespace.scala
NeilKleistGao: Thank you for your advice! I think I can also add some comments after the generated type definitions so that we can find them easily.

NeilKleistGao: So far, the output would be like this:
```ts
def XX'f[T]: (T) -> ((anything) -> (string)) /* warning: the overload of function f may be not supported yet */
```
whose form is the most general one

ts2mls/js/src/main/scala/ts2mls/TSCompilerInfo.scala
LPTK: So why not use it here too? Do you remember any specific problem while doing so? If possible that would be better, as it would make the code less brittle.

ts2mls/js/src/main/scala/ts2mls/TSNamespace.scala
LPTK: Ok, looks good! Except "may not be" -> "is not".

ts2mls/js/src/main/scala/ts2mls/TSCompilerInfo.scala
LPTK: I'm still quite confused by both the new name and the comment. What is the value of this thing supposed to be when the symbol does not correspond to a builtin type?!

ts2mls/js/src/main/scala/ts2mls/TSNamespace.scala
NeilKleistGao:
```ts
function f<T>(x: T, n: number): string;
function f<U>(x: U, n: boolean): string;
function f<T>(x: T, n) {
    return "";
}
```

Should I output the warning for code like this? Or merely generate the type definition? If no warning should be outputted, guess I need to add some logic to check if two type parameters list are compatible with each other.

ts2mls/js/src/main/scala/ts2mls/TSCompilerInfo.scala
NeilKleistGao: That would be a string of type in TypeScript style.
For example:
```ts
function add(x: number, y: number): number {
  return x + y
}
```
If we invoke this API, it would return `(x: number, y: number) => number`.

But we don't want another string parser, so I merely use it when it is a built-in type.

ts2mls/js/src/main/scala/ts2mls/TSCompilerInfo.scala
NeilKleistGao: ok I will try. 😸

ts2mls/js/src/main/scala/ts2mls/TSNamespace.scala
LPTK: Of course you should warn here too. Trying to see if overloads can be merged without loss of precision is a hugely complex problem, and it'd be totally useless to work on it since the whole point of overload is to handle situations where they can't be merged.

ts2mls/js/src/main/scala/ts2mls/TSCompilerInfo.scala
LPTK: Then the name is completely inappropriate and doesn't describe what it is at all ;)

I'd prefer if you used a specialized method in the specialized cases where this makes sense (such as directly accessing the builtin's name from the appropriate field), instead of polluting _all_ symbols with a method which doesn't make much sense for most of them.

Stringly-typed approaches, based on converting things to strings, are a recipe for unmaintainable code and disastrous maintenance.

ts2mls/js/src/main/scala/ts2mls/TSCompilerInfo.scala
NeilKleistGao: I just found another two API that can be used to check optional parameters and optional members in interfaces seperately. These operations are not manipulated on `TypeObject` but I think they can  at least improve the robuness of codes.

ts2mls/js/src/main/scala/ts2mls/TSCompilerInfo.scala
NeilKleistGao: I used another approach to get the `TypeObject` of objects, and removed this ambiguous member. 😄

ts2mls/js/src/main/scala/ts2mls/TSCompilerInfo.scala
LPTK: Cool 👍

ts2mls/js/src/main/scala/ts2mls/types/TSType.scala
NeilKleistGao: Since the logic is not correct in this function, it has been removed.

ts2mls/js/src/main/scala/ts2mls/JSWriter.scala
NeilKleistGao: I still prefer full names. So I guess I can keep them. 😄

ts2mls/js/src/main/scala/ts2mls/TSNamespace.scala
NeilKleistGao: So I just raise a warning when the TypeScript code tries to overload a function with type parameters. Like this:
```ts
def XX'f[T]: (T) -> ((anything) -> (string)) /* warning: the overload of function f is not supported yet. */
```

LPTK: Sure 👍

ts2mls/js/src/main/scala/ts2mls/TSSourceFile.scala
```diff
@@ -0,0 +1,174 @@
...
+    )
+    TSFunctionType(pList, getObjectType(node.returnType), getTypeParametes(node))
+  }
+
+  private def getStructuralType(types: TSTypeArray, isUnion: Boolean): TSStructuralType =
+    types.foldLeft[Option[TSType]](None)((prev, cur) => prev match {
+      case None => Some(getObjectType(cur))
+      case Some(p) =>
+        if (isUnion) Some(TSUnionType(p, getObjectType(cur))) else Some(TSIntersectionType(p, getObjectType(cur)))
+    }).get.asInstanceOf[TSStructuralType]
```
LPTK: `asInstanceOf` should be avoided. Why is it even important to return a `TSStructuralType` here?

ts2mls/js/src/main/scala/ts2mls/TSCompilerInfo.scala
```diff
@@ -0,0 +1,174 @@
...
+  lazy val types = TSTypeArray(obj.types)
+  lazy val properties = TSSymbolArray(checker.getPropertiesOfType(obj))
+  lazy val node = TSNodeObject(checker.typeToTypeNode(obj))
+  lazy val elementTypeOfArray = TSTypeObject(checker.getElementTypeOfArrayType(obj))
+
+  lazy val isTupleType = node.isTupleTypeNode
+  lazy val isArrayType = node.isArrayTypeNode
+  lazy val isEnumType = (flags & TypeScript.typeFlagsEnumLike) > 0
+  lazy val isUnionType = obj.isUnion().asInstanceOf[Boolean]
+  lazy val isIntersectionType = obj.isIntersection().asInstanceOf[Boolean]
```
LPTK: All unsafe cases (`asInstanceOf`) should be removed before the merge. Find a better/cleaner way.

ts2mls/js/src/main/scala/ts2mls/TSData.scala
```diff
@@ -0,0 +1,75 @@
...
+
+object IsUndefined {
+  def apply(v: js.Dynamic): Boolean = js.isUndefined(v)
+}
+
+// array for information object in tsc
+abstract class TSArray[T <: TSAny](arr: js.Dynamic) extends TSAny(arr) {
+  def get(index: Int): T = ???
+  lazy val length: Int = arr.length.asInstanceOf[Int]
```
LPTK: Same here.

LPTK (changes requested):

ts2mls/js/src/main/scala/ts2mls/TSCompilerInfo.scala
NeilKleistGao: All `asInstanceOf` have been removed.

LPTK (approved):
Thanks, this is a great first step for the project towards TypeScript interoperability!

PR #136, opened 2022-09-06
Fix function parameters parsing without implementations

NeilKleistGao:


NeilKleistGao:
Also, fix `=>` right-association

NeilKleistGao:
So far so good, but I'm not very sure about what to do next step. Any ideas please? :open_mouth:

LPTK:
Let's merge it ASAP. No need to let it linger any longer. Request a review and mark it as ready when you are.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -422,8 +422,10 @@ trait TermImpl extends StatementImpl { self: Term =>
       case p: TypeName => AppliedType(p, rhs.toType_! :: Nil)
       case _ => throw new NotAType(this)
     }
+    case Tup(fields) => Tuple(fields.map(fld => (fld._1, fld._2 match {
+      case Fld(_, _, v) => Field(N, v.toType_!)
```
LPTK: If the field is mutable we should make it like this:
```suggestion
      case Fld(m, s, v) => val ty = v.toType_!; Field(Option.when(m)(ty), ty)
```

LPTK (approved):
Great, thanks for the quick fixes!

LPTK:
After addressing my comment, you can go ahead and use the "squash and merge". Don't forget to clean up the commit message (remove the useless lines).

NeilKleistGao:
> After addressing my comment, you can go ahead and use the "squash and merge". Don't forget to clean up the commit message (remove the useless lines).

But I got no write access because I'm not in the organization yet. 🤣

PR #138, opened 2022-09-08
Code generation for by-value and by-name use of bindings

andongfan:
All tests are green.
Also tried in the web demo.

WeChat1d1caa46986bf9675a0165732d98b3b3


LPTK (approved):
Ok that seems good enough for now, thanks!

Though, note that in principle we should also be able to support things like `let rec foo = (0, () => foo)`, but code-generating them would be slightly more tricky (requiring a mutable local variable) so let's forget it for now.

andongfan:
Oh right, we are only rejecting _unguarded_ by-value recursive use.

Issue #134, opened 2022-09-02
TypeScript Libraries Support


NeilKleistGao:
Implementing ts2mls to allow using TypeScript libraries in MLscript.

Candidate TypeScript Libraries:
- https://github.com/gcanti/elm-ts
- https://github.com/gcanti/fp-ts/blob/master/src/Option.ts
- https://github.com/ionic-team/ionic-framework
- https://github.com/AykutSarac/jsoncrack.com
- https://www.npmjs.com/package/@types/lodash

NeilKleistGao:
TODO:
- [x] Generate MLscript declarations in new syntax for classes
- [x] Generate correct namespace structure by extending namespaces in `NewParser.scala`

chengluyu:
Is it from `.ts` to `.mls`, or `.d.ts` to `.mls`?

LPTK:
I think it should work with both. Right?

NeilKleistGao:
I think so. I can add some `.d.ts` tests later.

NeilKleistGao:
- [x] Generate type alias declaration

NeilKleistGao:
All interfaces are from [TypeScript types.ts](https://github.com/microsoft/TypeScript/blob/main/src/compiler/types.ts)

NeilKleistGao:
- [ ] Add `private`/`protected`

```ts
class A {
    private a: number
    protected b: string
}

```

NeilKleistGao:
- [ ] deal with `export` in ts

Example:
```ts
export namespace N1 {
    interface IF {
        foo()
    }

    class Base {
        a: number = 10
    }

    export class MyClass extends Base implements IF {
        foo() {}
    }
}
```

In this case, `Base` will not be exported so users can't access to it. However, we still need to consider that `MyClass` extends the `Base` so we need to add `a: number` into `MyClass`

NeilKleistGao:
> * [ ]  deal with `export` in ts
> 
> Example:
> 
> ```ts
> export namespace N1 {
>     interface IF {
>         foo()
>     }
> 
>     class Base {
>         a: number = 10
>     }
> 
>     export class MyClass extends Base implements IF {
>         foo() {}
>     }
> }
> ```
> 
> In this case, `Base` will not be exported so users can't access to it. However, we still need to consider that `MyClass` extends the `Base` so we need to add `a: number` into `MyClass`

Also we can create an instance by calling the exported function, but we can't use `new`:
```ts
export namespace N1 {
[...]

PR #139, opened 2022-09-14
Export TypeScript Declarations in New Syntax

NeilKleistGao:
related issue: #134 
also:
- Add `namespace` keyword
- Add some types in `toType_!`
- Add comments parsing
- Fix nested type arguments
- Add inner class translation
- Add constructor translation
- Add type parameters for functions
- Use parameter names in TS


shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -137,7 +137,10 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], raiseFun: D
   
   private def cur(implicit l: Line, n: Name) = {
     if (dbg) printDbg(s"? ${n.value}\t\tinspects ${summarizeCur}    [at l.${l.value}]")
-    _cur
+    _cur.dropWhile(_._1 match {
+      case COMMENT(_) => true
+      case _ => false
+    })
```
LPTK: I think it would be better as:
```suggestion
    while (_cur match {
      case COMMENT(_) => true
      case _ => false
    }) consume
    _cur
```

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -459,6 +459,7 @@ class JSBackend {
         traits += topLevelScope.declareTrait(name, tparams map { _.name }, body, methods)
       case TypeDef(Cls, TypeName(name), tparams, baseType, _, members, _) =>
         classes += topLevelScope.declareClass(name, tparams map { _.name }, baseType, members)
+      case TypeDef(Nms, TypeName(name), _, _, _, members, _) => () // TODO: deal with namespace
```
LPTK: Silent errors are the worst! Better raise a code-gen exception.

shared/src/main/scala/mlscript/codegen/Scope.scala
```diff
@@ -200,6 +200,8 @@ class Scope(name: Str, enclosing: Opt[Scope]) {
       declareTrait(name, tparams map { _.name }, body, mthdDefs)
     case TypeDef(Cls, TypeName(name), tparams, baseType, _, members, _) =>
       declareClass(name, tparams map { _.name }, baseType, members)
+    case TypeDef(Nms, TypeName(name), _, body, _, _, _) =>
+      declareTypeAlias(name, Ls(), body) // TODO: deal with namespaces
```
LPTK: Silent errors are the worst! Better raise a code-gen exception.

ts2mls/js/src/main/scala/ts2mls/types/Converter.scala
```diff
@@ -13,65 +13,82 @@ object Converter {
...
+    //   val lst = allRecs.filter(_.startsWith("  "))
+    //   if (lst.isEmpty) ""
+    //   else "\n" + lst.reduceLeft((bd, m) => s"$bd\n$m")
+    // }
     
     if (typeName === "trait ") body // anonymous interfaces
     else { // named interfaces and classes
-      val bodyWithParents = parents.foldLeft(body)((b, p) => s"$b & ${convert(p)}")
-      if (typeVars.isEmpty) s"$typeName: $bodyWithParents$methods"
+      val herirage =
```
LPTK:
```suggestion
      val inheritance =
```
?

ts2mls/js/src/test/diff/Array.d.mls
```diff
@@ -1,42 +1,23 @@
...
-//│    = <missing implementation>
+:NewParser
+:ParseOnly
+let first: MutArray<string> => string
+let getZero3: unit => MutArray<number>
+let first2: MutArray<number => number> => number => number
+let doEs: MutArray<int> => MutArray<int>
+class C {}
+trait I {
+  let i: number
```
LPTK, line 9: Note: I think we should make this just `i: number`. Maybe the parser needs to be changed for it. I can look into it later.

LPTK (commented):
Cool!

LPTK:
Also, `def f: T` in the old parser corresponds to `fun f: T` in the new one. Functions in TS should generate that.

NeilKleistGao:
I also wonder if `fun add(_: number, _: number): number` is acceptable. Or `fun add(x: number, y: number): number`, where `x` and `y` are original names from TypeScript codes, is better?

And what about `fun add(v0: number, v1: number): number`?

LPTK:
Doesn't `fun add(_: number, _: number): number` currently work?

NeilKleistGao:
So far, it works. But I guess a form like `fun add(v0: number, v1: number): number` is better?

LPTK:
I don't understand the question. Why not just reuse the names from TS?

NeilKleistGao:
> I don't understand the question. Why not just reuse the names from TS?

ok! I will.

ts2mls/js/src/test/diff/Array.d.mls
NeilKleistGao: I can give it a try if necessary :smile:

LPTK: It's fine for now. A lot of these things is going to change sooner than later anyway, once the new class typing arrives!

NeilKleistGao: Then I will prepare for the formal PR.

shared/src/main/scala/mlscript/JSBackend.scala
NeilKleistGao: I've replaced all silent errors with `throw CodeGenError("Namespaces are not supported yet.")` or `throw new Exception("Namespaces are not supported yet.")`. I used general `java.lang.Exception` when the code is not about the code generation.

LPTK:
Oh no... I didn't realized `>>` was going to be misparsed! Like C++ in the old days 🤦‍♂️

I guess your solution is good enough for now...

shared/src/main/scala/mlscript/JSBackend.scala
LPTK: Ok. You can also simply use `assert` and `require`, depending on the case. More straightforward and clear than explicitly throwing an `Exception`.

LPTK (approved):
Good work! Seems like this can be merged now. You can go ahead and squash-merge it with a clean commit message.

shared/src/main/scala/mlscript/JSBackend.scala
NeilKleistGao: Since `assert`'s and `require`'s return type is `Unit`, is it ok to append `null` or `td.bodyTy` to make the code pass the type checking?

shared/src/main/scala/mlscript/MLParser.scala
```diff
@@ -204,7 +204,7 @@ class MLParser(origin: Origin, indent: Int = 0, recordLocations: Bool = true) {
           ms.collect { case R(md) => md }, ms.collect{ case L(md) => md }, Nil)
       }
       case (k @ Als, id, ts) => "=" ~ ty map (bod => TypeDef(k, id, ts, bod, Nil, Nil, Nil))
-      case (k @ Nms, id, _) => throw new Exception("Namespaces are not supported yet.")
+      case (k @ Nms, _, _) => assert(false, "Namespaces are not supported yet."); null
```
LPTK: WHAT! Uuuuugh, no. `null` is an absolute anti-pattern. At this point just use `throw new AssertionError(...)`.

shared/src/main/scala/mlscript/TyperDatatypes.scala
```diff
@@ -234,7 +234,7 @@ abstract class TyperDatatypes extends TyperHelpers { self: Typer =>
         else TopType
       subst(td.kind match {
         case Als => td.bodyTy
-        case Nms => throw new Exception("Namespaces are not supported yet.")
+        case Nms => assert(false, "Namespaces are not supported yet."); td.bodyTy
```
LPTK: Same here. `assert(false, ...)` is quite smelly.

NeilKleistGao: Use `throw new AssertionError("Namespaces are not supported yet.")` instead.

LPTK: Actually, the more logical thing to throw here would rather be `new NotImplementedError(...)`.

LPTK: Sorry for the back and forth 😂

Assertions are for things that should never happen if the code is correct. Not the same as something not yet implemented.

NeilKleistGao: Updated.

PR #140, opened 2022-09-28
Add polymorphic variant examples from literature

andongfan:


andongfan:
Cleaned up. Both `PolyVariantCodeReuse.mls` and `PolyVariant.mls` tests are green in MLstruct.

LPTK (approved):
Great, thanks!

PR #143, opened 2022-10-04
Add Type Alias and Fix Record Type Translation in `ts2mls`

NeilKleistGao:
Related Issue: #134


NeilKleistGao:
I think we may separate `namespace` and `alias` from `class` and `trait` parsing(`case (KEYWORD(k @ ("class" | "trait" | "type" | "namespace")), l0) :: c`), because their syntax should be different. e.g. a namesapce should not have a parent namespace. :thinking:

LPTK:
I find it's better to be very lax and do semantic validation afterwards, because parsing errors are usually worse than semantic errors.

NeilKleistGao:
OK! And to keep `NewParser.scala` unchanged, I genetate type alias like this `type Option: (None) | (Some<A>)`. Is this syntax acceptable?

LPTK:
Not really no, we need `=` here. It's fine to extend `NewParser.scala` for that. But it should check that `=` is used for aliases and `:` for classes.

NeilKleistGao:
Also fixed a typo `getTypeParametes`.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -429,17 +429,18 @@ trait TermImpl extends StatementImpl { self: Term =>
     case Tup(fields) => Tuple(fields.map(fld => (fld._1, fld._2 match {
       case Fld(m, s, v) => val ty = v.toType_!; Field(Option.when(m)(ty), ty)
     })))
-    case Bra(rcd, trm) if (!rcd) => trm.toType_!
+    case Bra(_, trm) => trm.toType_!
```
LPTK: I don't think that's a good idea. Why not match specifically on `{ <record> }` and `( <something> )`? I don't think `( <record> )` makes sense for example. Better be strict in what we accept to avoid surprises.

ts2mls/js/src/main/scala/ts2mls/JSWriter.scala
```diff
@@ -7,7 +7,8 @@ import mlscript.utils._
 
 class JSWriter(filename: String) {
   import JSWriter._
-  private val out = fs.openSync(filename, "rs+")
+  private val out =
+    if (fs.existsSync(filename)) fs.openSync(filename, "rs+") else fs.openSync(filename, "w+")
```
LPTK: What does this do? Can you add some comments to explain?

LPTK (approved):
Otherwise LGTM!

LPTK:
After addressing the comments don't forget to use "squash and merge" and clean up the commit message.

ts2mls/js/src/main/scala/ts2mls/JSWriter.scala
NeilKleistGao: the mode `rs+` would fail when the file does not exist, so it can not be used to generate new files or tests

the mode `w+` would lead to modification of test files even though nothing has benn changed, thus the diffTests would failed too

LPTK: I just want to know what these things mean! I had to google this myself, which was pretty painful, to find this page: https://nodejs.org/api/fs.html#file-system-flags. You could have included the link.

Why do you need synchronous uncached writing? The doc emphasizes it has bad performance. _These_ are what I'm looking for in code documentation. Think about the people who'll read this code after you wrote it!

NeilKleistGao: You are right! I removed `s` in `rs+` and added the link in the comment.

ts2mls/js/src/main/scala/ts2mls/JSWriter.scala
```diff
@@ -7,10 +7,13 @@ import mlscript.utils._
...
-  // the mode `rs+` would fail when the file does not exist, so it can not be used to generate new files or tests
-  // the mode `w+` would lead to modification of test files even though nothing has benn changed, thus the diffTests would failed too
-  private val out =
-    if (fs.existsSync(filename)) fs.openSync(filename, "rs+") else fs.openSync(filename, "w+")
+  
+  // r+: Open file for reading and writing. An exception occurs if the file does not exist.
+  // Create an empty file if it does not exists.
+  // See https://nodejs.org/api/fs.html#file-system-flags to get more details.
+  if (!fs.existsSync(filename)) fs.writeFileSync(filename, "")
+  private val out = fs.openSync(filename, "r+")
```
LPTK: The comments' order is mixed up. Put every comment block above the thing it is actually supposed to comment on.

shared/src/main/scala/mlscript/helpers.scala
NeilKleistGao: Agree! But we now have `val bra = if (bk === Curly) Bra(true, Rcd(res.map { ...` in `NewParser.scala`. Isn't it enough? I'm not sure how to make it so far.

LPTK: No, it's not enough that this code exists in another file. The helper function in question could be used from other places than "NewParser".

NeilKleistGao: Is this way ok? 🤔

LPTK: Should be fine.

PR #144, opened 2022-10-04
Add nested-class lifter prototype

yuankeyu:
I created a new crossProject named compiler, isolated my classLifter from the mono project, and put it into the new project. Several tests were also added to the new project for the lifter.


LPTK:
Thanks! But you don't seem to have any tests where classes are defined within functions. Isn't it supported? 🤔 

Please make sure that is supported and test that feature in a new test file (the existing one is big enough). This support is crucial for using lifting to convert lambdas into nested classes into top-level classes.

compiler/shared/main/scala/mlscript/compiler/ClassLifter.scala
```diff
@@ -0,0 +1,482 @@
+package mlscript.compiler
+
+import mlscript.*
+import mlscript.utils.StringOps
+import scala.collection.mutable.StringBuilder as StringBuilder
+import scala.collection.mutable.Map as MMap
+import scala.collection.mutable.Set as MSet
+import mlscript.codegen.Helpers.inspect as showStructure
+
+class ClassLifter { self: ClassLifter =>
```
LPTK: This is useless
```suggestion
class ClassLifter {
```

compiler/shared/main/scala/mlscript/compiler/ClassLifter.scala
```diff
@@ -0,0 +1,482 @@
...
+    override def toString(): String = "(" ++ vSet.mkString(", ") ++ "; " ++ tSet.mkString(", ") ++ ")"
+  }
+  private def asContext(v: Var) = LocalContext(Set(v), Set())
+  private def asContextV(vS: IterableOnce[Var]) = LocalContext(vS.toSet, Set())
+  private def asContext(t: TypeName) = LocalContext(Set(), Set(t))
+  private def asContextT(tS: IterableOnce[TypeName]) = LocalContext(Set(), tS.toSet)
+  private def emptyCtx = LocalContext(Set(), Set())
+
+  case class ClassInfoCache(
+      originNm: TypeName,
```
LPTK: Indentation should be 2 spaces everywhere.

compiler/shared/main/scala/mlscript/compiler/ClassLifter.scala
```diff
@@ -0,0 +1,482 @@
...
+      originNm: TypeName, 
+      liftedNm: TypeName, 
+      var capturedParams: LocalContext, 
+      fields: MSet[Var], 
+      innerClses: MMap[TypeName, ClassInfoCache], 
+      supClses: Set[TypeName],
+      outerCls: Option[ClassInfoCache],
+      body: NuTypeDef, 
+      depth: Int){
+        def ++(other: ClassInfoCache) = this.copy(
```
LPTK: More legible:
```suggestion
  depth: Int
){
  def ++(other: ClassInfoCache) = this.copy(
```

compiler/shared/main/scala/mlscript/compiler/ClassLifter.scala
```diff
@@ -0,0 +1,482 @@
...
+
+  var retSeq: List[NuTypeDef] = Nil
+  var anonymCnt: Int = 0
+  var clsCnt: Int = 0
+  val logOutput: StringBuilder = new StringBuilder
+  val primiTypes = new mlscript.Typer(false, false, false).primitiveTypes
+
+  private def log(str: String): Unit = {
+    logOutput.append(str)
+    // println(str)
```
LPTK: Leftover comment

compiler/shared/main/scala/mlscript/compiler/ClassLifter.scala
```diff
@@ -0,0 +1,482 @@
...
+  private def tupleEntityToVar(fld: (Option[Var], Fld)): Option[Var] = fld match{
+    case (None, Fld(_, _, v: Var)) => Some(v)
+    case (Some(v: Var), _) => Some(v)
+    case _ => None
+  }
+  private def tupleEntityMapTrm(f: Term => Term)(fld: (Option[Var], Fld)): (Option[Var], Fld) = fld match {
+    case (None, snd@Fld(_, _, trm)) => (None, snd.copy(value = f(trm)))
+    case _ => fld
+  }
+  private def tupleEntityRenameVar(f: Var => Var)(fld: (Option[Var], Fld)): (Option[Var], Fld) = fld match {
```
LPTK: This seems to be unused. Probably the same for the other similar functions.

LPTK (changes requested):
This looks better, but it still has correctness problems. Consider:

```scala
// FIXME: `g` is extruded
fun f(x) =
  let g(x) = x + 1
  class Foo(x) {
    fun h = g(x)
  }
  Foo(x).h
//│ |#fun| |f|(|x|)| |#=|→|#let| |g|(|x|)| |#=| |x| |+| |1|↵|#class| |Foo|(|x|)| |{|→|#fun| |h| |#=| |g|(|x|)|←|↵|}|↵|Foo|(|x|)|.h|←|
//│ Parsed: {fun f = x, => {let g = x, => + (x,) (1,); class Foo(x,) {fun h = g (x,)}; (Foo (x,)).h}}
//│ Parsed:
//│ TypingUnit(NuFunDef(None, f, [], Lam(Tup(_: Var(x)), Blk(...))))
//│ Lifted:
//│ TypingUnit {
//│   class Foo(x,) {fun h = g ((this).x,)}
//│   fun f = x, => {let g = x, => + (x,) (1,); (Foo (x,)).h}
//│ }
```

Also, after you're done with your changes please read this before requesting my review: https://github.com/hkust-taco/mlscript/blob/mlscript/CONTRIBUTING.md

compiler/shared/main/scala/mlscript/compiler/ClassLifter.scala
```diff
@@ -183,6 +182,10 @@ class ClassLifter { self: ClassLifter =>
     case Blk(stmts) => 
       val newVs = getFields(stmts)
       stmts.map(getFreeVars(_)(using ctx.addV(newVs))).fold(emptyCtx)(_ ++ _)
+    case Let(isRec, name, rhs, body) => 
+      getFreeVars(rhs)(using ctx.addV(name)) ++ getFreeVars(body)(using ctx.addV(name))
+    // case Rcd(flds) =>
```
LPTK: Why this commented line?

compiler/shared/main/scala/mlscript/compiler/ClassLifter.scala
yuankeyu: Sorry I didn’t receive notifications from Github so I missed these comments… this method only serves for previous version of the code and is indeed unused not. I will remove it in a later commit

compiler/shared/main/scala/mlscript/compiler/ClassLifter.scala
yuankeyu: This is only for helping with debugging…if needed, as the whole “log” method doesn’t serve for the main logic, it can be removed as well. Should I also do that?

LPTK: No, of course not. Having the ability of performing debugging is useful. Leftover comments however are not.

LPTK (approved):
Ok, let's merge this now. We will still probably improve it later on as part of the UROP.

LPTK:
Please make sure that all my comments are addressed. When you are sure they are, you can close the corresponding conversations. Otherwise, leave a follow-up message and I'll answer.

Please also remove the warnings. You can use `""+_` instead of `_.toString`.

PR #146, opened 2022-10-12
Add `.d.ts` Tests, Literal Type Parsing, and Variables Parsing

NeilKleistGao:
Related Issue: #134 

A function-like variable is like this:
```ts
const some = <A>(a: A): Option<A> => ({ _tag: 'Some', value: a })
```

Also:
- Reorder `ts2mls` tests list


NeilKleistGao:
I think we may need to export other non-function-like object, or things like this would not be available in MLScript:
```ts
const none: Option<never> = { _tag: 'None' }
```

NeilKleistGao:
<del>I think features are ok so far. But the code is quite ugly. Let me find a better way to implement it. 😶‍🌫️ </del>
OK now it is better I guess.

LPTK (approved):
LGTM

PR #149, opened 2022-10-27
Fix The Error When Execute `fastOptJS`

NeilKleistGao:
Removed `scalaJSUseMainModuleInitializer := true` in `ts2mls`.


LPTK (approved):
Thanks!

PR #150, opened 2022-11-09
Ultimate conditional syntax

chengluyu:


shared/src/test/diff/nu/SimpleUCS.mls
```diff
@@ -223,20 +252,25 @@ fun f(u, v, w) =
...
-//│ ║  l.215: 	  if x - y > 0 then Some(x + y + z) else None()
+//│ ║  l.244: 	  if x - y > 0 then Some(x + y + z) else None()
 //│ ║         	                         ^^^^^^^^^
 //│ ╟── Note: constraint arises from reference:
-//│ ║  l.221: 	    Some(x) and x is
-//│ ╙──       	                ^
+//│ ║  l.250: 	    Some(x) and x is
+//│ ║         	                ^
+//│ ╟── from field selection:
+//│ ║  l.249: 	  if foo(u, v, w) is
```
LPTK: Whoops, this is not a field selection!

shared/src/test/diff/nu/InterleavedLet.mls
```diff
@@ -235,29 +179,163 @@ fun tr(x) =
...
+//│ Pair: ('fst, 'snd,) -> (Pair with {fst: 'fst, snd: 'snd})
+//│     = [Function: Pair1]
+//│ List: () -> List
+//│     = [Function: List1]
+//│ Nil: () -> Nil
+//│    = [Function: Nil1]
+//│ Cons: ('head, 'tail,) -> (Cons with {head: 'head, tail: 'tail})
+//│     = [Function: Cons1]
+
+// FIXME: Take a look...
```
LPTK: It should be `Nil then Pair(Nil(), Nil())`.

shared/src/test/diff/nu/SimpleUCS.mls
```diff
@@ -129,17 +116,20 @@ fun f(x, y) =
...
+//│ ║  l.117: 	    Some(xv) and y is Some(yv) then xv + yv
+//│ ║         	                 ^
+//│ ╟── [Missing Case 1/1] The case `None` is missing, which first appears here.
+//│ ║  l.118: 	    None() and y is None() then 0
+//│ ╙──       	                    ^^^^
 //│ f: (anything, anything,) -> error
 //│ Code generation encountered an error:
 //│   if expression has not been desugared
 
+// FIXME
```
LPTK: Yeah, this should be an exhaustiveness error (or at the very least a "useless case" warning for the `None() then xv * 2` part.

shared/src/test/diff/nu/SimpleUCS.mls
```diff
@@ -234,49 +208,48 @@ fun f(x, a, b) =
 //│ f: ((A with {value: 'value}) | (B with {value: 'value}) | ~A & ~B, anything, anything,) -> (0 | 'value)
 //│  = [Function: f14]
 
-fun f(x, y) =
+fun f(x, y, b) =
   if x is
     Some(xv) and y
       is Some(yv) then "bruh"
```
LPTK: Curious to see what happens if you make it
```suggestion
      is Some(yv) then yv
```
(please add this as an additional test)

shared/src/test/diff/nu/SplitOps.mls
```diff
@@ -79,37 +109,18 @@ fun f(x) =
...
-//│ ║  l.86: 	      None() then 0
-//│ ╙──      	^^^^^^^^^^^^^^^^^^^
-//│ sumOpt: (anything, anything,) -> error
-//│ Code generation encountered an error:
-//│   if expression has not been desugared
+//│ sumOpt: (None | Some & {value: int}, None | Some & {value: int},) -> int
+//│       = [Function: sumOpt]
 
-fun f(x, y) =
+// FIXME
```
LPTK: This is legit: `int & B` is `nothing`. I guess it would be nice if the UCS could warn that something's off. I think the UCS should always complain when there is an `is` test on some scrutinee with only one associated pattern like here.

LPTK:
Even though they're not perfect yet, the errors are pretty cool!

LPTK:
Beautiful!

shared/src/test/diff/nu/InterleavedLet.mls
chengluyu: This one is fixed.

shared/src/test/diff/nu/SimpleUCS.mls
chengluyu: It's added.

shared/src/test/diff/nu/SplitOps.mls
chengluyu: This is implemented but does not work on some cases. I made a note of this and will improve it later.

shared/src/test/diff/nu/SimpleUCS.mls
chengluyu: This is fixed.

shared/src/test/diff/nu/SimpleUCS.mls
chengluyu: This is done by adding a `sugaredTerm` to `TermImpl`. It's not perfect either.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -715,8 +721,29 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool)
...
+            println(s"- $scrutinee => " + patterns.keys.mkString(", "))
+          }
+          checkExhaustive(caseTree, N)(scrutineePatternMap, ctx, raise)
+          val trm = MutCaseOf.toTerm(caseTree)
+          println(s"Desugared term: ${trm.print(false)}")
+          iff.desugaredIf = S(trm)
+          typeTerm(trm)
+        } catch {
+          case e: DesugaringException => e.report(this)
+          case e: Throwable => throw e
```
LPTK: This line is useless.

shared/src/test/diff/nu/ParserFailures.mls
```diff
@@ -0,0 +1,16 @@
...
+//   let y = f(z)
+//   == y * y then 0
+
+// Interleaved let bindings are not implemented in `IfOpsApp`.
+// :pe
+// fun tt(x) =
+//   if x
+//     is A() then "A"
+//     let y = 0
+//     is B() then "B"
```
LPTK: The correct thing to do is NOT to comment those tests, but to add a `// FIXME` on top, which will silence the test failure.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -7,6 +7,7 @@ import scala.collection.immutable.SortedSet
 import math.Ordered.orderingToOrdered
 
 import mlscript.utils._, shorthands._
+import scala.collection.immutable
```
LPTK: This seems unused.

shared/src/main/scala/mlscript/ucs/Clause.scala
```diff
@@ -0,0 +1,89 @@
...
+
+  final case class MatchTuple(
+    scrutinee: Scrutinee,
+    arity: Int,
+    fields: Ls[Str -> Var]
+  ) extends Clause
+
+  final case class BooleanTest(test: Term) extends Clause
+
+  // Make it a class, and then include methods in `Conjunction`?
```
LPTK: Please either do this or remove the comment.

shared/src/main/scala/mlscript/ucs/Clause.scala
```diff
@@ -0,0 +1,89 @@
...
+    * Attach bindings to the first condition of a CNF.
+    *
+    * @param conditions the conditions
+    * @param interleavedLets the interleaved let buffer
+    * @return idential to `conditions`
+    */
+  def withBindings
+    (conditions: Conjunction)
+    (implicit interleavedLets: Buffer[(Bool, Var, Term)])
+  : Conjunction = {
```
LPTK:
```suggestion
  def withBindings(conditions: Conjunction)(implicit interleavedLets: Buffer[(Bool, Var, Term)])
        : Conjunction = {
```

shared/src/main/scala/mlscript/ucs/Conjunction.scala
```diff
@@ -0,0 +1,54 @@
...
+        : Opt[(Ls[Clause], MatchClass, Ls[Clause])] = {
+      upcoming match {
+        case Nil => N
+        case (head @ MatchClass(scrutinee, _, _)) :: tail =>
+          if (scrutinee === expectedScrutinee) {
+            S((past, head, tail))
+          } else {
+            rec(past :+ head, tail)
+          }
+        // TODO: Support tuples after merging `MatchClass` and `MatchTuple`.
```
LPTK: Aren't tuples supported? If not, add to https://github.com/hkust-taco/mlscript/issues/151

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -0,0 +1,605 @@
...
+    subPatterns.iterator.flatMap[Clause] {
+      case (scrutinee, subPattern) => destructPattern(scrutinee, subPattern)
+    }.toList
+  }
+
+  private val localizedScrutineeMap = MutMap.empty[Term, Var]
+
+  /**
+    * Create a `Scrutinee`. If the `term` is a simple expression (e.g. `Var` or
+    * `Lit`), we do not create a local alias. Otherwise, we create a local alias
```
LPTK, line 91: `Lit` doesn't actually seem to be handled.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -0,0 +1,605 @@
...
+    * @param matchRootLoc the caller is expect to be in a match environment,
+    * this parameter indicates the location of the match root
+    */
+  def makeScrutinee(term: Term, isMultiLineMatch: Bool)(implicit matchRootLoc: Opt[Loc]): Scrutinee =
+    trace(s"Making a scrutinee for $term") {
+      val res = term match {
+        case _: Var => Scrutinee(N, term)
+        case _ =>
+          Scrutinee(
+            S(localizedScrutineeMap.getOrElseUpdate(term, {
```
LPTK: This seems like a pretty bad idea. What of two subterms compare equal for some reason? You'll assign them the same variable! Please find a cleaner and **robust** way to handle this.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -0,0 +1,605 @@
...
+    *     A then "x is A!"
+    *     B then "x is B"
+    *   ```
+    *   is multi-line pattern match. Whereas `if x is Foo then 1 else 0` is not.
+    * @return a list of simple condition with bindings. This method does not
+    * return `ConjunctedCondition` because conditions built from nested patterns
+    * do not contain interleaved let bindings.
+    */
+  private def destructPattern
+      (scrutinee: Term, pattern: Term, isMultiLineMatch: Bool = false)
```
LPTK: Remove unnecessary default value. Defaults are error-prone so they should be reserved for easy-to-use APIs, not for core logic.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -0,0 +1,605 @@
...
+                  }.toList
+              })
+            }
+        }
+        // if (branches.length === 1 && scrutinee.isMultiLineMatch && default.isEmpty) {
+        //   import Message.MessageContext
+        //   raise(WarningReport({
+        //     msg"This scrutinee has only one case." -> scrutinee.matchRootLoc
+        //   } :: Nil))
+        // }
```
LPTK: What is this commented code doing here?

shared/src/main/scala/mlscript/ucs/Scrutinee.scala
```diff
@@ -0,0 +1,82 @@
...
+import mlscript.{Loc, Term, Var}
+import mlscript.utils.shorthands._
+
+// The point is to remember where does the scrutinee come from.
+// Is it from nested patterns? Or is it from a `IfBody`?
+// TODO: Replace `isMultiLineMatch` with `Scrutinee.Source`.
+final case class Scrutinee(local: Opt[Var], term: Term) {
+  def reference: Var = local.getOrElse(term match {
+    case v: Var => v
+    case _      => ???
```
LPTK: Is this a TODO? Either handle properly or add to https://github.com/hkust-taco/mlscript/issues/151

shared/src/main/scala/mlscript/ucs/Scrutinee.scala
```diff
@@ -0,0 +1,82 @@
+package mlscript.ucs
+
+import mlscript.{Loc, Term, Var}
+import mlscript.utils.shorthands._
+
+// The point is to remember where does the scrutinee come from.
```
LPTK:
```suggestion
// The point is to remember where the scrutinee comes from.
```

shared/src/main/scala/mlscript/ucs/Scrutinee.scala
```diff
@@ -0,0 +1,82 @@
...
+}
+
+object Scrutinee {
+  import mlscript._
+
+  /**
+    * Why we need a dedicated class for this?
+    * Because the scrutinee might comes from different structures.
+    * 1. Terms. E.g. `if x is A and y is B then ...`
+    * 2. `IfOpApp`. In this case, the scrutinee is indicate by the `IfOpApp`.
```
LPTK:
```suggestion
    * 2. `IfOpApp`. In this case, the scrutinee is indicated by the `IfOpApp`.
```

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -0,0 +1,605 @@
...
+        case _ =>
+          Scrutinee(
+            S(localizedScrutineeMap.getOrElseUpdate(term, {
+              Var(freshName).withLoc(term.toLoc)
+            })),
+            term,
+          )
+      }
+      res.isMultiLineMatch = isMultiLineMatch
+      res.matchRootLoc = matchRootLoc
```
LPTK: Do not use completely unnecessary mutation.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -0,0 +1,605 @@
...
+    * Destruct nested patterns to a list of simple condition with bindings.
+    *
+    * @param scrutinee the scrutinee of the pattern matching
+    * @param pattern the pattern we will destruct
+    * @param raise the `Raise` function
+    * @param aliasMap the field alias map
+    * @param matchRootLoc the location of the root of the pattern matching
+    * @param fragments fragment term that used to construct the given pattern.
+    *   It is used to tracking locations.
+    * @param isMultiLineMatch whether the scrutinee is in multi-line pattern match.
```
LPTK: I don't understand what this flag achieves. It doesn't seem to be really used anywhere.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -0,0 +1,605 @@
...
+      case classNameVar @ Var(className) =>
+        ctx.tyDefs.get(className) match {
+          case N => throw DesugaringException.Single({
+            import Message.MessageContext
+            msg"Cannot find the constructor `$className` in the context"
+          }, classNameVar.toLoc)
+          case S(_) => 
+            val clause = Clause.MatchClass(makeScrutinee(scrutinee, isMultiLineMatch), classNameVar, Nil)
+            println(s"Build a Clause.MatchClass from $scrutinee where pattern is $classNameVar")
+            clause.locations = collectLocations(scrutinee)
```
LPTK: Avoid unnecessary mutation. Why isn't this a parameter to `Clause`?

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -0,0 +1,605 @@
...
+            }
+          case S(Match(_, _, _)) => ???
+          case S(Consequent(_)) | S(MissingCase) | N => die
+        }
+      case IfThenElse(condition, whenTrue, whenFalse) =>
+        checkExhaustive(whenTrue, S(t))
+        checkExhaustive(whenFalse, S(t))
+      case Match(scrutinee, branches, default) =>
+        scrutineePatternMap.get(getScurtineeKey(scrutinee)) match {
+          // Should I call `die` here?
```
LPTK: `lastWords(s"...")` would be appropriate.

LPTK (changes requested):
This is great! But it still needs a little more polish.

It seems much of the logic hasn't been thoroughly cleaned up yet. Please do so. Try to remove unnecessary/gratuitious uses of mutable fields and variables.

I think too much info is currently printed in debug mode. Please add a `:ucsd` debugging flag to request more info. With only `:d`, I think we should basically only see two messages:

 * "The mutable CaseOf tree"

 * "Desugared term: ..."

shared/src/test/diff/nu/SimpleUCS.mls
LPTK: Seems good enough for now!

shared/src/test/diff/nu/SplitOps.mls
LPTK: Ok, please add the missing case description to https://github.com/hkust-taco/mlscript/issues/151

shared/src/main/scala/mlscript/ucs/Desugarer.scala
chengluyu: It was for reporting pattern matching with only one case. The commented code here caused some false positive. I will remove it.

shared/src/main/scala/mlscript/ucs/Conjunction.scala
chengluyu: I didn't know how to check tuples when I wrote this. I though I could merge `MatchClass` with `MatchTuple`. As what we discussed, we don't need to merge them anymore. So, this is no longer in need.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
chengluyu: I'm adding `Loc` to the key of the map. Now it's `MutMap.empty[(Opt[Loc], Term), Var]`.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
chengluyu: Removed.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
chengluyu: `isMultiLineMatch` is removed.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
chengluyu: Removed.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
chengluyu: Moved.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
LPTK: I added this to https://github.com/hkust-taco/mlscript/issues/151

shared/src/test/diff/nu/ParserFailures.mls
LPTK: You forgot to uncomment the tests 🤦‍♂️

shared/src/main/scala/mlscript/ucs/Desugarer.scala
chengluyu: Thanks!

shared/src/test/diff/ucs/NestedPattern.mls
```diff
@@ -0,0 +1,135 @@
...
+    (0, 0) then "zeros"
+    (1, 1) then "ones"
+    (y, 1) then x
+    _      then "que?"
+//│ ╔══[ERROR] type identifier not found: Tuple#2
+//│ ╙──
+//│ f: error -> error
+
+:e
+:d
```
LPTK: BTW we shouldn't commit debugging info.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -60,7 +60,7 @@ class Desugarer extends TypeDefs { self: Typer =>
     }.toList
   }
 
-  private val localizedScrutineeMap = MutMap.empty[Term, Var]
+  private val localizedScrutineeMap = MutMap.empty[(Opt[Loc], Term), Var]
```
LPTK: I still hate this. What if a scrutinee doesn't have a location for some reason (eg. maybe it's synthetic)? The robustness of your code SHOULD NOT depend on locations being present in the trees!!

shared/src/test/diff/nu/ParserFailures.mls
chengluyu: They will raise `???`. Can DiffTests catch `NotImplementedError`?

shared/src/main/scala/mlscript/ucs/Desugarer.scala
chengluyu: Sure. Let me think what I can do.

shared/src/test/diff/nu/ParserFailures.mls
LPTK: What part of "`// FIXME` will silence the test failure" don't you understand? 😜

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -375,7 +375,8 @@ class Desugarer extends TypeDefs { self: Typer =>
         case L(IfBlock(lines)) =>
           lines.foreach { desugarMatchBranch(scrutinee, _, partialPattern, collectedConditions) }
         // This case is rare. Let's put it aside.
-        case L(IfLet(_, _, _, _)) => ???
+        case L(IfLet(_, _, _, _)) =>
+          lastWords("please add this rare case to test files")
```
LPTK: Better leave `???` here. `lastWords` and `die` are for things that _cannot_ happen unless there's a bug in the compiler or in your reasoning.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -424,7 +425,8 @@ class Desugarer extends TypeDefs { self: Typer =>
         case IfOpApp(lhs, op, body) =>
           desugarIfBody(body, expr.addTermOp(lhs, op), acc)
         // This case is rare. Let's put it aside.
-        case IfLet(isRec, name, rhs, body) => ???
+        case IfLet(isRec, name, rhs, body) =>
```
LPTK, line 469: Same.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -495,10 +497,11 @@ class Desugarer extends TypeDefs { self: Typer =>
             if (whenFalse === t) {
               throw DesugaringException.Single(msg"Missing the otherwise case of test ${test.toString}", test.toLoc)
             } else {
-              ???
+              lastWords("`MissingCase` are not supposed to be the true branch of `IfThenElse`")
```
LPTK, line 539: This one is a good use. (You can resolve this comment.)

shared/src/main/scala/mlscript/ucs/Desugarer.scala
LPTK: If you really can't think of anything else, you can always try a `java.util.IdentityHashMap` https://docs.oracle.com/javase/7/docs/api/java/util/IdentityHashMap.html

shared/src/test/diff/nu/ParserFailures.mls
chengluyu: Oh. That's pretty cool.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
chengluyu: Actually `IdentityHashMap` doesn’t work in my case.

chengluyu: The goal is to localize non-`Var` scrutinees by creating let bindings so that we can reduce unnecessary computations. For example, we'd like to translate
```
if foo(x, y, z)
  < 0 then left
  > 0 then right
  = _ then this
```
to
```
let t = foo(x, y, z) in
case t < 0 of
  true -> left
  false -> case t > 0 of
    true -> right
    false -> this
```
rather than
```
case foo(x, y, z) < 0 of
  true -> left
  false -> case foo(x, y, z) > 0 of
    true -> right
    false -> this
```

LPTK: Why wouldn't `IdentityHashMap` work? I believe it does precisely what you want.

LPTK: Actually I'm not really clear on why you need a map at all, but if you really do `IdentityHashMap` should work as we're talking about a specific non-duplicated term here. But my feeling is that you're doing something in a suboptimal way.

chengluyu: This is good. I deleted two comments because there was a bug preventing it from working. Now it works.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -77,16 +78,18 @@ class Desugarer extends TypeDefs { self: Typer =>
...
-          Scrutinee(
-            S(localizedScrutineeMap.getOrElseUpdate((term.toLoc, term), {
-              Var(freshName).withLoc(term.toLoc)
-            })),
-            term
-          )
+          val localName = if (localizedScrutineeMap.containsKey(term)) {
+            localizedScrutineeMap.get(term)
+          } else {
+            val v = Var(freshName).withLoc(term.toLoc)
```
LPTK: Shouldn't this use `desugaredFrom`? In fact, maybe wee should _always_ use `desugaredFrom` instead of `withLocOf(term)` and we should make the latter also update the loc. (`withLocOf(term)` is currently like `withLocOf(term.toLoc)`).

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -375,8 +375,7 @@ class Desugarer extends TypeDefs { self: Typer =>
         case L(IfBlock(lines)) =>
           lines.foreach { desugarMatchBranch(scrutinee, _, partialPattern, collectedConditions) }
         // This case is rare. Let's put it aside.
-        case L(IfLet(_, _, _, _)) =>
-          lastWords("please add this rare case to test files")
+        case L(IfLet(_, _, _, _)) => ???
```
LPTK: You know it's a pity that `???` does not have more info attached to it. You could add a method `def TODO(msg: Str) = throw new NotImplementedError(msg)` in `utils/package.scala`, and make this:
```diff
- case L(IfLet(_, _, _, _)) => ???
+ case L(IfLet(_, _, _, _)) => TODO("please add this rare case to test files")
```

chengluyu: This is really cool. I added them.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
chengluyu: I didn't get "we should make the latter also update the loc". Doesn't `withLocOf(term)` update the loc?

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -336,7 +336,7 @@ trait TermImpl extends StatementImpl { self: Term =>
 
   def desugaredFrom(term: Term): this.type = {
     sugaredTerm = S(term)
-    this
+    withLoc(term.toLoc)
```
```suggestion
    withLocOf(term)
```

shared/src/main/scala/mlscript/ucs/Desugarer.scala
LPTK: Oh yeah I meant that we should make the _former_ update the loc, which is what you did.

shared/src/main/scala/mlscript/ucs/Conjunction.scala
```diff
@@ -0,0 +1,86 @@
...
+      case Nil => this
+      case head :: _ =>
+        head.bindings = trailingBindings ::: head.bindings
+        Conjunction(clauses ::: suffix, Nil)
+    }
+  }
+
+  def separate(expectedScrutinee: Scrutinee): Opt[(MatchClass, Conjunction)] = {
+    def rec(past: Ls[Clause], upcoming: Ls[Clause])
+        : Opt[(Ls[Clause], MatchClass, Ls[Clause])] = {
```
LPTK:
```suggestion
    def rec(past: Ls[Clause], upcoming: Ls[Clause]): Opt[(Ls[Clause], MatchClass, Ls[Clause])] = {
```

shared/src/main/scala/mlscript/ucs/Conjunction.scala
```diff
@@ -0,0 +1,86 @@
...
+      case head :: _ =>
+        head.bindings = interleavedLets.toList
+        this
+    }
+  }
+}
+
+object Conjunction {
+  def empty: Conjunction = Conjunction(Nil, Nil)
+}
```
LPTK:
```suggestion
}

```

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -0,0 +1,582 @@
...
+  */
+class Desugarer extends TypeDefs { self: Typer =>
+  var dbgUCS: Bool = false
+  private def printlnUCS(msg: => Any): Unit = if (dbgUCS) println(msg)
+  private def traceUCS[T](pre: => String)(thunk: => T)(post: T => String = noPostTrace) =
+    if (dbgUCS) trace(pre)(thunk)(post) else thunk
+
+  import Clause.{MatchClass, MatchTuple, BooleanTest}
+
+  type FieldAliasMap = MutMap[Term, MutMap[Str, Var]]
```
LPTK: This should probably be `MutMap[Var, MutMap[Str, Var]]` or `MutMap[SimpleTerm, MutMap[Str, Var]]`?

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -0,0 +1,582 @@
...
+  }
+
+  import MutCaseOf.{MutCase, IfThenElse, Match, MissingCase, Consequent}
+
+  /**
+    * A map from each scrutinee term to all its cases and the first `MutCase`.
+    */
+  type ExhaustivenessMap = Map[Str \/ Int, Map[Var, MutCase]]
+
+  type MutExhaustivenessMap = MutMap[Str \/ Int, MutMap[Var, MutCase]]
```
LPTK: Unused.

shared/src/main/scala/mlscript/ucs/DesugaringException.scala
```diff
@@ -0,0 +1,21 @@
...
+    override def report(typer: Typer)(implicit raise: Diagnostic => Unit): typer.SimpleType = {
+      typer.err(message, location)
+    }
+  }
+  final case class Multiple(messages: Ls[Message -> Opt[Loc]]) extends DesugaringException {
+    override def report(typer: Typer)(implicit raise: Diagnostic => Unit): typer.SimpleType = {
+      typer.err(messages)
+    }
+  }
+}
```
LPTK:
```suggestion
}

```

shared/src/main/scala/mlscript/ucs/DesugaringException.scala
```diff
@@ -0,0 +1,21 @@
...
+abstract class DesugaringException extends Throwable {
+  def report(typer: Typer)(implicit raise: Diagnostic => Unit): typer.SimpleType
+}
+
+object DesugaringException {
+  final case class Single(message: Message, location: Opt[Loc]) extends DesugaringException {
+    override def report(typer: Typer)(implicit raise: Diagnostic => Unit): typer.SimpleType = {
+      typer.err(message, location)
+    }
+  }
```
LPTK: This seems useless. Just have one.

shared/src/main/scala/mlscript/ucs/MutCaseOf.scala
```diff
@@ -0,0 +1,370 @@
...
+      locations ++= locs
+      this
+    }
+  }
+
+  import Clause.{MatchClass, MatchTuple, BooleanTest}
+
+  // A short-hand for pattern matchings with only true and false branches.
+  final case class IfThenElse(condition: Term, var whenTrue: MutCaseOf, var whenFalse: MutCaseOf) extends MutCaseOf {
+    override def describe: Str =
```
LPTK: Remove all the unnecessary `override`s.

shared/src/main/scala/mlscript/ucs/PartialTerm.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  final case class Total(val term: Term, override val fragments: Ls[Term]) extends PartialTerm(fragments) {
+    override def addTerm(term: Term): Total =
+      throw new PartialTermError(this, s"expect an operator but term $term was given")
+    override def addOp(op: Var): Half = Half(term, op, op :: fragments)
+  }
+
+  final case class Half(lhs: Term, op: Var, override val fragments: Ls[Term]) extends PartialTerm(fragments) {
+    override def addTerm(rhs: Term): Total = op match {
+      // TODO: Why are they look the same? How did the code here evolve?
```
LPTK: Please figure this out. Why *are* they look the same?

shared/src/main/scala/mlscript/ucs/PartialTerm.scala
```diff
@@ -0,0 +1,58 @@
...
+        val leftmost = mkBinOp(lhs, op, realRhs)
+        extraExprOpt match {
+          case N            => Total(leftmost, fragments)
+          case S(extraExpr) => Total(mkBinOp(leftmost, Var("and"), extraExpr), extraExpr :: fragments)
+        }
+    }
+    override def addOp(op: Var): Half =
+      throw new PartialTermError(this, s"expect a term but operator $op was given")
+  }
+}
```
LPTK:
```suggestion
}

```

shared/src/main/scala/mlscript/ucs/Scrutinee.scala
```diff
@@ -0,0 +1,20 @@
+package mlscript.ucs
+
+import mlscript.{Loc, Term, Var}
+import mlscript.utils.lastWords
+import mlscript.utils.shorthands._
+
+// The point is to remember where the scrutinee comes from.
+// Is it from nested patterns? Or is it from a `IfBody`?
+final case class Scrutinee(local: Opt[Var], term: Term)(implicit val matchRootLoc: Opt[Loc]) {
```
LPTK: Remove unnecessary implicits unless they really reduce duplication.

shared/src/test/diff/basics/Errors.fun
```diff
@@ -126,6 +135,23 @@ log / false + 1
...
+
+
+
+
+
+
+
+
+
+
```
LPTK: Is all this space necessary?

shared/src/test/diff/codegen/Terms.mls
```diff
@@ -341,39 +342,14 @@ case A { a = 0 } of
 //│    = [ 0, 1 ]
 
 
+// FIXME
 // With
 {} with {}
-//│ res: anything
-//│    = {}
```
LPTK: Is this change on purpose?

LPTK (changes requested):
LGTM except that the test files are borked. Please proofread your diff!

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -80,7 +80,7 @@ class Desugarer extends TypeDefs { self: Typer =>
   def makeScrutinee(term: Term)(implicit matchRootLoc: Opt[Loc]): Scrutinee =
```
LPTK: There's still an implicit here... is it useful?

shared/src/test/diff/basics/Errors.fun
LPTK: Please close when/if resolved.

shared/src/test/diff/codegen/Terms.mls
LPTK: Please close when/if resolved.

LPTK:
Please make one last proofreading of your diff and then request the review again.

LPTK (approved):
Ok, it finally LGTM! 🥳

LPTK:
This time I'll create a merge commit because there's a lot of history in this branch.

PR #153, opened 2022-11-24
Add `readonly` and `const` Parsing in `ts2mls`

NeilKleistGao:
Related Issue: #134


NeilKleistGao:
<del>I wonder if types like `mut string` acceptable. I made some changes in `NewParser.scala` because I found that we can do like this in OCaml:
```ocaml
val x : int ref = {contents = 42}
```
</del>

NeilKleistGao:
We will continue to work on this PR after this syntax's implementation:
```ts
class Foo { x: T; mut y: S; ... }
```

PR #155, opened 2022-12-05
Replace triggerBy with dependsOn

NeilKleistGao:
It seems that `triggerBy` in `build.sbt` would not exit with an error code even if sub tasks fail. I updated the `build.sbt` by using `dependsOn`, which could generate correct results in GitHub Actions.


LPTK (approved):
Looks good to me. Did you check that the process now correctly returns an error code when the tests fail?

NeilKleistGao:
So far it works on my device.

PR #157, opened 2023-03-12
Improvements and fixes for UCS

### LINKED ISSUE #151: UCS/UPS tracking issue
Issue to track the parts of the UCS implementation that are still incomplete or wrong – please update with more cases as you find them.

### Translation

 - [x] Support virtual constructors for pattern matching against tuples. **Note as of Jan 2024**: We implemented simple tuple pattern in #194 and we will support advanced tuple patterns, for example, splice tuple patterns, in the near future.
 - [x] Do not duplicate default cases and insert local function definitions instead.
   - **Note as of Jan 2024**: We paved the road for this in #194 and we will implement this feature in the near future.
   - **Note as of Nov 2025**: This has been solved in #336. Instead of using local functions, we use `Block` and `Break`.

### Locations

 - ~~Track scrutinee locations with `Scrutinee.Source`~~. (Note: it was removed in new desugarer.)
 - [x] No location reported in:

    ```scala
    fun testF(x) = if x is
      Foo(a) then a
      Foo(a) then a
    //│ ╔══[WARNING] duplicated branch
    //│ ╙──
    ```

### Reporting

 - [ ] Warn when an `is` test has a single pattern
    Though maybe we shouldn't warn when it's conjuncted with other things, as then the `is` may be used to bind expressions, as in `if xs is x :: xs and mapPartition(f, xs) is (as, bs) and ...`

### String patterns overhaul

- [ ] Make a PR that overhauls string patterns; currently, they misbehave in some cases; eg
    <details>
    <summary>Example</summary>
    
    ```fs
    pattern Digits = ("0"..="9") ~ (Digits | "")
    pattern DigitsCommas = Digits | Digits ~ "," ~ DigitsCommas
    pattern Integer = "0" | ("1"..="9") ~ (DigitsCommas | "")
    pattern FractionalPart = "." ~ Digits
    pattern ExponentPart = ("e" | "E") ~ ("+" | "-" | "") ~ Integer
    
    // FIXME: does not work!
    // pattern NumberStr = Integer ~ (FractionalPart | "") ~ (ExponentPart | "")
    
    // Works
    pattern NumberStr = Integer | Integer ~ FractionalPart ~ (ExponentPart | "")
    
    // FIXME: `Number` doesn't recognize commas
    pattern AsNumber = (NumberStr & n) => Number(n)
    ```
    
    </details>

### Fix pattern extraction semantics

- [ ] Patterns like `Pair(0 => 1, 0 => 1)` should rebuild the outer ctors
- [ ] Patterns like `C(_, 0 => 1)` should fail when the first param of `C` is private, as the thing can't be rebuilt
- [ ] Patterns like `P(Transf)` should warn about a discarded transformation when `P` is a HOP – users should write `P(Transf as _)` instead

### Fixed point patterns

- [ ] Avoid retraversing rec pattern prefixes when no change happened (see: EvaluationContexts2, StackParse)

chengluyu:
## Change Logs

- Add a complete test cases in UCS: untyped lambda calculus.
- WIP test cases in UCS: simply typed lambda calculus.
- Support alias clauses (e.g. `... and f(x) is y and ...`).
- No longer report inexhaustiveness when there is a default branch.
- Fix missing interleaved computations in nested branches (examples added in d93e0d6).
- Fix unhygienically extracting positional class parameters as described in #151.
- Mark the source of let bindings so no we won't promote let bindings which is only accessible in a case branch to the level of its parent case expression. (Mentioned in [this reply](https://github.com/hkust-taco/mlscript/issues/151#issuecomment-1453047328) of #151)
- Fix empty warnings of duplicated branches as described in #151.
    ```scala
    fun testF(x) = if x is
      Foo(a) then a
      Foo(a) then a
    //│ ╔══[WARNING] duplicated branch
    //│ ╙──
    ```

## Goals

- Align with the paper.
- Fix #151.

## Non-Goals

- Error messages and location reporting. This PR focuses more on making the UCS feature fully functional and minimizing errors.


chengluyu:
## Virtual Constructors

Two ideas for implementing virtual constructors (e.g. tuples).

1. Treat `Tuple#n` as an alias of tuple types in the typer. Then, translate `Tuple#n` as JavaScript predicates in the JS backend.
2. Type predicate function `isTuple(t, n)` in the typer and desugar tuple patterns to terms with reference to the function. Implement `isTuple` as a polyfill function in the JS backend.

shared/src/test/diff/ucs/LitUCS.mls
```diff
@@ -4,35 +4,37 @@
...
+// The following example can better reveal the essence of the problem.
 fun test(x: 0 | A) = if x is
   0 then 0
   A then A
-//│ ╔══[ERROR] Type mismatch in application:
-//│ ║  l.8: 	fun test(x: 0 | A) = if x is
-//│ ║       	                        ^
+//│ fun test: (x: 0 | A,) -> (0 | A)
+
+// FIXME
```
LPTK: Why is this a FIXME?

shared/src/test/diff/ucs/WeirdIf.mls
```diff
@@ -79,20 +79,17 @@ fun f(x) =
 //│ f: anything -> "bruh"
 //│  = [Function: f3]
 
-:e
-:ge
 // Hmmmmmm, this one is valid but how to get it work?
```
LPTK:
```suggestion
```

shared/src/test/diff/ucs/LitUCS.mls
chengluyu: The example below still doesn't work.

shared/src/test/diff/ucs/WeirdIf.mls
chengluyu: Fixed in a later commit.

shared/src/test/diff/ucs/InterleavedLet.mls
```diff
@@ -261,60 +260,38 @@ fun mapPartition(f, xs) =
...
-//│   unresolved symbol l
-
-// FIXME `b` should be bound here!
-// The example is actually wrong. `b` belongs to the consequent of `Left(b)`.
-// `let y = ...` has no access to `b` because it is not in scope.
-fun mn(a) =
-  if a is
-    Some(x) and x is
-      Left(b)  then "left-defined"
-      let y = b + 1
```
LPTK: Don't forget to add the example I suggested in place of this one.

shared/src/test/diff/ucs/SplitAroundOp.mls
```diff
@@ -73,6 +73,186 @@ fun f(x, b) =
 //│ ╟── string literal of type `"2"` is not an instance of type `number`
 //│ ║  l.14: 	      "2" then "s2"
 //│ ╙──      	      ^^^
+//│ ╔══[ERROR] Type mismatch in operator application:
```
LPTK: You could try this with `:NewDefs` and:

```scala
fun f(x, b) =
  if x is
    number and x ===
      0 and b then "n0"
      1 and b then "n1"
      2 then "n2"
    string and x ===
      "0" then "s0"
      "1" then "s1"
      "2" then "s2"
  else ":p"  
```

shared/src/test/diff/ucs/Wildcard.mls
```diff
@@ -0,0 +1,54 @@
...
+fun w1(x, e_0, e_1) =
+  if x is
+    Left(None) then 1
+    Right(None) then 2
+    _ and e_0 is y_0 and x is
+      Left(Some(lv)) then 3
+      _ and e_1 is y_1 and x is
+        Right(Some(rv)) then 4
+//│ ╔══[WARNING] Found duplicated branch
+//│ ╟── This decision path tries to fit
```
LPTK: Note: the "tries to fit" wording is confusing. I don't understand it.

shared/src/test/diff/ucs/SplitAroundOp.mls
chengluyu: Thanks!

shared/src/test/diff/ucs/InterleavedLet.mls
chengluyu: I added in commit [d972ee3](https://github.com/hkust-taco/mlscript/pull/157/commits/d972ee30385523a1934d11b49f5915a2ac984e44).

shared/src/test/diff/ucs/Wildcard.mls
chengluyu: Updated. How do you think of this?

```
// It should report duplicated branches.
:w
fun testF(x) = if x is
  Foo(a) then a
  Foo(a) then a
//│ ╔══[WARNING] Found a duplicated branch
//│ ╟── This branch
//│ ║  l.21: 	  Foo(a) then a
//│ ║        	              ^
//│ ╟── is subsumed by the branch here.
//│ ║  l.20: 	  Foo(a) then a
//│ ╙──      	              ^
//│ testF: (Foo & {x: 'x}) -> 'x
//│      = [Function: testF]
```

LPTK: Oh, that's much clearer! 👍

js/src/main/scala/Main.scala
```diff
@@ -133,7 +133,7 @@ object Main {
           implicit val extrCtx: Opt[typer.ExtrCtx] = N
 
           val vars: Map[Str, typer.SimpleType] = Map.empty
-          val tpd = typer.typeTypingUnit(tu, allowPure = true)(ctx.nest, raise, vars)
```
LPTK: The parameter name was changed. We should keep setting it to `true`.

shared/src/test/diff/ucs/LitUCS.mls
LPTK: This error is expected, as I've explained you on Telegram... `===` isn't meant to compare unrelated types and you're comparing `0 | A` to `0`, which implies comparing `A` with `0`.

chengluyu: I see! Thanks!

js/src/main/scala/Main.scala
chengluyu: Thanks. It's fixed.

chengluyu:
We have several `FIXME`s left to fix. Let’s go through them now.

- `shared/src/test/diff/ucs/InterleavedLet.mls`:
  - [x] Line 302: `// FIXME: Code generation problem...`

    `JSBackend` generates something like `let a = a`...

- `shared/src/test/diff/ucs/MultiwayIf.mls`:
  - [x] Line 13: `// FIXME`

    What's wrong with this test case?

    ```
    // FIXME
    fun f(x) =
      if
        x > 0  and
          x % 2 === 0 then true
          _           then false
        x == 0 then true
        _      then false
    //│ f: int -> bool
    //│  = [Function: f1]
    ```

  - [x] Line 24: `// FIXME`

    What's wrong with this test case?

    ```
    // FIXME
    fun f(x) =
      if
        x > 0  and
          x % 2 === 0 then true
          else false
        x == 0 then true
      else false
    //│ f: int -> bool
    //│  = [Function: f2]
    ```

- `shared/src/test/diff/ucs/SplitOps.mls`:
  - [x] Line 89: `// * FIXME: the missing otherwise is for `a == 0``

    As explained in the text, we should report the correct location.

- `shared/src/test/diff/ucs/TrivialIf.mls`:
  - [ ] Line 65: `// FIXME`

    We should optimize the error message here. Right?

    ```
    // FIXME
    if 42 is n then n + 1 else 0
    //│ ╔══[WARNING] Found a duplicated branch
    //│ ╟── This branch
    //│ ║  l.66: 	if 42 is n then n + 1 else 0
    //│ ║        	                           ^
    //│ ╟── is subsumed by the branch here.
    //│ ║  l.66: 	if 42 is n then n + 1 else 0
    //│ ╙──      	                ^^^^^
    //│ res: int
    ```

LPTK:
Great progress!


> `shared/src/test/diff/ucs/InterleavedLet.mls`:
> 
> * [x]  Line 302: `// FIXME: Code generation problem...`
>   `JSBackend` generates something like `let a = a`...

What's the status of hygiene in the UCS desugaring? If you're not using unique identifiers using this variable field we set up earlier, then it's likely that other similar cases will be wrong, because the UCS merges scopes and moves bindings around AFAIK.

> * `shared/src/test/diff/ucs/MultiwayIf.mls`:
>   
>   * [x]  Line 13: `// FIXME`
>     What's wrong with this test case?
>     ```
>     // FIXME
>     fun f(x) =
>       if
>         x > 0  and
>           x % 2 === 0 then true
>           _           then false
>         x == 0 then true
>         _      then false
>     //│ f: int -> bool
>     //│  = [Function: f1]
>     ```
>   * [x]  Line 24: `// FIXME`
>     What's wrong with this test case?
>     ```
>     // FIXME
>     fun f(x) =
>       if
>         x > 0  and
>           x % 2 === 0 then true
>           else false
>         x == 0 then true
>       else false
>     //│ f: int -> bool
>     //│  = [Function: f2]
>     ```

When there's nothing obviously wrong below a FIXME it usually means the problem was fixed but we omitted/forgot to remove the FIXME. Just add a couple of test cases to make sure it's compiled right and remove the FIXMEs.

chengluyu:
> What's the status of hygiene in the UCS desugaring? If you're not using unique identifiers using this variable field we set up earlier, then it's likely that other similar cases will be wrong, because the UCS merges scopes and moves bindings around AFAIK.

Yes, you're right. We need more test cases to test hygiene specifically.

As for this test case, it is caused by `JSBackend`. `JSBackend` makes temporary names in the alphabetical order. But it's not aware `a` is already used.

chengluyu (commented):
I have checked all the test files and updated the tracking issue.

shared/src/main/scala/mlscript/NewLexer.scala
```diff
@@ -55,6 +55,34 @@ class NewLexer(origin: Origin, raise: Diagnostic => Unit, dbg: Bool) {
   def takeWhile(i: Int, cur: Ls[Char] = Nil)(pred: Char => Bool): (Str, Int) =
     if (i < length && pred(bytes(i))) takeWhile(i + 1, bytes(i) :: cur)(pred)
     else (cur.reverseIterator.mkString, i)
+
+  @tailrec final
+  def str(i: Int, escapeMode: Bool = false, cur: Ls[Char] = Nil): (Str, Int) =
```
LPTK: This default is completely redundant.

Besides, how many times will I have to tell people not to use default in internal core logic?

js/src/main/scala/Main.scala
```diff
@@ -134,7 +134,7 @@ object Main {
 
           val vars: Map[Str, typer.SimpleType] = Map.empty
           val tpd = typer.typeTypingUnit(tu, topLevel = true)(ctx.nest, raise, vars)
-
+
```
LPTK, line 137: This is the best part of the PR. 👍

shared/src/main/scala/mlscript/ucs/Clause.scala
```diff
@@ -9,64 +9,70 @@ import scala.collection.mutable.Buffer
...
+sealed abstract class MatchClause extends Clause {
+  val scrutinee: Scrutinee
 }
 
 object Clause {
+  final case class MatchLiteral(
+    override val scrutinee: Scrutinee,
+    literal: SimpleTerm
+  )(override val locations: Ls[Loc]) extends MatchClause {
+    override def toString(): String = s"«$scrutinee is $literal" + bindingsToString
```
LPTK: `toString` is idiomatic in Scala, not `toString()`.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -545,75 +689,367 @@ class Desugarer extends TypeDefs { self: Typer =>
...
-        case IfThenElse(_, whenTrue, whenFalse) =>
-          rec(whenTrue)
-          rec(whenFalse)
-        case Match(scrutinee, branches, _) =>
-          val key = getScurtineeKey(scrutinee)
-          branches.foreach { mutCase =>
-            val patternMap = m.getOrElseUpdate( key, MutMap.empty)
-            if (!patternMap.contains(mutCase.patternFields._1)) {
-              patternMap += ((mutCase.patternFields._1, mutCase))
+  // private def summarizePatterns(t: MutCaseOf)(implicit ctx: Ctx, raise: Raise): ExhaustivenessMap = traceUCS("[summarizePatterns]") {
```
LPTK: Why the commented code? Does it serve a purpose? Maybe just make a tag to the commit or copy-paste this in your own notes/in the issue.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -545,75 +689,367 @@ class Desugarer extends TypeDefs { self: Typer =>
...
+  private def getClassHierarchy()(implicit ctx: Ctx): SuperClassMap =
+    traceUCS("[getClassHierarchy]") {
+      // ctx.tyDefs
+      val superClassMap = ctx.tyDefs.iterator
+        .filter(_._2.toLoc.isDefined)
+        .map { case (className, td) =>
+          className -> td.baseClasses.iterator.map(_.name).toList
+        } |> Map.from
+      Desugarer.transitiveClosure(superClassMap)
+      // ctx.tyDefs2
```
LPTK: What's the purpose of this commented line?

shared/src/test/diff/nu/BadUCS.mls
```diff
@@ -72,14 +72,8 @@ fun foo(x) = if x is M() then 0
...
 
-:e
 fun foo0(x, y) = if x is y then 0
-//│ ╔══[ERROR] Cannot find constructor `y` in scope
-//│ ║  l.76: 	fun foo0(x, y) = if x is y then 0
-//│ ╙──      	                         ^
-//│ fun foo0: (anything, anything,) -> error
-//│ Code generation encountered an error:
-//│   if expression was not desugared
+//│ fun foo0: (anything, anything,) -> 0
```
LPTK, line 76: Wait what? That's not right. How does it work?

shared/src/test/diff/nu/repro_EvalNegNeg.mls
```diff
@@ -29,34 +29,25 @@ module TestLang extends EvalBase
...
 fun mk(n) = if n is
   0 then Lit(0)
   1 then Neg(mk(n))
   _ then Add(mk(n), mk(n))
-//│ fun mk: forall 'E. number -> 'E
+//│ fun mk: forall 'E. anything -> 'E
 //│   where
 //│     'E :> Add['E] | Lit | Neg['E]
 
 // TODO support this in UCS
```
LPTK: Remove TODO?

shared/src/test/diff/ucs/DirectLines.mls
```diff
@@ -63,11 +56,8 @@ fun f(a, b) =
       2 then 2
       _ then 7
     else 3
-//│ ╔══[WARNING] Found a duplicated else branch
-//│ ║  l.65: 	    else 3
-//│ ║        	         ^
-//│ ╟── The first else branch was declared here.
```
LPTK, line 69: Regression in warning? This hint was helpful. You can just add it to the tracking issue.

shared/src/test/diff/ucs/Exhaustiveness.mls
```diff
@@ -23,11 +20,47 @@ fun f(x, y) =
...
+module Empty {
+  fun contains(wanted) = false
+}
+class Node[A](value: int, left: Tree[A], right: Tree[A]) {
+  fun contains(wanted) = if wanted
+    <= value then left.find(wanted)
+    >= value then right.find(wanted)
+    == value then true
+}
+//│ ╔══[ERROR] The case when this is false is not handled: == (wanted,) (value,)
```
LPTK, line 51: To the tracking issue: showing terms properly, not using the internal debug string. You could just print the corresponding location?

shared/src/test/diff/tapl/NuSimplyTyped.mls
```diff
@@ -0,0 +1,203 @@
+:NewParser
+:NewDefs
+
+:escape
+// You can push debug messages to this magic array.
```
LPTK: This code is duplicated many times. What's its point? You can just use `log`... Please simplify and keep only one instance if this code as a self-contained test (because it's interesting).

LPTK (changes requested):
Wow, impressive changes! But they still need a little polish.

shared/src/test/diff/nu/BadUCS.mls
chengluyu: This is desugared into `let y = x in 0` since the expression doesn't indicate any branches.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
chengluyu: Removed.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
chengluyu: No. It's completely useless. I removed it.

shared/src/test/diff/tapl/NuSimplyTyped.mls
chengluyu: Cool. Thanks.

shared/src/test/diff/tapl/SimplyTyped.mls
```diff
@@ -239,91 +221,91 @@ fun typeTerm(t, ctx) =
...
 //│           <: 'lhs & (PrimitiveType & {name: string} | 'a & (FunctionType & 'b | FunctionType & ~#FunctionType))
 //│     'b <: {lhs: 'rhs2, rhs: 'rhs2}
 //│     'rhs2 <: FunctionType & 'b | PrimitiveType | ~FunctionType & ~PrimitiveType
 //│     'ty :> 'lty | FunctionType & {lhs: 'lty, rhs: 'ty} | 'rhs1
-//│     'rhs0 <: FunctionType & {lhs: 'rhs0, rhs: 'rhs0} | PrimitiveType | ~FunctionType & ~PrimitiveType
 //│     'lhs <: FunctionType & {lhs: 'lhs & (PrimitiveType & {name: string} | ~#PrimitiveType), rhs: 'lhs} | PrimitiveType & {name: string}
+//│     'rhs0 <: FunctionType & {lhs: 'rhs0, rhs: 'rhs0} | PrimitiveType | ~FunctionType & ~PrimitiveType
 //│         = [Function: typeTerm]
 
 // FIXME
```
LPTK: Can you specify what needs to be fixed here? Same for the other FIXMEs without an actual error.

shared/src/test/diff/nu/BadUCS.mls
LPTK: Ah, got it! Seems like this would warrant a warning (whose precise specification should be determined later) because of how confusing it looks. Plz add to the backlog.

shared/src/test/diff/parser/NuParser.mls
```diff
@@ -0,0 +1,14 @@
+:NewParser
```
LPTK: This file's name should be something like `NegativeLits.mls`.

shared/src/test/diff/ucs/HygienicBindings.mls
```diff
@@ -0,0 +1,164 @@
...
+// FIXME: Precise scrutinee identification (easy)
+// This seems fine. But the subtrees are not merged.
+fun h1(a) =
+  if
+    a is Some(x) and x is Left(y) then y
+    a is Some(y) and y is Right(z) then z
+    a is None then 0
+//│ fun h1: forall 'leftValue. (None | Some[Right['leftValue]]) -> (0 | 'leftValue)
+
+// FIXME: This is the desugared version of the test case above.
```
LPTK: Does this one need to be fixed?! I don't think one should expect the UCS to merge explicit let bindings.

LPTK (approved):
Thanks, otherwise this looks good to me!

I'm a bit uncomfortable with the large code duplication introduced in your examples, but we can probably move these examples to the new compiler diff-tests later, where we'll be able to share some code in library files.

shared/src/test/diff/nu/BadUCS.mls
chengluyu: Sure. I added this to the backlog.

LPTK:
Ah, the build fails due to warnings. There were 34 before this PR and 35 now. Could you please try to solve the most obvious warnings, including specifically the one warning you introduced?

chengluyu:
Sure. I will handle them later.

LPTK:
Thanks! If there are some you're not sure how to deal with, leave them and I'll address them myself.

shared/src/test/diff/ucs/HygienicBindings.mls
chengluyu: We don't need to merge them, but I think the exhaustiveness checker should be aware of explicit let bindings which refer to the same class parameter.

chengluyu: I've added this to backlog issue.

chengluyu:
I tried to fix some warnings and non-exhaustive pattern errors. Now there remains 30 of them. It seems to go beyond the scope of what this PR should be responsible for if all of them need to be fixed. For example, the warning about `asInstanceOf`. So, I'll just do this much for now.

shared/src/test/diff/tapl/SimplyTyped.mls
LPTK: You didn't handle this part:

> Same for the other FIXMEs without an actual error.

LPTK:
Yes that's fine, we'll deal with the other warnings later. Let me know when I can merge.

LPTK:
@chengluyu can I merge soon? One conversation is still open.

PR #159, opened 2023-04-10
Defunctionalizer prototype

yuankeyu:


compiler/shared/main/scala/mlscript/compiler/mono/utils.scala
```diff
@@ -0,0 +1,3 @@
+package mlscript.compiler.mono
```
LPTK, line 1: This file seems to be unneeded.

LPTK:
Subsumed by https://github.com/hkust-taco/mlscript/pull/185.

PR #161, opened 2023-05-08
Make class parameter list optional and improve class construction

NeilKleistGao:
In this PR:
- Add `abstract` keyword
```ts
abstract class Foo(x: int) {
  fun f(y: int) = x + y
}
//│ class Foo(x: int) {
//│   fun f: (y: int,) -> int
//│ }
```
- Distinguish between `class Foo` and `class Foo()`: the former's instantiation requires `new` keyword and the latter's doesn't:
```ts
class A(x: int) {}
class C {
  constructor(x: int)
  fun f = x + 1
}
let c = new C(1)
```
- Change `New` codegeration: it always generates a `new` keyword in JS. The check will be done in type checker:
```ts
class D(x: int)
let dd = new D(42) // let dd = new D.class(42) in JS
```

- Override constructor in classes:
```ts
class D(x: int) {
  constructor(y: int) {
    x = y + 1
  }
  log(x)
}
let dd = new D(41)
dd.x // 42
```

- Freeze object before returning:
```ts
class A(x: int)
// this.#A = ((x) => Object.freeze(new A(x)));
```

- Ignore traits in inheritance clauses
- Fix other warnings


LPTK:
I just realized that the JS backend generates `A()` for a `new A` expression. This PR would be a good time to fix this. (It should generate `new A.class()` for MLscript classes and `new A()` for TS/JS ones.

It would be even nicer if you could also implement the fact that `new A(...) { ... }` should code-gen into `new class extends A { ... }`, but to do this well you'll need to reuse the class/mixin implementation, which may require some refactoring. What do you think?

NeilKleistGao:
I think I can finish the fixing in this PR and submit another for the `new A(...) { ... }`.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -389,7 +389,7 @@ trait NuDeclImpl extends Located { self: NuDecl =>
   def kind: DeclKind
   val declareLoc: Opt[Loc]
   val abstractLoc: Opt[Loc]
-  def isDecl: Bool = declareLoc.nonEmpty
+  def isDecl: Bool = declareLoc.nonEmpty || isAbstract
```
LPTK: Why?! This doesn't make sense.

shared/src/main/scala/mlscript/NewLexer.scala
```diff
@@ -281,8 +281,7 @@ object NewLexer {
     "in",
     "out",
     "null",
-    "undefined",
-    "abstract"
```
LPTK: Why revert the implementation of `abstract`?

LPTK:
This solution allows representing illegal states: when `has_params` (which should be `hasParams` BTW) is false and the parameter list is non-empty. Much better make the parameter list an `Option`.

https://fsharpforfunandprofit.com/posts/designing-with-types-making-illegal-states-unrepresentable/

Note: I only propose to do this change in `NuTypeDef`. Better leave the `TypedNu*` representations alone for now due to changes in other PRs.

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -402,11 +402,10 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], raiseFun: D
               }
               else ctor.headOption match {
                 case Some(ctor @ Constructor(tup)) =>
-                  if (has_params) err(msg"more than one constructor" -> S(l0) :: Nil)
+                  if (params.isDefined) err(msg"more than one constructor" -> S(l0) :: Nil)
```
LPTK: The error needs to be improved.

shared/src/test/diff/codegen/OptionalParam.mls
```diff
@@ -355,15 +354,24 @@ n.ll
...
-//│   TypeError: Class constructor L cannot be invoked without 'new'
+//│   = L {}
 //│ res
+//│     = 3
+
+// TODO reject
+// :e
+:re
+class M()
+let mm = new M()
```
LPTK: Why can't we make this work? The logic is the same as the logic to determine whether we generate `extends C.class` or `extends C`.

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -276,7 +277,13 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], raiseFun: D
         consume
         yeetSpaces
         go(acc.copy(acc.mods + ("declare" -> l0)))
+      case (KEYWORD("abstract"), l0) :: c =>
+        consume
+        yeetSpaces
+        go(acc.copy(acc.mods + ("abstract" -> l0)))
       case _ if acc.mods.isEmpty => acc
+      case (KEYWORD("constructor"), l0) :: _ =>
```
LPTK: Merge this with the case below!

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -369,8 +389,26 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], raiseFun: D
...
-            val res = NuTypeDef(kind, tn, tparams, params, sig, ps, N, N, body)(isDecl)
+            val tu = curlyTypingUnit
+            val (ctor, body) = tu.entities.partition {
+              case _: Constructor => true
+              case _ => false
+            }
+
+            val (real_params, ctorLoc) =
+              if (ctor.length > 1) {
+                err(msg"more than one constructor" -> S(l0) :: Nil)
```
LPTK: Improve

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -297,10 +304,23 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], raiseFun: D
       case (SPACE, _) :: _ => consume; block
       case c =>
         val t = c match {
+          case ModifierSet(mods, (KEYWORD("constructor"), l0) :: c) =>
```
LPTK: I don't understand why `constructor` is considered a modifier. It would only make sense if we wrote `constructor fun` or other modified declaration...

shared/src/main/scala/mlscript/helpers.scala
NeilKleistGao: Removed.

shared/src/main/scala/mlscript/NewLexer.scala
NeilKleistGao: Reimplemented.

shared/src/main/scala/mlscript/NewParser.scala
NeilKleistGao: Removed.

shared/src/main/scala/mlscript/NewParser.scala
NeilKleistGao: Removed.

LPTK:
Note: it's overridden, not overrided 😉

shared/src/main/scala/mlscript/NewParser.scala
NeilKleistGao: Replaced with `A class may only have one constructor`.

shared/src/test/diff/codegen/OptionalParam.mls
NeilKleistGao:
```ts
class M()
let mm = new M()
//│ globalThis.mm = new M.class();
```

shared/src/main/scala/mlscript/NewParser.scala
NeilKleistGao: Changed in another discussion.

shared/src/test/diff/codegen/OptionalParam.mls
```diff
@@ -435,3 +435,19 @@ new hh(1)
...
 //│ 2
+
+:e
+module O {
+  constructor(x: int)
+}
+mixin P {
+  constructor(x: int)
+}
+//│ ╔══[ERROR] module constructor overriden is not supported
```
LPTK:
```suggestion
//│ ╔══[ERROR] Explicit module constructors are not supported.
```

shared/src/test/diff/codegen/OptionalParam.mls
```diff
@@ -435,3 +435,19 @@ new hh(1)
...
+//│ ║  l.441: 	  constructor(x: int)
+//│ ╙──       	  ^^^^^^^^^^^^^^^^^^^
+//│ ╔══[ERROR] mixin constructor overriden is not supported
```
LPTK: Same.

LPTK:
Is this cleaned up and ready to review?

NeilKleistGao:
> Is this cleaned up and ready to review?

Yes. Review please. 😄

shared/src/main/scala/mlscript/syntax.scala
```diff
@@ -81,6 +81,7 @@ final case class Where(body: Term, where: Ls[Statement])             extends Ter
 final case class Forall(params: Ls[TypeVar], body: Term)             extends Term
 final case class Inst(body: Term)                                    extends Term
 final case class Super()                                             extends Term
+final case class Constructor(params: Tup, body: Ls[Statement])       extends Term // constructor(...) { ... }
```
LPTK: Why make this a Term instead of a statement?! Does something like `constructor(){} + 1` make sense to you?

NeilKleistGao: Use `final case class Constructor(params: Tup, body: Ls[Statement]) extends Statement // constructor(...) { ... }` instead.

LPTK: Okay, but the body should just be a `Term`, for simplicity and consistency.

shared/src/main/scala/mlscript/syntax.scala
```diff
@@ -181,12 +183,15 @@ final case class NuTypeDef(
   nme: TypeName,
   tparams: Ls[(Opt[VarianceInfo], TypeName)],
   params: Tup, // the specialized parameters for that type
+  ctor: Opt[Constructor],
+  isPlain: Bool,
```
LPTK:
```suggestion
  isPlainJSClass: Bool,
```

Please use this more understandable name everywhere.

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -369,8 +392,21 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], raiseFun: D
...
-            val res = NuTypeDef(kind, tn, tparams, params, sig, ps, N, N, body)(isDecl)
+            val tu = curlyTypingUnit
+            val (ctors, body) = tu.entities.partitionMap {
+              case c: Constructor => L(c)
+              case t => R(t)
+            }
+
+            val ctor =
+              if (ctors.lengthIs > 1) {
+                err(msg"A class may only have one constructor" -> S(l0) :: Nil)
```
```suggestion
                err(msg"A class may only have at most one explicit constructor" -> S(l0) :: Nil)
```

LPTK (commented):
I am not sure you satisfactorily handled my comment at https://github.com/hkust-taco/mlscript/pull/161#issuecomment-1543301563. Are there any situations where `isPlain`/`isPlainJSClass` is `true` and we *also* have class parameters? If not, why not just make the class parameters optional, as I already suggested, and make `isPlainJSClass` a derived method?

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1311,15 +1311,14 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool)
     def goDecl(d: NuMember)(implicit ectx: ExpCtx): NuDecl = d match {
       case TypedNuAls(level, td, tparams, body) =>
         ectx(tparams) |> { implicit ectx =>
-          NuTypeDef(td.kind, td.nme, td.tparams, Tup(Nil), N, td.isPlain, S(go(body)), Nil, N, N, TypingUnit(Nil))(
+          NuTypeDef(td.kind, td.nme, td.tparams, S(Tup(Nil)), N, S(go(body)), Nil, N, N, TypingUnit(Nil))(
```
```suggestion
          NuTypeDef(td.kind, td.nme, td.tparams, N, N, S(go(body)), Nil, N, N, TypingUnit(Nil))(
```

LPTK: Also it would be good to make sure type aliases were parsed into `NuTypeDef` instances with `N` as params.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1329,9 +1328,8 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool)
       case TypedNuCls(level, td, ttu, tparams, params, members, thisTy) =>
         ectx(tparams) |> { implicit ectx =>
           NuTypeDef(td.kind, td.nme, td.tparams,
-            Tup(params.map(p => N -> Fld(false, false, Asc(p._1, go(p._2.ub))))),
+            S(Tup(params.map(p => N -> Fld(false, false, Asc(p._1, go(p._2.ub)))))),
```
LPTK: The correct thing to do here is actually `Option.when(td.params.isDefined)(Tup(...))`.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -733,17 +733,19 @@ trait StatementImpl extends Located { self: Statement =>
...
           dataDefs.map(td => AppliedType(td.nme, td.tparams)).reduceOption(Union).getOrElse(Bot), Nil, Nil, Nil
         ).withLocOf(hd) :: cs)
-    case NuTypeDef(Nms, nme, tps, tup @ Tup(fs), ctor, plain, sig, pars, sup, ths, unit) =>
+    case NuTypeDef(Nms, nme, tps, tup, ctor, sig, pars, sup, ths, unit) =>
       ??? // TODO
-    case NuTypeDef(k @ Als, nme, tps, tup @ Tup(fs), ctor, plain, sig, pars, sup, ths, unit) =>
+    case NuTypeDef(k @ Als, nme, tps, tup, ctor, sig, pars, sup, ths, unit) =>
       // TODO properly check:
-      require(fs.isEmpty, fs)
+      require(tup.getOrElse(Tup(Nil)).fields.isEmpty, tup)
```
```suggestion
      require(tup.forall(tup => tup.fields.isEmpty, tup))
```

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -136,13 +136,13 @@ trait TypeLikeImpl extends Located { self: TypeLike =>
...
       // })).mkString("", "\n", "\n")
       })).mkString
-    case NuTypeDef(kind @ Als, nme, tparams, params, ctor, plain, sig, parents, sup, ths, body) =>
+    case NuTypeDef(kind @ Als, nme, tparams, params, ctor, sig, parents, sup, ths, body) =>
       s"type ${nme.name}${tparams.map(_._2.showIn(ctx, 0)).mkStringOr(", ", "[", "]")} = ${
         sig.getOrElse(die).showIn(ctx, 0)}"
-    case NuTypeDef(kind, nme, tparams, params, ctor, plain, sig, parents, sup, ths, body) =>
+    case NuTypeDef(kind, nme, tparams, params, ctor, sig, parents, sup, ths, body) =>
       val bodyCtx = ctx.indent
       s"${kind.str} ${nme.name}${tparams.map(_._2.showIn(ctx, 0)).mkStringOr(", ", "[", "]")}(${
```
LPTK: Let's refactor this bit so that a class with no parameter lists actually is _not_ printed as having an empty parameter list!

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -879,8 +881,8 @@ trait StatementImpl extends Located { self: Statement =>
     case Super() => Nil
     case Constructor(params, body) => params :: body :: Nil
     case Ass(lhs, rhs) => lhs :: rhs :: Nil
-    case NuTypeDef(k, nme, tps, ps, ctor, plain, sig, pars, sup, ths, bod) =>
-      nme :: tps.map(_._2) ::: ps :: pars ::: ths.toList ::: bod :: Nil
+    case NuTypeDef(k, nme, tps, ps, ctor, sig, pars, sup, ths, bod) =>
+      nme :: tps.map(_._2) ::: ps.getOrElse(Tup(Nil)) :: pars ::: ths.toList ::: bod :: Nil
```
```suggestion
      nme :: tps.map(_._2) ::: ps.toList ::: pars ::: ths.toList ::: bod :: Nil
```

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -849,11 +849,11 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], raiseFun: D
         }).withLoc(acc.toLoc.fold(some(loc))(_ ++ loc |> some))
         exprCont(res, prec, allowNewlines)
         
-      case (br @ BRACKETS(Square, toks), loc) :: _ => // * Currently unreachable because we match Square brackets as tparams
-        consume
-        val idx = rec(toks, S(br.innerLoc), "subscript").concludeWith(_.expr(0))
-        val res = Subs(acc, idx.withLoc(S(loc)))
-        exprCont(res, prec, allowNewlines)
+      // case (br @ BRACKETS(Square, toks), loc) :: _ => // * Currently unreachable because we match Square brackets as tparams
```
LPTK: Let's use `/*`/`*/` comments on the lines before/after to minimize useless diffs (please amend the commit and force-push).

shared/src/main/scala/mlscript/helpers.scala
NeilKleistGao: added a pattern matching to check if parameter lists are none. should I also apply this on the `compiler` subproject?

LPTK: What do you mean?

NeilKleistGao: in `compiler/shared/main/scala/mlscript/compiler/PrettyPrinter.scala`:
```scala
def showTypeDef(tyDef: NuTypeDef, indent: Int = 0): String =
  s"${tyDef.kind.str} ${tyDef.nme.name}"
    + (if tyDef.tparams.isEmpty
       then ""
       else tyDef.tparams.map(_._2.name).mkString("[", ",", "]"))
    + "(" + tyDef.params.getOrElse(Tup(Nil)) + ")"
    + (if tyDef.parents.isEmpty
       then ""
       else ": " + tyDef.parents.map(_.toString).mkString(", "))
    + showTypingUnit(tyDef.body, indent + 1)
```
Should we also check if `tyDef.params` is none?

LPTK: Sure, but I wasn't aware that the type definition pretty-printing functionality was duplicated in this sub-project (probably because it predates the implementation in the main project). Do you think we could just use the main project implementation here?

LPTK: Never mind, I just noticed that this is for debugging.

https://github.com/hkust-taco/mlscript/blob/703200f0cf9f16f654a85df326e8a32bb7e2728f/compiler/shared/main/scala/mlscript/compiler/PrettyPrinter.scala#L6

So it's probably fine to just update it as you described and otherwise leave it alone!

LPTK (approved):
Great job! Please resolve the last conversation and then let me know so I make the merge.

shared/src/main/scala/mlscript/helpers.scala
NeilKleistGao: Updated.

PR #164, opened 2023-05-13
Preliminary Implementation: Metaprogramming with Quasiquote

kkdlau:
## Description

This pull request introduces a preliminary implementation of metaprogramming using the quasiquote concept.

## Changes
Here's a brief summary of the changes:
- `Token.scala`: Introduce two new `BracketKind` sub classes: `BracketKind.Quasiquote` and `BracketKind.QuasiquoteTriple`.
- `NewLexer.scala`: Update the lexer to support lexing quasiquote syntax, including `code"..."` and `code"""..."""`.
- `NewParser.scala`: Revise the parser to facilitate the generation of ASTs using the quasiquote tokens.
- `Codegen.scala`: Add a new class, `JSQuasiquoteRunFunctionBody`, which contains the source code of the quasiquote interpreter (the run function).
- `JSBackend.scala`: Extend `translateTerm` to support quasiquote translation to s-expressions.
- `Polyfill.scala`: Introduce two extra built-in functions - `Const` and `run`.
- `Scope.scala`: Modify `Scope` class to support tracing both bound and free variables in quasiquotes during the code generation process.
- `quasiquote/*.mls`: Add new test cases to validate the quasiquote implementation.
- `typer.scala`: Implement a new primitive, `Code[T]`, which represents the type of a quasiquote. Here, `T` indicates the return type of the quasiquote execution.

## Testing

Unit tests have been added to ensure the correct functionality of the quasiquote operations. These tests cover both basic use cases and edge cases. All the test cases can be found at `shared/src/test/diff/quasiquote/`.

## Example

Here are some examples of using the new quasiquote feature.

Define a quasiquote:
```ts
code"1"
```

Define a quasiquote with a string inside:
```ts
code""" "github" """ // works without putting the space between """ and ", but we formatted for better readability
```

Define a quasiquote with a free variable:
```ts
code"a" // error - free variable is not allowed in this version
```

Execute a quasiquote:
```ts
code"1 + 1"
run(res) // res: 2
```


LPTK:
The CI does not pass as the code has many warnings...

Question: what happened to the "context-tracking" part of this branch called kkdlau:quasiquote/*context-tracking*?

LPTK:
Ok so when the CI does not pass you just remove the CI? Great job!

Let's revisit these changes later on. We'll probably selectively import what we need in a clean way.

Thanks for your work on this project anyway!

kkdlau:
For the clarification, I didn't remove the CI, instead as the CI runs only if I push something to branch mlscript / make a pull request on branch mlscript, I update the trigger condition to every push. This can help me understand why the CI by running it on my repo and get the full report of it. We will make sure the CI file is aligned with the parent repo

kkdlau:
> Ok so when the CI does not pass you just remove the CI? Great job!
> 
> Let's revisit these changes later on. We'll probably selectively import what we need in a clean way.
> 
> Thanks for your work on this project anyway!

For the clarification, I didn't remove the CI, instead as the CI runs only if I push something to branch mlscript / make a pull request on branch mlscript, I update the trigger condition to every push so everytime I push to my repo, I can see if the CI passes or not.

This also  can help me understand why the CI by running it on my repo and get the full report of it. We will make sure the CI file is aligned with the parent repo

LPTK:
I see, thanks for the clarification.

PR #167, opened 2023-06-05
Fix top-level mutually-recursive functions

NeilKleistGao:


LPTK (approved):
LGTM, thanks for the quick response time!

PR #169, opened 2023-06-08
Fix `this` capture for mixins in `globalThis`

NeilKleistGao:


shared/src/test/diff/codegen/MixinCapture.mls
```diff
@@ -0,0 +1,65 @@
...
+
+module TestLang extends EvalAddLit
+//│ module TestLang {
+//│   fun eval: Lit -> Int
+//│ }
+
+TestLang.eval(Lit(0))
+//│ Int
+//│ res
+//│     = 0
```
LPTK, line 65:
```suggestion
//│     = 0


```

LPTK (approved):
LGTM

PR #171, opened 2023-06-26
Support OCaml-style ADTs and pattern matching, port OCaml's List module, and add occurs-check

twitu:


shared/src/main/scala/mlscript/MLParser.scala
```diff
@@ -293,9 +294,9 @@ class MLParser(origin: Origin, indent: Int = 0, recordLocations: Bool = true) {
   })
   def litTy[p: P]: P[Type] = P( lit.map(l => Literal(l).withLocOf(l)) )
   
-  def toplvl[p: P]: P[Statement] =
-    P( defDecl | tyDecl | termOrAssign )
-  def pgrm[p: P]: P[Pgrm] = P( (";".rep ~ toplvl ~ topLevelSep.rep).rep.map(_.toList) ~ End ).map(Pgrm)
+  def toplvl[p: P]: P[Ls[Statement]] =
+      P(adtTyDecl) | P( defDecl | tyDecl | termOrAssign ).map(_ :: Nil)
```
LPTK: Why increase the indentation?

shared/src/main/scala/mlscript/MLParser.scala
```diff
@@ -304,7 +305,186 @@ class MLParser(origin: Origin, indent: Int = 0, recordLocations: Bool = true) {
...
+  ///////////////////////////////////////////////////////
+  /// ADT types
+  ///////////////////////////////////////////////////////
+
+  // Param defined like 'a
+  def adtTyParam[p: P]: P[TypeName] = locate(P(("'" ~~ lowercase.!).map(param =>
+    TypeName("'" + param)
+  )))
+
+  // Params defined like ('a, 'b) or 'a
```
LPTK: Can we use the existing type parameter syntax, for uniformity? As in: `List[A]` instead of `'a List`.

shared/src/main/scala/mlscript/MLParser.scala
```diff
@@ -304,7 +305,186 @@ class MLParser(origin: Origin, indent: Int = 0, recordLocations: Bool = true) {
...
+      case (id, None) => TypeDef(Cls, id, Nil, Record(Ls.empty), Nil, Nil, Nil)
+    })
+
+  // https://v2.ocaml.org/manual/typedecl.html#ss:typedefs
+  // TODO: handle record style type representation
+  // type 'a lst = [Null | Cons of 'a * 'a lst]
+  def adtDataCtor[p: P]: P[List[TypeDef]] = adtCtorDecl.rep(1, "|").map(_.toList)
+
+  def adtTyDecl[p: P]: P[Ls[Statement]] =
+    P((kw("typeadt") ~ adtTyParams ~ ctorName ~ "=" ~/
```
LPTK: That keyword looks quite strange. Why not `datatype`?

shared/src/main/scala/mlscript/MLParser.scala
```diff
@@ -304,7 +305,186 @@ class MLParser(origin: Origin, indent: Int = 0, recordLocations: Bool = true) {
...
+            case (t, i) => Var(s"_$i") -> Field(N, t._2)
+          }.toList
+          (tparams, Record(rcdBody))
+      }
+
+  def ctorName[p: P]: P[TypeName] =
+    locate(P(uppercase ~~ (letter | digit | "_" | "'").repX).!.filter(!keywords(_)).map(TypeName))
+
+  /** data constructor declaration
+    * type 'a, 'b tup = [Tup of 'a * 'b]
```
LPTK, line 320: We already support type synonyms! There is no need to introduce an alternative syntax for them.

On the other hand, I am lost as to why you think `type ('a, 'b) tup = [Tup of 'a * 'b]` should be a valid type declaration. The `of` makes no sense in this position.

shared/src/main/scala/mlscript/syntax.scala
```diff
@@ -19,7 +22,9 @@ final case class TypeDef(
   mthDecls: List[MethodDef[Right[Term, Type]]],
   mthDefs: List[MethodDef[Left[Term, Type]]],
   positionals: Ls[Var],
-) extends Decl
+) extends Decl {
+  var adtInfo: Opt[AdtInfo] = N
```
LPTK: Does this need to be mutable? Note that you can parameterize parser definitions with arbitrary parameters, such as passing in an `Option[AdtInfo]` when parsing cases.

shared/src/test/diff/mlscript/AdtStyle.mls
```diff
@@ -0,0 +1,59 @@
...
+//│ Plus: (Exp, Exp,) -> Exp
+//│     = <missing implementation>
+//│ Times: (Exp, Exp,) -> Exp
+//│      = <missing implementation>
+//│ Skip: Stmt
+//│     = <missing implementation>
+//│ Assign: (string, Exp,) -> Stmt
+//│       = <missing implementation>
+
+typeadt FCPoly = ('a -> 'a) List
```
LPTK: I don't get what this tries to do.

LPTK (commented):
Cool, looks promising!

Now we're only missing some new type checking code to type check pattern matching as expected. I suggest we import the `match ... with ...` syntax from the other project and reproduce how it's typed here.

shared/src/main/scala/mlscript/MLParser.scala
twitu: Well the problem with using the existing syntax `tyName` is that `Int` and `A` are the same. So a constructor like `Tup of Int * A` doesn't know by construction that it uses a type variable in it's body. This information is needed to fill the `paramPos` field in `AdtInfo`.

For e.g. in `datatype Tup[A, B] = IntTup of Int * B`, the `paramPos` for `IntTup` will be `List(1)`.

```
final case class AdtInfo(alsName: TypeName, paramPos: Ls[Int])
```

Maybe I can add a separate step that traverses the term tree and finds all the type variables by checking that it's a single capitalized alphabetic character.

shared/src/main/scala/mlscript/MLParser.scala
twitu: The bracket [] shows the part of the statement that is parsed. I'll remove the syntax for type alias.

shared/src/test/diff/mlscript/AdtStyle.mls
twitu: That's an older commit without forall.

```
typeadt 'a Poly = ('a -> 'a) List
//│ Defined type alias Poly[='a]

typeadt FCPoly = (forall 'a. 'a -> 'a) List
```

shared/src/main/scala/mlscript/MLParser.scala
twitu: On the other hand I can invent new syntax and simplify the whole thing with something like this

```
datatype Either[A, B] = Left[A] of A | Right[B] of B
datatype Tup[A, B] = IntTup[B] of Int * B | StrTup[A] of A * Str
```

This way the constructor has the information of what is parameterized.

LPTK: I don't understand why you think supporting `datatype Either[A, B]` is harder than supporting `('a, 'b) either`. The two syntaxes are isomorphic and contain the same amount of information. What's the problem?



> I can invent new syntax and simplify the whole thing with something like this
> 
> ```
> datatype Either[A, B] = Left[A] of A | Right[B] of B
> ```

I don't think we need that. Why would `Left` and `Right` need their own type parameter lists? That's not what happens in the current error messages prototype.

To be clear, we _do not want_ for `Left` and `Right` to introduce their own types. Remember we want ML-style typing of these ADTs, meaning the only type here is `Either[A, B]` and `Left`/`Right` are just constructors, to be used in expressions and patterns.

OTOH, we still want to generate `Left`/`Right` classes but that's just for the purpose of code generation. You can give them the same tparams as `Either`. But their ctors need to be shadowed with traditional the ML signature.

LPTK: So I think the generated code should look like:

```
class Either[A, B]
class Left[A, B](value: A) extends Either[A, B]
class Right[A, B](value: B) extends Either[A, B]
def Left[A, B] = Left : A -> Either[A, B]
def Right[A, B] = Right : B -> Either[A, B]
```
Notice the upcasts in the defs shadowing the case ctors.

LPTK:
I see you're trying to support several forms of ADTs, such as ones with records:

```
datatype List[A] = Cons[A] of {head: A; tail: List[A]} || Nil
```

I am not sure it is very useful. We could just stick with only supporting the following ADT definition format:

```
datatype List[A] = Cons[A](A, List[A]), Nil
```

shared/src/main/scala/mlscript/MLParser.scala
twitu: data declarations don't create new types only new constructor functions
The information is still captured in the `Either` typedef `AdtInfo` so that it can be used for codegen

LPTK (changes requested):
Thanks, this is great!

However, pattern matching on literals is currently typed too restrictively, which I expect to be an issue in actual code. For example:

```fs
foo x = match x with
  | 0 -> 0
  | 1 -> 1
//│ foo: nothing -> (0 | 1)

foo 0
//│ ╔══[ERROR] Type mismatch in application:
//│ ║  l.92: 	foo 0
//│ ║        	^^^^^
//│ ╟── integer literal of type `0` does not match type `1`
//│ ║  l.92: 	foo 0
//│ ║        	    ^
//│ ╟── Note: constraint arises from integer literal:
//│ ║  l.89: 	  | 1 -> 1
//│ ║        	    ^
//│ ╟── from reference:
//│ ║  l.87: 	foo x = match x with
//│ ╙──      	              ^
//│ res: 0 | 1 | error
```

Thankfully there is a workaround:

```fs
foo x = match x with
  | (0 : int) -> 0
  | (1 : int) -> 1
//│ foo: int -> (0 | 1)

foo 0
//│ res: 0 | 1
```

However I was curious how this could work. It turns out you type check patterns as expressions, which allows nonsensical things like:

```fs
foo x = match x with
  | (id 0) -> 0
```

Adding the `:d` flag crashed because there are unimplemented cases in debug printing.

To proceed, it'd be great if you could please:

 - [x] Implement all missing cases. For JS code-gen, you can use `???` for now, as we don't need to support JS immediately.
 - [x] Make sure there are no warnings left in the project.
 - [x] Type check literal patterns less strictly. It would probably be sufficient for now to just type check a `match`-`with` pattern like `1` as `int`

Later, if you have time:
 - [x] Reject nonsensical patterns or (harder) make them not representable and discard them during parsing.

LPTK (approved):
Thanks guy for the contributions!

PR #172, opened 2023-06-28
Add `driver` subproject

NeilKleistGao:
- Implement multiple-file compilation
- Add `import` and `export` keywords
- Update `ts2mls` DiffTests and test files
- Import `libes5.d.ts` and cherry-pick from `dom.d.ts`
- Add `UnusableLike` and `Unsupported` types
- Add ident escaping
- Implement invocation of ts libraries in mlscript
- Allow overriding built-in types from `Predef.mlsi`
- Optimize `NewParser.block` with tail-recursion
- Fix literal typing
- Handle selections in inheritance
- Update the implementation check for `declare` keyword
- Enhance `ts2mls`:
  - Mark complex signatures as unsupported
  - Fix generation of overloading, optional parameters and var-length parameters
  - Support `import`/`require` and `export` parsing
  - Support `export as namespace` and triple slashes
  - Fix flag checking
  - Fix private members' name


LPTK:
Wow impressive, that looks like a big leap forward!

```
[info] - ./mlscript/Lodash.mls *** FAILED ***
[info]   false equaled false failed when compiling ./mlscript/Lodash.mls. (DriverDiffTests.scala:21)
````

What's with the weird way tests fail?

Also, the test should be green. If there are tests that are expected to fail, there should be a way of specifying so.

**Comment by NeilKleistGao** [comment]
> Wow impressive, that looks like a big leap forward!
> 
> ```
> [info] - ./mlscript/Lodash.mls *** FAILED ***
> [info]   false equaled false failed when compiling ./mlscript/Lodash.mls. (DriverDiffTests.scala:21)
> ```
> 
> What's with the weird way tests fail?
> 
> Also, the test should be green. If there are tests that are expected to fail, there should be a way of specifying so.

The test failed because I forgot to update the CI script to install necessary libraries. 

> 

Sure! I will try to fix the CI before tomorrow's meeting. If there are any other problems, we can also discuss about them in the meeting.
(The CI has been canceled due to some JVM problems. I will also handle this.)

File: driver/js/src/test/esprojects/mlscript/MLS2TheMax.mls [hunk 1]
```diff
@@ -0,0 +1,44 @@
+import "../ts/ReadLine.ts"
+import "./tools/Concat.mls"
+
+fun ask(question: Str) =
+  console.log(question)
+  ReadLine.getStrLn()
+
+class Some<A>(value: A)
```
LPTK: I think we'll move away from angle brackets for now. May revisit later.
```suggestion
class Some[A](value: A)
```

driver/js/src/test/esprojects/mlscript/MLS2TheMax.mls
```diff
@@ -0,0 +1,44 @@
...
+class None()
```
LPTK:
```suggestion
module None
```

driver/js/src/test/esprojects/mlscript/MLS2TheMax.mls
```diff
@@ -0,0 +1,44 @@
...
+
+fun parse(s: Str) =
+  let i = parseInt(s, undefined)
+  if
+    isNaN(i) then None()
+    _ then Some(i)
```
LPTK: That's a funny way of writing
```suggestion
  if isNaN(i)
    then None()
    else Some(i)
```

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -34,8 +34,9 @@ class ConstraintSolver extends NormalForms { self: Typer =>
     val info = ctx.tyDefs2.getOrElse(clsNme, die/*TODO*/)
     
     if (info.isComputing) {
-      
-      ??? // TODO support?
+
+      L(ErrorReport(
+        msg"${info.decl.kind.str.capitalize} `${info.decl.name}` is not supported yet." -> fld.toLoc :: Nil))
```
LPTK: How could this be the appropriate error here? Do you have an example?

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -759,6 +766,7 @@ class ConstraintSolver extends NormalForms { self: Typer =>
           case (p @ ProvType(und), _) => rec(und, rhs, true)
           case (_, p @ ProvType(und)) => rec(lhs, und, true)
           case (_: TypeTag, _: TypeTag) if lhs === rhs => ()
+          case (lhs: Unsupported, rhs: Unsupported) => ()
```
LPTK, line 795: That looks very fishy. Do you have an example?

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -1040,6 +1048,7 @@ class ConstraintSolver extends NormalForms { self: Typer =>
       
       def doesntMatch(ty: SimpleType) = msg"does not match type `${ty.expNeg}`"
       def doesntHaveField(n: Str) = msg"does not have field '$n'"
+      def doesntSupport(uns: Unsupported) = msg"(TypeScript type ${uns.showTSType}) is not supported"
```
LPTK, line 1051: Why the strange parenthesization?

shared/src/main/scala/mlscript/ConstraintSolver.scala
LPTK: I now see that you meant to fix this example:

```ts
module M {
  class A
  fun a: M.A
  fun a = 1
}
//│ /!!!\ Uncaught error: scala.NotImplementedError: an implementation is missing
//│ ╔══[ERROR] Module `M` is not supported yet.
//│ ║  l.170: 	  fun a: M.A
//│ ╙──       	          ^^
```

That's clearly not the correct error. For now you could make it:
```suggestion
      L(ErrorReport(
        // TODO improve
        msg"${info.decl.kind.str.capitalize} `${info.decl.name}` has a cyclic type dependency that is not supported yet." -> fld.toLoc :: Nil))
```

shared/src/main/scala/mlscript/ConstraintSolver.scala
NeilKleistGao:
```ts
interface ErrorConstructor {
    new(message?: string): Error;
    (message?: string): Error;
    readonly prototype: Error;
}

interface EvalErrorConstructor extends ErrorConstructor {
    new(message?: string): EvalError;
    (message?: string): EvalError;
    readonly prototype: EvalError;
}
```

In this case, `new(message?: string): Error` and `new(message?: string): EvalError` are construction signatures and not supported. Since `EvalErrorConstructor` extends `ErrorConstructor`, the constraint solver will try to solve there two construction signatures, which leads to type errors in `es5.mlsi`. So here I marked that two `Unsupported` types are always the same. But you are right, maybe we can apply a better solution in `ts2mls`.

LPTK: Ok that's fine, but explain why this is needed in a comment, ie "required to allow signature overriding checks in declarations"

driver/js/src/test/scala/driver/DriverDiffTests.scala
```diff
@@ -0,0 +1,133 @@
...
+    tsconfig: Option[String] = None,
+    ignoreTypeError: Boolean = false,
+    expectError: Boolean = false
+  ) = driverEntry(entryModule, tsconfig, s"${diffPath}cjsprojects/",
+    s"${diffPath}cjsprojects/js/", ignoreTypeError, expectError, true)
+
+  private def esEntry(
+    entryModule: String,
+    tsconfig: Option[String] = None,
+    ignoreTypeError: Boolean = false,
```
LPTK:
```suggestion
    expectTypeError: Boolean = false,
```

driver/js/src/test/scala/driver/DriverDiffTests.scala
```diff
@@ -0,0 +1,133 @@
...
+        case None => ()
+      }
+    }
+  }
+}
+
+object DriverDiffTests {
+  // For local environment, we may change the driver so forcing compiling is necessary
+  // but we can ban it during CI
+  private val forceCompiling = !sys.env.get("CI").isDefined
```
LPTK: For now:
```suggestion
  private val forceCompiling = true
```

LPTK (changes requested):
Change requests discussed in the meeting:
 - [x] fail tests when errors are expected but not there
 - [x] redirect error, warning, and debugging messages to the mlsi file as comments
 - [x] use the same logic as the classical difftests regarding which tests to rerun based on git

LPTK:
> > What's with the weird way tests fail?
> 
> The test failed because I forgot to update the CI script to install necessary libraries.
> 

What I meant is that when a test fails, it reports a meaningless message "false equaled false" instead of an informative explanation of the failure!

NeilKleistGao:
- [x] Improve error messages when test cases failed.
- [x] Generate selection parents
- [x] Remove `tsc`-generated files and add corresponding scripts

driver/js/src/main/scala/driver/Driver.scala
```diff
@@ -30,6 +30,9 @@ class Driver(options: DriverOptions) {
       println(msg)
   }
 
+  // errors in imported files should be printed in their own files to avoid redundant
```
LPTK:
```suggestion
  // errors in imported files should be printed in their own files to avoid redundancy
```

driver/js/src/test/esprojects/.interfaces/mlscript/Self.mlsi
```diff
@@ -0,0 +1,4 @@
+declare module Self() {
+  class Foo()
+}
+//| can not import driver/js/src/test/esprojects/mlscript/Self.mls itself
```
LPTK:
```suggestion
//| Cannot import driver/js/src/test/esprojects/mlscript/Self.mls itself
```
Does `itself` mean this is a cyclic import? It should be rephrased because currently it's unclear.

NeilKleistGao: Yes. In `Self.mls`:
```ts
import "./Self.mls"
```

ts2mls/js/src/test/scala/ts2mls/TSTypeGenerationTests.scala
```diff
@@ -8,40 +8,18 @@ class TSTypeGenerationTest extends AnyFunSuite {
...
+      false, None, (file: FileInfo, writer: JSWriter) => ()) // no need for builtin
     program.generate
   })
 }
 
 object TSTypeGenerationTest {
+  // we only generate type information for builtin declarations here.
+  // user-defined scripts may contain errors and printing error messages can lead to test failure
+  // if we use the two-pass test.
+  // builtin declarations should never contain an error.
```
LPTK: Plz start comment sentences with capital letters. Makes them easier to parse.

driver/js/src/test/esprojects/.interfaces/mlscript/Self.mlsi
LPTK:
```suggestion
//| Cannot import driver/js/src/test/esprojects/mlscript/Self.mls from itself
```

ts2mls/js/src/main/scala/ts2mls/types/Converter.scala
```diff
@@ -115,7 +115,7 @@ object Converter {
   private def convertRecord(members: Map[String, TSMemberType], typeVars: List[TSTypeParameter], parents: List[TSType],
     statics: Map[String, TSMemberType], constructorList: List[TSType], anonymous: Boolean)(implicit indent: String) = {
     val allRecs = members.toList.map((m) => m._2.modifier match {
-      case Public =>
+      case Public if !m._1.contains("#") =>
```
LPTK: Is this meant to detect private fields? If so, it should use `startsWith` not `contains`.

NeilKleistGao: Yes. But so far we are reading the member's symbol name, which may be renamed by the `tsc`(e.g., `#bar` -> `__#1@#bar`). I will move the detection logic to the parsing part to get the untouched variable name rather than check during translation.

LPTK: Oh... so does TS make the fields non-private, then?

NeilKleistGao: No. This is just because in `tsc`, `private field: type` and `#field: type` have different structures. The first one uses a private keyword and an identifier, while the second one uses a private identifier(and renames the symbol).  Here is [an example](https://ts-ast-viewer.com/#code/MYGwhgzhAEBiD29oG8BQ1oGIBGYBOAXNAHYCuAttgKZ7rQAOeAlgG5gAuV0uAXkRO2bEA5qgC+qIA).
So far the parsing logic can only handle the identifiers, and that's why:
>  But so far we are reading the member's symbol name, which may be renamed by the tsc(e.g., #bar -> __#1@#bar). I will move the detection logic to the parsing part to get the untouched variable name rather than check during translation.

LPTK: All I can see from your link is that there's an `escapedName` field with containing `"__#1@#bar"`. This is obviously not something we should care about. By the way, my expectation is that this escaped name probably _is_ for the purpose of desugaring the private field into a non-private one somehow.

Note: weirdly, the TS palyground desugars your `Foo` example:
```ts
class Foo {
  #bar: number
  private baz: string
}
```
to:
```js
class Foo {
    constructor() {
        _Foo_bar.set(this, void 0);
    }
}
_Foo_bar = new WeakMap();
```
https://www.typescriptlang.org/play?#code/MYGwhgzhAEBiD29oG8CwAoa0DEAjMATgFzQB2ArgLa4CmBGWADgQJYBuYALjdPgF4kInVqQDmGAL4YgA

driver/npm/README.md
```diff
@@ -1 +1,43 @@
 # MLscript Driver
+
+**This is a pre-release of MLscript driver.**
```
LPTK: Why do you call this MLscript driver instead of MLscript? The "driver" name means nothing for people outside you and me...

ts2mls/js/src/main/scala/ts2mls/types/Converter.scala
NeilKleistGao: Replaced with correct private identifiers.

shared/src/main/scala/mlscript/syntax.scala
```diff
@@ -122,11 +122,11 @@ trait IdentifiedTerm
...
-final case class LetS(isRec: Bool, pat: Term, rhs: Term)  extends Statement
-final case class DataDefn(body: Term)                     extends Statement
-final case class DatatypeDefn(head: Term, body: Term)     extends Statement
-final case class Constructor(params: Tup, body: Blk)      extends Statement // constructor(...) { ... }
-class Import(val path: Str)                               extends Statement
+final case class LetS(isRec: Bool, pat: Term, rhs: Term) extends Statement
+final case class DataDefn(body: Term)                    extends Statement
+final case class DatatypeDefn(head: Term, body: Term)    extends Statement
+final case class Constructor(params: Tup, body: Blk)     extends Statement // constructor(...) { ... }
+class Import(val path: Str)                              extends Statement
```
LPTK: Not sure about these changes...

NeilKleistGao: Oh. I found I misaligned the original code in the previous commits, so I undid it in the refactor.

LPTK:
Thanks for extensive work!

As discusses, there are _a lot_ of changes in here that will need to be ported to the new compiler incrementally. I believe a lot of the implementation effort can be saved, but this will still be a painstaking process.

In the meantime, I will close this with a "to revisit" label.

PR #174, opened 2023-08-09
Named args

mbroughani81:
- rename the variables used in desugaring.
- delete unnecessary logs
- add test


.vscode/settings.json
```diff
@@ -12,5 +12,6 @@
             "other": "off",
             "strings": "off"
         }
-    }
+    },
+    "files.autoSave": "off"
```
chengluyu, line 17: What's this change for?

shared/src/main/scala/mlscript/NuTypeDefs.scala
```diff
@@ -536,7 +536,10 @@ class NuTypeDefs extends ConstraintSolver { self: Typer =>
     }
     ctx ++= infos
     ctx ++= sigInfos
-    
+    println("infos => " + infos)
+    println("sigInfos => " + sigInfos)
+    println("ctx => " + ctx)
+
```
chengluyu: Remember to remove changes only for debugging when the PR is ready to be merged.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -8,6 +8,8 @@ import scala.util.chaining._
 import scala.annotation.tailrec
 import mlscript.utils._, shorthands._
 import mlscript.Message._
+import mlscript.codegen.Helpers
+import mlscript.ucs.helpers
```
chengluyu: Remember to remove these imports after this PR is done.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -993,37 +995,158 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, var ne
...
+          val fun_ty = mkProxy(f_ty, funProv)
+            // ^ This is mostly not useful, except in test Tuples.fun with `(1, true, "hey").2`
+          val resTy = con(fun_ty, FunctionType(arg_ty, res)(
+            prov
+            // funProv // TODO: better?
+            ), res)
+          resTy
+        }
+        /////////////////////////////
+        def desugarNamedArgs(term: Term, f: Var, a: Tup): SimpleType = {
```
chengluyu: Consider extracting this method to class level?

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -993,37 +995,158 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, var ne
...
-              })(as match { // TODO dedup w/ general Tup case
-                case Nil | ((N, _) :: Nil) => noProv
-                case _ => tp(tup.toLoc, "argument list")
-              })
-            case _ => // can happen in the old parser
-              typeArg(a)
+        println("applying-to-function => " + f + "<" + f.getClass() + ">" + a + "<" + a.getClass() + ">")
+        println("now we should desugar in case of named args!")
+        /////////////////////////////
+        def typeUnnamedApp(f: Term, a: Term) = {
```
chengluyu: Consider extract this method to class level?

shared/src/test/diff/wander/Wander.mls
```diff
@@ -0,0 +1,254 @@
+:NewDefs
+
+fun fff(x, y, z) = (x - y) * z
+//│ fun fff: (Int, Int, Int,) -> Int
+
+:d
```
chengluyu: Don't forget to remove all debug flags after this PR is done.

chengluyu (changes requested):

NeilKleistGao:
Before making a formal PR(i.e., non-draft), please:
- read the [contribution guide](https://github.com/hkust-taco/mlscript/blob/mlscript/CONTRIBUTING.md)
- make sure your code passes all regression tests
- remove all compilation warning messages and unnecessary debug information

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -993,37 +995,158 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, var ne
...
+            // ^ This is mostly not useful, except in test Tuples.fun with `(1, true, "hey").2`
+          val resTy = con(fun_ty, FunctionType(arg_ty, res)(
+            prov
+            // funProv // TODO: better?
+            ), res)
+          resTy
+        }
+        /////////////////////////////
+        def desugarNamedArgs(term: Term, f: Var, a: Tup): SimpleType = {
+          val Var(name) = f;
```
NeilKleistGao:
```suggestion
          val Var(name) = f
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -993,37 +995,158 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, var ne
...
+          ctx.get(name) match {
+            case Some(value) => 
+              println("value => " + value  + " " + value.getClass())
+              value match {
+                case VarSymbol(ty, definingVar) => 
+                  println("definingVar => " + definingVar)
+                  println("ty => " + ty + " " + ty.getClass())
+                  ty match {
+                    case FunctionType(lhs, rhs) => 
+                      println("lhs, rhs => " + lhs + " " + rhs + " " + lhs.getClass() + " " + rhs.getClass())
```
NeilKleistGao:
```suggestion
                      println(s"lhs, rhs => $lhs $rhs ${lhs.getClass()} ${rhs.getClass()}")
```
You can use string templates to avoid unnecessary string concat.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -993,37 +995,158 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, var ne
...
+                  desugarNamedArgs(term, f1, a1)
+                } else {
+                  println("!!unnamed args case!!")
+                  typeUnnamedApp(f, a)
+                }
+            }
+          case _ =>
+            println("f type is => " + codegen.Helpers.inspect(f))
+            typeUnnamedApp(f, a)
         }
```
NeilKleistGao:
```suggestion
        (f, a) match {
          // in this case, we should check the arguments.
          case (f: Var, a @ Tup(fields)) if fields.exists(x => x._1.isDefined) =>
            // check arguments, if 
            println("f is var")
            println("!!named args case!!")
            desugarNamedArgs(term, f, a)
          case _ =>
            println("f type is => " + codegen.Helpers.inspect(f))
            typeUnnamedApp(f, a)
        }
```

shared/src/main/scala/mlscript/Typer.scala
NeilKleistGao: It seems that we can not desugar the program like this:
```ts
fun print(x: Int) = (y: Int, z: Int) => log([x, y, z])
let p = print(0)
p(z: 1, y: 2)

class Baz() {
  fun f(x: Int, y: Int) = log([x, y])
}
Baz().f(y: 1, x: 2)
```

NeilKleistGao (changes requested):

shared/src/main/scala/mlscript/Typer.scala
chengluyu: Nice catch!

.vscode/settings.json
mbroughani81: It is for turning the autosave off only for the current project.

NeilKleistGao: I think @chengluyu may mean that `files.autoSave` might be enabled by others, so this should not be committed to the repo.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -992,38 +994,64 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, var ne
...
-              typeArg(a)
-          }
+      case App(f: Term, a @ Tup(fields)) if (fields.exists(x => x._1.isDefined)) =>
+        val f_ty = typeTerm(f)
+        val fun_ty: SimpleType = f_ty.unwrapProxies match {
+          case tv: TypeVariable =>
+            def getLowerboundFuns(tv: TypeVariable): List[FunctionType] = {
+              val res: List[FunctionType] = tv.lowerBounds.map(x => 
+                x.unwrapProvs match {
+                  case PolymorphicType(_, ProvType(fun_ty @ FunctionType(_, _))) =>
```
LPTK: Use the `AliasOf` extractor instead.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -992,38 +994,64 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, var ne
...
+                  case tvv: TypeVariable =>
+                    getLowerboundFuns(tvv) 
+                  case _ =>
+                    err(s"unexpected type for lowerbound members! ${x.unwrapProvs.getClass()}", f.toLoc)
+                    Nil
+                }).flatten
+              res
+            }
+            val funs = getLowerboundFuns(tv)
+            if (funs.sizeCompare(1) < 0) {
```
LPTK: Use pattern matching on `List` with `Nil`, `f :: Nil` and `xs`.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1340,7 +1368,132 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, var ne
...
+          ))
+          App(f, y)
+      }
+    }
+    if (a.fields.exists(x => x._1.isDefined) &&
+        a.fields.exists(x => x._1.isEmpty) && 
+        a.fields.indexWhere(x => x._1.isDefined) < a.fields.lastIndexWhere(x => x._1.isEmpty)) {
+      err("the unnamed args should appear first when using named args!", a.toLoc) 
+    } else
+    if (a.fields.sizeCompare(argsList) > 0 || a.fields.sizeCompare(argsList) < 0) {
```
LPTK: Avoid Boolean blindness (https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/), avoid `sizeCompare` – use pattern matching again.

shared/src/test/diff/nu/NamedArgs.mls
```diff
@@ -1,79 +1,241 @@
...
 
-
 fun test(x: 'a) = if x is undefined then 0 else x + 1
 //│ fun test: (x: Int | undefined,) -> Int
 
-test(x: 0)
-//│ Int
+:e
+test(x: 0, 1)
+//│ ╔══[ERROR] the unnamed args should appear first when using named args!
```
LPTK: Please don't use informal language such as 'args' and '!' in error messages

shared/src/test/diff/nu/NamedArgs.mls
```diff
@@ -1,79 +1,241 @@
...
+//│ ║  l.8: 	test(x: 0, 1)
+//│ ╙──     	    ^^^^^^^^^
+//│ error
 //│ res
 //│     = 1
 
+// FIXME "dummy"
```
LPTK: Remove the FIXME

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1340,7 +1389,107 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, var ne
...
+    // we check all possibe prefix_num combination, till we found one that is not in the nonValidVars
+    val ints = LazyList.from(1)
+    val result = ints.find(index => {
+      !nonValidVars.contains(Var(prefix + "_" + index))
+    })
+    result match {
+      case Some(index) => 
+        prefix + "_" + index
+      case N => 
+        "ERROR" // unreachable, cause there must be an possible NewVar
```
NeilKleistGao: You can call `die` for unreachable branches:
```suggestion
        die // unreachable, cause there must be an possible NewVar
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1013,17 +1061,18 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, var ne
...
         val fun_ty = mkProxy(f_ty, funProv)
           // ^ This is mostly not useful, except in test Tuples.fun with `(1, true, "hey").2`
-        val resTy = con(fun_ty, FunctionType(arg_ty, res)(
+        val res = freshVar(prov, N)
+        val res_ty = con(fun_ty, FunctionType(arg_ty, res)(
           prov
           // funProv // TODO: better?
           ), res)
-        resTy
+        res_ty
```
NeilKleistGao: It seems that these changes are not necessary.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1340,7 +1389,107 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, var ne
...
+  (implicit ctx: Ctx, raise: Raise, vars: Map[Str, SimpleType]): SimpleType = {
+    def rec (as: List[(String -> Fld) -> Boolean], acc: Map[String, Either[Var, Term]]): Term = {
+      as match {
+        case ((v, fld), isNamed) :: tail =>
+          if (isNamed) {
+            fld.value match {
+              case lit: Lit =>
+                rec(tail, acc + (v -> R(fld.value)))
+              case varr: Var =>
+                rec(tail, acc + (v -> R(fld.value)))
```
NeilKleistGao:
```suggestion
              case _: Lit | _: Var =>
                rec(tail, acc + (v -> R(fld.value)))
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1340,7 +1389,107 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, var ne
...
+              case _ =>
+                val newVar = Var(getNewVarName(v, freeVars(ctx, a)))
+                Let(false, newVar, fld.value, rec(tail, acc + (v -> L(newVar))))
+            }
```
NeilKleistGao:
```suggestion
            }
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1340,7 +1389,107 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, var ne
...
+    })
+    result match {
+      case Some(index) => 
+        prefix + "_" + index
+      case N => 
+        "ERROR" // unreachable, cause there must be an possible NewVar
+    }
+  }
+
+  def freeVars(ctx: Ctx, t: Term): Set[Var] = {
```
NeilKleistGao: The parameter `ctx` is never used.
```suggestion
  def freeVars(t: Term): Set[Var] = {
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1340,7 +1389,107 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, var ne
         RecordType.mk(fs)(prov)
       } else TupleType(fields.reverseIterator.mapValues(_.toUpper(noProv)))(prov)
   }
-  
+
+  def getNewVarName(prefix: String, nonValidVars: Set[Var])(implicit raise: Raise): String = {
```
NeilKleistGao: The implicit parameter `raise` is never used.
```suggestion
  def getNewVarName(prefix: String, nonValidVars: Set[Var]): String = {
```

NeilKleistGao (commented):
Some minor suggestions for you to improve the code. 😉

LPTK:
Seems this PR doesn't yet take into account the change to `Fld`'s definition.

mbroughani81:
> Seems this PR doesn't yet take into account the change to `Fld`'s definition.

I resolved the suggested changes, and also merge this branch with the current new-definition-typing(on local), to resolve possible conflicts.

shared/src/test/diff/nu/NamedArgs.mls
```diff
@@ -1,79 +1,241 @@
...
 
 :e
-Bar(y: 1)
-//│ ╔══[ERROR] Wrong tuple field name: found 'y' instead of 'x'
-//│ ║  l.72: 	Bar(y: 1)
-//│ ╙──      	   ^^^^^^
-//│ Bar | error
+fun print(x) = (y, z) => log([x, y, z])
+let p = print(0)
+p(z: 1, y: 2)
```
LPTK, line 234: Why do we get _four_ repetitions of the same error here? Can't we avoid this with a little refactoring?

shared/src/test/diff/nu/NamedArgs.mls
```diff
@@ -1,79 +1,241 @@
...
-//│ Int
+//│ ╔══[ERROR] number of parameters dosen't match with the function signature!
+//│ ║  l.31: 	test(y: 0)
+//│ ╙──      	    ^^^^^^
+//│ error
 //│ res
-//│     = 1
+//│     = [ 0, undefined ]
+
+// FIXME extra error "name y is missed in function call"
```
LPTK: Remove the FIXME now that this is fixed

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1506,7 +1554,105 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+              ((value.name, x._2), true)
+            case N =>
+              ((argsList(idx).name, x._2), false)
+          }}
+          val asGroupedByVarName = as.groupBy(x => x._1._1)
+          if (asGroupedByVarName.sizeCompare(argsList) < 0) {
+            asGroupedByVarName.foreach(
+              x =>
+                x._2 match {
+                  case x1 :: y1 :: xs => err(s"parameter ${x._1} is duplicate!", a.toLoc)
```
```suggestion
                  case x1 :: y1 :: xs => err(s"Argument for parameter ${x._1} is duplicated", a.toLoc) 
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1506,7 +1554,105 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+            acc.get(x.name) match {
+              case Some(v) =>
+                v match {
+                  case Left(v) =>
+                    (None, Fld(FldFlags(false, false), v))
+                  case Right(t) => 
+                    (None, Fld(FldFlags(false, false), t))
+                }
+              case None =>
+                err(s"name ${x} is missed in function call", a.toLoc)
```
```suggestion
                err(s"Argument named ${x} is missing from this function call", a.toLoc)
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1506,7 +1554,105 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+    val ints = LazyList.from(1)
+    val result = ints.find(index => {
+      !nonValidVars.contains(Var(prefix + "_" + index))
+    })
+    result match {
+      case Some(index) => 
+        prefix + "_" + index
+      case N => 
+        die
+    }
```
```suggestion
    }).get + "_" + index
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1506,7 +1554,105 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+      case App(lhs, rhs) => 
+        freeVars(lhs) ++ freeVars(rhs)
+      case v @ Var(_) => 
+        Set(v)
+      case Tup(fields) =>
+        fields.map(f => freeVars(f._2.value))
+                    .flatMap(x => x)
+                    .toSet
+      case _ =>
+        Set() // TODO
```
LPTK: There's already a `freeVars` method you can call on `Term`!

LPTK (changes requested):
Ok it's starting to look better now! Please address the comments ~~and also merge the main branch into this branch to update it~~ (I just did it).

LPTK:
To solve the compilation failures, just use `FldFlags.empty` instead of `FldFlags(false, false)`.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1021,6 +1021,70 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+                    Nil
+                }).flatten
+              res
+            }
+            val funs = getLowerboundFuns(tv)
+            funs match {
+              case x :: Nil => 
+                funs.head
+              case Nil =>
+                err("Cannot extract any function", f.toLoc)
```
LPTK: This error is never exercised in the test. Besides, it should give some **context** as to what's happening! How do you expect *end users* to understand the source of the error with such message?

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1021,6 +1021,70 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+              case _ =>
+                err(s"More than one fun type found: ${funs}", f.toLoc)
```
LPTK: Same comment as above.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1021,6 +1021,70 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+          case tv: TypeVariable =>
+            def getLowerboundFuns(tv: TypeVariable): List[FunctionType] = {
+              val res: List[FunctionType] = tv.lowerBounds.map(x => 
+                x.unwrapProvs match {
+                  case PolymorphicType(_, AliasOf(fun_ty @ FunctionType(_, _))) =>
+                    List(fun_ty)
+                  case tvv: TypeVariable =>
+                    getLowerboundFuns(tvv) 
+                  case _ =>
+                    err("match error", f.toLoc)
```
LPTK: Why is this **user-facing error** reported? First, it doesn't make sense to report an error if one of the lower bounds happens not to be a function; second, the error message is completely meaningless!

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1021,6 +1021,70 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+                err("Cannot extract any function", f.toLoc)
+              case _ =>
+                err(s"More than one fun type found: ${funs}", f.toLoc)
+            }
+          case PolymorphicType(_, AliasOf(fun_ty @ FunctionType(_, _))) =>
+            fun_ty
+          case FunctionType(_, _) =>
+            f_ty
+          case _ =>
+            err("match error", f.toLoc)
```
LPTK: Why is this **user-facing error** reported?

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1021,6 +1021,70 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+            f_ty
+        }
+        // check if all
```
LPTK: Comment seems incomplete and doesn't make sense.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1021,6 +1021,70 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+        val hasUntypedArg = fun_ty.unwrapProxies match {
+          case FunctionType(TupleType(fields), _) =>
+            fields.exists(x => x._1 match {
+              case Some(arg) =>
+                false
+              case N =>
+                true
+            })
+          case _ => 
+            err("match error", f.toLoc)
```
LPTK: Why is this **user-facing error** reported?

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1021,6 +1021,70 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+            true
+        }
+        if (hasUntypedArg) {
+          err("Cannot use named args if there is untyped argument", a.toLoc)
```
```suggestion
          err("Cannot use named args as the function type has untyped arguments", a.toLoc)
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1021,6 +1021,70 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+        }
+        // check if all
+        val hasUntypedArg = fun_ty.unwrapProxies match {
+          case FunctionType(TupleType(fields), _) =>
+            fields.exists(x => x._1 match {
+              case Some(arg) =>
+                false
+              case N =>
+                true
+            })
```
```suggestion
            fields.exists(_._1.isEmpty)
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1021,6 +1021,70 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+        if (hasUntypedArg) {
+          err("Cannot use named args if there is untyped argument", a.toLoc)
+        } else {
+          val argsList = fun_ty.unwrapProxies match {
+            case FunctionType(TupleType(fields), _) =>
+              fields.map(x => x._1 match {
+                case Some(arg) =>
+                  arg
+                case N =>
+                  Var("error")
```
LPTK: Add a comment here to explain that this is used as a dummy term referring to the `error` builtin.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1021,6 +1021,70 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+              })
+            case _ => 
+              err("match error", f.toLoc)
```
LPTK: Why is this **user-facing error** reported?

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1554,7 +1618,84 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+        case Nil =>
+          val y: Term = Tup(argsList.map(x => 
+            acc.get(x.name) match {
+              case Some(v) =>
+                v match {
+                  case Left(v) =>
+                    (None, Fld(FldFlags.empty, v))
+                  case Right(t) => 
+                    (None, Fld(FldFlags.empty, t))
+                }
```
```suggestion
              case Some(Left(v)) => (None, Fld(FldFlags.empty, v))
              case Some(Right(t)) => (None, Fld(FldFlags.empty, t))
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1554,7 +1618,84 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+    } else 
+      a.fields.sizeCompare(argsList) match {
+        case 0 =>
+          val as = a.fields.zipWithIndex.map{ case(x, idx) =>
+          x._1 match {
+            case Some(value) => 
+              ((value.name, x._2), true)
+            case N =>
+              ((argsList(idx).name, x._2), false)
+          }}
```
LPTK: Indentation is missing for this block.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1554,7 +1618,84 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+          val asGroupedByVarName = as.groupBy(x => x._1._1)
+          if (asGroupedByVarName.sizeCompare(argsList) < 0) {
+            asGroupedByVarName.foreach(
+              x =>
+                x._2 match {
+                  case x1 :: y1 :: xs => err(s"Argument for parameter ${x._1} is duplicated", a.toLoc) 
+                  case _ =>     
+                }
+            )
```
```suggestion
            asGroupedByVarName.foreach(_._2 match {
              case x1 :: y1 :: xs => err(s"Argument for parameter ${x._1} is duplicated", a.toLoc)
              case _ =>
            })
```

LPTK (changes requested):
Actually there are still a bunch of things to improve before we can merge this.

You need to understand that `err` reports an error message to the programmer. Doing things like `err("match error", f.toLoc)` makes zero sense.

shared/src/main/scala/mlscript/Typer.scala
mbroughani81: Because of the use of x in error message, i think the former one is better.

shared/src/main/scala/mlscript/Typer.scala
mbroughani81: I think this case should not happen(it is not a user error, but rather some exceptional case not considered). Changing to ``die``

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1021,6 +1021,50 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+        }
+        val f_ty = typeTerm(f)
+        val fun_tys: List[FunctionType] = getLowerBoundFunctionType(f_ty)
+
+        fun_tys match {
+          case fun_ty :: Nil =>
+            val hasUntypedArg = fun_ty match {
+              case FunctionType(TupleType(fields), _) =>
+                fields.exists(_._1.isEmpty)
+              case _ => die // FIXME happens when lhs of function type is not TupleType(possible?)
```
LPTK: Please don't crash for no reason here. Just return the safe option that makes sense, which is `false`, since this is meant as a conservative test (it's not too bad if we reject more than strictly needed).

This case actually _can_ happen, although it's hard to trigger.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1021,6 +1021,50 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+            if (hasUntypedArg) {
+              err("Cannot use named arguments as the function type has untyped arguments", a.toLoc)
+            } else {
+              val argsList = fun_ty match {
+                case FunctionType(TupleType(fields), _) =>
+                  fields.map(x => x._1 match {
+                    case Some(arg) => arg
+                    case N => die // cannot happen, because already checked with the hasUntypedArg
+                  })
+                case _ => die // FIXME happens when lhs of function type is not TupleType(possible?)
```
LPTK: Same comment.

LPTK:
I noticed that your latest commits *removed* some test cases, for example https://github.com/hkust-taco/mlscript/pull/174/commits/84995d033b9a52c196e238b54c0e82697d13c691. There are already very few tests, so please do not make it worse by reducing the coverage further! ;^)

LPTK (approved):
Ok, I think we can merge now. Please address the remaining comments and I'll make the squash-merge later.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1041,26 +1041,20 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
-              }
+              val argsList = fields.map(x => x._1 match {
+                case Some(arg) => arg
+                case N => die // cannot happen, because already checked with the hasUntypedArg
+              })
               desugarNamedArgs(term, f, a, argsList, f_ty)
             }
-          case x :: y :: _ => 
+          case _ :: Nil =>
+            err("Function type is not handled", f.toLoc)
```
LPTK: Come on. You know this error message is not good. Let's just remove this case and make the last case match `Nil | _ :: Nil` instead of just `Nil`.

LPTK (approved):

Issue #173, opened 2023-07-13
Implement virtuality and constructor restrictions


LPTK:
- [x] allow `fun` and `val` members to be declared `virtual`; a signature without an implementation is considered virtual automatically
 - [x] check that overridden `fun`s and `val`s are virtual
 - [x] forbid uses of `this` in constructors, including a class' initialization statements in the body and `val` RHSs
 - [x] make an exception to the above for each _final_ (ie non-virtual) member which either is a `val` _or_ is a `lazy val` or a `fun` that doesn't itself transitively use this

NeilKleistGao:
I think all tasks are finished.

Issue #166, opened 2023-06-01
Iron out semantics of new object definition parameters and fields


LPTK:
* [x] Do not generate accessors for `let` bindings (unlike `val` fields)
 * [x] Only generate accessors for parameters marked `val`
 * [x] Figure out whether we really should insert `this` implicitly when accessing members defined in the same scope. Currently we do it for members but not for parameters. We need something consistent and not too surprising.

Definitions:

**Implicit `this` insertion**: Accesses to members in scope are implicitly prefixed with `this.` in JS
**Direct member access**: Accesses to members in scope refer to the specific member definition does _not_ virtually dispatch to possibly overridden ones. This might be a bit annoying to implement for methods.

Currently, we use direct access for parameters, which is (currently) not very consistent but has advantages:
 * It allows referring to these in the class ctor (where access to `this` should be forbidden for safety). Note that the ctor includes the initializers of all class fields. Making ctor semantics different from method body semantics is possible but would be surprising.
 * Consistent with parameters that do _not_ generate an accessor. And it would be strange if adding `val` to make the parameter generate an accessor would change the semantics of references to this parameter.

Pros of direct accesses for methods and fields:
 * Allows using them in the ctor – currently one would need to use a let binding instead.
 * Allows naturally defining efficient recursive functions, and converting tail-recursive ones to loops (`this` insertion would make these calls virtual)

Cons of direct accesses for methods:
 * It doesn't match what happens in most Java-style language, which, unlike JS, allow omitting `this`.

LPTK:
- [x] Note: the current semantics of `val`/`fun` implicit `this` insertion is inconsistent. It is performed for `abstract class C { val x = 1; fun f = x }` whereas `abstract class C { val x : int; fun f = x }` fails with "`Code generation encountered an error: unresolved symbol x`".
  The second one should work as well.

LPTK:
After some consideration, I figured the best approach may be to:
 * require the use of a `virtual` keyword for members that may be overridden later; and
 * reject unqualified (i.e., `this`-less) accesses to virtual members.

This has lots of benefits:
 * All accesses to non-virtual members in scope can be unqualified, and treated as though they were just variables in scope, which is morally fine because their targets won't unexpectedly be rewired by derived classes, helping to reason about semantics.
[...]

PR #175, opened 2023-08-22
Implement New object definition

NeilKleistGao:
Related Issues:
- #166 
- #173 

In this PR:
- Only generate fields for `val`s rather than `let`s
  - If a let binding is used in a method, we still generate a private field
- Only generate public getters for constructor parameters with `val`s
  - If it is in `constructor(...)` and has no `val`, the parameter is available only in the constructor
  - Add static function `$unapply` to non-plain classes(i.e., `class ClassName(...)`)
  - Unfold let-bindings generated by UCS for `$unapply` method invocations
- Fix `Code generation encountered an error: unresolved symbol` if members only have type signatures
- Add `virtual` keyword for `fun`s and `val`s.
  - If a member only has the type signature, it is virtual automatically
  - Require `this` when accessing the virtual members
  - Overriding non-virtual members is not allowed
- Add restrictions on constructors
  - Using `this` is forbidden in the constructor statements and RHS of `val`s, including directly and indirectly
  - Accessing non-virtual members via `this` is still allowed if the members do not refer to `this`


shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -195,6 +195,18 @@ class JSBackend(allowUnresolvedSymbols: Boolean) {
       JSRecord(fields map { case (key, Fld(_, _, _, value)) =>
         key.name -> translateTerm(value)
       })
+    case Sel(Var(cls), Var(u: "unapply")) =>
```
LPTK: Why? Why wouldn't the access to `unapply` be uniform across different kinds of classes? Users may want to call that method explicitly. It should always just be `ClassName.unapply`.

LPTK: By the way this code is crazy. You're messing up with any static class field incidentally named `unapply` here.

LPTK: To elaborate on my suggestion: just use a field in the `ClassName` value, be it a class or a function.

Is there any advantage to using an actual static method here?

NeilKleistGao: If we treat `unapply` as a non-static method, we need to consider how to handle the overriding of it.

LPTK: I think you misunderstand my suggestion. How would you override a method that's called as MyClass.foo

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -195,17 +178,11 @@ class JSBackend(allowUnresolvedSymbols: Boolean) {
...
-        case S(sym: NuTypeSymbol with RuntimeSymbol) =>
-          if (sym.isPlainJSClass) translateNuTypeSymbol(sym).member(u)
-          else translateNuTypeSymbol(sym).member("class").member(u)
-        case S(sym @ CapturedSymbol(out, cls: NewClassSymbol)) =>
-          if (cls.isPlainJSClass)
-            translateCapture(sym).member(u)
-          else
-            translateCapture(sym).member("class").member(u)
-        case _ => JSField(translateTerm(Var(cls)), u)
+        case S(sym: NewClassSymbol) if !sym.isPlainJSClass =>
```
LPTK: Why is this case needed? I don't know why you make a special case here and in `translateVar`:

```scala
      case S(sym: NuTypeSymbol with RuntimeSymbol) =>
        if (sym.isPlainJSClass || !isCallee) translateNuTypeSymbol(sym)
        else translateNuTypeSymbol(sym).member("class")
```

What's the justification? Please provide an example where the behavior we want is not just to generate a plain reference to the symbol.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -517,13 +484,12 @@ class JSBackend(allowUnresolvedSymbols: Boolean) {
     JSClassDecl(classSymbol.runtimeName, fields, base, members, traits)
   }
 
-  protected def translateSelfDeclaration(self: ValueSymbol): Ls[JSStmt] =
-    if (visitedSymbols(self)) {
-      visitedSymbols -= self
-      JSConstDecl(self.runtimeName, JSIdent("this")) :: Nil
+  protected def translateQualifierDeclaration(qualifier: ValueSymbol): Ls[JSStmt] =
+    if (visitedSymbols(qualifier)) {
+      visitedSymbols -= qualifier
```
LPTK: What's going on here? Please justify this line with a comment or use a cleaner approach. Think about what would happen if the symbol was truly visited first, and then later removed by this line.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -734,24 +683,45 @@ class JSBackend(allowUnresolvedSymbols: Boolean) {
   )(implicit scope: Scope): JSClassNewDecl = {
     val nuTypeScope = scope.derive(sym.toString)
     val constructorScope = nuTypeScope.derive(s"${sym.lexicalName} constructor")
-    val selfSymbol = constructorScope.declareThisAlias()
 
     val memberList = ListBuffer[RuntimeSymbol]() // pass to the getter of nested types
     val typeList = ListBuffer[Str]()
 
+    final case class Pack(memberScope: Scope, qualifier: Str)
```
LPTK: Use a more descriptive name?

shared/src/main/scala/mlscript/JSBackend.scala
NeilKleistGao: The unnecessary case has been removed.
When we generate code for `a instanceof  ...` or `new ...`, we should use the actual class (i.e., `ClassName.class`) rather than the constructor function. I indicated all the cases where we need `.class` by commenting.

shared/src/main/scala/mlscript/JSBackend.scala
NeilKleistGao: For each scope, we create a qualifier symbol (i.e., an alias for `this`). We generate the actual `const qualifier = this;` only if the symbol is used in the body. This method would be called at the end of the member generation, so removing or not does not effect the result. Since this method can also be called somewhere else and it may lead to unexpected results, this line has been removed.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -776,22 +765,43 @@ class JSBackend(allowUnresolvedSymbols: Boolean) {
...
+          )
+        }
+        else {
+          val sym = nuTypeScope.resolveValue(nme) match {
+            case Some(sym: NewClassMemberSymbol) => sym
+            case _ => throw new AssertionError(s"error when handling $nme")
+          }
+          if (visitedSymbols.contains(sym) || ctorParams.contains(nme)) { // This field is used in other methods, or it overrides the ctor parameter
+            privateMems += nme
+            visitedSymbols -= sym
```
LPTK: Same remark as the other occurrence of this pattern earlier.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1068,15 +1068,29 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, var ne
                 rcdSel(realPrefix, fieldName)
               }
           }
+        def fallback =
```
LPTK: Document that this is a temporary solution to the lack of proper static `unapply` method desugaring from class definitions.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -424,6 +424,19 @@ trait NuDeclImpl extends Located { self: NuDecl =>
         sps.getOrElse(Tup(Nil))})${sig.fold("")(": " + _.showDbg2)}${
           if (parents.isEmpty) "" else if (k === Als) " = " else ": "}${parents.mkString(", ")}"
   }
+  def genUnapply: Opt[NuFunDef] = this match {
+    case NuTypeDef(Cls, _, _, S(Tup(fields)), _, _, _, _, _, _) =>
```
LPTK: This pattern is an anti-pattern. Just use `td: NuTypeDef  if td.kind is Cls`

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -424,6 +424,19 @@ trait NuDeclImpl extends Located { self: NuDecl =>
         sps.getOrElse(Tup(Nil))})${sig.fold("")(": " + _.showDbg2)}${
           if (parents.isEmpty) "" else if (k === Als) " = " else ": "}${parents.mkString(", ")}"
   }
+  def genUnapply: Opt[NuFunDef] = this match {
```
LPTK: Make it a lazy val.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -785,9 +786,39 @@ class Desugarer extends TypeDefs { self: Typer =>
       xs match {
         case MutCase.Constructor(className -> fields, cases) :: next =>
           printlnUCS(s"• Constructor pattern: $className(${fields.iterator.map(x => s"${x._1} -> ${x._2}").mkString(", ")})")
-          // TODO: expand bindings here
           val consequent = rec(cases)(defs ++ fields.iterator.map(_._2))
-          Case(className, mkLetFromFields(scrutinee, fields.toList, consequent), rec2(next))
+          val unapplyMtd = ctx.get(className.name) match {
+            case S(CompletedTypeInfo(cls: TypedNuCls)) => cls.td.genUnapply
+            case S(ti: DelayedTypeInfo) => ti.decl.genUnapply
+            case _ => N
```
LPTK: Enumerate what this is matching.

shared/src/main/scala/mlscript/JSBackend.scala
NeilKleistGao: Removed the unnecessary branch for `unapply`

shared/src/main/scala/mlscript/Typer.scala
NeilKleistGao: Added comments for the temporary solution.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
NeilKleistGao: Added comments for matching branches.

shared/src/main/scala/mlscript/JSBackend.scala
LPTK: In fact, why is `visitedSymbols` a global variable? It has no business being reused all over the place and storing every symbol... that's just smelly design. At this point, why not just store a Boolean flag in the symbol itself? It's equally smelly, but at least it will be much more efficient.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
LPTK: You didn't enumerate anything... I've already told you in situations like these to avoid `case _ =>` and prefer `case _: OtherA | _: OtherB | ... =>`.

shared/src/main/scala/mlscript/NuTypeDefs.scala
```diff
@@ -460,6 +470,71 @@ class NuTypeDefs extends ConstraintSolver { self: Typer =>
...
+      case TyApp(lhs, targs) => visit(lhs)
+      case Var(name) =>
+        if (name === "this") true
+        else {
+          refs.add((name, false))
+          false
+        }
+      case Where(body, where) => where.foldLeft(visit(body))((r, s) => r || visit(s))
+      case With(trm, fields) => visit(trm) || visit(fields)
+      case _ => false
```
LPTK: Why false?! Which cases are these?

shared/src/main/scala/mlscript/NuTypeDefs.scala
```diff
@@ -460,6 +470,71 @@ class NuTypeDefs extends ConstraintSolver { self: Typer =>
...
+      case Tup(fields) => fields.foldLeft(false)((r, s) => r || visit(s._2.value))
+      case TyApp(lhs, targs) => visit(lhs)
+      case Var(name) =>
+        if (name === "this") true
+        else {
+          refs.add((name, false))
+          false
+        }
+      case Where(body, where) => where.foldLeft(visit(body))((r, s) => r || visit(s))
+      case With(trm, fields) => visit(trm) || visit(fields)
```
LPTK: It would be better to add a `childrenTerms` to `TermImpl`, which could be used here and possibly reused elsewhere, and would avoid this huge pattern matching.

shared/src/main/scala/mlscript/syntax.scala
```diff
@@ -211,10 +211,19 @@ final case class NuFunDef(
   nme: Var,
   tparams: Ls[TypeName],
   rhs: Term \/ Type,
-)(val declareLoc: Opt[Loc], val signature: Opt[NuFunDef], val outer: Opt[Outer]) extends NuDecl with DesugaredStatement {
+)(
+  val declareLoc: Opt[Loc],
+  val virtualLoc: Opt[Loc],
```
LPTK: Plz document what `virtualLoc` means.

shared/src/main/scala/mlscript/NuTypeDefs.scala
```diff
@@ -389,6 +389,11 @@ class NuTypeDefs extends ConstraintSolver { self: Typer =>
       this
   }
   
+  // Store referred members. indirect ref: ref by using `this`
```
LPTK: Comment is impossible to understand. Please clearly state what this stores and the meaning of the two `Bool` uses.

shared/src/main/scala/mlscript/NuTypeDefs.scala
```diff
@@ -389,6 +389,11 @@ class NuTypeDefs extends ConstraintSolver { self: Typer =>
...
+  case class RefMap(useThis: Bool, refs: Set[(Str, Bool)])
```
LPTK: Isn't `useThis` supposed to be `usesThis`?...

shared/src/main/scala/mlscript/NuTypeDefs.scala
```diff
@@ -985,6 +1060,54 @@ class NuTypeDefs extends ConstraintSolver { self: Typer =>
...
+                        msg"Declared here:" -> nf.fd.toLoc
+                      :: Nil)
+                    case _ => ()
+                  })
+
+                members.foreach {
+                  case tf @ TypedNuFun(_, fd, _) =>
+                    val refs = tf.getFunRefs
+                    if (fd.isLetRec.isDefined && checkThisInCtor(refs, S(tf.name), tf.name :: Nil)(true)) // not a function && access `this` in the ctor
+                      err(msg"Can not access `this` when initializing field ${tf.name}", fd.toLoc)
```
LPTK: The location here is wrong. Instead of returning a Boolean, `checkThisInCtor` should return the `Var` of the offending reference, so we can use its location.

shared/src/main/scala/mlscript/NuTypeDefs.scala
```diff
@@ -985,6 +1060,54 @@ class NuTypeDefs extends ConstraintSolver { self: Typer =>
...
+
+                def getMember(name: Str) =
+                  members.find(_.name === name).fold(baseAndSigs.find(_.name === name))(m => S(m))
+
+                // Return true if it is invalid
+                def checkThisInCtor(refs: RefMap, name: Opt[Str], stack: Ls[Str])(expection: Bool): Bool = {
+                  def run: Bool = {
+                    refs.useThis || (
+                      refs.refs.foldLeft(false)((res, p) => res || (getMember(p._1) match {
+                        case S(nf: TypedNuFun) if p._1 =/= name.getOrElse("") && !stack.contains(p._1) =>
```
LPTK: Use of `""` here is a code smell. Document or clean up logic.

shared/src/main/scala/mlscript/NuTypeDefs.scala
```diff
@@ -985,6 +1060,54 @@ class NuTypeDefs extends ConstraintSolver { self: Typer =>
               
             case td: NuTypeDef =>
               
+              /** Check no `this` access in ctor statements or val rhs and reject unqualified accesses to virtual members.. */
+              def qualificationCheck(members: Ls[NuMember], stmts: Ls[Statement], baseAndSigs: Ls[NuMember]): Unit = {
+                val cache = mutable.HashMap[Str, Bool]()
+
+                def getMember(name: Str) =
+                  members.find(_.name === name).fold(baseAndSigs.find(_.name === name))(m => S(m))
```
LPTK: This is extremely inefficient. It seems you really like linear traversals! Why not create a map for fast retrieval?

LPTK (changes requested):
Looks pretty solid! But I think there's a bug (added to `Virtual.mls`):

```fs
// FIXME `x` should be considered virtual here
class C() {
  virtual fun x : Int
  fun x = 1
  fun foo0 = x // should error
}
C().x
//│ class C() {
//│   fun foo0: 1
//│   fun x: Int
//│ }
//│ Int
//│ res
//│     = 1

module D extends C {
  fun x = 2
}
D.x
//│ module D extends C {
//│   fun foo0: 1
//│   fun x: 2
//│ }
//│ 2
//│ res
//│     = 2
```

LPTK:
BTW in case you don't know @NeilKleistGao you can automatically fix diff-test conflicts by just re-running the tests.

shared/src/main/scala/mlscript/NuTypeDefs.scala
NeilKleistGao: Updated the comment here.

shared/src/main/scala/mlscript/syntax.scala
NeilKleistGao: Added comments for `virtualLoc` here.

shared/src/test/diff/nu/NoThisCtor.mls
```diff
@@ -64,8 +64,8 @@ class Foo4() extends Foo() {
   let bb = bar(0) // call `this` indirectly
 }
 //│ ╔══[ERROR] Cannot access `this` while initializing field bb
-//│ ║  l.64: 	  let bb = bar(0) // call `this` indirectly
-//│ ╙──      	      ^^^^^^^^^^^
+//│ ║  l.63: 	  fun bar(x) = this.foo + x // ok
```
LPTK, line 70: Hmm this case shows that we should still sometimes show the place where the problem arises (here the let binding location that was previously shown) in addition to the place where `this` occurs. We should omit the former when the latter is included in it.

shared/src/main/scala/mlscript/NuTypeDefs.scala
NeilKleistGao: Updated `checkThisInCtor`. Now it returns `Option[Var]`: `Some(v: Var)` means there is an invalid access where the `v` indicates the corresponding reference, and `None` means the access is safe.

shared/src/main/scala/mlscript/NuTypeDefs.scala
NeilKleistGao: Actually, we have a similar code for this:
```scala
trait Located {
  def children: List[Located]
  // ...
}
```

So far I replaced the huge pattern matching with a `foldLeft` on the `children`.

shared/src/main/scala/mlscript/ucs/Desugarer.scala
NeilKleistGao: Oh, I misunderstood that... I replaced the `case _ =>` branch.

shared/src/main/scala/mlscript/NuTypeDefs.scala
NeilKleistGao: Updated. Similar to the above conversation.

shared/src/main/scala/mlscript/JSBackend.scala
NeilKleistGao: Actually, here I just reused the existing code: [shared/src/main/scala/mlscript/JSBackend.scala](https://github.com/hkust-taco/mlscript/blob/mlscript/shared/src/main/scala/mlscript/JSBackend.scala#L22).

shared/src/test/diff/nu/NoThisCtor.mls
NeilKleistGao: Is this one better:
```ts
:e
class Foo4() extends Foo() {
  val foo: Int
  virtual val foo = 4
  fun bar(x) = this.foo + x // ok
  let bb = bar(0) // call `this` indirectly
}
//│ ╔══[ERROR] Cannot access `this` while initializing field bb
//│ ║  l.67: 	  let bb = bar(0) // call `this` indirectly
//│ ║        	      ^^^^^^^^^^^
//│ ╟── The access to `this` is here
//│ ║  l.66: 	  fun bar(x) = this.foo + x // ok
//│ ╙──      	                   ^^^^
//│ class Foo4() extends Foo {
//│   fun bar: Int -> Int
//│   let bb: Int
//│   let foo: Int
//│ }
```

shared/src/test/diff/gadt/ThisMatching.mls
```diff
@@ -8,6 +8,9 @@ module Dummy {
 }
 //│ ╔══[ERROR] Cannot access `this` during object initialization
 //│ ║  l.7: 	  log(if this is Dummy then "duh!" else "huh?")
+//│ ║       	  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+//│ ╟── The access to `this` is here
+//│ ║  l.7: 	  log(if this is Dummy then "duh!" else "huh?")
```
LPTK: You can check whether the `this` location is included in the first location (there are helper methods for that), in which case you should just output "Cannot access `this` during object initialization" with just the precise location of `this`.

shared/src/test/diff/nu/NoThisCtor.mls
LPTK: Much better! But why is `.foo` underlined instead of `this`? 🤔

NeilKleistGao: I replaced the boolean flag with the `Opt[Var]` and now we can store the location of `this`.

shared/src/main/scala/mlscript/JSBackend.scala
LPTK: Sure. Still, while we're at it, could you plz make this a mutable flag in the symbol instead?

LPTK: In general the principle is that it's better to leave the code in a better state than we found it in ;)

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1080,6 +1080,19 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, var ne
...
+          // * The code below is only a temporary solution to type `ClassName.unapply`.
+          // * It removed when static methods can be typed properly.
+          case Var(nuCls) =>
+            if (fieldName.name === "unapply") (ctx.get(nuCls) match {
+                case S(CompletedTypeInfo(cls: TypedNuCls)) => cls.td.genUnapply
+                case S(ti: DelayedTypeInfo) => ti.decl.genUnapply
+                case _ => N
+              }) match {
+                case S(NuFunDef(_, _, _, L(unapplyMtd))) =>
+                  typePolymorphicTerm(unapplyMtd)
```
LPTK: This would better be
```suggestion
                  typeMonomorphicTerm(unapplyMtd)
```
because the result is applied immediately (no need to have it be generalized/polymorphic).

shared/src/main/scala/mlscript/ucs/Desugarer.scala
```diff
@@ -785,9 +786,39 @@ class Desugarer extends TypeDefs { self: Typer =>
       xs match {
         case MutCase.Constructor(className -> fields, cases) :: next =>
           printlnUCS(s"• Constructor pattern: $className(${fields.iterator.map(x => s"${x._1} -> ${x._2}").mkString(", ")})")
-          // TODO: expand bindings here
           val consequent = rec(cases)(defs ++ fields.iterator.map(_._2))
-          Case(className, mkLetFromFields(scrutinee, fields.toList, consequent), rec2(next))
+          val unapplyMtd = ctx.get(className.name) match {
+            case S(CompletedTypeInfo(cls: TypedNuCls)) => cls.td.genUnapply // Classes from other typing units
+            case S(ti: DelayedTypeInfo) => ti.decl.genUnapply // Classes in the same typing units
```
LPTK: I wonder why you check that the thing is a `Cls` in the line above but not in this one. Better be consistent!

shared/src/test/diff/ecoop23/PolymorphicVariants.mls
```diff
@@ -198,16 +198,16 @@ Test2.eval(Cons(("a", Abs("d", Var("d"))), Nil), Add(Numb(1), Var("a")))
 
 module Test3 extends EvalVar, EvalExpr, EvalLambda
 //│ module Test3 {
-//│   fun eval: forall 'a 'A. (Cons[{_1: anything, _2: 'result}] | Nil, Abs['b] | App['A] | Object & 'c & ~#Abs & ~#App) -> ('A0 | 'a)
+//│   fun eval: forall 'a. (Cons[{_1: anything, _2: 'b}] | Nil, Abs['c] | App['d] | Object & 'e & ~#Abs & ~#App) -> ('A | 'a)
```
LPTK, line 201: I need to investigate this change. Why would your modifications make one less variable be generalized in this signature? (It means it'll be generalized at the module level instead!)

shared/src/test/diff/nu/BadClassInherit.mls
```diff
@@ -25,15 +25,13 @@ class C2 extends C1(y) {
 //│ class C2 extends C1 {
 //│   let y: Int
 //│ }
-//│ Code generation encountered an error:
-//│   unresolved symbol y
```
LPTK: Do you know why this is no longer a code-gen error? What does it generate?

shared/src/test/diff/nu/ImplicitMethodPolym.mls
```diff
@@ -102,14 +102,20 @@ module M extends Mx {
 //│ }
 
 
-
+:e
```
LPTK: Let's add a FIXME here. (And put back the empty line you deleted!)
```suggestion

:e // FIXME
```

shared/src/main/scala/mlscript/NuTypeDefs.scala
LPTK: Yeah I know, but this traverses every single located thing as opposed to every term. But it's okay to use this for now. (You can close this.)

shared/src/main/scala/mlscript/ucs/Desugarer.scala
NeilKleistGao: Removed class check in the first branch. It would be checked inside the `genUnapply`.

shared/src/test/diff/nu/BadClassInherit.mls
NeilKleistGao: It accesses the signature unexpectedly. Now it is fixed.

LPTK: I don't understand your fix. It doesn't cover other similar situations, such as:
```scala
class C2 extends C1(y) {
  val y: Int = 1
}
```

BTW why are you not introducing a new scope for the part of the class between the `{...}`?

NeilKleistGao: You are right! I adjusted the scopes in `translateNewTypeDefinition`.

shared/src/main/scala/mlscript/JSBackend.scala
LPTK: @NeilKleistGao Are we going to remove this `visitedSymbols`? It would make sense, since you removed the other one.

NeilKleistGao: I think this `visitedSymbols` has been removed 🤔

shared/src/main/scala/mlscript/NuTypeDefs.scala
```diff
@@ -389,6 +389,13 @@ class NuTypeDefs extends ConstraintSolver { self: Typer =>
       this
   }
   
+  // The field `thisRef` is defined when the member accesses to `this` object without field selection
+  // e.g., val x = this
+  // The field `refs` contains all `Var`s accessed by the member with their qualifiers (None if it is an unqualifier access)
```
```suggestion
  // Field `thisRef` is defined when the member refers to `this` selecting a field on it
  // e.g., val x = this
  // Field `refs` contains all `Var`s accessed by the member with their possible `this` qualifiers (`None` if it is an unqualified access)
```

LPTK:
I think a good test to add would be this:
```scala
abstract class Foo: Int -> Int {
  val x = f
  fun f = this(0)
}
//│ ╔══[ERROR] Cannot access `this` while initializing field x
//│ ║  l.6: 	  val x = f
//│ ║       	      ^^^^^
//│ ╟── The access to `this` is here
//│ ║  l.7: 	  fun f = this(0)
//│ ╙──     	          ^^^^
//│ abstract class Foo: Int -> Int {
//│   fun f: nothing
//│   let x: nothing
//│ }
```
Since it accesses `this` in an unusual way!

shared/src/main/scala/mlscript/JSBackend.scala
LPTK: Oh yeah sorry I looked at the wrong diff!

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -437,6 +437,20 @@ trait NuDeclImpl extends Located { self: NuDecl =>
         sps.getOrElse(Tup(Nil))})${sig.fold("")(": " + _.showDbg2)}${
           if (parents.isEmpty) "" else if (k === Als) " = " else ": "}${parents.mkString(", ")}"
   }
+  lazy val genUnapply: Opt[NuFunDef] = this match {
+    case td: NuTypeDef if td.kind is Cls => td.params.fold[Opt[NuFunDef]](N)(tup => {
```
LPTK: Note: `.fold(N)(x => S(...))` is the same as `.map(x => ...)`!

shared/src/main/scala/mlscript/Typer.scala
LPTK: Never mind, I realized that `genUnapply` actually returns a lambda with an unbound type variable occurrence, and the latter must be generalized before the result can be used.

shared/src/test/diff/ecoop23/PolymorphicVariants.mls
LPTK: Ok, I guess what happens here is some obscure interaction due to the fact that the new typing of patterns now involves intermediate type variables. I think it's likely a minor thing and this can be closed.

LPTK (approved):
Ok, I think I'm happy enough with the changes now!

LPTK:
@NeilKleistGao please fix the conflicts and close the remaining conversations, then please merge (with a merge commit).

PR #177, opened 2023-09-12
Implement user-defined symbolic operators & uncurry all operator applications

LPTK:
Examples:
```scala
fun (>>) compose(f, g) = x => g(f(x))
//│ fun (>>) compose: forall 'a 'b 'c. ('a -> 'b, 'b -> 'c) -> 'a -> 'c

:js 
let r = succ >> succ
//│ let r: Int -> Int
//│ // Prelude
//│ function succ(x) {
//│   return x + 1;
//│ }
//│ class TypingUnit1 {}
//│ const typing_unit1 = new TypingUnit1;
//│ // Query 1
//│ globalThis.r = compose(succ, succ);
//│ // End of generated code
//│ r
//│   = [Function (anonymous)]

r(3)
//│ Int
//│ res
//│     = 5
```
At the type level:
```scala
type *[A, B] = [A, B]
//│ type *[A, B] = [A, B]


fun x : Int * Int
fun x = [0, 1]
//│ fun x: [0, 1]
//│ fun x: *[Int, Int]
```


NeilKleistGao:
It seems that there are some parsing problems:
```ts
fun (<<|+_+|>>) robot(a, b) = a + b
//│ fun (<<|+) robot: (Int, Int) -> Int
// should be `<<|+_+|>>`, but we got `<<|+`

fun (:-D) dd(a, b) = a + b
//│ fun (:-) dd: (Int, Int) -> Int
// should be `:-D`, but we got `:-`
```

The parsed operators are not complete. If there is any unsupported symbol in the defined operators, maybe we can reject it in the parser.

LPTK:
@NeilKleistGao Good catch! I've fixed the parsing bug (it was swallowing some erroneous tokens) and improved the error messages.

chengluyu:
When will we support emoji? 😁😄😃 Can't wait to use real magic wind 🪄 instead of `-*`.

LPTK:
@chengluyu Good suggestion! But the lexer currently says "unexpected character" :^(

LPTK:
@NeilKleistGao @chengluyu Some parts of `JSBackend.scala` are really hairy, with strange/fishy logic. For example see all the `// TODO Improve` comments I added here: https://github.com/hkust-taco/mlscript/blob/157c3e7117d8e34d8cb671618f3007736c5c821a/shared/src/main/scala/mlscript/JSBackend.scala#L1364-L1403

Because of this I couldn't easily tell why this one does not currently work:

```scala
val (-->) f(x, y) = [x, y]
//│ let (-->) f: forall 'a 'b. ('a, 'b) -> ['a, 'b]
//│ f
//│   = [Function: f]

:ge // FIXME
12 --> 34
//│ [12, 34]
//│   unresolved symbol -->
```

NeilKleistGao (approved):
LGTM.

PR #179, opened 2023-09-19
Fix type checking of classes, constructors, and fields

LPTK:
Supersedes https://github.com/hkust-taco/mlscript/pull/178.


shared/src/main/scala/mlscript/NuTypeDefs.scala
```diff
@@ -1138,24 +1225,32 @@ class NuTypeDefs extends ConstraintSolver { self: Typer =>
...
+                            }` is not virtual and cannot be overridden" -> m.toLoc ::
+                          msg"Originally declared here:" -> fun.toLoc ::
+                          Nil)
+                      case p: NuParam if (!p.isVirtual && !clsSigns.contains(p)) =>
+                        err(msg"Inherited parameter named `${m.name
+                            }` is not virtual and cannot be overridden" -> m.toLoc ::
+                          msg"Originally declared here:" -> fun.toLoc ::
                           Nil)
                       case _ =>
+                        // println(s"?! $fun")
```
NeilKleistGao: Is this a commented debug print? 🤔

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1563,19 +1599,30 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
+            mkTypingUnit(thisTy, members)
+          )(td.declareLoc, td.abstractLoc)
         }
-      case TypedNuCls(level, td, tparams, params, members, thisTy, sign, ihtags, ptps) =>
+      case TypedNuCls(level, td, tparams, params, acParams, members, thisTy, sign, ihtags, ptps) =>
         ectx(tparams) |> { implicit ectx =>
           NuTypeDef(td.kind, td.nme, td.tparams,
-            Opt.when(td.params.isDefined)(Tup(params.map(p => N -> Fld(FldFlags(false, false, false), Asc(p._1, go(p._2.ub)))))),
+            // Opt.when(td.params.isDefined)(Tup(params.map(p => N -> Fld(FldFlags(false, false, false), Asc(p._1, go(p._2.ub)))))),
+            params.map(ps => Tup(ps.map(p => N -> Fld(FldFlags(false, false, false), Asc(p._1, go(p._2.ub)))))),
```
NeilKleistGao: Maybe using `FldFlags.empty` is better because we will have another flag pretty soon.
```suggestion
            params.map(ps => Tup(ps.map(p => N -> Fld(FldFlags.empty, Asc(p._1, go(p._2.ub)))))),
```

LPTK: Yeah in fact I replaced all occurrences of `FldFlags(false, false, false)` by `FldFlags.empty` except for the defining one. There were a lot!

shared/src/main/scala/mlscript/NuTypeDefs.scala
LPTK: Yes, good catch! Fixed.

NeilKleistGao (approved):
LGTM.

PR #181, opened 2023-09-23
Fix that execution errors of a term are swallowed when there are functions before the term

chengluyu:


shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -258,7 +258,7 @@ class DiffTests
           case "dv" => mode.copy(debugVariance = true)
           case "ge" => mode.copy(expectCodeGenErrors = true)
           case "re" => mode.copy(expectRuntimeErrors = true)
-          case "r" | "showRepl" => mode.copy(showRepl = true)
+          case "ShowRepl" => mode.copy(showRepl = true)
```
LPTK: Why would you do that? There's a reason I changed it to lower case... Only commands that modify the global file state are upper case!

chengluyu: This is not a deliberate change. It was accidentally caused when I switched branches.

LPTK:
Thanks! But it looks like there are uncommited diff-test changes. And you left debugging comments (the "what" things).

chengluyu:
It's removed!

LPTK (changes requested):
Thanks but please read https://github.com/hkust-taco/mlscript/blob/mlscript/CONTRIBUTING.md again and try to avoid wasting my time too much 😓

LPTK:
@chengluyu Let me just be clear here: what upsets me is when someone asks for a review on their PR and it turns out they haven't even reviewed their own diff before doing so. **Please review your own diff.**

shared/src/test/diff/codegen/Mixin.mls
```diff
@@ -256,7 +256,7 @@ TestLang.eval(mk(0))
 //│ res = TestLang.eval(mk(0));
 //│ // End of generated code
 //│ res
-//│     = [Function: mk]
```
chengluyu: I guess the `:js` flag for this block is to see why `res` is `mk`. Right? If so, I will remove `:js` flag. P.S. Expand to see the `:js` flag above.

chengluyu (commented):
I have read all changes. I left one comment. Please have a look.

shared/src/test/diff/codegen/NuFuns.mls
```diff
@@ -26,7 +26,7 @@ foo
 //│ fun foo: Int
 //│ Int
 //│ res
-//│     = [Function: foo1]
+//│     = 111
 
 
 // FIXME: returns a thunk
```
LPTK:
```suggestion
```

shared/src/test/diff/codegen/Mixin.mls
LPTK: Yeah probably. Let's remove it!

LPTK (approved):
Thanks. When the comments are addressed, feel free to go ahead and squash-merge this (with a clean commit message).

PR #183, opened 2023-09-28
Optional args

mbroughani81:


shared/src/main/scala/mlscript/TyperDatatypes.scala
```diff
@@ -274,6 +274,7 @@ abstract class TyperDatatypes extends TyperHelpers { Typer: Typer =>
       fields.map(f => s"${f._1.fold("")(_.name+": ")}${f._2},").mkString(" ")
     override def toString = s"($showInner)"
     // override def toString = s"(${fields.map(f => s"${f._1.fold("")(_+": ")}${f._2},").mkString(" ")})"
+    def isSubtype(that: TupleType): Boolean = fields.sizeCompare(that.fields) <= 0 && fields.sizeCompare(that.fields.filter(x => !x._2.opt)) >= 0
```
LPTK: This method is mis-named. It doesn't actually perform a subtype comparison!

LPTK: A better name could be `isLengthCompatibleWith`.

LPTK:
It seems that the branch you merged earlier was still a wildly outdated version. That's why it would help you to use a git GUI, which would allow you to better visualize what you're doing.

mbroughani81:
> It seems that the branch you merged earlier was still a wildly outdated version. That's why it would help you to use a git GUI, which would allow you to better visualize what you're doing.

I will give it a try!
BTW, I resolved the conflict. Almost all was because of the FldFlags new fields.

LPTK:
You messed up the merge a bit. File [QML_exist_Classes_CT.mls](https://github.com/hkust-taco/mlscript/pull/183/files#diff-947043354654cfa2c4a854d5ce44fda948d2daab7102bdfe1006d332157bde60) was deleted from the main branch and you mistakenly reinstated it.

LPTK:
## Test cases to add

As discussed on Discord, here are expected usages of optional parameters:

```fs
fun foo0(x: Int?) = if x is undefined then 0 else x + 1

fun foo1(x?) = if x is undefined then 0 else x + 1
```
The following should not type check
```fs
:e
fun foo0(x: Int?) = x + 1

:e
fun foo1(x?) = x + 1
```

To do this, you have to adapt the typing of tuples when in pattern positions.

And some interesting cases to add, which contain signature checks:

```fs
fun foo: (x: Int?) -> Int
fun foo(x?) = if x is undefined then 0 else x + 1

// Ok: function with optional parameter can be widened to function with mandatory parameter
fun foo: (x: Int) -> Int
fun foo(x?) = if x is undefined then 0 else x + 1

:e // The definition's parameter spec is too strong
fun foo: (x: Int?) -> Int
fun foo(x) = x + 1
```

And some subtype tests:

```
fun f(x) = (x : [Int, Int?]) : ['a, 'b]
```
I expect we could accept the case above by constraining `Int | undefined <: 'b`

LPTK:
Thanks for looking into this at the time @mbroughani81. We now have a new compiler implementation (in branch `hkmc2`) where all the current work is going. Optional arguments will be implemented there directly.

PR #184, opened 2023-09-30
Annotate RegionDSL to simplify types

andongfan:


LPTK (changes requested):
Wow the file reads much nicer now :^)

But let's keep the original unannotated tests in a separate file – call the original `SimpleRegionDSL_raw.mls` and the new one `SimpleRegionDSL_annot.mls`, for example. Letting the inferred types through can be useful to spot changes and regression, and also compare the performance.

andongfan:
Done! Interestingly I found that on my machine the annotated version run slightly faster than the raw version by about 30ms.

LPTK (approved):
Thanks!

> I found that on my machine the annotated version run slightly faster than the raw version by about 30ms

TBH that's likely just noise from the test setup, although the effect might be real but small.

PR #186, opened 2023-10-01
Fix codegen error for global functions

NeilKleistGao:
Related Issue: https://github.com/hkust-taco/mlscript/issues/65#issuecomment-1741167915


shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -1381,6 +1381,16 @@ class JSTestBackend extends JSBackend(allowUnresolvedSymbols = false) {
       case _ => die
     }
 
+    otherStmts.foreach {
+      case fd @ NuFunDef(isLetRec, Var(nme), symNme, _, L(body)) if isLetRec.getOrElse(true) =>
```
LPTK: Why are we using this super weird condition `if isLetRec.getOrElse(true)`? It would deserve some documentation. It seems to somehow filter out specifically non-recursive let bindings, but for some reason include recursive let bindings. 🧐

shared/src/test/diff/nu/Refinements.mls
```diff
@@ -85,15 +85,17 @@ let r = new {}
 //│ Code generation encountered an error:
 //│   Unsupported `new` class term: Bra(rcd = true, Rcd())
 
+:ge
 r : Object
 //│ Object
-//│ res
-//│     = {}
+//│ Code generation encountered an error:
```
LPTK, line 91: Actually is this happening because some mutable state is not reset when the previous code-gen error is thrown? This could be fixed easily by using a `try`/`finally`. If not and it's harder to fix, pls disregard my question and let's merge.

NeilKleistGao: I tried and found it hard to fix for now: we added symbols into the scope for top-level mutually-recursion and it is hard to undo it (the previous symbol in `lexicalValueSymbols` is gone). I will trace this in issue #65 and fix it when we refactor the `JSBackend`.

PR #188, opened 2023-10-06
Fix unary minus behaviour

HarrisL2:
Breaks one test in `NegativeLits.mls` when using `:NewParser`, but doesn't produce error if changed to `:NewDefs`.
Not sure what that test is supposed to represent.

Behaviour could be further modified to allow for unary addition (+) and negation (!)


shared/src/main/scala/mlscript/NewLexer.scala
```diff
@@ -168,10 +168,10 @@ class NewLexer(origin: Origin, raise: Diagnostic => Unit, dbg: Bool) {
         // go(j, if (keywords.contains(n)) KEYWORD(n) else IDENT(n, isAlphaOp(n)))
         lex(j, ind, next(j, if (keywords.contains(n)) KEYWORD(n) else IDENT(n, isAlphaOp(n))))
       case _ if isOpChar(c) =>
-        if (c === '-' && isDigit(bytes(i + 1))) {
-          val (str, j) = takeWhile(i + 1)(isDigit)
-          lex(j, ind, next(j, LITVAL(IntLit(-BigInt(str)))))
-        } else {
+        // if (c === '-' && isDigit(bytes(i + 1))) {
```
LPTK: Let's delete this commented code.

shared/src/main/scala/mlscript/NewLexer.scala
```diff
@@ -180,7 +180,7 @@ class NewLexer(origin: Origin, raise: Diagnostic => Unit, dbg: Bool) {
           }
           // else go(j, if (isSymKeyword.contains(n)) KEYWORD(n) else IDENT(n, true))
           else lex(j, ind, next(j, if (isSymKeyword.contains(n)) KEYWORD(n) else IDENT(n, true)))
-        }
+        //}
```
LPTK: Same.

shared/src/test/diff/parser/NegativeLits.mls
```diff
@@ -1,14 +1,18 @@
 :NewParser
 :ParseOnly
 
+:e // FIXME: What does this test represent?
```
LPTK: Oh, this test makes me realize why @chengluyu added support for negative literals in the lexer in the first place – it was to support negative singleton types like `-1`, as demonstrated here.

A better solution is to add a case in the code you added to the parser: check when the rhs is an integer literal, and create a new negative integer literal as a result.

shared/src/test/diff/nu/UnaryMinus.mls
```diff
@@ -0,0 +1,37 @@
+:NewDefs
```
LPTK, line 1: You forgot to test `- 1`.

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -787,6 +787,17 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], newDefs: Bo
       case ((KEYWORD(";;") /* | NEWLINE */ /* | BRACKETS(Curly, _) */, l0) :: _) =>
         R(UnitLit(true).withLoc(S(l0)))
         // R(errExpr) // TODO
+      case (IDENT("-", true), l0) :: _ /*if opPrec("-")._1 > prec*/ =>
+        consume
+        val v = Var("-").withLoc(S(l0))
+        exprOrIf(opPrec("-")._2) match {
+          case L(rhs) => ???
```
LPTK: If you don't intend to handle this case, use `expr` instead of `exprOrIf`.

LPTK (changes requested):
Thanks! This is definitely an improvement, but plz see the PR comments.

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -787,16 +787,16 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], newDefs: Bo
       case ((KEYWORD(";;") /* | NEWLINE */ /* | BRACKETS(Curly, _) */, l0) :: _) =>
         R(UnitLit(true).withLoc(S(l0)))
         // R(errExpr) // TODO
-      case (IDENT("-", true), l0) :: _ /*if opPrec("-")._1 > prec*/ =>
+      case (IDENT("-", true), l0) :: _ /*if opPrec("-")._1 > prec*/ => // Unary subtraction
```
```suggestion
      case (IDENT("-", true), l0) :: _ /*if opPrec("-")._1 > prec*/ => // Unary minus
```

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -787,16 +787,16 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], newDefs: Bo
...
         consume
         val v = Var("-").withLoc(S(l0))
         expr(opPrec("-")._2) match {
-          case IntLit(i) =>
-            exprCont(IntLit(-i), prec, false)
-          case rhs: Term => 
+          case IntLit(i) => // Special case for negative literals
+            exprCont(IntLit(-i), prec, false)
```
```suggestion
            exprCont(IntLit(-i), prec, false)
```

LPTK (approved):
LGTM, thanks! Plz just address the two comments, then I'll squash-merge this.

PR #189, opened 2023-10-09
Fix parenthesization for binary operators

HarrisL2:
A small modification that shouldn't impact other code, fixes part of [#65](https://github.com/hkust-taco/mlscript/issues/65#issuecomment-1752113845)


LPTK (changes requested):
Thanks for the quick fix!

Only thing is that there's already `shared/src/test/diff/codegen/Parentheses.mls`, so better move your test cases there.

Also, let's add this one:

```fs
fun (--) minusminus(a, b) = a - b
//│ fun (--) minusminus: (Int, Int) -> Int

:js
1 -- (3 -- 5)
//│ Int
//│ // Prelude
//│ class TypingUnit3 {}
//│ const typing_unit3 = new TypingUnit3;
//│ // Query 1
//│ res = minusminus(1, minusminus(3, 5));
//│ // End of generated code
//│ res
//│     = 3
```

It doesn't have a problem because custom operators are desugared to plain methods. It was not clear to me that this is why your condition `JSBinary.operators.contains(op)` is sufficient in `JSBackend.scala`. Better add a comment to that effect there!

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -277,7 +277,9 @@ class JSBackend(allowUnresolvedSymbols: Boolean) {
           name -> translateTerm(value)
         }) :: Nil
       )
-    case Bra(false, trm) => trm match { // Only parenthesize binary operatiors
+    // Only parenthesize binary operators
+    // Custom operators are also handled since they are desugared to plain methods
```
```suggestion
    // Custom operators do not need special handling since they are desugared to plain methods
```

shared/src/test/diff/codegen/NuParenthesis.mls
```diff
@@ -0,0 +1,179 @@
...
+//│ // Query 1
+//│ res = minusminus(1, minusminus(3, 5));
+//│ // End of generated code
+//│ res
+//│     = 3
+
+fun (-+-) complex(a, b) = a - 2*b
+//│ fun (-+-) complex: (Int, Int) -> Int
+
+:p
```
LPTK: The test is not clean. Please remove this and finally get this PR merged...

LPTK: Never mind, I did it myself.

LPTK (approved):

PR #194, opened 2023-11-23
New UCS desugarer and rudimentary `PreTyper`

chengluyu:
## TL;DR

This PR adds new UCS desugarer and a rudimentary implementation of `PreTyper`.

### Desugarer

The new UCS desugarer replaces the old desugarer. The facilities old desugarer were also removed. The new desugarer has following facilities.

- **Two Sets of Syntax.** We use two sets of syntax tree nodes: source abstract syntax and core abstract syntax. Both of them are implement according to our paper precisely in package `mlscript.ucs.syntax`.
- **Stage Classes.** The new desugarer consists of five stages, which correspond five classes in package `mlscript.ucs.stages`. The stages are the following.
	- `Transformation`: transforming parsed AST to source abstract syntax.
	- `Desugaring`: generating let bindings and expand patterns.
	- `Normalization`: removing backtracking and convert core abstract syntax to case branches.
	- `PostProcessing`: merging case branches with the same scrutinee.
	- `CoverageChecking`: detecting unreachable cases and duplicated cases.
- **Scrutinee Management.** We tracks useful information of each scrutinee's patterns by `Context`, `Scrutinee` and `Pattern`.
	- Each `If` term is associated with a unique `Context`. A `Context` keeps track of all scrutinee and is in charge of allocating temporary variable names.
	- `Scrutinee` represents a term that is being matched in UCS expressions. It keeps the tally of all patterns the scrutinee has been matched against as well as occurrences and sub-scrutinees of each pattern.

We support two kinds of new patterns in this PR.

- **_Simple_ Tuple Pattern** is a syntax sugar for selecting fields. It cannot be used together with other patterns.
- **Alias Pattern** allows people name the entire term represented by a pattern.

### PreTyper

- `PreTyper` traverses the AST and associate each variable with symbol it refers to.
- **A Hierarchy of `Symbol`s.** Each symbol class maintains certain information useful to desugarer.
- `Scope` maintains an association from variables to symbols.

### DiffTests

- Local flag `:ducs` controls the new `Desugarer` right now. It supports sub-flags.
	- `:ducs:[stage]` where `[stage]` is one of `transform`, `desugar`, `normalize`, `postprocess`, `coverage` will display the debug messages from the specified stage.
	- `:ducs:[stage].result` shows the pretty-printed outcome of the specified stage.
	- `:ducs:desugared` shows the desugared MLscript term.
	- `:ducs:A,B,C` shows debug messages for all `A`, `B`, and `C`.
- Global flag `:ShowPreTyperErrors` emits warnings and errors reported by `PreTyper`.

### Others

- The utility trait `Traceable` abstracts debugging methods used in `Typer`, including `trace`, `println`, etc.
- Implement `freeVars` for `Located`. The original implementation is broken. But my implementation is not perfect either.
- Support parsing `_ then ...` and `else` in `IfOpsApp`.
- Fix code generation for case expression of `Num`, `Int`, `Bool`, and `Str`.
- Fix display of escaped characters in `StrLit`.
- Add utility methods `foldLeft` and `foreach` to `CaseBranchesImpl`.

## Future Works

This PR does not implement all features, but it lays the foundation for some of the following features.

- Branch deduplication. Currently, we duplicate branches during normalization, which generates many code in some examples (e.g., JSON parser). We can reduce branches by abstracting duplicated branches into local functions.
- _Better_ tuple patterns. As mentioned before, we only implement "simple" tuple patterns. It is not real tuple patterns because the underlying MLscript `CaseOf` doesn't support tuples right now. However, it paves the road for complicated tuple patterns even tuple splices.

Besides, we can still improve location reporting in many places.


shared/src/main/scala/mlscript/ucs/stages/Transformation.scala
```diff
@@ -105,26 +105,27 @@ trait Transformation { self: mlscript.pretyper.Traceable =>
...
     case App(classNme @ Var(_), Tup(parameters)) =>
       ClassPattern(classNme, S(parameters.map {
         case (_, Fld(_, Var("_"))) => N // Consider "_" as wildcard.
-        case (_, Fld(_, t)) => S(rec(t))
+        case (_, Fld(_, t)) => S(transformPattern(t))
       }))
     case Tup(fields) => TuplePattern(fields.map {
       case _ -> Fld(_, Var("_")) => N // Consider "_" as wildcard.
-      case _ -> Fld(_, t       ) => S(rec(t))
+      case _ -> Fld(_, t       ) => S(transformPattern(t))
```
LPTK: Maybe document this part of the code:
```suggestion
      case _ -> Fld(_, t       ) => S(transformPattern(t))
      //                ^^^^^^^  space reserved to park my car
```

shared/src/test/diff/pretyper/ucs/DualOption.mls
```diff
@@ -1,13 +1,13 @@
 :NewDefs
 :PreTyper
 
-class Some[T](value: T)
-module None
-type Option[T] = Some[T] | None
```
LPTK: Why this change? It doesn't work with the type alias?

chengluyu: It works with the type alias. I just wanted to see if there's any difference if I use abstract classes. I'll keep both versions in different test files.

shared/src/test/diff/pretyper/ucs/SpecilizationCollision.mls
```diff
@@ -41,10 +43,25 @@ fun example3(t) =
...
 //│ fun example3: Base -> Int
 
-// FIXME
 fun example4(t, x) =
   if t is
     Base(x) and p1(x) then x
-    Derived(y) then y + x // Oh no, x is captured by Base! Because this branch is absorbed by the previous one.
+    Derived(y) then y + x
+    //                  ^
+    // Oh no, x is captured by x from Base!
```
LPTK: Isn't this comment outdated?

shared/src/main/scala/mlscript/ucs/Context.scala
```diff
@@ -0,0 +1,65 @@
...
+  def toMatchRegistry: MatchRegistry =
+    scrutineeBuffer.iterator.flatMap { scrutinee =>
+      val caseSet = scrutinee.toCaseSet
+      scrutinee.aliasesIterator.map(alias => (alias -> scrutinee) -> caseSet)
+    }.toMap
+}
+
+object Context {
+  // TODO: Generate fresh prefix in a determinstic way. I tried to use a counter,
+  // but the produced value is not stable across different runs.
```
LPTK: Counters are stable... Your implementation may not be

shared/src/main/scala/mlscript/pretyper/PreTyper.scala
```diff
@@ -4,6 +4,7 @@ import collection.mutable.{Set => MutSet}
 import mlscript.ucs.DesugarUCS
 import symbol._
 import mlscript._, utils._, shorthands._
+import mlscript.{Cls, Trt, Mxn, Als, Mod}
 import scala.annotation.tailrec
```
LPTK: Redundant, you already import `mlscript._`.

shared/src/test/diff/pretyper/ucs/patterns/Literals.mls
```diff
@@ -10,7 +10,7 @@ class Pair[A, B](x: A, y: B)
 //│ class Pair[A, B](x: A, y: B)
 
 fun f(x) = if x is Some(1) then true else false
```
LPTK, line 12: What happens if you remove the `else` clause? Add that test case?

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -1590,4 +1593,17 @@ object JSBackend {
 
   def isSafeInteger(value: BigInt): Boolean =
     MinimalSafeInteger <= value && value <= MaximalSafeInteger
+
+    // Temporary measurement until we adopt the new tuple index.
```
```suggestion
  // Temporary measure until we adopt the new tuple index.
```

LPTK (commented):
It would be better to merge the changes to term printing in a separate prior PR, so the diff of this one is not utterly polluted by this change.

shared/src/test/diff/pretyper/ucs/SpecilizationCollision.mls
chengluyu: Yes. You're right.

chengluyu:
🥱 I merged (rather than rebase) `new-definition-typing` because there are just too many commits. The good news is that the changed test files are all related to UCS now.

shared/src/main/scala/mlscript/pretyper/PreTyper.scala
```diff
@@ -219,6 +220,16 @@ object PreTyper {
     rec(Nil, parents)
   }
 
+  /**
+    * Extract types in class signatures. For example, for this piece of code
+    * ```mls
+    * abstract class Option[A]: Some[A] | None
+    * ```
+    * this function returns, `Some` and `None`.
```
LPTK: Again, could you explain to me why we need seal-ness info in the pretyper/UCS?

shared/src/main/scala/mlscript/ucs/stages/Desugaring.scala
```diff
@@ -285,20 +292,30 @@ trait Desugaring { self: PreTyper =>
...
-              term = mkBinOp(scrutineeVar, Var("==="), literal, true),
-              tail = c.Branch(
-                scrutinee = test,
-                pattern = truePattern,
-                continuation = desugarTermSplit(head.continuation)(PartialTerm.Empty, scope + test.symbol, context)
-              ) :: rec(scrutineeVar, tail)
-            )
-          case s.ConcretePattern(nme) => 
+            scrutineeVar.getOrCreateScrutinee
+                        .withAlias(scrutineeVar)
```
LPTK: No horizontal alignment plz. We should be able to rename variables without having to change other code's indentation.

shared/src/main/scala/mlscript/ucs/stages/Transformation.scala
```diff
@@ -169,15 +237,17 @@ trait Transformation { self: mlscript.pretyper.Traceable =>
 }
 
 object Transformation {
-  private object OperatorIs {
+  private def rare: Nothing = lastWords("found a very rare case during desugaring UCS terms")
```
LPTK: What does it mean for something to be "rare"? And do you think it looks serious if the compiler tells the user they hit a """rare""" case? Why not just use `die`?

shared/src/test/diff/pretyper/ucs/patterns/Literals.mls
```diff
@@ -10,25 +10,91 @@ class Pair[A, B](x: A, y: B)
...
+//│     = [ 1, 2, 2 ]
 
 fun g(x) = if x && true is true then 1 else 2
 //│ fun g: Bool -> (1 | 2)
 
 fun h(x) = if (x : Bool) then 1 else 2
 //│ fun h: Bool -> (1 | 2)
 
+// Currently post-processing cannot hanlde this case. The desugared term is not
+// perfect. Also, is the inferred type wrong?
```
LPTK, line 59: What's the desugared term?

shared/src/test/diff/pretyper/ucs/patterns/Literals.mls
```diff
@@ -10,25 +10,91 @@ class Pair[A, B](x: A, y: B)
...
+fun mix(x) = if x is
+  true then "true"
+  Some(value) then "Some"
+  0 then "zero"
+//│ fun mix: (0 | Some[anything]) -> ("Some" | "true" | "zero")
+
+[mix(true), mix(Some(1)), mix(0)]
```
LPTK, line 66: This shouldn't type check according to the inferred signature above. Maybe a type simplification bug 🤔

LPTK:
Is it on purpose that there are several deleted files in the "migration" commit? https://github.com/hkust-taco/mlscript/pull/194/commits/d6bf3a38642274f7384f78d5a53ac1a69491a6d2

chengluyu:
> Is it on purpose that there are several deleted files in the "migration" commit? [d6bf3a3](https://github.com/hkust-taco/mlscript/commit/d6bf3a38642274f7384f78d5a53ac1a69491a6d2)

Yes, that's on purpose. Actually, there are only two examples, each with two files, which are the ones with `NewParser` and `NewDefs` activated respectively. One example had already been migrated in a previous commit, so I deleted the old files.

shared/src/main/scala/mlscript/ucs/stages/Desugaring.scala
chengluyu: Done.

shared/src/main/scala/mlscript/ucs/stages/Transformation.scala
chengluyu: So far, this message is for my own reference, because I found some unconsidered transformation cases while I was migrating test cases. However, I can also remove it because I found that I can still view stack trace.

shared/src/test/diff/pretyper/ucs/patterns/Literals.mls
chengluyu:
```
case x of {
  true => "true";
  _ => case x of {
    Some =>
      let ucs$args_x$Some = (Some).unapply(x,) in
      let value = (ucs$args_x$Some).0 in
      "Some";
    _ => case x of {
        0 => "zero"
    }
  }
}
```

shared/src/test/diff/pretyper/ucs/patterns/Literals.mls
chengluyu: I posted the desugared term at the previous comment. I think it's a type simplification bug.

shared/src/test/diff/pretyper/ucs/examples/ULC.mls
```diff
@@ -320,20 +320,46 @@ fun search(f: 'A -> Option['B], xs: List['A]): Option['B] =
 //│   where
 //│     'B <: 'A0
 
-// The removal of type annotations will cause running out of fuel.
+// =============================================================================
+// TO BE INVESTIGATED: The following version does not have a stable type.
```
LPTK, line 326: Does the type oscillate between different possible results? While the desugared term doesn't change?

chengluyu: Yes. Just hit <kbd>Command + S</kbd> to see it changes.

chengluyu: It sometimes switches between the two types below.

```diff
-//│ fun combinations: forall 'T 'T0 'A. (n: Int, acc: List['T0], alphabet: List['T], xs: List[Str]) -> Option[Str]
+//│ fun combinations: forall 'T 'A 'T0. (n: Int, acc: List['T], alphabet: List['T0], xs: List[Str]) -> Option[Str]
 //│   where
-//│     'T <: 'T0 & 'A
-//│     'T0 :> 'A
-//│     'A := 'T0
+//│     'T0 <: 'T & 'A
+//│     'T :> 'A
+//│     'A := 'T
```

chengluyu: It also oscillates if I remove type annotations. Moreover, if I continue to remove the type annotations of some functions above (like `reverse`), the fuel will run out, but I'm not sure which functions to remove right now.

chengluyu: Here is a slightly simpler code whose type also oscillates. I'm sure it has a stable desugared term.

```
fun combinations(n: Int, acc: List['T], alphabet: List['T], xs: List[Str]): Option[Str] =
  if n <= 0 then
    let x = reverse(acc) |> join("")
    if contains(xs, x) then None else Some(x)
  else
    search((x) => combinations(n - 1, x :: acc, alphabet, xs), alphabet)
```

Here is a video showing how it changed.

https://github.com/hkust-taco/mlscript/assets/2239547/c45a11b5-074d-4a89-9f6c-c5fcd162a40b

shared/src/test/diff/pretyper/ucs/patterns/Literals.mls
LPTK: We should add a way of pretty-printing the code in the new syntax so we can easily try it out by copy-pasting it...

chengluyu: Do you want me to do that? I'd happy to work on it. 😈

shared/src/test/diff/pretyper/ucs/examples/LispInterpreter.mls
```diff
@@ -504,32 +503,30 @@ showEval("(def id (lambda (x) x) (id 42))")
...
-//│ Str
-//│ res
-//│     = 'Ok: 120'
-//│ res
-//│     = 'Ok: 55'
-//│ res
-//│     = 'Ok: 1275'
-//│ res
-//│     = 'Ok: 29'
+// showEval("(def fact (lambda (n) (if (eq n 0) 1 (* n (fact (- n 1))))) (fact 5))")
```
LPTK: Don't forget to uncomment

shared/src/test/diff/pretyper/ucs/patterns/Literals.mls
LPTK: Sure, if you feel like it!

shared/src/test/diff/pretyper/ucs/examples/JSON.mls
```diff
@@ -419,7 +419,7 @@ fun parseObjectEntry(state: ParserState): ParseResult[[Str, JsonValue]] =
       None then Failure("expected ':' instead of end of input")
     else Failure("expected ':' instead of end of input")
 fun parseObject(state: ParserState): ParseResult[ListMap[Str, JsonValue]] =
-  let rec parseObjectTail(acc, state) =
+  let rec parseObjectTail(acc: ListMap[Str, JsonValue], state: ParserState) =
```
chengluyu, line 421: This line now needs type annotation. Otherwise, the typer will exceed the recursion limit.

shared/src/test/diff/pretyper/ucs/examples/LispInterpreter.mls
chengluyu: Thanks for the reminder!

shared/src/test/diff/pretyper/ucs/examples/JSON.mls
chengluyu: This typing unit contains many functions, because they all reference each other. Each function has auxiliary functions, so constraint solver may go into deeper recursion. I guess, maybe, the recursion limit should be a dynamic value which varies with the scale of the typing unit.

LPTK: If you increase it too much you just run into stack overflows.

shared/src/test/diff/ucs/HygienicBindings.mls
```diff
@@ -28,14 +28,42 @@ fun h0(a) =
     a is None then 0
 //│ fun h0: forall 'a. (None | Some[Left['a] | Right['a]]) -> (0 | 'a)
 
-// FIXME: Precise scrutinee identification (easy)
-// This seems fine. But the subtrees are not merged.
+// Precise scrutinee identification (easy)
+// =======================================
+// The desugarer should be able to identify pattern matching on the first
+// parameter of `Some` even if they are bound to different variables.
+:dpt:postprocess.result
```
LPTK: TODO rm

shared/src/test/diff/ucs/HygienicBindings.mls
```diff
@@ -55,54 +83,36 @@ fun h1'(a) =
     None then 0
 //│ fun h1': forall 'leftValue. (None | Some[Right['leftValue]]) -> (0 | 'leftValue)
 
-// FIXME This seems fine but the desugared term does not merge the cases.
-// See the example below.
-fun h1''(a) =
```
LPTK, line 60: Why remove this interesting slight variation?

shared/src/test/diff/ucs/HygienicBindings.mls
chengluyu: rm'ed

shared/src/test/diff/codegen/Nested.mls
```diff
@@ -1262,9 +1262,6 @@ fun main =
     else g(x - 1)
   let g(x: Int): Int = f(x)
   f
-//│ ╔══[ERROR] identifier `g` not found
```
LPTK: Why did this error disappear?

chengluyu: This error is from `PreTyper`, not from `Typer`. I have muted the output of `PreTyper` in all tests, as `PreTyper` is not yet fully implemented. Currently, `PreTyper` can only complete the symbol resolution required by UCS.

By using the global flag `:ShowPreTyperErrors`, you can turn on the error output of `PreTyper`. I have already copied the use cases that will cause `PreTyper` to fail to `pretyper/Errors.mls` for future research when implementing `PreTyper`.

todo.md
chengluyu, line 1: This file should be deleted.

shared/src/test/diff/ucs/Wildcard.mls
```diff
@@ -196,7 +206,18 @@ fun w5(y) =
...
+//│ case y*‡ of
+//│   Alpha*◊ -> "alpha"
+//│   Gamma*◊ -> "gamma"
+//│   Delta*◊ -> "delta"
+//│   Beta*◊ -> "beta"
+//│   _ ->
+//│     case y*‡ of
+//│       _ ->
+//│         case y*‡ of
+//│           _ -> "unknown"
```
chengluyu: We can coalesce the nested wildcard branches.

shared/src/test/diff/ucs/WeirdIf.mls
```diff
@@ -1,96 +1,93 @@
...
 else 1
-//│ ╔══[WARNING] Found a duplicated branch
-//│ ╟── This branch
-//│ ║  l.7: 	  else 0
+//│ ╔══[WARNING] unreachable case
+//│ ║  l.6: 	  else 0
 //│ ║       	       ^
-//│ ╟── is subsumed by the branch here.
-//│ ║  l.6: 	  _ then 0
+//│ ╟── because this branch covers the case
```
chengluyu: The old message looks better.

shared/src/test/diff/ucs/WeirdIf.mls
```diff
@@ -1,96 +1,93 @@
...
-//│ error
-//│ Code generation encountered an error:
-//│   if expression was not desugared
+//│ ╔══[ERROR] missing else branch
+//│ ║  l.51: 	  then 0
+//│ ╙──      	       ^
+//│ ╔══[ERROR] Type mismatch in `case` expression:
+//│ ║  l.51: 	  then 0
+//│ ║        	       ^
+//│ ╙── expression of type `Bool` is not an instance of type `true`
```
chengluyu, line 58: This location can be improved.

shared/src/test/diff/ucs/WeirdIf.mls
```diff
@@ -102,7 +99,9 @@ fun boolToStr(x) =
   if x is
     true then "yah"
     false then "nah"
-//│ fun boolToStr: nothing -> ("nah" | "yah")
+//│ ╙──
+//│ ╙──
```
chengluyu: Strange.

shared/src/test/diff/codegen/NewDefsBug.mls
chengluyu, line 1: This file should be deleted. Because it was renamed to NewMutualRef.mls in the main branch.

shared/src/test/diff/gadt/Exp1.mls
```diff
@@ -33,12 +33,18 @@ fun f(e) = if e is
...
+//│ ╔══[ERROR] identifier not found: l
+//│ ║  l.35: 	  Pair['a, 'b](l, r) then [l, r]
+//│ ╙──      	                           ^
+//│ ╔══[ERROR] identifier not found: r
+//│ ║  l.35: 	  Pair['a, 'b](l, r) then [l, r]
+//│ ╙──      	                              ^
+//│ fun f: anything -> [error, error]
 //│ Code generation encountered an error:
-//│   if expression was not desugared
+//│   unresolved symbol l
```
chengluyu: We can actually accept this kind of pattern with a warning message: type parameters `'a` and `'b` will be ignored.

shared/src/test/diff/nu/ArrayProg.mls
```diff
@@ -189,34 +189,44 @@ fun add(e) =
...
     Pair(Numbr(n), Numbr(m))    then 0
     Pair(Vectr(xs),  Vectr(ys)) then 1
-//│ ╔══[ERROR] The match is not exhaustive.
-//│ ║  l.+2: 	  if e is
-//│ ║        	     ^^^^
-//│ ╟── The scrutinee at this position misses 1 case.
+//│ ╔══[ERROR] When scrutinee `e` is `Pair`,
+//│ ║  l.+3: 	    Pair(Numbr(n), Numbr(m))    then 0
+//│ ║        	    ^^^^
+//│ ╟── scrutinee `e$Pair_0` is `Numbr`, and
```
chengluyu: `e$Pair_0` is not meaningful and can be improved.

shared/src/main/scala/mlscript/pretyper/Symbol.scala
```diff
@@ -0,0 +1,101 @@
...
+
+    def isDeclaration: Bool = defn.rhs.isRight
+
+    def operatorAlias: Opt[Var] = defn.symbolicNme
+  }
+
+  class LocalTermSymbol(val nme: Var) extends TermSymbol {
+    override def name: Str = nme.name
+  }
+}
```
chengluyu: Missing a trailing newline.

shared/src/main/scala/mlscript/pretyper/package.scala
chengluyu, line 1: This file can be deleted as well.

shared/src/main/scala/mlscript/ucs/context/CaseSet.scala
```diff
@@ -0,0 +1,95 @@
...
+}
+
+object Pattern {
+  final case class ClassLike(symbol: TypeSymbol) extends Pattern
+  // Currently, there is only simple tuple pattern, so we cannot differentiate
+  // between tuple patterns of different arity. That's why the parameter list
+  // is empty for now.
+  final case class Tuple() extends Pattern
+  final case class Literal(literal: Lit) extends Pattern
+}
```
chengluyu: Can we reuse `PatternInfo`?

shared/src/main/scala/mlscript/ucs/package.scala
```diff
@@ -0,0 +1,46 @@
...
+        case (0, line) :: lines if lines.forall(_._1 > 0) => (0, s"$prefix $line") :: lines
+        case lines => (0, prefix) :: lines.indent
+      }
+    }
+    def @:(prefix: Str): Lines = {
+      lines match {
+        case (_, line) :: Nil => (0, prefix + " " + line) :: Nil
+        case lines => (0, prefix) :: lines.indent
+      }
+    }
```
chengluyu: These operators need documentation.

shared/src/main/scala/mlscript/ucs/stages/PostProcessing.scala
```diff
@@ -0,0 +1,223 @@
...
+          rest -> S(body)
+        } else {
+          val (n1, y1) = disentangleTerm(body)
+          val (n2, y2) = disentangleMatchedCaseBranches(rest)
+          (kase.copy(body = n1, rest = n2)(kase.refined), mergeTerms(y1, y2))
+        }
+      // case kase @ Case(otherClassName, body, rest) =>
+      //   println(s"found another case branch matching against $otherClassName")
+      //   val (n, y) = disentangleMatchedCaseBranches(rest)
+      //   kase.copy(rest = n) -> y
```
chengluyu: rm?

shared/src/main/scala/mlscript/ucs/syntax/core.scala
```diff
@@ -0,0 +1,102 @@
...
+    }
+  }
+
+  object Split {
+    final case class Cons(head: Branch, tail: Split) extends Split
+    final case class Let(rec: Bool, name: Var, term: Term, tail: Split) extends Split
+    final case class Else(default: Term) extends Split
+    final case object Nil extends Split
+  }
+}
```
chengluyu: Missing the trailing newline.

shared/src/main/scala/mlscript/ucs/syntax/package.scala
chengluyu, line 1: Rm this file.

shared/src/test/diff/codegen/NewMatching.mls
```diff
@@ -264,13 +250,17 @@ fun c(x) =
...
-//│ ╙──       	    ^^^^^^^^^^^^
+//│ ╔══[ERROR] Type mismatch in field selection:
+//│ ╟── tuple literal of type `{0: ?#v, 1: ?#vc}` does not have field '2'
+//│ ║  l.248: 	class VVC(v: Int, vc: Int)
+//│ ║         	          ^^^^^^^^^^
+//│ ╟── but it flows into operator application with expected type `{2: ?a}`
+//│ ║  l.250: 	  if x is
+//│ ║         	     ^^^^
+//│ ║  l.251: 	    VVC(x, y, z) then x + y + z
+//│ ╙──       	^^^^^^^
```
chengluyu, line 261: The old desugarer generates better messages. We should polish the error message.

shared/src/test/diff/pretyper/Declarations.mls
chengluyu, line 1: rm

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -766,6 +795,18 @@ trait Located {
...
+    case Let(false, nme, rhs, body) => body.freeVars - nme ++ rhs.freeVars
+    case Lam(tup: Tup, body) => body.freeVars -- tup.freeVars
+    case Tup(fields) => fields.iterator.flatMap(_._2.value.freeVars.iterator).toSet
+    case Blk(stmts) => stmts.iterator.foldRight(Set.empty[Var]) {
+      case (NuFunDef(isLetRec, nme, _, _, L(rhs)), fvs) => fvs - nme ++ (isLetRec match {
+        case N | S(true) => rhs.freeVars - nme
+        case S(false) => rhs.freeVars
+      })
+      case (statement, fvs) => fvs ++ statement.freeVars
+    }
```
chengluyu: Note that the implementation of `freeVars` is still incomplete.

shared/src/test/diff/nu/CaseExpr.mls
```diff
@@ -63,21 +63,24 @@ map(succ) of None
...
 //│ ║  l.65: 	  None as n then n
 //│ ╙──      	       ^^
-//│ fun map: anything -> anything -> error
+//│ ╔══[ERROR] type identifier not found: as
+//│ ║  l.65: 	  None as n then n
+//│ ╙──      	       ^^
+//│ fun map: forall 'a 'A. ('a -> 'A) -> Some['a] -> (Some['A] | error)
 //│ Code generation encountered an error:
-//│   if expression was not desugared
+//│   unresolved symbol as
```
chengluyu: Can we support `as` pattern? I ignored this.

shared/src/test/diff/nu/CaseExpr.mls
```diff
@@ -129,37 +132,40 @@ case then 1 else 0
...
-//│   if expression was not desugared
-
-:e // * FIXME[UCS]
-case [x, y] then x + y
-//│ ╔══[ERROR] type identifier not found: Tuple#2
-//│ ╙──
 //│ nothing -> error
 //│ Code generation encountered an error:
-//│   unknown match case: Tuple#2
+//│   unresolved symbol ,
```
chengluyu, line 152: Code generator should support `,` operators.

shared/src/test/diff/nu/FilterMap.mls
```diff
@@ -28,11 +28,10 @@ fun filtermap(f, xs) = if xs is
...
     [true, z] then Cons(y, filtermap(f, ys))
-//│ ╔══[ERROR] type identifier not found: Tuple#2
-//│ ╙──
-//│ fun filtermap: ((Cons[nothing] | Nil) -> nothing, Cons[anything] | Nil) -> (Cons[nothing] | Nil | error)
-//│ Code generation encountered an error:
-//│   unknown match case: Tuple#2
+//│ ╔══[ERROR] unsupported pattern
+//│ ║  l.30: 	    [true, z] then Cons(y, filtermap(f, ys))
+//│ ╙──      	     ^^^^
+//│ fun filtermap: ((Cons[nothing] | Nil) -> (Object & {1: anything} & ~false & ~true | false | true), Cons[anything] | Nil) -> (Cons[nothing] | Nil)
```
chengluyu: This should be supported!

shared/src/test/diff/nu/LitMatch.mls
```diff
@@ -22,17 +22,21 @@ b : true | false
...
+//│ ╙──
+//│ fun foo: false -> 0
 
 fun foo(x) = if x is
   false then 0
   true  then 1
-//│ fun foo: nothing -> (0 | 1)
+//│ ╙──
+//│ ╙──
+//│ fun foo: Bool -> (0 | 1)
```
chengluyu: Weird.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -116,6 +116,7 @@ abstract class JSBackend(allowUnresolvedSymbols: Bool) {
         else
           throw new UnimplementedError(sym)
       case S(sym: ValueSymbol) =>
+        // Temporary disable this line of code because it invalidates many working test cases.
```
chengluyu: rm

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -1594,4 +1596,17 @@ object JSBackend {
...
+    def unapply(fieldName: Var): Opt[Int] = {
+      val name = fieldName.name
+      if (name.startsWith("_") && name.forall(_.isDigit))
+        name.drop(1).toIntOption match {
+          case S(n) if n > 0 => S(n - 1)
+          case _ => N
+        }
+      else N
+    }
+  }
```
chengluyu: rm

shared/src/test/diff/nu/CaseExpr.mls
LPTK: Weird, I thought it already did.

shared/src/test/diff/pretyper/ucs/DualOption.mls
```diff
@@ -0,0 +1,164 @@
+:NewDefs
+:NewDefs
```
NeilKleistGao:
```suggestion
:NewDefs

```

shared/src/main/scala/mlscript/ucs/stages/Normalization.scala
```diff
@@ -0,0 +1,346 @@
...
+import pretyper.{Diagnosable, Traceable}
+
+trait Normalization { self: Desugarer with Traceable =>
+  import Normalization._
+
+  // TODO: We might not need the case where `deep` is `false`.
+  private def fillImpl(these: Split, those: Split, deep: Bool)(implicit
+      scope: Scope,
+      context: Context,
+      generatedVars: Set[Var],
```
NeilKleistGao: Avoid using implicit set parameters to prevent the following implicit conversion problem:
```scala
def foo(implicit v: Set[String]) = {
  if ("abc") 42 else 0
}
```

shared/src/main/scala/mlscript/ucs/context/CaseSet.scala
chengluyu: Done.

shared/src/test/diff/nu/CaseExpr.mls
chengluyu: This is parsed as `if case$scrut is (x, y) then x + y`. So, it's not a problem of new UCS desugarer, we can address problems of parsing and desugaring UCS shorthands after we finish UCS paper.

LPTK:
So can  `shared/src/test/diff/pretyper/Errors.mls` be removed now? It says:

```fs
// This file is used to test the error messages of the _unfinished_ `PreTyper`.
// The goal is to document `PreTyper` behavior and to make sure that the error
// messages are clear and helpful. We can delete this file after it is done.
```
Or is the pretyper still not being used by default?

chengluyu:
The `PreTyper` is being used now. But the error messages and warnings it generates will be hidden, because they are redundant with the information generated by `Typer`. The purpose of retaining this file is to allow us to observe its changes when improving the `PreTyper` in the next PR.

shared/src/test/diff/pretyper/ucs/SpecilizationCollision.mls
```diff
@@ -89,10 +90,11 @@ fun example3(t) =
 //│ | | | | | | | | | remaining cases should be covered by the wildcard
 //│ | | | | | | | | | checkCoverage <== TERM 42
 //│ | | | | | | | | checkCoverage ==> 0 diagnostics
-//│ | | | | | | | | remaining cases are not covered: []
+//│ | | | | | | | | remaining cases should be covered by the wildcard
+//│ | | | | | | | | checkCoverage <== TERM 42
 //│ | | | | | | | checkCoverage ==> 0 diagnostics
 //│ | | | | | | | Coverage checking result: 0 errors
-//│ fun example3: forall 'a. (Base & {#x: Num & 'a}) -> (Int | 'a)
+//│ fun example3: forall 'a. (Base & {#x: Num & 'a} | Object & ~#Base) -> (Int | 'a)
```
LPTK, line 64: I wonder why we don't see `Derived` and its field `y` in this inferred type. Seems like a bug.

LPTK:
It seems there's something wrong in the handling of `refined` with multiple branches. I just hit this:

```fs

module None
class Some[out A](val value: A)
//│ module None
//│ class Some[A](value: A)


// OK
:ducs:desugar.result,postprocess.result
x => if x is
  refined(None) then x
//│ Desugared UCS term:
//│ if x*‡ is refined None then x
//│ Post-processed UCS term:
//│ case x*‡ of
//│   refined None*† -> x
//│ forall 'a. (None & 'a) -> (None & 'a)
//│ res
//│     = [Function: res]

// OK
:ducs:desugar.result,postprocess.result
x => if x is
  refined(Some) then x
//│ Desugared UCS term:
//│ if x*‡ is refined Some then x
//│ Post-processed UCS term:
//│ case x*‡ of
//│   refined Some*◊ -> x
//│ forall 'a. (Some[anything] & 'a) -> (Some[anything] & 'a)
//│ res
//│     = [Function: res]

// NOT OK
:ducs:desugar.result,postprocess.result
x => if x is
  refined(None) then x
  refined(Some) then x
//│ Desugared UCS term:
//│ if
//│   x*‡ is refined None then x
//│   x*‡ is refined Some then x
//│ Post-processed UCS term:
//│ case x*‡ of
//│   refined None*† -> x
//│   Some*◊ -> x
//│ (None | Some[anything]) -> Some[nothing]
//│ res
//│     = [Function: res]
```

shared/src/test/diff/pretyper/ucs/Refinement.mls
```diff
@@ -0,0 +1,49 @@
...
+//│ if x*‡ is refined Some then x
+//│ Post-processed UCS term:
+//│ case x*‡ of
+//│   refined Some*◊ -> x
+//│ forall 'a. (Some[anything] & 'a) -> (Some[anything] & 'a)
+//│ res
+//│     = [Function: res]
+
+// NOT OK
+:ducs:desugar.result,postprocess.result
```
LPTK, line 60: It's OK now 😉

shared/src/main/scala/mlscript/pretyper/PreTyper.scala
```diff
@@ -0,0 +1,345 @@
...
+      case literal: Lit =>
+        raiseWarning(msg"literal patterns are ignored" -> literal.toLoc)
+        Iterator.empty
+      case App(Var(name), parameters) if name.isCapitalized => rec(parameters)
+      case Tup(arguments) =>
+        arguments.iterator.flatMap {
+          case (S(nme: Var), Fld(_, _)) => Iterator.single(new LocalTermSymbol(nme))
+          case (N, Fld(_, term)) => rec(term)
+        }
+      case PlainTup(arguments @ _*) => arguments.iterator.flatMap(extractParameters)
```
LPTK, line 32: This raises a warning.

shared/src/main/scala/mlscript/pretyper/PreTyper.scala
```diff
@@ -0,0 +1,345 @@
...
+      transitiveClosure(edges).foreachEntry { (self, bases) =>
+        self.baseTypes = bases
+        println(s"base types of `${self.name}`: ${bases.iterator.map(_.name).mkString(", ")}")
+      }
+      // Pass 1.2: Resolve signature types for collecting sealed derived types.
+      println("Resolve sealed signature types")
+      typeSymbols.foreach {
+        case _: MixinSymbol | _: TypeAliasSymbol | _: ModuleSymbol => ()
+        case symbol => symbol.defn.sig.foreach { unions =>
+          val derivedTypes = try extractSignatureTypes(unions) catch { case _: NotImplementedError => Nil }
```
LPTK: :poop:

shared/src/main/scala/mlscript/pretyper/PreTyper.scala
```diff
@@ -0,0 +1,345 @@
...
+        case head :: tail =>
+          raiseWarning(msg"unknown type in parent types: ${head.showDbg}" -> head.toLoc)
+          rec(acc, tail)
+      }
+    rec(Nil, parents)
+  }
+}
+
+object PreTyper {
+  def transitiveClosure[A](graph: Map[A, List[A]])(implicit ord: Ordering[A]): SortedMap[A, List[A]] = {
```
LPTK: Apparently this needs to be removed?

chengluyu: No. I realized that the function computes base classes for each class so that we can know two pattern matching branches are overlapped or not.

shared/src/main/scala/mlscript/ucs/context/Scrutinee.scala
```diff
@@ -2,12 +2,9 @@ package mlscript.ucs.context
...
 import collection.mutable.{Buffer, SortedMap => MutSortedMap, SortedSet => MutSortedSet}
 import mlscript.{Lit, Loc, Var}
-import mlscript.pretyper.symbol.TypeSymbol
+import mlscript.pretyper.symbol.{ClassLikeSymbol, TypeSymbol}
 import mlscript.utils._, shorthands._
-import mlscript.DecLit
-import mlscript.IntLit
-import mlscript.StrLit
-import mlscript.UnitLit
+import mlscript.{DecLit, IntLit, StrLit, UnitLit}
```
LPTK: Instead of making all these useless imports, which are basically just noise, just make the package declaration:
```scala
package mlscript
package ucs.context
````

Comment (chengluyu) [reply shared/src/main/scala/mlscript/ucs/context/Scrutinee.scala]: Great suggestion! I've already rewritten most of the files in this way!

File: shared/src/main/scala/mlscript/pretyper/PreTyper.scala (hunk 44 above)
Comment (chengluyu) [reply shared/src/main/scala/mlscript/pretyper/PreTyper.scala]: They've already been removed.

**Comment by chengluyu** [comment]
I'm fixing not being able to report some unreachable cases and conflicting cases. I'm handling the situation we discussed before: some duplicated branches can also trigger these checks. The solution I take is to mark the duplicated syntax tree nodes, so we will not trigger warnings if these nodes are detected.

File: compiler/shared/test/diff/Defunctionalize/OldMonoList.mls [hunk 46]
```diff
@@ -42,17 +42,23 @@ fun add2(x) = x+2
...
-//│ 	at: mlscript.VarImpl.symbol_$eq(helpers.scala:870)
-//│ 	at: mlscript.VarImpl.symbol_$eq$(helpers.scala:864)
-//│ 	at: mlscript.Var.symbol_$eq(syntax.scala:67)
-//│ 	at: mlscript.ucs.Desugarer$VarOps.resolveTermSymbol(Desugarer.scala:124)
-//│ 	at: mlscript.ucs.Desugarer$VarOps.withResolvedTermSymbol(Desugarer.scala:129)
-//│ 	at: mlscript.ucs.stages.Desugaring.desugarPatternSplit(Desugaring.scala:432)
-//│ 	at: mlscript.ucs.stages.Desugaring.$anonfun$desugarTermBranch$2(Desugaring.scala:101)
-//│ 	at: mlscript.pretyper.Traceable.trace(Traceable.scala:39)
-//│ 	at: mlscript.pretyper.Traceable.trace$(Traceable.scala:36)
+//│ ╔══[ERROR] the `if` expression has already been desugared
```
LPTK: This should still be an internal error (with a more helpful error message), not a user-facing type error! The hint doesn't make any sense for a user.

LPTK:
> The solution I take is to mark the duplicated syntax tree nodes, so we will not trigger warnings if these nodes are detected.

That's a good enough short-term solution, but after this PR is merged we need to think of a more principled and correct way of handling this!

chengluyu:
> That's a good enough short-term solution, 

This solution is more limited than I thought. I expected it to be able to detect all the duplicated cases. Because a split might be concatenated as a fallback in the specialization for different scrutinees, a Boolean flag is not enough. I tried to turn this boolean flag into a map, but I need to modify the map in many places. I haven't been able to make it work well until now, and the code will also become very messy.

I feel it can figure out most of the duplicated cases, but it can't detect some of the covered cases, such as the one below.

```
fun f(x) = if
  x is A and
    x is B then 1
  x is B then 2
```

> but after this PR is merged we need to think of a more principled and correct way of handling this!

I think a neater approach is to check on the desugared (rather than normalized) syntax tree. Although we can know which branches are removed during specialization, we don't know if they are removed due to duplication and which branches they overlap with because they need non-local information.

Lastly, I've also improved the warning message for different duplicated cases at the same time. There're three types of duplication.

1. Simple duplication (mainly caused by identical patterns)

		```
		:w
		fun f(x) = if x is
			"a" then 1
			"b" then 2
			"a" then 3
		//│ ╔══[WARNING] found a duplicated case
		//│ ║  l.10: 	  "a" then 3
		//│ ║        	  ^^^
		//│ ╟── there is an identical pattern "a"
		//│ ║  l.8: 	  "a" then 1
		//│ ╙──     	  ^^^
		//│ fun f: ("a" | "b") -> (1 | 2)
		```

2. Duplication caused by inheritance

		```
		:w
		fun f(x) = if
			x is A then 1
			x is B then 2
		//│ ╔══[WARNING] found a duplicated case
		//│ ║  l.27: 	  x is B then 2
		//│ ║        	       ^
		//│ ╟── the case is covered by pattern A
		//│ ║  l.26: 	  x is A then 1
		//│ ║        	       ^
		//│ ╟── due to the subtyping relation
		//│ ║  l.7: 	class B() extends A()
		//│ ╙──     	^^^^^^^^^^^^^^^^^^^^^
		//│ fun f: A -> 1
		```

3. Always-true pattern (mainly due to subtyping)

		```
		:w
		fun f(x) = if
			x is B and
				x is A then 1
				p(x) then 2
			x is A then 31
			x is B then 3
			else 4
		//│ ╔══[WARNING] the pattern always matches
		//│ ║  l.67: 	    x is A then 1
		//│ ║        	         ^
		//│ ╟── the scrutinee was matched against B
		//│ ║  l.66: 	  x is B and
		//│ ║        	       ^
		//│ ╟── which is a subtype of A
		//│ ║  l.4: 	class B() extends A()
		//│ ╙──     	^^^^^^^^^^^^^^^^^^^^^
		//│ fun f: (B | Object & ~#B) -> (1 | 31 | 4)
		```

4. Always-false pattern (mainly due to matching a scrutinee against unrelated constructors)

	  ```
	  :w
		fun f(x) = if
			x is X and
				x is Y and
					x is Z then 1
			else 2
		//│ ╔══[WARNING] possibly conflicted patterns
		//│ ║  l.105: 	    x is Y and
		//│ ║         	         ^
		//│ ╟── the scrutinee was matched against X
		//│ ║  l.104: 	  x is X and
		//│ ║         	       ^
		//│ ╟── which is unrelated with Y
		//│ ║  l.105: 	    x is Y and
		//│ ╙──       	         ^
		//│ fun f: Object -> 2
	  ```

shared/src/main/scala/mlscript/ucs/stages/Normalization.scala
```diff
@@ -197,57 +205,117 @@ trait Normalization { self: Desugarer with Traceable =>
...
+                  raiseDesugaringWarning(
+                    msg"found a duplicated case" -> thatPattern.toLoc,
+                    msg"the case is covered by pattern ${pattern.toString}" -> pattern.toLoc,
+                  )
+                }
+                specializeImpl(tail)(
+                  `-`(if (samePattern) continuation.isFull else full),
+                  scrutineeVar, scrutinee, pattern, context)
+              } else {
+                println(s"Case 1.2.2: $pattern are unrelated with $thatPattern")
```
LPTK: Sounds more natural:
```suggestion
                println(s"Case 1.2.2: $pattern are unrelated to $thatPattern")
```
https://textranch.com/c/unrelated-to-or-unrelated-with/

shared/src/main/scala/mlscript/ucs/stages/Normalization.scala
```diff
@@ -197,57 +205,117 @@ trait Normalization { self: Desugarer with Traceable =>
...
+                      msg"the pattern always matches" -> thatPattern.toLoc,
+                      msg"the scrutinee was matched against ${pattern.toString}" -> pattern.toLoc,
+                      msg"which is a subtype of ${thatPattern.toString}" -> (pattern match {
+                        case Pattern.Class(_, symbol, _) => symbol.defn.toLoc
+                        case _ => thatPattern.getLoc
+                      }))
+                    continuation :++ specializeImpl(tail)
+                  } else {
+                    raiseDesugaringWarning(
+                      msg"possibly conflicted patterns" -> thatPattern.toLoc,
```
LPTK:
```suggestion
                      msg"possibly conflicting patterns" -> thatPattern.toLoc,
```

shared/src/test/diff/pretyper/ucs/coverage/ConflictedPatterns.mls
```diff
@@ -0,0 +1,121 @@
...
+    x is Y then 1
+    x is Y then 2
+  else 3
+//│ Normalized UCS term:
+//│ case x*‡ of
+//│   X*◊ -> 3
+//│   _ -> 3
+//│ ╔══[WARNING] possibly conflicted patterns
+//│ ║  l.43: 	    x is Y then 1
+//│ ║        	         ^
```
LPTK: This location is repeated, which is a bit weird/confusing. It would be more helpful to show something like:
```suggestion
//│ ╔══[WARNING] possibly conflicted patterns for this scrutinee
//│ ║  l.43: 	    x is Y then 1
//│ ║        	    ^
```

LPTK:
> I think a neater approach is to check on the desugared (rather than normalized) syntax tree. Although we can know which branches are removed during specialization, we don't know if they are removed due to duplication and which branches they overlap with because they need non-local information.

Makes sense. This can go in a future PR!

LPTK:
> It seems there's something wrong in the handling of `refined` with multiple branches.

https://github.com/hkust-taco/mlscript/pull/194#issuecomment-1951634075

BTW was this fixed or at least reproduced in the tests?

chengluyu:
> BTW was this fixed or at least reproduced in the tests?

Yes, this was fixed.

shared/src/test/diff/pretyper/ucs/SpecilizationCollision.mls
chengluyu: Any clue on this?

shared/src/main/scala/mlscript/ucs/context/Context.scala
```diff
@@ -0,0 +1,64 @@
...
+    val scrutinee = new Scrutinee(this, S(parent))
+    scrutineeBuffer += scrutinee
+    scrutinee
+  }
+
+  def scrutinees: Iterator[Scrutinee] = scrutineeBuffer.iterator
+}
+
+object Context {
+  private def freshPrefix(): Str = "ucs"
```
LPTK:
```suggestion
  private val freshPrefix: Str = "ucs"
```

shared/src/test/diff/pretyper/ucs/SpecilizationCollision.mls
LPTK: Actually not a bug (see test we added in meeting).

LPTK (approved):
LGTM

PR #195, opened 2023-11-28
Support for GADTs

Meowcolm024:
Add support for GADTs

- GADT-reasoning in pattern matching (type annotation needed)
- Typing for type selections
- Parsing the wildcard type `?` in type arguments
- use `as` keyword for upcast (same as `:`)
- test cases `GADT[1-6].mls` in `gadt` directory, and `AsOp.mls`
- refactored wildcard type arguments
- correct extruded type args

Some TODOs (added by @LPTK):
- [x] Merge branch `local-gadt-wip` after we figure out the new errors
- [ ] Merge my local changes to simplify away compositions with `??X` types into main branch before merging this PR

---

The PR has become way too big and it's difficult to keep track of what changes cause what results or regressions. Let's split it up and merge each of these things separately:
- [ ] Typing for type selections
- [ ] Parsing the wildcard type `?` in type arguments, use `as` keyword for upcast (same as `:`)
- [ ] Refactored wildcard type arguments, correct extruded type args
- [ ] Rethink the way GADT reasoning is triggered: only introduce skolems when they correspond to user-annotated pattern variable type arguments (as in `App[f, a]`); _and_ in this case don't try to refine these type arguments in the scrutinee.


LPTK:
Cool! Will try to take a look later. In the meantime, plz read this again and make sure all the changes are clean: https://github.com/hkust-taco/mlscript/blob/mlscript/CONTRIBUTING.md

Meowcolm024:
Known issue: recursion depth exceeded (infinite extrusion loop) when typing certain terms (see `GADT3.mls` and `GADT5.mls` etc)
Observation: it seems that the loop follows a pattern of: `a''# & S <: b'`

shared/src/test/diff/nu/AsOp.mls
```diff
@@ -0,0 +1,54 @@
...
+//│ ╟── Note: constraint arises from type reference:
+//│ ║  l.29: 	"hello" as Int
+//│ ╙──      	           ^^^
+//│ Int
+//│ res
+//│     = 1
+//│ res
+//│     = 'hello'
+
+// what ?
```
LPTK: ?

shared/src/test/diff/parser/IfThenElse.mls
```diff
@@ -447,13 +447,7 @@ let Some(x) = v
 // TODO support?
```
LPTK:
```suggestion
```

shared/src/main/scala/mlscript/syntax.scala
```diff
@@ -234,6 +235,9 @@ final case class NuFunDef(
 final case class Constructor(params: Tup, body: Blk) extends DesugaredStatement with ConstructorImpl // constructor(...) { ... }
 
 
+final case class TypeParamInfo(varinfo: Option[VarianceInfo], visible: Boolean) {
```
LPTK:
```suggestion
final case class TypeParamInfo(varinfo: Opt[VarianceInfo], visible: Bool) {
```

shared/src/main/scala/mlscript/syntax.scala
```diff
@@ -234,6 +235,9 @@ final case class NuFunDef(
...
+  def getVarOr(default: VarianceInfo): VarianceInfo = varinfo.getOrElse(default)
```
LPTK: This method seems entirely redundant.

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -33,10 +33,9 @@ class ConstraintSolver extends NormalForms { self: Typer =>
         (implicit ctx: Ctx, raise: Raise)
         : Either[Diagnostic, NuMember]
         = {
-    val info = ctx.tyDefs2.getOrElse(clsNme, ???/*TODO*/)
+    val info = ctx.tyDefs2.getOrElse(clsNme, lastWords(s"Can't find member `$clsNme`"))
```
LPTK: This is a compiler crash, not a proper compiler error. Please fix or leave a TODO.

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -33,10 +33,9 @@ class ConstraintSolver extends NormalForms { self: Typer =>
...
     
     if (info.isComputing) {
-
```
LPTK, line 39: Remove whitespace change.

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -657,7 +659,8 @@ class ConstraintSolver extends NormalForms { self: Typer =>
           : SimpleType =
     {
       val originalVars = ty.getVars
-      val res = extrude(ty, lowerLvl, pol, upperLvl)(ctx, MutMap.empty, MutSortMap.empty, reason)
+      val res = extrude(ty, lowerLvl, pol, upperLvl)(ctx, ctx.extrCache, ctx.extrCache2, reason)
+      // val res = extrude(ty, lowerLvl, pol, upperLvl)(ctx, MutMap.empty, MutSortMap.empty, reason)
```
LPTK: Question: does using this line instead make any difference anywhere? What differences?

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -851,8 +854,8 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], newDefs: Bo
                 else App(App(v, PlainTup(acc)), PlainTup(rhs))
             }, prec, allowNewlines)
         }
-      case (KEYWORD(":"), l0) :: _ if prec <= outer.prec(':') =>
-        consume
+      case (KEYWORD("as" | ":"), l0) :: _ if prec <= outer.prec(':') =>
+        consume // use both "as" and ":"
```
```suggestion
        consume // * Accept both "as" and ":" for now (maybe remove `:` ascriptions later?)
```

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -1071,15 +1081,17 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], newDefs: Bo
...
         }
       case _ =>
-        vinfo match {
-          case S((_, loc)) =>
+        (visinfo, vinfo) match {
+          case (_, (S(_), S(loc))) =>
             err(msg"dangling variance information" -> S(loc) :: Nil)
-          case N =>
+          case ((true, S(loc)), _) =>
+            err(msg"dangling visible type member" -> S(loc) :: Nil)
```
LPTK, line 1335: We need a test to exercise this.

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -1051,18 +1054,25 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], newDefs: Bo
...
   // TODO support line-broken param lists; share logic with args/argsOrIf
-  def typeParams(implicit fe: FoundErr, et: ExpectThen): Ls[(Opt[VarianceInfo], TypeName)] = {
+  def typeParams(implicit fe: FoundErr, et: ExpectThen): Ls[(TypeParamInfo, TypeName)] = {
+    val visinfo = yeetSpaces match {
+      case (KEYWORD("type"), l0) :: _ =>
+        consume
+        (true, S(l0))
+      case _ => (false, N)
+    }
     val vinfo = yeetSpaces match {
```
LPTK, line 1291: Why did you make this a tuple of options? You just made some illegal states representable...

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -1071,15 +1081,17 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], newDefs: Bo
...
       case _ =>
-        vinfo match {
-          case S((_, loc)) =>
+        (visinfo, vinfo) match {
+          case (_, (S(_), S(loc))) =>
             err(msg"dangling variance information" -> S(loc) :: Nil)
-          case N =>
+          case ((true, S(loc)), _) =>
+            err(msg"dangling visible type member" -> S(loc) :: Nil)
+          case _ =>
```
LPTK: Don't use a default case. List all possibilities.

shared/src/main/scala/mlscript/NuTypeDefs.scala
```diff
@@ -945,7 +946,7 @@ class NuTypeDefs extends ConstraintSolver { self: Typer =>
     }
     
     lazy val explicitVariances: VarianceStore =
-      MutMap.from(tparams.iterator.map(tp => tp._2 -> tp._3.getOrElse(VarianceInfo.in)))
+      MutMap.from(tparams.iterator.map(tp => tp._2 -> tp._3.getVarOr(VarianceInfo.in)))
```
LPTK: This is unnecessary. Just use `.varinfo.getOrElse`

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -36,7 +36,7 @@ class TypeDefs extends NuTypeDefs { Typer: Typer =>
   case class TypeDef(
     kind: TypeDefKind,
     nme: TypeName,
-    tparamsargs: List[(TypeName, TypeVariable)],
+    tparamsargs: List[(TypeName, TypeVariable)],  // TODO add type member mark
```
LPTK: What does the TODO mean?

shared/src/main/scala/mlscript/TypeDefs.scala
```diff
@@ -343,7 +345,8 @@ class TypeDefs extends NuTypeDefs { Typer: Typer =>
                 case _ =>
                   val fields = fieldsOf(td.bodyTy, paramTags = true)
                   val tparamTags = td.tparamsargs.map { case (tp, tv) =>
-                    tparamField(td.nme, tp) -> FieldType(Some(tv), tv)(tv.prov) }
+                    // TODO fix type member name mangling
```
LPTK: What needs to be fixed? Why did you not fix it?

shared/src/main/scala/mlscript/TypeSimplifier.scala
```diff
@@ -245,7 +245,7 @@ trait TypeSimplifier { self: Typer =>
                 
                 // * Reconstruct a TypeRef from its current structural components
                 val typeRef = TypeRef(td.nme, td.tparamsargs.zipWithIndex.map { case ((tp, tv), tpidx) =>
-                  val fieldTagNme = tparamField(clsTyNme, tp)
+                  val fieldTagNme = tparamField(clsTyNme, tp, false) // default to `C#A`
```
LPTK: The use of `false` seems wrong. Just use the correct info here.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -16,7 +16,7 @@ import mlscript.Message._
  */
 class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val newDefs: Bool)
     extends ucs.Desugarer with TypeSimplifier {
-  
+
```
LPTK: Remove whitespace change.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1495,32 +1525,44 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
...
-                case Cls => val t = clsNameToNomTag(td)(tp(pat.toLoc, "class pattern"), ctx); t -> t
-                case Trt => val t = trtNameToNomTag(td)(tp(pat.toLoc, "trait pattern"), ctx); t -> t
+                case Als | Mod | Mxn => val t = err(msg"can only match on classes and traits", pat.toLoc)(raise); (t, t, t)
+                case Cls => val t = clsNameToNomTag(td)(tp(pat.toLoc, "class pattern"), ctx); (t, t, t)
+                case Trt => val t = trtNameToNomTag(td)(tp(pat.toLoc, "trait pattern"), ctx); (t, t, t)
               }
           }
       }
-      val newCtx = ctx.nest
+      val newCtx = ctx.nest // TODO refactor
```
LPTK: Note: we probably want to refactor this. (I'll do it)

shared/src/main/scala/mlscript/TyperHelpers.scala
```diff
@@ -1150,7 +1150,8 @@ abstract class TyperHelpers { Typer: Typer =>
       lazy val tparamTags =
         if (paramTags) RecordType.mk(td.tparamsargs.map { case (tp, tv) =>
             val tvv = td.getVariancesOrDefault
-            tparamField(defn, tp) -> FieldType(
+            // default to `C#A`
```
LPTK: Why? Just use the correct value.

shared/src/test/diff/nu/RawTypes.mls
```diff
@@ -0,0 +1,121 @@
...
+//│ ║  l.34: 	fun foo(f: Foo[?]) = f.a : f.A
+//│ ╙──      	               ^
+//│ fun foo: forall 'A. (f: Foo['A]) -> 'A
+
+foo(Foo(1))
+//│ 1
+//│ res
+//│     = 1
+
+// FIXME
```
LPTK: Explain why this is a FIXME.

LPTK (changes requested):
There are large inferred type regressions, for example in `ucs/NestedBranches.mls`, `SimpleRegionDSL_raw.mls`, and `JSON.mls`. Did you figure out why? Could you please give a high-level intuituion?
**EDIT**: also `NuUntyped.mls`, `NuSimplyTyped.mls`

We either need to solve these or add a GADT flag, so these don't just happen by default.

LPTK:
So, we need to fix the problems in https://github.com/Meowcolm024/mlscript/tree/local-gadt-wip

Looking at one example: the error you get [here](https://github.com/Meowcolm024/mlscript/blob/9c5e2c5795fbeb3f3c405c57bff58f676bb19b85/shared/src/test/diff/gadt/GADT1.mls#L16-L27) is **correct**:

```scala
abstract class Vec[type L, type T]: Nil[T] | Cons[S, T]
class Nil[T] extends Vec[Z, T]
class Cons[L, T](val h: T, val t: Vec[L, T]) extends Vec[S[L], T]
//│ ╔══[ERROR] Type mismatch in type declaration:
//│ ║  l.18: 	class Cons[L, T](val h: T, val t: Vec[L, T]) extends Vec[S[L], T]
//│ ║        	^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//│ ╟── type `L` is not an instance of type `S`
//│ ║  l.18: 	class Cons[L, T](val h: T, val t: Vec[L, T]) extends Vec[S[L], T]
//│ ║        	           ^
//│ ╟── Note: constraint arises from class tag:
//│ ║  l.16: 	abstract class Vec[type L, type T]: Nil[T] | Cons[S, T]
//│ ╙──      	                      
```

`Cons[L, T]` is **indeed** not a `Cons[S, T]` because `L` is not the same type as `S`.

What you really want to write here is `: Nil[T] | Cons[?, T]`. So it seems you need to add support for wildcards.

On the other hand, I think `: Nil | Cons` should *in principle* work, but it currently does not (the map function yields an error) and we could try to find out why.

**EDIT to add:**
For the record, the reason `map` doesn't work with the `: Nil | Cons` self-annotation is that we infer scrutinee type `#Nil | #Cons & { Cons#L = ?L0..?L1; Cons#T = ?T0..?T1 }` with `?T0 <: ?T1 <: ‘A | ~⊥(??T)`, which leads to derived constraint `Vec[A, B] <: {Cons#T}`, which yields an extruded `??T` for the `Cons#T` field, from the `Cons` raw type part of the annotation.
Note: the `⊥(??T)` is really weird; feels like it should be `⊤(??T)` – perhaps a pretty-printing issue — EDIT actually no, a plain old bug https://github.com/hkust-taco/mlscript/commit/abe5e62d891fb3844d96c87135fff64fa1ecc19f
Note: at least in situations like these, we could maybe see that the `Cons#T` refinement in this case could be turned into an equivalent `Vec#T` refinement?

LPTK:
See: https://github.com/LPTK/mlscript/commit/2f193c87bbe7d70c9da1d4cb793ee6471a091416

The goal is to allow:

```scala
abstract class Vec[type L, type T]: Nil[T] | Cons[?, T]
class Nil[T] extends Vec[Z, T]
class Cons[L, T](val h: T, val t: Vec[L, T]) extends Vec[S[L], T]
```

And to elaborate on the previous message, this is what happens if we use `: Nil | Cons`:
* we get constraint `Vec['L, 'A] <: #Nil | #Cons & { Cons#T: ..'A }` for the scrutinee
* ie `#Vec & { L = 'L; T = 'A } & (#Nil | #Cons) <: #Nil | #Cons & { Cons#T: ..'A }`
* leading to constraint `#Cons & { L = 'L; T = 'A } <: #Cons & { Cons#T: ..'A }`
* leading to looking up `Cons#T` in `#Cons`, which yields an extruded skolem not compatible with `'A`

LPTK:
So the way type selections/projections are typed is currently broken, as it falls foul of the very same problem they had in Sala. Here's an example:

```scala
class Wrap[type A]
class Test[type B]
//│ class Wrap[A] {
//│   constructor()
//│ }
//│ class Test[B] {
//│   constructor()
//│ }

fun oops(w: Wrap[Test[Int] & Test[Str]]) = 1 : w.A.B
//│ fun oops: (w: Wrap[in Test[out Int | Str] out Test[in Int | Str out nothing]]) -> nothing

oops(new Wrap)
//│ nothing
//│ res
//│     = 1
```

And here's a Scala 3 issue that's similar: https://github.com/lampepfl/dotty/issues/15578

The root of the issue is being able to use an abstract type that has bad bounds as though it really existed (had a valid instance), whereas we have no evidence that it does. Indeed, the only guarantee we have is that the types of _values_ in the context (and their `val` fields) have good bounds. Types nested in _types_ in the context have no such guarantees.

This means the Vector-tail example is broken. We could solve this by adding a way of doing inversion reasoning, or by some other technique.

shared/src/test/diff/nu/NuScratch.mls
```diff
@@ -1,3 +1,302 @@
 :NewDefs
 
+class Ty[type T]
```
LPTK: We need to remember to clean this file before merging.

shared/src/main/scala/mlscript/syntax.scala
```diff
@@ -90,6 +90,7 @@ final case class Forall(params: Ls[TypeVar], body: Term)             extends Ter
 final case class Inst(body: Term)                                    extends Term
 final case class Super()                                             extends Term
 final case class Eqn(lhs: Var, rhs: Term)                            extends Term // equations such as x = y, notably used in constructors; TODO: make lhs a Term
+final case class WildcardType()                                      extends Term // TODO use Var("?") instead
```
LPTK: Why? What does this do?

Meowcolm024: It seems that  `WildcardType` actually does nothing, so I decide to remove it (just using `Var("?")` should be fine.

LPTK: But it's still there...

Meowcolm024: changes not pushed yet

shared/src/main/scala/mlscript/syntax.scala
Meowcolm024: removed

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -648,7 +648,7 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], newDefs: Bo
         exprCont(Super().withLoc(S(l0)), prec, allowNewlines = false)
       case (IDENT("?", true), l0) :: _ =>
```
LPTK, line 773: Why do you need a special case for this?

shared/src/test/diff/gadt/Wildcard.mls
```diff
@@ -411,42 +420,77 @@ if x is ? then 1
...
 //│   Invalid or unexpected token
 
+:e
+type ? = Int
+//│ ╔══[ERROR] Type name '?' is reserved
+//│ ║  l.460: 	type ? = Int
+//│ ╙──       	^^^^^^
+//│ type ? = Int
+
+// TODO should we allow this
```
LPTK: We should not. `?` is only for class/trait type argument positions.

shared/src/main/scala/mlscript/ConstraintSolver.scala
LPTK: You misunderstood me. I'm not saying you SHOULD crash. Do you want your compiler to crash when it does not find a type member or do you want it to produce a proper error??

shared/src/main/scala/mlscript/TyperHelpers.scala
```diff
@@ -1287,8 +1285,8 @@ abstract class TyperHelpers { Typer: Typer =>
         if (pol =/= S(true)) apply(S(false))(lb)
         if (pol =/= S(false)) apply(S(true))(ub)
       case WildcardArg(lb, ub) =>
-        if (pol =/= S(true)) apply(S(false))(lb)
-        if (pol =/= S(false)) apply(S(true))(ub)
+         apply(pol.map(!_))(lb);
```
LPTK: Why the weird indentation and semicolon? Please try to be more careful.

shared/src/main/scala/mlscript/NewParser.scala
Meowcolm024: removing it will cause unexpected operator in parsing

shared/src/main/scala/mlscript/ConstraintSolver.scala
Meowcolm024: fixed crash

shared/src/main/scala/mlscript/NuTypeDefs.scala
```diff
@@ -1446,12 +1446,15 @@ class NuTypeDefs extends ConstraintSolver { self: Typer =>
                       case _ =>
                     }
                     
-                    val ttu = typeTypingUnit(td.body, S(td))
+                    val ttu = typeTypingUnit(td.body, S(td))
```
```suggestion
                    val ttu = typeTypingUnit(td.body, S(td))
```

shared/src/test/diff/gadt/AliasGADT.mls
```diff
@@ -5,31 +5,54 @@ type Test[T] = LitInt | LitStr
...
+//│ ║        	                 ^
+//│ ╟── type `Str` does not match type `T`
+//│ ║  l.6: 	class LitStr(s: Str) { type T = Str }
+//│ ║       	                                ^^^
+//│ ╟── Note: constraint arises from method type parameter:
+//│ ║  l.17: 	fun foo[T](x: Test[T]): T = if x is
+//│ ╙──      	        ^
+//│ fun foo: forall 'T. (x: Test['T]) -> 'T
 
+// * somehow we need to find a way to associate T in Foo and T in MkF
```
LPTK, line 40: Right. It should probably be `type Foo[T0] = MkF & { type T = T0 }`.

(Or in the future, using a type refinement syntax, `MkF { T = T0 }`.)

LPTK: Problem is we currently don't support parsing `{ type T = T0 }`.

LPTK: So maybe let's not try to implement this in the current PR. It would require too many additional changes/features.

shared/src/test/diff/nu/TypeMembers.mls
```diff
@@ -0,0 +1,64 @@
...
+trait Test { type T = Int }
+//│ trait Test
+
+1 : Test.T
+//│ Int
+//│ res
+//│     = 1
+
+class Foo[T](x: T) { 
+  type T = T
```
LPTK, line 25: Doesn't `T` refer to itself here? 🤔

shared/src/main/scala/mlscript/NewLexer.scala
```diff
@@ -442,7 +442,8 @@ object NewLexer {
     "undefined",
     "abstract",
     "constructor",
-    "virtual"
+    "virtual",
+    "restricts"
```
LPTK:
```suggestion
    "restricts",
```

shared/src/test/diff/gadt/AliasGADT.mls
```diff
@@ -5,31 +5,54 @@ type Test[T] = LitInt | LitStr
 class LitInt(n: Int) { type T = Int }
 class LitStr(s: Str) { type T = Str }
 //│ type Test[T] = LitInt | LitStr
-//│ class LitInt(n: Int) {
-//│   type T = Int
```
LPTK: TODO: remember to make this pretty-printed again.

shared/src/main/scala/mlscript/TypeSimplifier.scala
Meowcolm024: I think these are in the logic of using the old defs (`TypeDef`), which doesn't support type members. So it should probably be fine here 🤔

LPTK: Ah, then please add a comment why you pick false here.

LPTK:
So regarding https://github.com/hkust-taco/mlscript/pull/195#pullrequestreview-1805379488, I minimized the regression to:

```scala
type List[out A] = Cons[A]
module Nil
class Cons[A](head: A, tail: List[A])
//│ type List[A] = Cons[A]
//│ module Nil
//│ class Cons[A](head: A, tail: List[A])

fun listConcat(xs) =
  if xs is
    Cons(x, xs2) then Cons(x, listConcat(xs2))
//│ fun listConcat: forall 'A 'A0 'A1. Cons['A & ('A0 | ~??A)] -> Cons[in 'A0 out 'A1]
//│   where
//│     'A0 :> 'A1 | ??A & 'A
//│     'A1 :> 'A0 | ??A & 'A
```
Previously, we didn't have these duplicated bounds.

I confirmed that they are simply due to extruding things several times. When we either cache the extrusion context or don't increase the level of patmat branches, then we get something like the original inferred type back.

Caching the extrusion context:
```scala
//│ fun listConcat: forall 'A 'A0. Cons['A & ('A0 | ~??A)] -> Cons['A0]
//│   where
//│     'A0 :> ??A & 'A
```

Not increasing the level of patmat branches:
```scala
//│ fun listConcat: forall 'A 'A0. Cons['A] -> Cons['A0]
//│   where
//│     'A0 :> ??A & 'A
```

So we should cache the extrusion context. But to do this properly, we should only cache extrusions _to the same level_, instead of having one single cache that's used for extrusions to any level, as is currently implemented.

shared/src/test/diff/nu/InferredInheritanceTypeArgs.mls
```diff
@@ -174,12 +174,8 @@ B.foo
 
 
 // * TODO: when
```
LPTK: I guess we can remove the TODO.

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -1147,13 +1147,13 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], newDefs: Bo
         }
         val lb = getTypeName("restricts")
         val ub = getTypeName("extends")
-
+        // TODO update `TypeParamInfo` to use lb and ub
```
LPTK, line 1320: TODO

shared/src/main/scala/mlscript/NormalForms.scala
```diff
@@ -241,7 +241,14 @@ class NormalForms extends TyperDatatypes { self: Typer =>
...
         b2.forall(b2 => b1.exists(_ <:< b2)) &&
-          ts2.forall(ts1) && rt1 <:< rt2 &&
+          ts2.forall {
+            case sk: SkolemTag => ts1(sk)
+            case tt: TraitTag => ts1(tt)
+            case Extruded(pol, sk) => !pol || ts1.exists { // find ? <: bot
+              case Extruded(true, _) => true
+              case _ => false
+            }
+          } && rt1 <:< rt2 &&
```
LPTK: This thing looks very fishy. What is going on here? Why don't we use the normal set functions?

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -332,6 +332,21 @@ class ConstraintSolver extends NormalForms { self: Typer =>
...
       }()
+
+      // * finish when a type appears at both sides
+      val both = lhsCs.exists(s => s.lnf match {
+        case LhsTop => false
+        case LhsRefined(lbs, ltt, lr, ltr) =>
+          rhs.cs.exists(c => c.lnf match {
+            case LhsTop => false
+            case LhsRefined(rbs, rtt, rr, rtr) => 
+              !ltt.intersect(rtt).collect{ case sk: SkolemTag => sk }.isEmpty
```
LPTK: Why not like this? It's much more efficient.
```suggestion
              ltt.exists(rtt.contains)
```
Also I think this could be generalized to all abstract tags, not just skolem tags, no?

shared/src/test/diff/gadt/GADT5.mls
```diff
@@ -157,22 +157,11 @@ fun g[T](x: F[T, T]): T = if x is
 
 // * We used to be able to manually refine it
 // * but this now fails... not sure why
```
LPTK: Comment is outdated.

shared/src/test/diff/gadt/Misc.mls
```diff
@@ -300,18 +299,75 @@ u(MkU(1))
...
 fun u[T](x: U[T, 'b]): T = if x is
-    MkU(s) then s : x.A
-//│ ╔══[ERROR] Type mismatch in `case` expression:
-//│ ║  l.304: 	fun u[T](x: U[T, 'b]): T = if x is
-//│ ║         	                              ^^^^
-//│ ║  l.305: 	    MkU(s) then s : x.A
-//│ ║         	^^^^^^^^^^^^^^^^^
-//│ ╙── expression of type `??S & ~T & ~??S0` does not match type `nothing`
-//│ fun u: forall 'T 'b. (x: U['T, 'b]) -> (error | 'T)
+    MkU(s) then s :x.A
```
LPTK:
```suggestion
    MkU(s) then s : x.A
```

shared/src/test/diff/gadt/Misc.mls
```diff
@@ -300,18 +299,75 @@ u(MkU(1))
...
+//│ Int | error
+//│ res
+//│     = true
+
+abstract class U[type A, type B]: MkU
+class MkU[S](x: S) extends U[S, Int]
+//│ abstract class U[A, B]: MkU[anything]
+//│ class MkU[S](x: S) extends U
+
+// TODO no variable leaks now ?
```
LPTK: Not sure what the TODO mean here. Elaborate?

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -527,11 +527,11 @@ class ConstraintSolver extends NormalForms { self: Typer =>
             case (LhsRefined(_, ts, _, trs), RhsBases(pts, _, _)) if ts.exists(pts.contains) => ()
             
             case (LhsRefined(bo, ts, r, trs), _) if trs.nonEmpty =>
-              annoying(trs.valuesIterator.map { tr => tr.expand }.toList,
+              annoying(trs.valuesIterator.map { tr => tr.expand(true) }.toList,
                 LhsRefined(bo, ts, r, SortedMap.empty), Nil, done_rs)
             
             case (_, RhsBases(pts, bf, trs)) if trs.nonEmpty =>
-              annoying(Nil, done_ls, trs.valuesIterator.map(_.expand).toList, RhsBases(pts, bf, SortedMap.empty))
+              annoying(Nil, done_ls, trs.valuesIterator.map(_.expand(true)).toList, RhsBases(pts, bf, SortedMap.empty))
```
LPTK: This should be `false`... The RHS is on the right-hand side, a negative type.

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -992,22 +992,22 @@ class ConstraintSolver extends NormalForms { self: Typer =>
                   }
               }
             } else {
-              if (tr1.mayHaveTransitiveSelfType) rec(tr1.expand, tr2.expand, true)
+              if (tr1.mayHaveTransitiveSelfType) rec(tr1.expand(true), tr2.expand(true), true)
```
LPTK: Why would you use `true` everywhere? It's true/false here...

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -992,22 +992,22 @@ class ConstraintSolver extends NormalForms { self: Typer =>
...
               else (tr1.mkClsTag, tr2.mkClsTag) match {
                 case (S(tag1), S(tag2)) if !(tag1 <:< tag2) =>
                   reportError()
                 case _ =>
-                  rec(tr1.expand, tr2.expand, true)
+                  rec(tr1.expand(true), tr2.expand(true), true)
```
LPTK: Be more careful. true/false again here.

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -992,22 +992,22 @@ class ConstraintSolver extends NormalForms { self: Typer =>
...
-          case (tr: TypeRef, _) => rec(tr.expand, rhs, true)
+          case (tr: TypeRef, _) => rec(tr.expand(true), rhs, true)
           case (err @ ClassTag(ErrTypeId, _), tr: TypeRef) =>
             // rec(tr.copy(targs = tr.targs.map(_ => err))(noProv), tr, true)
             // * ^ Nicely propagates more errors to the result,
             // * but can incur vast amounts of unnecessary constraining in the context of recursive types!
             ()
           case (_, tr: TypeRef) =>
-            rec(lhs, tr.expand, true)
+            rec(lhs, tr.expand(true), true)
```
LPTK: false

shared/src/main/scala/mlscript/NuTypeDefs.scala
```diff
@@ -585,7 +585,7 @@ class NuTypeDefs extends ConstraintSolver { self: Typer =>
                 fd.copy()(fd.declareLoc, fd.virtualLoc, fd.mutLoc, fd.signature, outer, fd.genField)
             }
           case td: NuTypeDef =>
-            if (td.nme.name in reservedTypeNames)
+            if (td.nme.name.in(reservedTypeNames) || td.nme.name === "?")
```
LPTK: Why not put "?" in `reservedTypeNames` instead?

shared/src/main/scala/mlscript/TypeSimplifier.scala
```diff
@@ -92,7 +92,7 @@ trait TypeSimplifier { self: Typer =>
       case ProvType(ty) => process(ty, parent, canDistribForall = canDistribForall)
       
       case tr @ TypeRef(defn, targs) if builtinTypes.contains(defn) && tr.canExpand =>
-        process(tr.expandOrCrash, parent)
+        process(tr.expandOrCrash(true), parent)
```
LPTK, line 99: This will be incorrect if the type has type parameters.
```suggestion
        assert(tr.targs.isEmpty)
        process(tr.expandOrCrash(true), parent)
```

shared/src/main/scala/mlscript/TypeSimplifier.scala
```diff
@@ -268,7 +268,7 @@ trait TypeSimplifier { self: Typer =>
                 })(noProv)
                 println(s"typeRef ${typeRef}")
                 
-                val clsFields = fieldsOf(typeRef.expandWith(paramTags = true, selfTy = false), paramTags = true)
+                val clsFields = fieldsOf(typeRef.expandWith(paramTags = true, selfTy = false, pol.getOrElse(true)), paramTags = true)
```
LPTK: I don't know how to fix it yet but this is definitely wrong.
```suggestion
                val clsFields = fieldsOf(typeRef.expandWith(paramTags = true, selfTy = false, pol.getOrElse(die/*TODO*/)), paramTags = true)
```

shared/src/main/scala/mlscript/TypeSimplifier.scala
```diff
@@ -356,7 +356,7 @@ trait TypeSimplifier { self: Typer =>
                 })(noProv)
                 println(s"typeRef ${typeRef}")
                 
-                val clsFields = fieldsOf(typeRef.expandWith(paramTags = true, selfTy = false), paramTags = true)
+                val clsFields = fieldsOf(typeRef.expandWith(paramTags = true, selfTy = false, pol = pol.getOrElse(true)), paramTags = true)
```
LPTK: Same.

shared/src/main/scala/mlscript/TyperHelpers.scala
```diff
@@ -554,9 +554,9 @@ abstract class TyperHelpers { Typer: Typer =>
         case (_, NegType(und)) => (this & und) <:< BotType
         case (NegType(und), _) => TopType <:< (that | und)
         case (tr: TypeRef, _)
-          if (primitiveTypes contains tr.defn.name) && tr.canExpand => tr.expandOrCrash <:< that
+          if (primitiveTypes contains tr.defn.name) && tr.canExpand => tr.expandOrCrash(true) <:< that
         case (_, tr: TypeRef)
-          if (primitiveTypes contains tr.defn.name) && tr.canExpand => this <:< tr.expandOrCrash
+          if (primitiveTypes contains tr.defn.name) && tr.canExpand => this <:< tr.expandOrCrash(true)
```
LPTK: false

shared/src/main/scala/mlscript/TyperHelpers.scala
```diff
@@ -638,7 +638,7 @@ abstract class TyperHelpers { Typer: Typer =>
       case ct: ConstrainedType => ct
     }
     def unwrapAll(implicit ctx: Ctx): SimpleType = unwrapProxies match {
-      case tr: TypeRef if tr.canExpand => tr.expandOrCrash.unwrapAll
+      case tr: TypeRef if tr.canExpand => tr.expandOrCrash(true).unwrapAll
```
LPTK: Wrong!

shared/src/main/scala/mlscript/TyperHelpers.scala
```diff
@@ -647,7 +647,7 @@ abstract class TyperHelpers { Typer: Typer =>
       case ComposedType(true, l, r) => l.negNormPos(f, p) & r.negNormPos(f, p)
       case ComposedType(false, l, r) => l.negNormPos(f, p) | r.negNormPos(f, p)
       case NegType(n) => f(n).withProv(p)
-      case tr: TypeRef if !preserveTypeRefs && tr.canExpand => tr.expandOrCrash.negNormPos(f, p)
+      case tr: TypeRef if !preserveTypeRefs && tr.canExpand => tr.expandOrCrash(true).negNormPos(f, p)
```
LPTK: Wrong!

shared/src/main/scala/mlscript/TyperHelpers.scala
```diff
@@ -1052,7 +1052,7 @@ abstract class TyperHelpers { Typer: Typer =>
           println(s"  where: ${res.showBounds}")
           if (cannotBeDistribbed.isEmpty) S(res)
           else S(PolymorphicType(polymLevel, res))
-        case tr: TypeRef if !traversed.contains(tr.defn) => go(tr.expand, traversed + tr.defn, polymLevel)
+        case tr: TypeRef if !traversed.contains(tr.defn) => go(tr.expand(true), traversed + tr.defn, polymLevel)
```
LPTK: Wrong!

shared/src/main/scala/mlscript/TyperHelpers.scala
```diff
@@ -1077,8 +1077,8 @@ abstract class TyperHelpers { Typer: Typer =>
         // * Object types do not need to be completed in order to be expanded
         info.kind.isInstanceOf[ObjDefKind]
         || info.isComputed)
-    // TODO add pol: Bool
-    def expand(implicit ctx: Ctx, raise: Raise): SimpleType = {
+    // TODO provide correct pol at each expand call (we're assuming `pol = true` now)
```
LPTK: > (we're assuming `pol = true` now)

Just saw that. Really not a good idea as you'll forget to fix most of the places you put the wrong value...

shared/src/main/scala/mlscript/TyperHelpers.scala
```diff
@@ -1110,14 +1110,15 @@ abstract class TyperHelpers { Typer: Typer =>
             assert(td.tparams.size === targs.size)
             subst(td.body, td.tparams.lazyZip(targs).map {
               case (tp, w: WildcardArg) => 
-                // TODO raise type error here
-                SkolemTag(tp._2)(noProv) -> errType
+                // * FIXME we should NOT use TypeBounds here (somehow causes infinite subtyping loop)
+                // *       but raising an error will break a lot of  code related to alias (maybe changling `subst`)
```
LPTK: This should be an error. We must make sure type aliases never get wildcard args.

shared/src/main/scala/mlscript/TyperHelpers.scala
```diff
@@ -1478,7 +1479,7 @@ abstract class TyperHelpers { Typer: Typer =>
       def go(ty: ST, traversed: Set[AnyRef]): Opt[PolymorphicType] = //trace(s"go $ty") {
           if (!distributeForalls) N else ty match {
         case poly @ PolymorphicType(plvl, bod) => S(poly)
-        case tr: TypeRef if !traversed.contains(tr.defn) => go(tr.expand, traversed + tr.defn)
+        case tr: TypeRef if !traversed.contains(tr.defn) => go(tr.expand(true), traversed + tr.defn)
```
LPTK: wrong

shared/src/main/scala/mlscript/TyperHelpers.scala
```diff
@@ -1500,7 +1501,7 @@ abstract class TyperHelpers { Typer: Typer =>
   object AliasOf {
     def unapply(ty: ST)(implicit ctx: Ctx): S[ST] = {
       def go(ty: ST, traversedVars: Set[TV]): S[ST] = ty match {
-        case tr: TypeRef if tr.canExpand => go(tr.expandOrCrash, traversedVars)
+        case tr: TypeRef if tr.canExpand => go(tr.expandOrCrash(true), traversedVars)
```
LPTK: wrong

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -521,6 +521,7 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
         }
       case tn @ TypeTag(name) => rec(TypeName(name.decapitalize)) // TODO rm this hack
       // case tn @ TypeTag(name) => rec(TypeName(name))
+      case wc @ TypeName("?") => err(msg"Invalid use of ? here", wc.toLoc)
```
```suggestion
      case wc @ TypeName("?") => err(msg"Invalid use of `?` here", wc.toLoc)
```

shared/src/main/scala/mlscript/TyperHelpers.scala
Meowcolm024: it seems that wildcards can get into type aliases via extrusion currently

shared/src/main/scala/mlscript/TyperHelpers.scala
Meowcolm024: now all fixed...

Meowcolm024: now all fixed...

shared/src/main/scala/mlscript/ConstraintSolver.scala
Meowcolm024: maybe not? class tags have type members associated with and need to be compared?

shared/src/test/diff/gadt/Misc.mls
Meowcolm024: there was a type variable leak here before we early finish the constrain solving for DNF (when we found skolem at both sides)

shared/src/main/scala/mlscript/TyperHelpers.scala
LPTK: Well extrusion must be changed so this doesn't happen. If the thing is a type alias, its type arguments can use the old extrusion approach (TypeBounds).

shared/src/test/diff/gadt/Misc.mls
LPTK: Ok then we can remove the TODO?

shared/src/main/scala/mlscript/ConstraintSolver.scala
LPTK: A class tag represents the type of a class instance whose type members are unknown. If you find the same one on both sides of a constraint, you can obviously finish the constraint right there.

shared/src/main/scala/mlscript/TyperHelpers.scala
Meowcolm024: done

shared/src/main/scala/mlscript/ConstraintSolver.scala
Meowcolm024: okay

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -1426,8 +1426,11 @@ class ConstraintSolver extends NormalForms { self: Typer =>
             WildcardArg(extrude(lb, lowerLvl, false, upperLvl), extrude(ub, lowerLvl, true, upperLvl))(tr.prov)
           case (N, targ: ST) =>
             // * Note: the semantics of TypeBounds is inappropriuate for this use (known problem; FIXME later)
-            // TypeBounds.mk(extrude(targ, lowerLvl, false, upperLvl), extrude(targ, lowerLvl, true, upperLvl)) // Q: ? subtypes?
-            WildcardArg(extrude(targ, lowerLvl, false, upperLvl), extrude(targ, lowerLvl, true, upperLvl))(tr.prov)
+            // * use TypeBounds for type alias
+            if (ctx.tyDefs2.get(d.name).map(_.decl.kind === Als).getOrElse(false))
```
LPTK: The correct thing to do is to assume we can't youse wildcards, to be on the safe side.
```suggestion
            if (ctx.tyDefs2.get(d.name).map(_.decl.kind === Als).getOrElse(true))
```

Meowcolm024: Fixed

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -46,6 +46,37 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
   
   type GenLambdas >: Bool
   def doGenLambdas(implicit gl: GenLambdas): Bool = gl === true
+
+  class ExtrCache(
+    val cache: MutMap[Level, MutMap[TypeVarOrRigidVar->Bool, TypeVarOrRigidVar]],
+  ) {
+      private var cycle: Int = 0
+      // TODO maybe find a better way
```
LPTK, line 54: Could you explain what this achieves and why it's needed?

LPTK:
Thanks again for the great work, @Meowcolm024! These changes will be useful for future reference, even though they are already obsolete (since we are moving to a new type checker implementation). So I will close this for now as it is not worth fixing the conflicts and cleaning up the history, similarly to https://github.com/hkust-taco/mlscript/pull/219

By the way, as I told you in person some time ago, you merged some commits on my WIP branch (the "W"-named commits) which were only temporary experimentation changes. These were not intended to be made part of your PR – or at least, not as is.

PR #196, opened 2023-12-07
Defunctionalization with base AST

HarrisL2:
In-progress refactoring of the previous defunctionalization code which used its own IR. This implementation uses the base AST instead and can be fed back into the typer.

Points of Interest:
Added new content to `shared/src/main/scala/mlscript/syntax.scala` to represent the bounds of evaluating a term
Added `var BoundedTerm` to `TermImpl` to store evaluation result
Added new flags `:nm`, `:lift` that allow the transformed AST to be used in the following stages
Use flag `:dd` to debug the defunctionalization process

`compiler/shared/test/diff/NuMono/Differentiation.mls`
A use case test that exposes many bugs with in the implementation, being worked on right now.


compiler/shared/test/diff/NuMono/Differentiation.mls
```diff
@@ -0,0 +1,92 @@
...
+//│   class Sum$4([lhs: Exp$1, rhs: Exp$1,]): Exp$1() {}
+//│   class Mul$5([lhs: Exp$1, rhs: Exp$1,]): Exp$1() {}
+//│   class Variable$3([nm: Str,]): Exp$1() {}
+//│   class Exp$1([]) {}
+//│   fun derive$Sum$4 = (this,) => Sum$4(let obj = let obj = this in if obj is ‹(Sum$4(lhs, rhs,)) then lhs› in if obj is ‹(Variable$3) then derive$Variable$3(obj,)›, let obj = let obj = this in if obj is ‹(Sum$4(lhs, rhs,)) then rhs› in if obj is ‹(Variable$3) then derive$Variable$3(obj,)›,)
+//│   fun main$$5 = () => let obj = Sum$4(Variable$3("y",), Variable$3("x",),) in if obj is ‹(Sum$4) then derive$Sum$4(obj,)›
+//│   fun derive$Variable$3 = (this,) => Numeric$2(1,)
+//│   Code(List(main$$5()))
+//│ }
+//│ ╔══[ERROR] Type mismatch in `case` expression:
```
HarrisL2: The generated pattern match of defunctionalization doesn't pass the type check because it doesn't match for all possible types of `Exp`. However in the full program no `Exp` flows into the pattern match other than `Variable`.
Should we:
\- Ignore the typing error?
\- Generate all possible cases?

compiler/shared/main/scala/mlscript/compiler/mono/Monomorph.scala
```diff
@@ -249,7 +250,7 @@ class Monomorph(debug: Debug = DummyDebug) extends DataTypeInferer:
 
   private def nuUpdateFunction(funcName: String): Unit = {
     val updateCount = nuEvalCnt.get(funcName).getOrElse(0)
-    if(updateCount <= 10){
+    if(updateCount <= 3){
```
LPTK: So the stack depth limit is... 3?

compiler/shared/test/diff/NuMono/Differentiation.mls
LPTK: For now I think it would be better to generate a catch-all case that throws an "unreachable" exception.

HarrisL2: Added an `else error` branch. Is there a way to throw exceptions with more information?

compiler/shared/main/scala/mlscript/compiler/mono/Monomorph.scala
HarrisL2: Was a temporary change while debugging a recursion bug.

compiler/shared/main/scala/mlscript/compiler/mono/Monomorph.scala
```diff
@@ -22,7 +22,7 @@ import mlscript.compiler.printer.ExprPrinter
 import mlscript.compiler.mono.specializer.BoundedExpr
 import mlscript.compiler.mono.specializer.{MonoValue, TypeValue, ObjectValue, UnknownValue, FunctionValue, VariableValue}
 import mlscript.{MonoVal, TypeVal, ObjVal, FuncVal, LiteralVal, PrimVal, VarVal, TupVal, UnknownVal, BoundedTerm}
-
+
```
LPTK: ??

LPTK: It's good practice to review your diffs before pushing your commits.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -1083,7 +1083,7 @@ trait MonoValImpl { self: MonoVal =>
     case UnknownVal() =>  s"UnknownVal"
     case PrimVal() => s"PrimVal()"
     case VarVal(vx, version, _) => s"VarVal(${vx})"
-  }
+      }
```
LPTK: ??

compiler/shared/main/scala/mlscript/compiler/mono/Monomorph.scala
HarrisL2: Noted. I've turned on show whitespace diffs in my IDE.

shared/src/main/scala/mlscript/helpers.scala
HarrisL2: Removed.

shared/src/test/diff/nu/NuScratch.mls
```diff
@@ -1,3 +1,12 @@
...
+class A() {
+    fun f = 6
+}
+A().f
+//│ class A() {
+//│   fun f: 6
+//│ }
+//│ 6
+//│ res
+//│     = 6
```
LPTK: This file should be clean.

compiler/shared/test/diff/scratch.mls
```diff
@@ -0,0 +1,40 @@
+:NewDefs
+
+:e // FIXME
```
LPTK: Address this or move this to some normal test file and document.

LPTK (commented):
Please address the items mentioned in the meeting. Notably, make sure all old test cases are ported and then delete them (and the old impl.).

compiler/shared/test/diff/Defunctionalize/OldMonoList.mls
```diff
@@ -0,0 +1,275 @@
...
+//│ fun main$$5: () -> Int
+//│ fun main$$4: () -> Int
+//│ Int
+//│ res
+//│     = 2
+//│ res
+//│ Runtime error:
+//│   RangeError: Maximum call stack size exceeded
+
+:e //FIXME: Class parameters not lifted
```
LPTK: Shouldn't this be fixed before we merge the PR?

compiler/shared/test/diff/Defunctionalize/Modules.mls
```diff
@@ -0,0 +1,40 @@
...
+//│ Mono:
+//│ TypingUnit {
+//│   module x$2([]) {}
+//│   class Foo$1([]) {}
+//│   let y$x$2 = (this,) => Foo$1()
+//│   fun f$Foo$1 = (this,) => 0
+//│   fun main$$2 = () => let obj = let obj = x$2() in if obj is ‹(x$2) then y$x$2(obj,); else error› in if obj is ‹(Foo$1) then f$Foo$1(obj,); else error›
+//│   Code(List(main$$2()))
+//│ }
+//│ ╔══[ERROR] Module parameters are not supported
```
LPTK: Why are these errors reported?

compiler/shared/test/diff/Defunctionalize/Modules.mls
```diff
@@ -0,0 +1,40 @@
+:NewDefs
+
+:e // FIXME: Change behaviour based on TypeDefKind
```
LPTK: Don't we want to fix this?

compiler/shared/test/diff/Defunctionalize/Differentiation.mls
```diff
@@ -0,0 +1,125 @@
+// A usecase test for defunctionalization
+
+:NewDefs
+
+:mono
+class Exp() {
```
LPTK: Why not make it an `abstract class`?

LPTK:
@waterlens can you please have a very quick look at whether you see anything unreasonable or broken? We want to merge this soon. It's temporary code anyway since the whole approach will be refactored. Probably the most interesting/important stuff about this PR is the test suite.

compiler/shared/test/diff/Defunctionalize/OldMonoList.mls
HarrisL2: Fixed.

compiler/shared/main/scala/mlscript/compiler/mono/specializer/Specializer.scala
```diff
@@ -7,7 +7,7 @@ import mlscript.compiler.mono.MonomorphError
 import mlscript.compiler.mono.Monomorph
 import mlscript.TypeName
 import mlscript.{App, Asc, Assign, Bind, Blk, Bra, CaseOf, Lam, Let, Lit,
-                 New, Rcd, Sel, Subs, Term, Test, Tup, With, Var, Fld, FldFlags, If, PolyType, 
+                 New, NuNew, Rcd, Sel, Subs, Term, Test, Tup, With, Var, Fld, FldFlags, If, PolyType,
```
LPTK: This whole slab of imports is really completely silly.

Just make the package declaration:

```
package mlscript
package compiler
package mono
```
(the `specializer` sub-package seems excessive)

Then you won't need any of these silly imports.

compiler/shared/main/scala/mlscript/compiler/ClassLifter.scala
```diff
@@ -299,17 +299,22 @@ class ClassLifter(logDebugMsg: Boolean = false) {
...
     target match {
     case v: Var => 
-      if(ctx.contains(v) || v.name.equals("this") || primiTypes.contains(v.name))   (v, emptyCtx)
+      if(globFuncs.contains(v)) {
+        (globFuncs.get(v).get)
+      }
       else if(cache.contains(TypeName(v.name))){
-        val ret = liftConstr(TypeName(v.name), Tup(Nil))
-        App(Var(ret._1.name), ret._2) -> ret._3
+        //val ret = liftConstr(TypeName(v.name), Tup(Nil))
```
waterlens: Maybe it would be better to not leave the old code in the comments because you can always find it from the history if necessary.

compiler/shared/main/scala/mlscript/compiler/mono/specializer/BoundedTerm.scala
```diff
@@ -0,0 +1,186 @@
...
+      val (restVals1, mergingVals1) = splitSpecifiedObjects(self.values, mergingValNms)
+      val (restVals2, mergingVals2) = splitSpecifiedObjects(other.values, mergingValNms)
+
+      val ret = mergingValNms.map(nm => (mergingVals1.get(nm), mergingVals2.get(nm)) match {
+        case (Some(x1: ObjVal), Some(x2: ObjVal)) => x1.merge(x2)(instackExps ++ Set(self.hashCode(), other.hashCode()))
+        case _ => ???
+      })
+
+      var ret2 = restVals1 ++ restVals2
+      // TODO: eliminate redundant values
```
waterlens, line 145: Ditto

compiler/shared/test/scala/mlscript/compiler/Test.scala
```diff
@@ -4,42 +4,60 @@ import mlscript.utils.shorthands.*
...
 import mlscript.compiler.mono.Monomorph
-import mlscript.compiler.printer.ExprPrinter
 import mlscript.compiler.mono.MonomorphError
 
 class DiffTestCompiler extends DiffTests {
   import DiffTestCompiler.*
-  override def postProcess(mode: ModeType, basePath: List[Str], testName: Str, unit: TypingUnit): List[Str] = 
+  override def postProcess(mode: ModeType, basePath: List[Str], testName: Str, unit: TypingUnit, output: Str => Unit): (List[Str], Option[TypingUnit]) = 
     val outputBuilder = StringBuilder()
+    //output("Parsed:")
```
waterlens: Ditto

waterlens (commented):
It looks good to me but please remove the legacy code in comments, especially those in `compiler/shared/test/scala/mlscript/compiler/Test.scala`. It may cause trouble for future merging and maintenance.

waterlens (approved):

PR #198, opened 2023-12-30
Fix code generation for local functions declared by `let rec`

chengluyu:
This PR fixes the problem that local functions declared by `let rec` doesn't work in code generation before.

### DIFF
--- a/shared/src/main/scala/mlscript/JSBackend.scala
+++ b/shared/src/main/scala/mlscript/JSBackend.scala
@@ -11,6 +11,15 @@ import scala.util.chaining._
 
 abstract class JSBackend(allowUnresolvedSymbols: Bool) {
   def oldDefs: Bool
+
+  protected implicit class TermOps(term: Term) {
+    def isLam: Bool = term match {
+      case _: Lam => true
+      case Bra(false, inner) => inner.isLam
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -11,6 +11,15 @@ import scala.util.chaining._
 
 abstract class JSBackend(allowUnresolvedSymbols: Bool) {
   def oldDefs: Bool
+
+  protected implicit class TermOps(term: Term) {
```
NeilKleistGao, line 15: Do we have a better name for this? :thinking:

LPTK: This name follows the standard pattern.

chengluyu: `XOps` generally provides extension methods for `X`.

LPTK:
@chengluyu in case you didn't notice: the PR needs to be fixed.

chengluyu:
Thanks for the reminder. Let me confirm. Besides conflicts, is there anything else I neglected?

LPTK (approved):
LGTM, thanks!

PR #200, opened 2023-12-30
Remove inspection function for syntax tree nodes

chengluyu:
~~This PR is extracted from #194. But I'm actually quite dubious about this PR, because manually creating a inspect function for _every_ syntax nodes is really troublesome. I wonder if we can do this by macros or is there anything builtin that can achieve the similar results.~~

### TL;DR

The most important change this PR makes is, if you want to:
- see a code representation of a syntax tree, please call `showDbg`;
- see how a syntax tree is constructed (the _raw_ structure), please call `toString`.

### Changes

- Add `showDbg` function to `Statement` (incl. `Term`, etc), `TypingUnit`, and `Pgrm`. The function `showDbg` was there but not available at all of them.
- Remove `toString` functions of these syntax nodes, so that the compiler will synthesize a default `toString` for them.
- Remove `mlscript.codegen.Helpers.inspect` and replace all its call with `toString`.

> [!Note]
> I spent some time to make sure that `showDbg` behaves in the same way as `toString` before. As a result, _only test files that involves the output from `mlscript.codegen.Helpers.inspect` should be changed in this PR_.


shared/src/test/diff/ucs/LeadingAnd.mls
```diff
@@ -23,7 +23,7 @@ fun f(a, b) = if a is Some(av)
     and b is Some(bv)
     then av + bv
 //│ |#fun| |f|(|a|,| |b|)| |#=| |#if| |a| |is| |Some|(|av|)|→|and| |b| |is| |Some|(|bv|)|↵|#then| |av| |+| |bv|←|
-//│ AST: TypingUnit(NuFunDef(None, f, None, [], Lam(Tup(_: Var(a), _: Var(b)), If(IfOpApp(Var(a), Var(is), I; f; O; p; s; A; p; p; (; A; p; p; (; V; a; r; (; S; o; m; e; ); ,;  ; T; u; p; (; _; :;  ; V; a; r; (; a; v; ); ); ); ,;  ; <; i; t; e; r; a; t; o; r; >, None))))
```
chengluyu, line 26: What did the previous implementation do...

LPTK: Lol wat

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -1051,7 +1051,7 @@ class DiffTests
               err.getStackTrace().take(
                 if (mode.fullExceptionStack) Int.MaxValue
                 else if (mode.fixme || err.isInstanceOf[StackOverflowError]) 0
-                else 10
+                else 50
```
LPTK: Why 50? Seems like a lot. Maybe 20?

LPTK (commented):
Nice simplification. But while we're at it, could you please remove the `toString` printing for all the compiler tests? We should only print this stuff when `:p` is on. It's far too verbose and noisy to be done by default.

shared/src/test/scala/mlscript/DiffTests.scala
chengluyu: I have added `???` in `toString`, and according to the stack trace, I could find the implicit `toString` call that could not be found by the search. 10 was indeed not enough, so I increased it to 20, and finally to 50, to complete the refactoring. However, I feel that 10 is sufficient normally, so I will change it back to 10.

shared/src/test/diff/ucs/LeadingAnd.mls
chengluyu: Some borked output: `Var(is), I; f; O; p; s; A; p; p; (; A; p; p; (; V; a; r; (; S; o; m; e; );`...

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -472,20 +473,28 @@ trait NuDeclImpl extends Located { self: NuDecl =>
...
+
+  def showDbgTop: Str = entities.iterator.map {
+    case t: Term => t.print(false)
     case d: NuDecl => d.showDbg
-    case c: Constructor => c.toString
+    case c: Constructor => c.showDbg
     case e => lastWords(s"Unexpected typing unit entity: $e")
   }.mkString("{", "; ", "}")
-  override def toString: String = s"‹${entities.mkString("; ")}›"
+
```
chengluyu: Regarding this, the pretty printing for `TypingUnit` used two types of brackets: `‹›` and `{}`, in order not to have a lot of irrelevant changes in DiffTests, I kept the two print outputs and named the version using `{}` as `showDbgTop`. I want to ask if we can only keep one version of `showDbg` function.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -1000,15 +1020,11 @@ trait StatementImpl extends Located { self: Statement =>
...
-    case _: Term => super.toString
-    case d: Decl => d.showDbg
-    case d: NuDecl => d.showDbg
+  def showDbg: Str = this match {
+    case LetS(isRec, name, rhs) => s"let${if (isRec) " rec" else ""} ${name.showDbg} = ${rhs.showDbg}"
+    case DatatypeDefn(head, body) => s"data type ${head.showDbg} of ${body.showDbg}"
+    case DataDefn(head) => s"data ${head.showDbg}"
+    case _: Constructor | _: Term | _: Decl | _: NuDecl => ??? // Implemented in subclasses
   }
 }
```
chengluyu: `LetS`, `DatatypeDefn`, and `Datadefn` doesn't have implementation traits. So this `showDbg` at `StatementImpl` must have a definition. I left an `???` because `Constructor`, `Term`, `Decl`, and `NuDecl` has their implementation traits. Is this okay?

chengluyu (commented):
Hello, I have gone through it myself, and left two comments on areas I'm not quite sure about.

shared/src/test/scala/mlscript/DiffTests.scala
chengluyu: Done.

shared/src/main/scala/mlscript/helpers.scala
LPTK: Of course not. This code is dangerous and not refactoring-safe. Just move the implementations here.

LPTK: In general you should never make the logic of a piece of code depend on some implementation detail in another piece of code located somewhere else in the code base. It's still needed sometimes (not here), and then it needs extensive documentation.

chengluyu: Thanks! I will fix this immediately.

shared/src/main/scala/mlscript/helpers.scala
LPTK: Would just using `‹›`  create too many changes? Anyway, it's better for it to be consistent, so let's use it.

chengluyu: Yes, it changes 55 files. See https://github.com/chengluyu/mlscript/commit/e4e03b1a0578e02e03072b281f61d92cea724a62 for details.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -1024,7 +1016,17 @@ trait StatementImpl extends Located { self: Statement =>
     case LetS(isRec, name, rhs) => s"let${if (isRec) " rec" else ""} ${name.showDbg} = ${rhs.showDbg}"
     case DatatypeDefn(head, body) => s"data type ${head.showDbg} of ${body.showDbg}"
     case DataDefn(head) => s"data ${head.showDbg}"
-    case _: Constructor | _: Term | _: Decl | _: NuDecl => ??? // Implemented in subclasses
+    case Constructor(params, body) => s"constructor(${params.showElems}) ${body.showDbg}"
+    case t: Term => t.print(false)
+    case d: Decl => d.showHead + (d match {
+      case TypeDef(Als, _, _, _, _, _, _, _) => " = ";
```
LPTK: This underscore fest is actually an anti-pattern (I know it's used in a few places currently). Just use `case td: TypeDef if td.kind is Als`. Same remark for code below.

shared/src/main/scala/mlscript/helpers.scala
LPTK: How many changes does picking the other one create? :P

chengluyu: > How many changes does picking the other one create? :P

Only a few.

```
M  compiler/shared/test/diff/Lifter.mls
M  compiler/shared/test/diff/LifterBlks.mls
M  shared/src/test/diff/parser/BasicSyntax.mls
M  shared/src/test/diff/parser/New.mls
```

Actually, I don't think guillemets are often used here. Most test cases looks like this.

```
new A { fun x = 1 }
//│ |#new| |A| |{| |#fun| |x| |#=| |1| |}|
//│ Parsed: {new A { ‹fun x = 1› }}
```

But it doesn't make much sense to have nested brackets as in `{ ‹fun x = 1› }`.

shared/src/main/scala/mlscript/helpers.scala
chengluyu: Ah, I also felt uncomfortable when I saw this, but I ignored it. Now it has been fixed.

shared/src/main/scala/mlscript/helpers.scala
chengluyu: So, let's just use `{}`? 🫢

LPTK: Sure.

chengluyu:
Thanks!

PR #202, opened 2024-01-03
Implement effect prototype

NeilKleistGao:


shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -285,7 +295,9 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
     TupleType(N -> ty1.toUpper(ty1.prov) :: N -> ty2.toUpper(ty2.prov) :: Nil)(noProv)
   val builtinBindings: Bindings = {
     val tv = freshVar(noProv, N)(1)
-    import FunctionType.{ apply => fun }
+    // import FunctionType.{ apply => fun }
```
LPTK, line 298:
```suggestion
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -67,6 +67,9 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
       tyDefs2: MutMap[Str, DelayedTypeInfo],
       inRecursiveDef: Opt[Var], // TODO rm
       extrCtx: ExtrCtx,
+      effVars: MutSet[ST],
+      handlers: MutMap[Str, SkolemTag],
+      penv: MutMap[Term, TypeVariable],
```
LPTK, line 72: Please document what these do and how they work.

LPTK:
I believe these changes have already mostly been reimplemented as part of the new compiler implementation and more specifically the BbML prototype, correct? @NeilKleistGao

There might still be test cases from this PR that would be interesting and helpful to port over to `hkmc2`, though. WDYT?

NeilKleistGao:
> I believe these changes have already mostly been reimplemented as part of the new compiler implementation and more specifically the BbML prototype, correct? @NeilKleistGao
> 
> There might still be test cases from this PR that would be interesting and helpful to port over to `hkmc2`, though. WDYT?

Yes, basic effects are supported in BbML. We still lack primitive effects, like `io`, and user-defined effects. I think these can be incrementally implemented in the later PRs and this PR can be marked as "to revisit".

PR #203, opened 2024-01-03
Constraint solving for function overloading

auht:


LPTK:
Note: normally, even for draft PRs, we try to maintain a state where all the tests are green. For test cases that currently don't work, we still use the corresponding `:e` (= expect a type error here) but together with a TODO/FIXME comment.

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -818,13 +820,23 @@ class ConstraintSolver extends NormalForms { self: Typer =>
             val newBound = (cctx._1 ::: cctx._2.reverse).foldRight(rhs)((c, ty) =>
               if (c.prov is noProv) ty else mkProxy(ty, c.prov))
             lhs.upperBounds ::= newBound // update the bound
+            lhs.lbtsc.foreach {
+              case (tsc, i) =>
+                tsc.filterUB(i, rhs)
+                if (tsc.constraints.isEmpty) reportError()
```
LPTK, line 826: Should probably raise a proper error here, which should refer to the failing TSC.

shared/src/main/scala/mlscript/TypeSimplifier.scala
```diff
@@ -76,7 +76,6 @@ trait TypeSimplifier { self: Typer =>
                 .reduceOption(_ &- _).filterNot(_.isTop).toList
             else Nil
         }
-
```
LPTK, line 79: ?

shared/src/test/diff/nu/HeungTung.mls
```diff
@@ -106,23 +116,30 @@ fun g = f
...
+//│ Int
 //│ res
 //│     = 0
 
 // f(0) : case 0 of { Int => Int; Bool => Bool } == Int
 
 
 x => f(x)
-//│ (Bool | Int) -> (Int | false | true)
+//│ anything -> nothing
```
LPTK, line 127: TODO: Type simplification should not turn these into top/bot. We need to update `getVarsPol` and treat those vars in TSCs as invariant.

LPTK (commented):
TODO:
 - [ ] update `getVarsPol`
 - [ ] update `Constrained` class with ew field to use for TSCs
 - [ ] update App typer case to do the TSC introduction there (and not in constraint solver)

LPTK:
@auht Please make sure to address every comment that was reported in this PR and then promptly close it (since https://github.com/hkust-taco/mlscript/pull/213 subsumes it). It's been needlessly lingering for a long time.

PR #204, opened 2024-01-10
Introduce comma operator and use `,`/`;` instead of `;`/`;;`

LPTK:
The use of the `;;` statement separator and a `;` sequencing operator, which was inspired by OCaml, just felt too weird an unnatural. Let's go back to something most programmers will more likely already be familiar with.

A nice thing with the comma operator is that it gives us a way of succinctly but clearly discarding return values, by writing `foo(123), ()` instead of `let _ = foo(123) in ()` or `discard(foo(123))`.

A cons is that the meaning of the comma kind of becomes overloaded.

Another cons is that because we want to be consistent, we want to parse `if ... then foo(), 1 else bar(), 2` as `if ... then (foo(), 1) else (bar(), 2)`, but this means `if`-`then`-`else`s now need parentheses when used in an argument list, as in `baz(A , (if ... then B1 else B2), C)`.

The PR also makes a couple of improvements here and there.


NeilKleistGao (commented):
Are the following behaviors expected? :thinking:
```ts
:pe
1, 
//│ ╔══[PARSE ERROR] Unexpected end of input; an expression was expected here
//│ ║  l.389: 	1, 
//│ ╙──       	   ^
//│ ()
//│ res
//│     = undefined

(1, )
//│ 1
//│ res
//│     = 1

fun bar(x, y, z) = x + y + z
bar(1, 2, 3, )
//│ fun bar: (Int, Int, Int) -> Int
//│ Int
//│ res
//│     = 6
```

LPTK:
Oh that's a good catch. We currently allow trailing commas in parenthesis sections but not outside of them...

We do want trailing commas as they're very convenient. Scala has them but forces a new line right afterwards. Maybe we could do that and only in argument positions...

shared/src/test/diff/nu/RightAssocOps.mls
```diff
@@ -9,7 +9,7 @@ fun (:+) post(xs, x: Int) = [xs, [x]]
 fun (++) conc(xs, ys) = [xs, ys]
 //│ fun (+:) pre: forall 'a. (x: Int, 'a) -> [[Int], 'a]
 //│ fun (:+) post: forall 'b. ('b, x: Int) -> ['b, [Int]]
-//│ fun (++) conc: forall 'c 'd. ('c, 'd) -> ['c, 'd]
+//│ fun (++) conc: forall 'c 'd. ('d, 'c) -> ['d, 'c]
```
chengluyu: Why has this been changed and why are there many type changes in other files?

shared/src/test/scala/mlscript/ReplHost.scala
```diff
@@ -31,7 +31,9 @@ final case class ReplHost() {
   private def collectUntilPrompt(): ReplHost.Reply = {
     val buffer = new StringBuilder()
     while (!buffer.endsWith("\n> ")) {
-      buffer.append(stdout.read().toChar)
+      val c = stdout.read()
+      if (c === -1) lastWords(s"ReplHost could not read more from NodeJS stdout.")
```
chengluyu, line 35: Does this indicate a unexpected termination of Node.js process?

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -16,6 +16,66 @@ object NewParser {
...
+  private val AppPrec = prec('.') - 1
+  
+  final def opCharPrec(opChar: Char): Int = prec(opChar)
+  final def opPrec(opStr: Str): (Int, Int) = opStr match {
+    case "is" => (4, 4)
+    case "and" => (3, 3)
+    case "or" => (2, 2)
+    case "=>" =>
+      // * The lambda operator is special:
+      // *  it should associate veyr strongly on the left and very loosely on the right
```
chengluyu:
```suggestion
      // *  it should associate very strongly on the left and very loosely on the right
```

shared/src/test/scala/mlscript/DiffTests.scala
chengluyu, line 1: When will these `TEST CASE FAILURE` happen?

shared/src/test/diff/mlscript/MutArray.mls
```diff
@@ -70,10 +70,8 @@ update0 0 ((mut 1,))
 update1 0 ((mut 1,))
 //│ = []
 
-:re // FIXME
 update2 0 ((mut 1,))
-//│ Runtime error:
-//│   TypeError: ((intermediate value) , []) is not a function
+//│ = []
```
chengluyu, line 74: Why has this error gone?

shared/src/test/diff/nu/CommaOperator.mls
```diff
@@ -0,0 +1,388 @@
...
+log(1), 2
+//│ 2
+//│ res
+//│     = 2
+//│ // Output
+//│ 1
+
+
+(),
+  1
```
chengluyu, line 64: What if

```
(),
1
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -338,6 +338,10 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
         PolymorphicType(MinLevel, fun(singleTup(v), fun(singleTup(v), BoolType)(noProv))(noProv))
       },
       "error" -> BotType,
+      "," -> {
+        val v = freshVar(noProv, N)(1)
+        PolymorphicType(MinLevel, fun(TupleType(N -> TopType.toUpper(provTODO) :: N -> v.toUpper(provTODO) :: Nil)(noProv), v)(noProv))
+      },
```
chengluyu: Is this line of code causing so many inferred type changes?

chengluyu (approved):
LGTM except some questions.

shared/src/test/diff/nu/RightAssocOps.mls
LPTK: It's a very sad state of affairs. Because I added a new primitive which uses a new type variable, then the unique IDs of all type variables are shifted and for obscure reasons this can lead to such results. It shouldn't, ideally, but I've never had the time to get to the bottom of this yet.

shared/src/test/scala/mlscript/ReplHost.scala
LPTK: Pretty much. The code you originally wrote here was very dangerous: it kept growing the buffer with `-1`s until the whole JVM ran out of memory!

LPTK: This happened to me when a brew upgrade broke NodeJS. I suddenly started getting OOM exceptions when trying to run any test. Good times debugging this.

shared/src/test/scala/mlscript/DiffTests.scala
LPTK: When there's a test case failure. Previously the failure was not explained, we just had its line number.

shared/src/test/diff/mlscript/MutArray.mls
LPTK: IDK... and I also don't know why this type checks, BTW! `((mut 1,))` is not supposed to be an array. Let me look into it...

shared/src/main/scala/mlscript/Typer.scala
LPTK: Yup. The `freshVar` call, to be specific.

shared/src/test/diff/nu/CommaOperator.mls
LPTK: Good question. Currently, `Unexpected newline in expression position`.

LPTK: Just pushed a commit to allow it.

shared/src/test/scala/mlscript/ReplHost.scala
chengluyu: True. I originally thought `read()` would cause an error if there's no characters left.

shared/src/main/scala/mlscript/Typer.scala
LPTK: All right, I changed it slightly so all the meaningless type changes are gone.

Calls for a better solution in the future, though.

shared/src/test/diff/nu/WeirdUnions.mls
```diff
@@ -11,23 +11,23 @@ fun f: [Str] | [Str, Int]
 fun f: (Str | [Str, Int])
 //│ fun f: Str | [Str, Int]
 
-:pe
+:e
 fun f: Str | (Str, Int)
-//│ ╔══[PARSE ERROR] Expected '=>' or '->' after this parameter section
+//│ ╔══[ERROR] type identifier not found: ,
```
chengluyu, line 16: Wow, so can `,` be a user-defined type operator right now?

shared/src/test/diff/mlscript/MutArray.mls
LPTK: OH never mind, I missed that this was in the old parser. In the old parser, this is indeed an array.

shared/src/test/diff/nu/WeirdUnions.mls
LPTK: Nah because `,` will parse as a COMMA token.

```scala
type , = 1
//│ ╔══[PARSE ERROR] Expected a type name; found comma instead
//│ ║  l.4: 	type , = 1
//│ ╙──     	     ^
//│ TEST CASE FAILURE: There was an unexpected parse error
//│ type <error> = 1
```

Anyway in the future terms and types will parse the same, so this will be moot.

PR #205, opened 2024-01-11
Fix/virtual type arg params

LPTK:
This change exposed some more limitations when typing cyclic definitions. In the future we should create a proper framework for correctly dealing with typing dependencies, allowing to type more things, to still infer refined types in the absence of cycles, and to give more specific error messages.


chengluyu (approved):
LGTM.

PR #206, opened 2024-01-11
Improve self-type handling

LPTK:
This PR fixes various issues with the handling of self types and adds self-type checking.
As a byproduct, it makes class constructors return proper type refs, which has the welcome benefit of simplifying some inferred recursive types by tying the knot earlier!

To deal with:
 - [x] Fix transitivity problems (see end of ClassSignatures.mls)
 - [x] Add warning when non-abstract class has an annot (did you mean to use `extends`?)


NeilKleistGao (commented):
I think we need a check for this:
```ts
abstract class Foo(): Foo
//│ /!!!\ Uncaught error: java.lang.StackOverflowError

abstract class Bar(): Bar | Bar
//│ /!!!\ Uncaught error: java.lang.StackOverflowError
```

LPTK:
@NeilKleistGao Wow, nice catch! Funnily the SOF happened in type simplification. It was an easy fix.

NeilKleistGao (approved):
LGTM.

LPTK:
Thanks for the quick review!

PR #208, opened 2024-01-31
Add imperative features

LPTK:
This is to add support for mutable local variables (`mut let`), mutable fields (`mut val`), and while loops (`while ... do ...`). Reassignment is given syntax `set x = t`, which clearly distinguishes it from equations, which use a single `=`.


shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -570,6 +574,7 @@ class ConstraintSolver extends NormalForms { self: Typer =>
             } else {
               val fty = lookupField(() => done_ls.toType(sort = true), S(nme), r.fields.toMap.get, ts, fldNme)
               rec(fty.ub, fldTy.ub, false)
+              println("!!1")
```
NeilKleistGao: Debug information?

LPTK: Good catch, thanks!

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -280,12 +280,16 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], newDefs: Bo
...
+        consume
+        yeetSpaces
+        go(acc.copy(acc.mods + ("mut" -> l0)))
       case (KEYWORD("abstract"), l0) :: c =>
         consume
         yeetSpaces
         go(acc.copy(acc.mods + ("abstract" -> l0)))
       case _ if acc.mods.isEmpty => acc
-      case (KEYWORD("class" | "infce" | "trait" | "mixin" | "type" | "namespace" | "module" | "fun" | "val"), l0) :: _ =>
+      case (KEYWORD("class" | "infce" | "trait" | "mixin" | "type" | "namespace" | "module" | "fun" | "val" | "let"), l0) :: _ =>
```
chengluyu, line 292: Unrelated question: will `infce` be actually used in the future?

shared/src/main/scala/mlscript/ConstraintSolver.scala
```diff
@@ -639,8 +643,12 @@ class ConstraintSolver extends NormalForms { self: Typer =>
         (lhs.lb, rhs.lb) match {
           case (Some(l), Some(r)) => rec(l, r, false)
           case (Some(l), None) =>
-            if (lhs.prov.loco.isEmpty || rhs.prov.loco.isEmpty) reportError()
-            else reportError(S(msg"is not mutable"))(
+            println(s"RHS not mutable! in $lhs <- $rhs")
```
chengluyu, line 646: Removing the extra exclamation mark?

```suggestion
            println(s"RHS not mutable in $lhs <- $rhs")
```

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -301,10 +302,12 @@ abstract class JSBackend(allowUnresolvedSymbols: Bool) {
       JSArray(terms map { case (_, Fld(_, term)) => translateTerm(term) })
     case Subs(arr, idx) =>
       JSMember(translateTerm(arr), translateTerm(idx))
+    case While(cond, body) =>
+      JSImmEvalFn(N, Nil, R(JSWhileStmt(translateTerm(cond), translateTerm(body)) :: Nil), Nil)
     case Assign(lhs, value) =>
       lhs match {
         case _: Subs | _: Sel | _: Var =>
-          JSCommaExpr(JSAssignExpr(translateTerm(lhs), translateTerm(value)) :: JSArray(Nil) :: Nil)
+          JSCommaExpr(JSAssignExpr(translateTerm(lhs), translateTerm(value)) :: JSIdent("undefined") :: Nil)
```
chengluyu: Generally, [void operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void) is used to discard the return value of an expression.

shared/src/main/scala/mlscript/NuTypeDefs.scala
```diff
@@ -954,22 +957,21 @@ class NuTypeDefs extends ConstraintSolver { self: Typer =>
...
               assert(!mut && !spec, "TODO") // TODO
-              value.toType match {
-                case R(tpe) =>
-                  implicit val newDefsInfo: Map[Str, (TypeDefKind, Int)] = Map.empty // TODO?
-                  val ty = typeType(tpe)
-                  nme -> FieldType(N, ty)(provTODO)
-                case _ => ???
-              }
+              val tpe = value.toTypeRaise
+              implicit val newDefsInfo: Map[Str, (TypeDefKind, Int)] = Map.empty // TODO?
```
chengluyu: TODO for what?

shared/src/test/diff/nu/LetRec.mls
```diff
@@ -176,3 +176,16 @@ foo()()
...
+:re
+let xf = yf()
+let yf() = xf
+//│ let xf: nothing
+//│ let yf: () -> nothing
+//│ xf
+//│ Runtime error:
+//│   ReferenceError: yf is not defined
+//│ yf
+//│    = [Function: yf]
```
chengluyu, line 189: If I understand correctly, `PreTyper` should be able to reject this before typing.

shared/src/test/diff/nu/MissingTypeArg.mls
```diff
@@ -57,7 +57,7 @@ test(p, p)
 //│ ╟── into field selection of type `#Eql`
 //│ ║  l.8: 	fun test(pt1, pt2) = pt1.color === pt1.color and
 //│ ╙──     	                     ^^^^^^^^^
-//│ error
+//│ error | false | true
```
chengluyu, line 60: Why has the type changed here?

shared/src/test/diff/nu/MutLet.mls
```diff
@@ -0,0 +1,143 @@
...
+
+mut let oops = None
+//│ let oops: None
+//│ oops
+//│      = None { class: [class None] }
+
+fun funny(x) =
+  set oops = Some(x)
+  x
+//│ fun funny: forall 'a. 'a -> 'a
```
chengluyu, line 50: Does the type of `oops` change after this function is declared?

ts2mls/js/src/test/typescript/Heritage.ts
```diff
@@ -13,7 +13,7 @@ class C<T> {
 
   private t: T
 
-  set(x: T) { this.t = x; }
+  sett(x: T) { this.t = x; } // FIXME support `set` methods...
```
chengluyu, line 16: We can either support escaped identifiers (`id"set"` as mentioned [here](https://arc.net/l/quote/gmklqgar)) or make `set` a soft keyword.

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -957,8 +961,8 @@ abstract class JSBackend(allowUnresolvedSymbols: Bool) {
     val bodyStmts = R(preDecs ++ tempDecs ++ qualifierStmts ++ (bodyResult :: Nil))
     // Returns members depending on what it is.
     memberParams match {
-      case S(memberParams) => JSClassMethod(name, memberParams, bodyStmts)
-      case N => JSClassGetter(name, bodyStmts)
+      case S(memberParams) => JSClassMethod(name, memberParams, bodyStmts) :: Nil
+      case N => JSClassGetter(name, bodyStmts) :: Nil
```
chengluyu: If it's just a list with one element, why change the return type to a list?

chengluyu (changes requested):
LGTM. Btw the `minRep` test is cool!

shared/src/main/scala/mlscript/NewParser.scala
LPTK: I guess not. That abbreviated keyword is really ugly. We might add `interface` later on, but traits can already serve this purpose so maybe not.

shared/src/main/scala/mlscript/ConstraintSolver.scala
LPTK: nope

shared/src/main/scala/mlscript/JSBackend.scala
LPTK: Interesting. Didn't know that one. So would you argue we should use the same in MLscript, making it:
```fs
fun foo(x) =
  void foo(x + 1)
  x - 1
```
instead of
```fs
fun foo(x) =
  foo(x + 1), ()
  x - 1
```
?

ts2mls/js/src/test/typescript/Heritage.ts
LPTK: I know, but this is not currently being used by ts2mls. Cc @NeilKleistGao

shared/src/test/diff/nu/MutLet.mls
LPTK: It just gets a new lower bound. But conceptually and in a real program this is just a detail of the type inference and the type can be considered fixed and inferred from all uses.

shared/src/test/diff/nu/MissingTypeArg.mls
LPTK: Due to some obscure type error reporting implementation detail? I wouldn't worry about it

shared/src/test/diff/nu/LetRec.mls
LPTK: Yes, indeed.

LPTK:
@chengluyu Thanks for the helpful review! I'll merge when the conversations are resolved. I've answered all of them.

shared/src/test/diff/ex/RepMin.mls
```diff
@@ -0,0 +1,288 @@
...
+      log of join of "Traversing leaf ", String of n
+      if n < min then set min = n else ()
+      () => Leaf of if n > 2 * min then min else n
+    Node(l, r) then
+      let l' = go(l)
+      let r' = go(r)
+      () => Node of l'(), r'()
+  let res = go(t)
+  [min, res()]
+//│ fun repmin: forall 'min. (Leaf[Int & 'min] | Node[Int & 'min]) -> [2147483648, Leaf[2147483648 | 'min] | Node[2147483648 | 'min]]
```
LPTK: I just noticed the `[2147483648,` part of the inferred type is wrong.

LPTK: Minimized:
```fs
fun repmin(t) =
  mut let min = 123
  let go(t) = set min = t.value
  go(t)
  min
//│ fun repmin: {value: anything} -> 123
```

LPTK: Lol the problem is that let bindings were still generalized even when mutable. Trying to solve this in a principled way now.

ts2mls/js/src/test/typescript/Heritage.ts
NeilKleistGao: > I know, but this is not currently being used by ts2mls. Cc @NeilKleistGao

It is finished in the driver PR. We can cherry-pick it if necessary.

shared/src/main/scala/mlscript/NewParser.scala
chengluyu: I see. 👍

shared/src/test/diff/nu/MutLet.mls
chengluyu: I see!

shared/src/main/scala/mlscript/JSBackend.scala
chengluyu: No. I suggest that we should generate code like `void expr`.

chengluyu: We can use `JSUnary("void", expr)` instead of `JSCommaExpr`.

LPTK: Ah _that's_ what you're suggesting! Sure, let me try to do it.

chengluyu (approved):
LGTM!

PR #210, opened 2024-02-10
Annotation System

CAG2Mark:
Adds an annotation system and basic type checker support (TODO). 

The parser supports adding arbitrary annotations to any expressions/statements. For example:
```fs
@tailrec
fun sum(n) = if n == 0 then 0 else @tailrec sum(n - 1)
```
This is in preparation for implementing the `@tailrec` annotation.

TODO:
- Currently it uses a mutable style `Annotated` trait similar to `Located`. Change it to use an `Ann` syntax that contains other terms and contains a list of annotations.
- Implement pretty-printing for the changes.
- Basic type checking to make sure all annotation names are in scope.


CAG2Mark:
Added an `Ann` case class that extends `Term` and modified the parser to parse to this.

TODO: 
- Figure out the best way to store annotations in non-Term statements that we still want to annotate such as `NuFunDef`.  Possible solutions: 
  - Just add an `annotations` field to each of these.
  - Create an `AnnStatement` class, though this is quite messy and also forces us to also consider the case when `AnnStatement` contains a `Term`, which feels unclean.
- (*need help*) Missing case for `Ann` in `ClassLifter` class `liftTerm`. Still need to figure out what this function does.
- (*need help*) In `compiler/Helpers:32`, need to figure out whether this is a conversion we actually want, i.e. whether we should preserve annotations when converting to `Expr`. If I'm understanding correctly, `Expr` represents nodes in a control flow graph, so we probably don't need to preserve annotations (?)
- Update the type checker.

LPTK:
> Figure out the best way to store annotations in non-Term statements that we still want to annotate such as `NuFunDef`. Possible solutions:
> * Just add an `annotations` field to each of these.

Yes, this is the way. In practice that means you only have to add the field to `NuFunDef` and `NuTypeDef`. Terms already support annotations through `Ann`, and other statements are either deprecated or won't support annotations.






> * (_need help_) Missing case for `Ann` in `ClassLifter` class `liftTerm`. Still need to figure out what this function does.

Just add it to the catch-all case of unsupported stuff, for now. This is experimental code that's already being changed anyway.

> * (_need help_) In `compiler/Helpers:32`, need to figure out whether this is a conversion we actually want, i.e. whether we should preserve annotations when converting to `Expr`. If I'm understanding correctly, `Expr` represents nodes in a control flow graph, so we probably don't need to preserve annotations (?)

Sam remark. This code is already obsolete.

> * Update the type checker.

Easy. Just defer the type checking of `Ann(Var(nme) :: anns, receiver)` to that of `Var(nme)` and `Ann(anns, receiver)`.

BTW I'm not convinced we need multiple annotations in the same `Ann` term. This has the downside of allowing the dummy `Ann(Nil, t)` and it's also kind of unnecessary. Class `Term` could be given a helper method `peelAnnotations` that would return a `(List[Var], Term)`, if that's even needed.

compiler/shared/main/scala/mlscript/compiler/Helpers.scala
```diff
@@ -29,6 +29,7 @@ object Helpers:
 
   def term2Expr(term: Term): Expr = {
       term match
+        case Ann(anns, receiver) => term2Expr(receiver) // TODO: Make sure this is actually what we want.
```
LPTK: Don't sweat it ;^)
```suggestion
        case Ann(anns, receiver) => term2Expr(receiver)
```

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -8,6 +8,8 @@ import utils._, shorthands._
 import mlscript.Message._
 import BracketKind._
 
+import mlscript.Annotation
+
```
LPTK: This import does nothing.
```suggestion
```

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -817,6 +817,9 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
       case v @ Var("_") =>
         if (ctx.inPattern || funkyTuples) freshVar(tp(v.toLoc, "wildcard"), N)
         else err(msg"Widlcard in expression position.", v.toLoc)
+      
+      // TODO: Check all annotations are in scope
```
LPTK: Don't forget to handle that.

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -985,6 +987,7 @@ trait StatementImpl extends Located { self: Statement =>
   }
   
   def children: List[Located] = this match {
+    case Ann(_, trm) => trm :: Nil
```
LPTK: The annotations should be part of the children!

shared/src/main/scala/mlscript/syntax.scala
```diff
@@ -47,6 +47,14 @@ final case class MethodDef[RHS <: Term \/ Type](
   val children: Ls[Located] = nme :: body :: Nil
 }
 
+// Annotations
+final case class Annotation(name: Var) extends Located {
+  def children: List[Located] = name :: Nil
+}
+
+// final case class Ann(anns: Ls[Annotation], receiver: Statement) extends Statement
+// final case class AnnDesugared(anns: Ls[Annotation], receiver: DesugaredStatement) extends DesugaredStatement
```
LPTK: Don't forget to clean up.

shared/src/test/diff/nu/Annotations.mls
```diff
@@ -0,0 +1,50 @@
...
+//│ Parsed: {fun hello = (x,) => {let rec x = @hello 2}}
+//│ AST: TypingUnit(List(NuFunDef(None,Var(hello),None,List(),Left(Lam(Tup(List((None,Fld(_,Var(x))))),Blk(List(NuFunDef(Some(true),Var(x),None,List(),Left(Ann(List(Annotation(Var(hello))),IntLit(2)))))))))))
+//│ Parsed: 
+//│ 
+
+
+:p
+fun x =
+    let y = 2
+    2
```
LPTK: Test case seems redundant?

shared/src/test/diff/nu/Annotations.mls
```diff
@@ -0,0 +1,50 @@
+:NewDefs
```
LPTK, line 1: Also test annotations on classes, modules, and vals.

CAG2Mark:
> BTW I'm not convinced we need multiple annotations in the same Ann term. This has the downside of allowing the dummy Ann(Nil, t) and it's also kind of unnecessary.

True, we can just store it as `Ann(annotation1, Ann(annotation2, ...))`.

CAG2Mark:
It seems that adding an `annotations` field to `NuTypeDef` or `NuFunDef` creates errors in 70+ other places. Is there a better way of doing this?

LPTK:
For such secondary fields, we normally add them to the second parameter list, which creates fewer errors as it does not affect the pattern syntax. Just add another `val` in there. You'll still need to initialize and propagate a bunch of them, but these are not very hard changes.

CAG2Mark:
TODO: Fix codegen -- probably ok to just discard annotations for now. Maybe throw an error if `@tailrec` is seen, since it is not yet implemented.

shared/src/test/diff/nu/UserDefinedAnnotations.mls
```diff
@@ -14,9 +14,48 @@ fun foo = 0
 class tailrec2() extends Annotation
 //│ class tailrec2() extends Annotation
 
-// :e // TODO should be rejected
+:e // In the future, this should be supported.
```
LPTK: Why should this be supported? The code is wrong.

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -621,8 +653,16 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], newDefs: Bo
     raise(WarningReport(msg"[${cur.headOption.map(_._1).mkString}] ${""+msg}" -> loco :: Nil,
       newDefs = true))
   
-  final def exprOrIf(prec: Int, allowSpace: Bool = true)(implicit et: ExpectThen, fe: FoundErr, l: Line): IfBody \/ Term = wrap(prec, allowSpace) { l =>
-    cur match {
+  final def exprOrIf(prec: Int, allowSpace: Bool = true, annotations: Ls[Annotation] = Nil)(implicit et: ExpectThen, fe: FoundErr, l: Line): IfBody \/ Term = wrap(prec, allowSpace) { l =>
+    val moreAnnotations: Ls[Annotation] = parseAnnotations(false)
+
+    if (!moreAnnotations.isEmpty) {
```
```suggestion
    if (moreAnnotations.nonEmpty) {
```

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -851,7 +891,24 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], newDefs: Bo
...
+    }
+    
+    if (allAnns.isEmpty) res
+    else res match {
+      case Left(body) => body match {
+        case IfThen(expr, rhs) =>
+          Left(IfThen(wrapAnns(expr, allAnns), rhs))
+        case _ =>
+          err(msg"Unexpected annotation" -> allAnns.head.toLoc :: Nil)
+          L(body) // discord annotations for now
```
```suggestion
          L(body) // discard annotations for now
```

shared/src/main/scala/mlscript/NuTypeDefs.scala
```diff
@@ -1127,6 +1127,13 @@ class NuTypeDefs extends ConstraintSolver { self: Typer =>
                 if (fd.tparams.nonEmpty)
                   err(msg"Type parameters are not yet supported in this position",
                     fd.tparams.head.toLoc)
+              
+              implicit val gl: GenLambdas =
```
LPTK: Let's not do that. The comment refers to a very specific call to `typeTerm`. There's no reason to type annotations with the same value of `GenLambdas`. For annotations, just use `false`.

shared/src/main/scala/mlscript/NuTypeDefs.scala
```diff
@@ -1404,6 +1413,13 @@ class NuTypeDefs extends ConstraintSolver { self: Typer =>
                   case _ => die
                 }
               }
+
+              // Check annotations
+              td.annotations.foreach(ann => {
+                implicit val gl: GenLambdas = false // TODO: what should this be?
```
LPTK: `false` is fine. It's irrelevant here.

shared/src/main/scala/mlscript/NuTypeDefs.scala
```diff
@@ -1404,6 +1413,13 @@ class NuTypeDefs extends ConstraintSolver { self: Typer =>
                   case _ => die
                 }
               }
+
+              // Check annotations
+              td.annotations.foreach(ann => {
```
```suggestion
              td.annotations.foreach { ann =>
```

shared/src/main/scala/mlscript/Typer.scala
LPTK: My remark was also meant to remind you to remove the comment! Unresolved.

shared/src/main/scala/mlscript/Typer.scala
```diff
@@ -1736,7 +1746,7 @@ class Typer(var dbg: Boolean, var verbose: Bool, var explainErrors: Bool, val ne
       case TypedNuAls(level, td, tparams, body) =>
         ectx(tparams) |> { implicit ectx =>
           NuTypeDef(td.kind, td.nme, td.tparams, N, N, S(go(body)), Nil, N, N, TypingUnit(Nil))(
-            td.declareLoc, td.abstractLoc)
+            td.declareLoc, td.abstractLoc, Nil)
```
LPTK: Why not `td.annotations`?

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -531,6 +531,7 @@ trait TermImpl extends StatementImpl { self: Term =>
   def describe: Str = sugaredTerm match {
     case S(t) => t.describe
     case N => this match {
+      case Ann(annotations, receiver) => receiver.describe
```
```suggestion
      case Ann(_, Ann(_, receiver)) => receiver.describe
      case Ann(_, receiver) => "annotated " + receiver.describe
```

shared/src/main/scala/mlscript/helpers.scala
```diff
@@ -985,6 +987,10 @@ trait StatementImpl extends Located { self: Statement =>
   }
   
   def children: List[Located] = this match {
+    case Ann(ann, trm) => trm match {
+      case t: Ann => ann :: t.children
+      case t => ann :: t :: Nil
+    }
```
LPTK: Strange idea. Why not just `case Ann(ann, trm) => ann :: trm :: Nil`?

shared/src/main/scala/mlscript/syntax.scala
```diff
@@ -47,6 +47,11 @@ final case class MethodDef[RHS <: Term \/ Type](
   val children: Ls[Located] = nme :: body :: Nil
 }
 
+// Annotations
+final case class Annotation(name: Var) extends Located {
```
LPTK: Didn't we conclude that this class is useless? Just use a term in `Ann`.

shared/src/main/scala/mlscript/syntax.scala
```diff
@@ -216,14 +222,15 @@ final case class NuFunDef(
   nme: Var,
   symbolicNme: Opt[Var],
   tparams: Ls[TypeName],
-  rhs: Term \/ Type,
```
LPTK: Please don't...

shared/src/test/diff/nu/UserDefinedAnnotations.mls
CAG2Mark: My bad, I should've stated it more clearly. I meant that we should add support for annotations with parameters in the future.

shared/src/main/scala/mlscript/helpers.scala
CAG2Mark: Ah, I tried to keep the semantics of `Ann(Ls[Annotation], term)` and have all the annotations in one list. This isn't necessary with the updated definition of `Ann`, so I'll change this.

shared/src/test/diff/nu/UserDefinedAnnotations.mls
LPTK: Got it. Then why not add a new test case using `@tailrec2()` and move the comment to that one?

CAG2Mark: Perhaps I should just leave it in a comment, and not as an actual test. Such a test case would actually pass, because it will add a `tailrec2` annotation to the unit literal `()`, which feels wrong. I'll need to make some non-trivial changes to the parser to implement this in the end.

EDIT: Seems it doesn't pass, so I'll add it as a proper test

shared/src/test/diff/nu/UserDefinedAnnotations.mls
```diff
@@ -59,3 +59,16 @@ module dummy
 //│ ║  l.56: 	@helloa
 //│ ╙──      	 ^^^^^^
 //│ module dummy
+
+// TODO: In the future, we should still support annotations that require a constructor.
+:e
```
LPTK: Comments starting with "TODO" are interpreted specially (a wart we should remove).
```suggestion
:e // TODO: In the future, we should still support annotations that require a constructor.
```

LPTK (approved):
Otherwise it looks good to me. Nice feature!

CAG2Mark:
> TODO: Fix codegen -- probably ok to just discard annotations for now. Maybe throw an error if `@tailrec` is seen, since it is not yet implemented.

I see you've approved the changes, but I suppose this should be done before merging into master?

shared/src/test/diff/nu/UserDefinedAnnotations.mls
CAG2Mark: In this case, is keeping it as a TODO fine?

LPTK:
Right. Forgot about that one, but it should be trivial.

I any case, I'll do the merge when it's ready.

CAG2Mark:
My bad, probably should have rebased instead of merging. Anyway, it should be ready to merge now.

shared/src/test/diff/nu/Annotations.mls
```diff
@@ -39,7 +48,7 @@ fun foo(n) =
   else ()
 //│ fun foo: Int -> ()
 
-// TODO
+:NoJS // recurses infinitely since `div` is float division. do not run it for now
```
LPTK: Why not just increase the threshold to `n > 0.5` instead of writing a full comment and disabling JS? :P

shared/src/main/scala/mlscript/JSBackend.scala
```diff
@@ -76,9 +76,9 @@ abstract class JSBackend(allowUnresolvedSymbols: Bool) {
       translatePattern(base)
     case Inst(bod) => translatePattern(bod)
     case Ann(ann, receiver) => translatePattern(receiver)
-    case _: Lam | _: App | _: Sel | _: Let | _: Blk | _: Bind | _: Test | _: With | _: CaseOf
-        | _: Subs | _: Assign | _: If | _: New  | NuNew(_) | _: Splc | _: Forall | _: Where
-        | _: Super | _: Eqn | _: AdtMatchWith | _: Rft | _: While =>
+    case _: Lam | _: App | _: Sel | _: Let | _: Blk | _: Bind | _: Test | _: With | _: CaseOf | _: Subs | _: Assign
+        | If(_, _) | New(_, _)  | NuNew(_) | _: Splc | _: Forall | _: Where | _: Super | _: Eqn | _: AdtMatchWith
+        | _: Rft | _: While | _: Quoted | _: Unquoted =>
```
LPTK: What happened here? Why all these unnecessary changes?

Note: I always prefer writing things like `_: If` instead of `If(_, _)` as the former will need to be adapted if/when `If` gets a new argument.

LPTK: Never mind I just realize this is probably because of the merge.

LPTK:
> My bad, probably should have rebased instead of merging. Anyway, it should be ready to merge now.

No, I know people rebase branches like this, but it's an anti-pattern. It is likely to make the commits in the middle broken, causes unnecessary conflicts, and is annoying for people (like me) who've already checked out your branch. I'm going to squash the PR on merge anyway, as its history isn't interesting/useful.

CAG2Mark:
> > My bad, probably should have rebased instead of merging. Anyway, it should be ready to merge now.
> 
> No, I know people rebase branches like this, but it's an anti-pattern. It is likely to make the commits in the middle broken, causes unnecessary conflicts, and is annoying for people (like me) who've already checked out your branch. I'm going to squash the PR on merge anyway, as its history isn't interesting/useful.

Got it. What's the usual procedure for merging into the main branch here?

LPTK:
It really depends. For short-lived PRs like this with many almost meaningless commits (like applying review suggestions), I prefer to squash. But if the branch has a lot of history by several authors or/and has been used as a base of other branches, I use a merge commit, to retain history details and reproducible commits (ie which compile and whose tests are usually green).

PR #214, opened 2024-03-07
Update index.html

Craig-Macomber:
I tried `window.alert "Hello!"` as recommended and was presented with:

```
╔══ [WARNING] Paren-less applications should use the 'of' keyword
║ l.26: window.alert "Hello!"
╙──
```

If the suggestion was intended to show off the very nice error message, then it was quite successful and feel free to close this.

I'm guessing `window.alert of "Hello!"` was intended, and this change updates the suggestion accordingly.


LPTK (approved):
Haha thanks, good catch. That's some old text from when we had a previous parser/syntax, like in https://hkust-taco.github.io/mlscript/

PR #216, opened 2024-03-11
Fix lack of stability in inferred types

LPTK:
So I finally went down the rabbit-hole to figure out why we kept having diffs in inferred types when internal unique type variables identifiers changes (even when they retained their same relative order). Turns out there were _many_ places where we were iterating on hash-backed collections and committing similar blunders, which rendered things unstable. Most of these should be fixed, now, which should make submitting, maintaining, and reviewing PRs much less annoying.


shared/src/test/diff/nu/NuScratch.mls
```diff
@@ -1,3 +1,3 @@
 :NewDefs
-
+:NoJS
```
LPTK:
```suggestion

```

PR #218, opened 2024-03-29
Tail Recursion Optimization

CAG2Mark:
Adds a pass to the IR which optimizes mutually tail-recursive and mutually tail-recursive modulo cons functions.

The full pass first optimizes modulo cons functions to be tail recursive, then optimizes those functions to be tail recursive.

This PR consists of the following features
# IR Field Assignment
Adds an assignment primitive to the IR. Given `x: CtorApp` and `val: TrvialExpr`, we now have the primitive
```
x.field = val in ...
```
Also modified the interpreter to no longer assume references to constructors are referentially transparent.

# Tail Recursion Modulo Cons
Extends the idea in [this paper](https://dl.acm.org/doi/abs/10.1145/3571233) to optimize strongly connected components of tail recursive functions, of which some calls may be within a constructor (guarded recursion).

Supports both modulo-cons calls and regular tail calls within the same strongly connected component.

# Rewrite mutually tail recursive functions
Suppose we have functions:
```
def f(n) = e1
def g(m) = e2
```
where `e1` and `e2` tail-call `f` or `g`. This function will be rewritten a single function:
```
def f_g(branch, n, m) =
  let join j(branch, m, n) = 
    if branch == 0 then e1'
    else e2'
  j(branch, m, n)

def f(n) = f_g(0, n, dummy)
def g(n) = f_g(1, dummy, m)
```
where `e1'` and `e2'` are `e1` and `e2` with their tail calls to `f` and `g` replaced with jumps to `j`.

In general, strongly connected components of mutually tail recursive functions will be optimized in this way. All join points used in each strongly connected component will be rewritten and merged into the same join point.


CAG2Mark:
TODO:
- [x] Make sure variable names in the stack frame do not clash
- [x] Make sure the merged function's name does not clash with other functions
- [x] Substitute dummy variables (may need extension to the IR). See [here](https://github.com/hkust-taco/mlscript/pull/218/files#diff-4e4040f006dd21e8c4f6e21d0b7cf497362ccb880a52fc3d432ce8c544fc1bdeR169)
- [x] Propagage `@tailrec` annotations to the IR.

compiler/shared/main/scala/mlscript/compiler/optimizer/TailRecOpt.scala
```diff
@@ -290,26 +289,31 @@ class TailRecOpt(fnUid: FreshInt, tag: FreshInt) {
         sccEdges = sccEdges.filter { c => sccIds.contains(c.getDefn.id)}
 
         sccs = DefnGraph(scc, sccEdges) :: sccs
-      }
-    }
+      
+    
 
-    for (v <- defns) {
+    for (v <- defns)
```
LPTK: Why not
```suggestion
    for v <- defns do
```
?

CAG2Mark:
Everything is mostly done. There's just a few things left to check, such as making sure function and parameter names don't clash

CAG2Mark:
I can't seem to get the checks to become green, since I created a testcases which purposefully errors and the order of items in the stacktrace is not always the same :/

HarrisL2:
If the contents of the stack trace are not essential to the test, you could try only displaying the stack trace if a debug flag is present.

LPTK:
Yes, stack traces should never be part of the test output. By the way, an exception being thrown means the compiler has a bug. If there's a program error to report, it should be reported properly.

compiler/shared/main/scala/mlscript/compiler/ir/Builder.scala
```diff
@@ -243,7 +245,9 @@ final class Builder(fresh: Fresh, fnUid: FreshInt, classUid: FreshInt, tag: Fres
     res
   
   private def buildDefFromNuFunDef(using ctx: Ctx)(nfd: Statement): Defn = nfd match
-    case NuFunDef(_, Var(name), None, Nil, L(Lam(Tup(fields), body))) =>
+    case nfd: NuFunDef =>
+      val NuFunDef(_, Var(name), None, Nil, L(Lam(Tup(fields), body))) = nfd
```
LPTK: Note that this partial destructuring is unsafe. Conceptually, the lambda `(...xs) => ...` will not have a tuple on its LHS. Though that syntax is currently not supported.

compiler/shared/main/scala/mlscript/compiler/optimizer/TailRecOpt.scala
```diff
@@ -66,37 +77,57 @@ class TailRecOpt(fnUid: FreshInt, classUid: FreshInt, tag: FreshInt):
       if names.contains(nme) then Some(nme)
       else None
 
-  // would prefer to have this inside discoverOptimizableCalls, but this makes scala think it's not tail recursive
+  // would prefer to have this inside discoverOptCalls, but scala does not support partially tail recursive functions directly
```
LPTK, line 159: I think you could just add `private val discoverOptCallsIndirect = discoverOptCalls` and call that instead.

LPTK:
Instead of writing `class True` and `class False` in _every single test case_, why not make it a built-in class info? It should be a trivial change and would remove much clutter. Could also do the same with `Cons`/`Nil` and `Some`/`None`.

shared/src/test/diff/nu/NuScratch.mls
LPTK, line 1: Remove these changes.

compiler/shared/main/scala/mlscript/compiler/optimizer/TailRecOpt.scala
```diff
@@ -0,0 +1,948 @@
...
+        case LetCall(names, defn, args, isTailRec, body) =>
+          if isTailCall(node) && defnInfoMap.contains(defn.expectDefn.id) then
+            Jump(jpDefnRef, transformStackFrame(args, defnInfoMap(defn.expectDefn.id))).attachTag(tag)
+          else LetCall(names, defn, args, isTailRec, transformNode(body)).attachTag(tag)
+        case AssignField(assignee, clsInfo, field, value, body) => AssignField(assignee, clsInfo, field, value, transformNode(body)).attachTag(tag)
+
+      // Tail calls to another function in the component will be replaced with a tail call
+      // to the merged function
+      def transformDefn(defn: Defn): Defn =
+        // TODO: Figure out how to substitute variables with dummy variables.
```
LPTK: This should be addressed or explained better.

LPTK:
Steps before this can be an official PR:
* Address the remaining TODOs or phrase them better so other people can address them later
* Make a proper error reporting system that points to the problem
* Make sure to document the nontrivial aspects/invariants of the approach
* Make sure no reordering of computations can occur; you may use a purity check to determine if something can be reordered, whose current implementation can be very primitive (it will be improved later)
* Document the semantics of `@tailrec` when ascribing calls and definitions
    We should in fact have both:
    - `@tailcall` to annotate calls, with the current semantics assigned to `@tailrec`
    - ~~`@tailrec`-annotated calls that ensure the recursion back to the current function is tail-recursive~~ from the meeting discussion: just drop these for now and only accept `@tailrec` on definitions

compiler/shared/test/diff-ir/IRRec.mls
```diff
@@ -1,6 +1,7 @@
 :NewParser
 :ParseOnly
 :UseIR
+:NoTailRec
```
LPTK, line 4: When this is enabled, don't duplicate the IR printout.

compiler/shared/main/scala/mlscript/compiler/ir/IR.scala
```diff
@@ -137,7 +138,8 @@ enum Node:
   case Case(scrut: Name, cases: Ls[(ClassInfo, Node)])
   // Intermediate forms:
   case LetExpr(name: Name, expr: Expr, body: Node)
-  case LetCall(names: Ls[Name], defn: DefnRef, args: Ls[TrivialExpr], body: Node)
+  case LetCall(names: Ls[Name], defn: DefnRef, args: Ls[TrivialExpr], isTailRec: Bool, body: Node)
+  case AssignField(assignee: Name, clsInfo: ClassInfo, fieldName: Str, value: TrivialExpr, body: Node)
```
LPTK: Why is this a `Node` and not an `Expr`? In principle, nodes are for control operations and `Expr`s are for expression-like things like this.

CAG2Mark: I don't see why `AssignField` should be an `Expr`, since that would make using it quite clunky, i.e. you would need `let dummy = (assign trivialExpr to foo.bar) in body`.

LPTK: What's the problem with this? It seems more regular than having an entirely different node type with its own body, requiring more code to deal with node shapes.

CAG2Mark: resolved in latest commit

compiler/shared/test/diff-ir/IRRec.mls
CAG2Mark: resolved in latest commit

CAG2Mark: resolved in latest commit

CAG2Mark:
Completed:
- Improved error reporting and propagation of locations. 
  - Added compilation errors to the diff tests. Note that this changes the output of some old tests.
  - Also, `postProcess` now takes a `raise: Diagnostic -> Unit` parameter, so extensions using `postProcess` can now raise errors in the same way as tests in the main project.

TODO:
- Purity check
- Remaining TODOs and unresolved comments 
- Document semantics and non-trivial aspects

After this it should be good to merge.

compiler/shared/main/scala/mlscript/compiler/optimizer/TailRecOpt.scala
```diff
@@ -5,13 +5,18 @@ import mlscript.compiler.ir.Node._
 import scala.annotation.tailrec
 import mlscript.IntLit
 import mlscript.utils.shorthands.Bool
+import mlscript.Diagnostic
+import mlscript.ErrorReport
+import mlscript.Message
+import mlscript.Message.MessageContext
+import mlscript.Loc
```
LPTK: Needless imports. Make it like this instead:
```scala
package mlscript
package compiler.optimizer

import utils.shorthands._
import Message.MessageContext

import compiler.ir._
import compiler.ir.Node._
```

CAG2Mark:
> Instead of writing `class True` and `class False` in _every single test case_, why not make it a built-in class info? It should be a trivial change and would remove much clutter. Could also do the same with `Cons`/`Nil` and `Some`/`None`.

resolved

compiler/shared/main/scala/mlscript/compiler/optimizer/TailRecOpt.scala
CAG2Mark: This doesn't seem to work due to the `implicit` parameters :/

Either way, I don't think it is a big deal so I'll resolve this for now. If you disagree you can re-open this comment.

CAG2Mark:
OK, should be good to merge. There's just one trivial merge conflict in `DiffTests.scala` to handle

compiler/shared/main/scala/mlscript/compiler/ir/IR.scala
```diff
@@ -160,7 +173,9 @@ enum Node:
     case Jump(defn, args) => Jump(defn, args.map(_.mapNameOfTrivialExpr(f)))
     case Case(scrut, cases) => Case(f(scrut), cases.map { (cls, arm) => (cls, arm.mapName(f)) })
     case LetExpr(name, expr, body) => LetExpr(f(name), expr.mapName(f), body.mapName(f))
-    case LetCall(names, defn, args, body) => LetCall(names.map(f), defn, args.map(_.mapNameOfTrivialExpr(f)), body.mapName(f))  
+    case x: LetCall => 
+      val LetCall(names, defn, args, isTailRec, body) = x
+      LetCall(names.map(f), defn, args.map(_.mapNameOfTrivialExpr(f)), isTailRec, body.mapName(f))(x.loc)
```
waterlens, line 178: Consider to change this into
```scala
case x @ LetCall(names, defn, args, isTailRec, body) =>
  LetCall(names.map(f), defn, args.map(_.mapNameOfTrivialExpr(f)), isTailRec, body.mapName(f))(x.loc)
```

compiler/shared/main/scala/mlscript/compiler/ir/IR.scala
```diff
@@ -169,9 +184,10 @@ enum Node:
     case LetExpr(name, expr, body) => 
       val name_copy = name.copy
       LetExpr(name_copy, expr.mapName(_.trySubst(ctx)), body.copy(ctx + (name_copy.str -> name_copy)))
-    case LetCall(names, defn, args, body) => 
+    case x: LetCall => 
+      val LetCall(names, defn, args, isTailRec, body) = x
       val names_copy = names.map(_.copy)
-      LetCall(names_copy, defn, args.map(_.mapNameOfTrivialExpr(_.trySubst(ctx))), body.copy(ctx ++ names_copy.map(x => x.str -> x)))
+      LetCall(names_copy, defn, args.map(_.mapNameOfTrivialExpr(_.trySubst(ctx))), isTailRec, body.copy(ctx ++ names_copy.map(x => x.str -> x)))(x.loc)
```
waterlens, line 190: Same as above

shared/src/test/diff/nu/Annotations.mls
```diff
@@ -40,6 +40,40 @@ Foo
...
+//│ ║  l.65: 	fun foo(x) = 1
+//│ ║        	    ^^^^^^^^^^
+//│ ╟── integer literal of type `1` is not an instance of type `Annotation`
+//│ ║  l.63: 	let x = 1
+//│ ╙──      	        ^
+//│ let x: 1
+//│ fun foo: anything -> 1
+//│ x
+//│   = 1
+
```
waterlens, line 76: Where are these new errors coming from?

CAG2Mark: It's because `@tailrec` and `@tailcall` can only be used on functions and function calls respectively. However, the type system and grammar allow annotations to be applied to any expression, typedef or function.

I suppose the indicated location of the error could be more helpful though

waterlens: OK I see. Thanks.

compiler/shared/main/scala/mlscript/compiler/ir/IR.scala
CAG2Mark: Ah. So that's the syntax for doing that.

I've done that in a lot of places. I'll refactor the code in those places to make it cleaner

LPTK:
> OK, should be good to merge.

Great!

> There's just one trivial merge conflict in `DiffTests.scala` to handle

So... do you intend to ever fix these conflicts? 🤔

compiler/shared/main/scala/mlscript/compiler/ir/Builder.scala
```diff
@@ -307,7 +307,10 @@ final class Builder(fresh: Fresh, fnUid: FreshInt, classUid: FreshInt, tag: Fres
 
       import scala.collection.mutable.{ HashSet => MutHSet }
 
-      val cls = grouped.getOrElse(0, Nil).map(buildClassInfo)
+      val cls = ClassInfo(classUid.make, "True", List()) // TODO: add "True" amd "False" at some pointgrouped.getOrElse(0, Nil).map(buildClassInfo)
```
LPTK: There seems to be an inadvertent copy-paste in this comment.

compiler/shared/main/scala/mlscript/compiler/optimizer/TailRecOpt.scala
```diff
@@ -9,6 +9,62 @@ import Message.MessageContext
 import compiler.ir._
 import compiler.ir.Node._
 
+/*
+
+DOCUMENTATION OF SEMANTICS OF @tailcall and @tailrec
+
+@tailcall: Used to annotate specific function calls. Calls annotated with @tailcall 
+must be tail calls or tail modulo-cons calls. These calls may be optimized to not
```
LPTK:
```suggestion
must be tail calls or tail modulo-cons calls. These calls must be optimized to not
```

compiler/shared/main/scala/mlscript/compiler/optimizer/TailRecOpt.scala
```diff
@@ -9,6 +9,62 @@ import Message.MessageContext
...
+
+fun foo() =
+  let b = bar()
+  let a = @tailcall foo()
+  A(a, b)
+
+because here, you are taking responsibility for the reordering of the computations.
+
+@tailrec: Used to annotate functions. When this annotation is used on a function, say
+@tailrec fun foo(), the compiler will ensure no sequence of recursive calls back to foo()
```
LPTK:
```suggestion
@tailrec fun foo(), the compiler will ensure no sequence of direct recursive calls back to foo()
```

compiler/shared/main/scala/mlscript/compiler/optimizer/TailRecOpt.scala
```diff
@@ -9,6 +9,62 @@ import Message.MessageContext
...
+@tailrec fun foo() = bar()
+fun bar() =
+  foo()
+  bar()
+
+is invalid. If we swap the position of foo() and bar() in the body of bar, it is still invalid.
+
+Equivalently, if fun foo() is annotated with @tailrec, let S be the largest strongly
+connected component in the call-graph of the program that contains foo. Then an error
+will be thrown if and only if all edges (calls) connecting the nodes of the strongly
```
LPTK: Do you mean "unless"?
```suggestion
connected component in the call-graph of the program that contains foo. Then an error
will be thrown unless all edges (calls) connecting the nodes of the strongly
```

compiler/shared/main/scala/mlscript/compiler/optimizer/TailRecOpt.scala
CAG2Mark: What is the difference between "recursive calls" and "direct recursive calls" in this case?

CAG2Mark:
> So... do you intend to ever fix these conflicts? 🤔

I figured doing it now will make the diff a bit harder to review. Once everything else is sorted, I'll fix the conflicts.

compiler/shared/main/scala/mlscript/compiler/ir/Builder.scala
```diff
@@ -79,8 +80,13 @@ final class Builder(fresh: Fresh, fnUid: FreshInt, classUid: FreshInt, tag: Fres
             val v = fresh.make
             
             ann match
-              case Some(Var(nme)) if nme == "tailcall" => 
-                LetCall(List(v), DefnRef(Right(g.str)), args, true, v |> ref |> sresult |> k)(f.toLoc).attachTag(tag)
+              case Some(ann @ Var(nme)) =>
+                if nme == "tailcall" then
```
LPTK:
```suggestion
                if nme === "tailcall" then 
```

LPTK: Please apply the same suggestion on all other occurrences of `==`.

compiler/shared/test/scala/mlscript/compiler/TestIR.scala
```diff
@@ -6,18 +6,40 @@ import mlscript.compiler.ir._
 import scala.collection.mutable.StringBuilder
 import mlscript.{DiffTests, ModeType, TypingUnit}
```
LPTK: Please fix the import gore by using:
```scala
package mlscript
package compiler
```
and removing the useless imports.

compiler/shared/test/scala/mlscript/compiler/TestIR.scala
```diff
@@ -6,18 +6,40 @@ import mlscript.compiler.ir._
...
+
+        val graph = if (!mode.noTailRecOpt) {
+          val tailRecOpt = new TailRecOpt(fnUid, classUid, tag, raise)
+          val (g, comps) = tailRecOpt.run_debug(graph_)
+          output("\nStrongly Connected Tail Calls:")
+          output(comps.toString)
+          g
+        } else {
+          graph_
+        }
```
LPTK: This subproject is on Scala 3, isn't it? It's more idiomatic, elegant, and easier to edit to make it:
```suggestion
        val graph = if !mode.noTailRecOpt
          then
            val tailRecOpt = new TailRecOpt(fnUid, classUid, tag, raise)
            val (g, comps) = tailRecOpt.run_debug(graph_)
            output("\nStrongly Connected Tail Calls:")
            output(comps.toString)
            g
          else
            graph_
```

compiler/shared/test/diff-ir/IR.mls
```diff
@@ -14,28 +15,9 @@ foo()
 //│ 
 //│
```
LPTK: Is there any reason for these two empty lines? One might be enough.

compiler/shared/test/diff-ir/IR.mls
```diff
@@ -14,28 +15,9 @@ foo()
...
-//│ x$3 -- #13
-//│ )
-//│ Def(2, foo, [],
-//│ 1, 
-//│ let* (x$4) = mktup2(1,2) in -- #22
-//│ x$4 -- #21
-//│ )
-//│ },
-//│ let* (x$5) = foo() in -- #26
-//│ x$5 -- #25)
```
LPTK, line 35: Why is this no longer printed? And if that's intended, shouldn't the "IR:" part also be removed?

CAG2Mark: It is intended (see https://github.com/hkust-taco/mlscript/pull/218#discussion_r1598011027); I forgot to remove the `IR:` printout.

compiler/shared/test/diff-ir/IR.mls
CAG2Mark: none that I can see, changed to one line

LPTK (approved):
Amazing change, thanks! Should we merge now? (I'll do it.)

CAG2Mark:
Go ahead

PR #219, opened 2024-04-09
Type Selection

Meowcolm024:
> **Note**: part of #195 

Changes:
- Parsing
  - type member declaration with bounds `Foo[type A extends L restricts U]`
  - type selection `x.A`
  - `as` operator
  - wildcard type `?`
- Type checking:
  - type selection typing


Meowcolm024:
**Not** included in this PR:
- pushing type alias member to type members i.e. `class Foo { type T }`

shared/src/test/diff/nu/Metaprog.mls
```diff
@@ -28,7 +28,7 @@ abstract class Test[C] {
   // * Represents what happens in "... ${input} ..." when a binding of C is in scope
   fun unquote: (input: Code['a, C | 'c]) -> Code[Int, 'c]
   fun getVar: Code[Int, C]
-  fun test0 = this.unquote of IntLit(1)
+  fun test0 = this.unquote of (IntLit(1) : Code['ia, 'ic]) // ? why ascription works here
```
LPTK, line 31: Why the weird change?

shared/src/test/diff/nu/NuScratch.mls
```diff
@@ -1,3 +1,292 @@
...
+//│ Code generation encountered an error:
+//│   if expression was not desugared
+
+
+
+class C(x: Int, y: Str)
+//│ class C(x: Int, y: Str)
+
+
+:d
```
LPTK, line 101: Debug trace left here...

LPTK (commented):
I just noticed that I have old review comments pending that I never submitted; releasing them now.

LPTK:
Thanks for the great work!

However, as you know, we've moved to a new version of the compiler, which will have a brand new type checker implementation. So while your PR is useful for future reference when we reimplement similar features in the new compiler (I will add a tag to that effect), it's probably not worth fixing the merge conflicts with the old type checker.

PR #220, opened 2024-04-22
Optional Arguments

noordahx:
Draft PR:
* Resolved merge conflicts of [original PR](https://github.com/hkust-taco/mlscript/pull/183)
* Tests failed at OptionalArgs.mls
```
fun f1(a: Int, b: Int) = a + b
//│ ╔══[ERROR] Type mismatch in operator application:
//│ ║  l.17: 	fun f1(a: Int, b: Int) = a + b
//│ ║        	                         ^^^^^
//│ ╟── type `()` is not an instance of type `Int`
//│ ║  l.17: 	fun f1(a: Int, b: Int) = a + b
//│ ║        	          ^^^
//│ ╟── but it flows into reference with expected type `Int`
//│ ║  l.17: 	fun f1(a: Int, b: Int) = a + b
//│ ╙──      	                         ^
//│ TEST CASE FAILURE: There was an unexpected type error
//│ fun f1: (a: Int | (), b: Int | ()) -> (Int | error)
```

### DIFF
--- a/compiler/shared/main/scala/mlscript/compiler/ir/Builder.scala
+++ b/compiler/shared/main/scala/mlscript/compiler/ir/Builder.scala
@@ -4,11 +4,12 @@ import mlscript.compiler.optimizer.FreeVarAnalysis
 import mlscript.utils.shorthands._
 import mlscript.utils._
 import mlscript._
+import mlscript.Message._
 import collection.mutable.ListBuffer
 
 final val ops = Set("+", "-", "*", "/", ">", "<", ">=", "<=", "!=", "==")
 
-final class Builder(fresh: Fresh, fnUid: FreshInt, classUid: FreshInt, tag: FreshInt):
+final class Builder(fresh: Fresh, fnUid: FreshInt, classUid: FreshInt, tag: FreshInt, raise: Diagnostic => Unit):
   import Node._
   import Expr._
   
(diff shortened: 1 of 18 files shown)


LPTK:
Hi @noordahx, thanks for the PR draft. Do you need some assistance in moving it forward?

shared/src/test/diff/nu/OptionalArgs.mls
```diff
@@ -0,0 +1,304 @@
...
+fun f1(a: Int, b: Int) = a + b
+//│ ╔══[ERROR] Type mismatch in operator application:
+//│ ║  l.17: 	fun f1(a: Int, b: Int) = a + b
+//│ ║        	                         ^^^^^
+//│ ╟── type `()` is not an instance of type `Int`
+//│ ║  l.17: 	fun f1(a: Int, b: Int) = a + b
+//│ ║        	          ^^^
+//│ ╟── but it flows into reference with expected type `Int`
+//│ ║  l.17: 	fun f1(a: Int, b: Int) = a + b
+//│ ╙──      	                         ^
```
noordahx: Hi @LPTK , I am having trouble with function parameter type parsing (seems all of the parameter types are '()' by default)

noordahx: NewParser.scala "?" parser -> [line](https://github.com/hkust-taco/mlscript/blob/7648ae331477f5c231c5c798b26d30bae68b5346/shared/src/main/scala/mlscript/NewParser.scala#L1428)

LPTK: That code does not make much sense to me. You may have mismerged the conflicts. The code is computing `(argName, argOpt)` by consuming an identifier, `?`, and `:` tokens. And then it redundantly checks optionality with `isOptional` – the logic is all messed up. I'm guessing the `()` might come from consuming the type tokens and then parsing an empty expression (which produces `undefined`, aka `()`), or it could be the later logic to make things optional that's broken. It might be better to start the parsing change from scratch.

PR #221, opened 2024-05-17
Fix statement desugar in `Unquote`

NeilKleistGao:
Even though quoted statements are not supported, we can still have statements in unquoted blocks:
```ts
x `=>
  let v = Ref(x)
  let _ = y `=>
    set v.value = y
    `0
  v.value
```
Note that `let v = Ref(x)` and `set v.value = y` are not quoted code fragments since we have not supported corresponding syntax.

This PR fixed the desugar problem of statements in unquoted blocks and we can generate JS code for following program:
```ts
:ne
:js
x `=>
  let v = Ref(x)
  let _ = y `=>
    set v.value = y
    `0
  v.value
//│ Code[forall 'a. 'a -> 'a, ??y & ~??x]
//│ // Prelude
//│ class TypingUnit30 {}
//│ const typing_unit30 = new TypingUnit30;
//│ // Query 1
//│ res = ((x1) => Lam(Var(x1), (() => {
//│   let v = Ref(Var(x1));
//│   let _ = ((y1) => Lam(Var(y1), (() => {
//│     void(v.value = Var(y1));
//│     return IntLit(0);
//│   })()))(freshName("y"));
//│   return v.value;
//│ })()))(freshName("x"));
//│ // End of generated code
```

More examples are in `shared/src/test/diff/qq/Codegen.mls`.


LPTK (approved):
LGTM

PR #223, opened 2024-07-08
Optional arguments

noordahx:
draft pr


shared/src/test/diff/nu/OptionalArgs.mls
```diff
@@ -9,8 +9,8 @@ fun (??) oops(a, b) = a + b
 //│     = 3
 
 
-fun f1(a?: Int, b?: Int) = a + b
-//│ fun f1: (a: (Int)?, b: (Int)?) -> Int
+fun f1(aOpt?: Int, bOpt?: Int) = aOpt + bOpt
```
noordahx, line 12: @LPTK how can I handle this case?
should it return an error requiring `a` and `b` defined before `a + b`?

LPTK: The idea of this approach is that if your parameter `x` of type `T` is optional (annotated with `?`), then while the corresponding parameter type remains `T` when viewed from the outside (but of course, the parameter is still marked as optional), as in `(? T) -> Int`, the *visible* type of `x` inside the body is `T | undefined`. That's all we need for the feature to work correctly.

LPTK:
Thanks for looking into this at the time @noordahx. We now have a new compiler implementation (in branch `hkmc2`) where all the current work is going. Optional arguments will be implemented there directly.

PR #225, opened 2024-07-23
Sync IR changes

waterlens:


CAG2Mark:
The trivial changes look fine. However, there is a pain point with `LetMethodCall`, which from what I know, is used for virtual dispatch.

We definitely want to optimize method calls. For instance, `List.map` and `List.foldLeft`, when implemented in a recursive manner, should definitely have the tail-rec (mod cons) optimizations applied. However, the optimization is quite aggressive, in the sense that it needs to analyze the code itself, pluck out certain bits of code, replace them with other things and then possibly merge several functions into one. This is fundamentally incompatible with `LetMethodCall`, which defers resolving the virtual method calls to a later stage.

## Temporary Solution

One **temporary** solution is to reduce `LetMethodCall`s to `LetCall`s when we can resolve the method call at compile-time and throw an error if a `LetMethodCall` appears inside a `@tailrec` strongly-connected component of the call graph. Specifically, if 
- `V <: T`,
- `foo` is a virtual method of `T`,
- `V` implements `foo`,
- `x: V` implies `x.foo` resolves to `V`'s implementation of `foo` (note that if `x: W`, then `W <: V` is possible),

then for any `x: V`, we can resolve `x.foo` to `V`'s implementation of `foo`.

## Possible Fix

I think the cleanest fix is to scrap `LetMethodCall` (or only have it in an intermediate processing step and not in the final IR) and implement virtual dispatch in the IR directly. From what I know, we plan to implement a `switch` statement in the IR which matches the scrutinee in constant time. Either that, or it is directly implemented into the current `Case` node. In either case, it will be adequately performant.

As an example, if we have
```fs
fun bar(x: T, y: Int): Int = x.foo // --#1

trait T {
    fun foo: Int  // --#2
}

class T1 extends T {
    fun foo = bar(this, 0)
}

class T2 extends T {
    fun foo = bar(this, 1)
}
```
Then we would have any instance of `T` have an extra parameter `clsId`. Let's say that `T1` has `clsId = 1` and `T2` has `clsId = 2`. In the IR, we would then have functions `T1_foo(this_: T)` and `T2_foo(this_: T)` as `T1` and `T2`'s implementations of `foo`.

Then `x.foo` at `#1` will be replaced with a call to `T_foo_lookup`, and `#2`'s "implementation" will be `T_foo_dispatch`, which would look something like
```fs
fun T_foo_dispatch(this_: T, clsId: Int) = switch clsId {
  case 1 => T1_foo(this_)
  case 2 => T2_foo(this_)
}
```
Now, note that every single function call in the strongly-connected component { `bar`, `T_foo_dispatch`, `T1_foo` and `T2_foo` } are tail-calls. Therefore, the tailrec optimizer can optimize them into one function.

Of course, this idea is provisional for now and I would like @LPTK's approval before going ahead with either of these ideas, if at all.

CAG2Mark (changes requested):
I've skimmed the changes to `TailRecOpt.scala` and the trivial changes look fine, but something needs to be done about `LetMethodCall`. See my bigger comment above.

My tailrec unit tests pass so it seems nothing existing has been broken.

LPTK:
@CAG2Mark Thanks for the comments.

What you're proposing is fine in principle, but it makes a close-world assumption. However, the IR should be able to represent and optimize open-world code, as we won't always have the entire program on hand (modules could even be loaded dynamically), or in other situations we _might_ have the entire program available but global whole-program optimizations might be too expensive (esp. for debug/test builds) if that program is very big.

It's better to cleanly separate these distinct concerns. `LetMethodCall` is meant to be for those calls we don't resolve statically, either because we can't or because we don't want to analyze/optimize them. If the source method call is to a static or final method, the elaborator or IR lifter might probably change it into a `LetCall` anyway before it even reaches the IR. Or there might be a separate IR pass that does this change. `@tailrec` should simply be blind to `LetMethodCall` and that's fine. It is a reasonable restriction that `@tailrec` only be specified to handle direct/concrete calls.

> I think the cleanest fix is to scrap `LetMethodCall` (or only have it in an intermediate processing step and not in the final IR) and implement virtual dispatch in the IR directly.

But `LetMethodCall` **_is_**  our direct implementation of virtual dispatch already...

What you're proposing here is not a virtual dispatch implementation, it's a _defunctionalized_ (or more specifically "_demethodized_") representation of it. Defunctionalization is an orthogonal IR transformation pass. Currently we plan to do it only internally to modules/compilation units. Virtual calls across compilation units will remain virtual by default to preserve separate compilation.

Besides, your defunctionalization proposal is not very good as it assumes all call sites have to assume all possible class implementations flow into the receiver. In practice we'll want to base this on a local flow analysis, so that each method call will become a switch of only those classes that can flow there, with possibly a default case resolving to the original virtual call, if unknown implementations may flow there.

CAG2Mark:
Got it, so now we have a question of semantics: If a function is annotated with @tailrec, and we have an unresolved call, which could possibly recurse back to the annotated function, should we ignore it or should we error?

I suppose the semantics should be the same as calls to anonymous functions (which can be implemented using virtual dispatch anyway) but I don't exactly remember the required semantics of those either.

compiler/shared/main/scala/mlscript/compiler/codegen/CppAst.scala
```diff
@@ -0,0 +1,198 @@
...
+
+enum Specifier:
+  case Extern
+  case Static
+  case Inline
+
+  def toDocument = this match
+    case Extern => "extern" |> raw 
+    case Static => "static" |> raw
+    case Inline => "inline" |> raw
```
LPTK:
```suggestion
  def toDocument = raw:
    this match
    case Extern => "extern"
    case Static => "static"
    case Inline => "inline"
```

compiler/shared/main/scala/mlscript/compiler/codegen/CppAst.scala
```diff
@@ -0,0 +1,198 @@
...
+
+  override def toString(): Str = toDocument.print
```
LPTK: In Scala, it's idiomatic that methods assumed to be without side effects do not take an extra pair of empty parens.
```suggestion
  override def toString: Str = toDocument.print
```
The same remark applies everywhere else you do this.

LPTK:
> Got it, so now we have a question of semantics: If a function is annotated with @tailrec, and we have an unresolved call, which could possibly recurse back to the annotated function, should we ignore it or should we error?

Yes, the intended semantics is to ignore them. Just like in Scala (in fact Scala ignores even more calls, as you know). `@tailrec` does not guarantee that the same function won't ever be called indirectly in non-tail position. In general it doesn't guarantee that the body won't overflow the stack due to calls to other methods, and for all intents and purposes these indirect calls aren't so different from other methods. Think of a tail-recursive implementation of `List.fold(f)` and a call that wants to fold sublists as part of `f`'s implementation. We certainly don't want that to be rejected as "not tail-recursive". Doing so would also be anti-modular anyway.

CAG2Mark:
Got it, the changes should be fairly simple then. I'll suggest them soon.

compiler/shared/main/scala/mlscript/compiler/codegen/CppAst.scala
```diff
@@ -0,0 +1,198 @@
...
+        if i == 0 then
+          x.toDocument(extraTypename)
+        else
+          sep <#> x.toDocument(extraTypename)
+    }.fold(raw(""))(_ <#> _)
+
+  def toDocuments(args: Ls[(Str, Type)], sep: Document): Document =
+    args.zipWithIndex.map {
+      case (x, i) =>
+        if i == 0 then
```
LPTK:
```suggestion
      case (x, 0) =>
```

compiler/shared/main/scala/mlscript/compiler/codegen/CppAst.scala
```diff
@@ -0,0 +1,198 @@
...
+  def toDocument = this match
+    case Extern => "extern" |> raw 
+    case Static => "static" |> raw
+    case Inline => "inline" |> raw
+
+  override def toString(): Str = toDocument.print
+
+object Type:
+  def toDocuments(args: Ls[Type], sep: Document, extraTypename: Bool = false): Document =
+    args.zipWithIndex.map {
```
LPTK: When transforming a list just for the purpose of folding it later, better use an iterator to avoid intermediate allocations.
```suggestion
    args.iterator.zipWithIndex.map {
```
The same remark applies everywhere else you do this, like in the next function.

compiler/shared/main/scala/mlscript/compiler/codegen/CppAst.scala
```diff
@@ -0,0 +1,198 @@
...
+      case Prim(name) => name |> raw
+      case Ptr(inner) => aux(inner) <#> raw("*") 
+      case Ref(inner) => aux(inner) <#> raw("&")
+      case Array(inner, size) => aux(inner) <#> raw("[") <#> size.fold(raw(""))(x => x.toString |> raw) <#> raw("]")
+      case FuncPtr(ret, args) => aux(ret) <#> raw("(") <#> Type.toDocuments(args, sep = raw(", ")) <#> raw(")")
+      case Struct(name) => raw(s"struct $name")
+      case Enum(name) => raw(s"enum $name")
+      case Template(name, args) => raw(s"$name") <#> raw("<") <#> Type.toDocuments(args, sep = raw(", ")) <#> raw(">")
+      case Var(name) => name |> raw
+      case Qualifier(inner, qual) => aux(inner) <:> raw(qual)
```
LPTK: Is this output correctly? If the qualifier is `virtual`, like [here](https://github.com/hkust-taco/mlscript/pull/225/files#diff-3f1c396f79b8b1e9904a56a8f922f74bb89351b6c5faf21e31f343ef67c91bfeR56), you put it _after_ the type?

compiler/shared/main/scala/mlscript/compiler/codegen/CppAst.scala
```diff
@@ -0,0 +1,198 @@
...
+  case For(init: Stmt, cond: Expr, update: Stmt, body: Stmt)
+  case ExprStmt(expr: Expr)
+  case Break
+  case Continue
+  case Block(decl: Ls[Decl], stmts: Ls[Stmt])
+  case Switch(expr: Expr, cases: Ls[(Expr, Stmt)])
+  case Raw(stmt: Str)
+
+  def toDocument: Document =
+    def aux(x: Stmt): Document = x match
```
LPTK, line 89: Functions like this with millions of `raw` calls make me think we might want to define a `Conversion[Str, Document]`. Maybe try it? Then you can replace all `raw(x)` by just `x`. There are over a hundred occurrences of `raw` in this file.

compiler/shared/main/scala/mlscript/compiler/codegen/CppAst.scala
```diff
@@ -0,0 +1,198 @@
...
+    args.zipWithIndex.map {
+      case (x, i) =>
+        if i == 0 then x.toDocument
+        else sep <#> x.toDocument
+    }.fold(raw(""))(_ <#> _)
+  
+enum Expr:
+  case Var(name: Str)
+  case IntLit(value: BigInt)
+  case FloatLit(value: Float)
```
LPTK: Why do we have Float and not Double?

compiler/shared/main/scala/mlscript/compiler/codegen/CppCodeGen.scala
```diff
@@ -0,0 +1,242 @@
+package mlscript.compiler.codegen.cpp
+
+import mlscript.compiler.ir.{Expr => IExpr, _}
+import mlscript.compiler.utils._
+import mlscript.utils._
+import mlscript.utils.shorthands._
+
+
+class CppCodeGen:
```
LPTK: What's the point of making everything `private` here? It seems this is only used as `CppCodeGen().codegen(graph)`.

Instead, just make the class itself private and use a top-level `codegen` function:

```
private class CppCodeGen:
  ...
def codegen(prog: Program) = 
```

compiler/shared/main/scala/mlscript/compiler/codegen/CppCodeGen.scala
```diff
@@ -0,0 +1,242 @@
...
+      Stmt.Block(Ls(), Ls(Stmt.Return(Expr.Call(Expr.Var(fn), (0 until n).map{x => Expr.Var(s"arg$x")}.toList)))), true)
+  private def mlsNeverValue(n: Int) = if (n <= 1) then Expr.Call(Expr.Var(s"_mlsValue::never"), Ls()) else Expr.Call(Expr.Var(s"_mlsValue::never<$n>"), Ls())
+
+  private case class Ctx(
+    val defnCtx: Set[Str],
+  )
+
+  private def codegenClassInfo(using ctx: Ctx)(cls: ClassInfo): (Opt[Def], Decl) =
+    val fields = cls.fields.map{x => (x |> mapName, mlsValType)}
+    val parents = if cls.parents.nonEmpty then cls.parents.toList.map{x => x |> mapName} else mlsObject :: Nil
```
LPTK:
```suggestion
    val parents = if cls.parents.nonEmpty then cls.parents.toList.map(mapName) else mlsObject :: Nil
```

compiler/shared/main/scala/mlscript/compiler/codegen/CppCodeGen.scala
```diff
@@ -0,0 +1,242 @@
...
+    case "==" => Expr.Binary("==", toExpr(args(0)), toExpr(args(1)))
+    case "!=" => Expr.Binary("!=", toExpr(args(0)), toExpr(args(1)))
+    case "<" => Expr.Binary("<", toExpr(args(0)), toExpr(args(1)))
+    case "<=" => Expr.Binary("<=", toExpr(args(0)), toExpr(args(1)))
+    case ">" => Expr.Binary(">", toExpr(args(0)), toExpr(args(1)))
+    case ">=" => Expr.Binary(">=", toExpr(args(0)), toExpr(args(1)))
+    case "&&" => Expr.Binary("&&", toExpr(args(0)), toExpr(args(1)))
+    case "||" => Expr.Binary("||", toExpr(args(0)), toExpr(args(1)))
+    case "!" => Expr.Unary("!", toExpr(args(0)))
+    case _ => ???
```
LPTK: Is this a TODO? Then use `mlscript.utils.TODO`. Is this an impossible case? Then instead of `op: Str` use `op: "+" | "-" | ...` or at least use `mlscript.utils.die`.

compiler/shared/main/scala/mlscript/compiler/codegen/CppCodeGen.scala
```diff
@@ -0,0 +1,242 @@
...
+      codegen(body, storeInto)(using decls, stmts2)
+    // Use method calls instead of apply
+    // 
+    // case Node.LetApply(names, fn, args, body) if fn.str == "builtin" =>
+    //   val stmts2 = stmts ++ codegenBuiltin(names, args.head.toString.replace("\"", ""), args.tail)
+    //   codegen(body, storeInto)(using decls, stmts2)
+    // case Node.LetApply(names, fn, args, body) =>
+    //   val call = mlsCall(fn.str |> mapName, args.map(toExpr))
+    //   val stmts2 = stmts ++ Ls(Stmt.AutoBind(names.map(mapName), call))
+    //   codegen(body, storeInto)(using decls, stmts2)
```
LPTK: Is this commented code useful as a future reference? If not, better delete it.

compiler/shared/main/scala/mlscript/compiler/codegen/CppCodeGen.scala
```diff
@@ -0,0 +1,242 @@
...
+      degree -= node
+      depgraph -= node
+      depgraph = depgraph.view.mapValues(_.filter(_ != node)).toMap
+      degree = depgraph.view.mapValues(_.size).toMap
+    var sorted = Ls.empty[ClassInfo]
+    var work = degree.filter(_._2 == 0).keys.toSet
+    while work.nonEmpty do
+      val node = work.head
+      work -= node
+      sorted = sorted :+ prog.classes.find(_.name == node).get
```
LPTK: This is terribly inefficient. Just use a `mutable.Buffer` (such as a `mutable.ListBuffer`) assigned to a `val` here, instead of a `List` assigned to a `var`.

compiler/shared/main/scala/mlscript/compiler/codegen/CppCodeGen.scala
```diff
@@ -0,0 +1,242 @@
...
+  private def codegenTopNode(node: Node)(using ctx: Ctx): (Def, Decl) =
+    val decls = Ls(mlsRetValueDecl)
+    val stmts = Ls.empty[Stmt]
+    val (decls2, stmts2) = codegen(node, mlsRetValue)(using decls, stmts)
+    val stmtsWithReturn = stmts2 :+ Stmt.Return(Expr.Var(mlsRetValue))
+    val theDef = Def.FuncDef(mlsValType, mlsMainName, Ls(), Stmt.Block(decls2, stmtsWithReturn))
+    val decl = Decl.FuncDecl(mlsValType, mlsMainName, Ls())
+    (theDef, decl)
+
+  private def sortClasses(prog: Program): Ls[ClassInfo] =
```
LPTK: Add documentation that this performs a topological sort of the classes based on their inherited parents.

shared/src/test/scala/mlscript/DiffTests.scala
```diff
@@ -483,13 +497,12 @@ class DiffTests
...
+            
             val newMode = if (useIR) { mode.copy(useIR = true) } else mode
             val newNewMode = if (noTailRec) { newMode.copy(noTailRecOpt = true) } else newMode
+            val (postLines, nuRes) = postProcess(newNewMode, basePath, testName, res, output, raise)
+            postLines.foreach(output)
 
-            val (postLines, nuRes) = 
-              postProcess(newNewMode, basePath, testName, res, output, raise)
-            postLines.foreach(output)            
-
```
LPTK, line 494: Please revert the needless whitespace changes. Why do you remove the empty-line separation between these lines and the previous block?

shared/src/main/scala/mlscript/NewParser.scala
```diff
@@ -1475,7 +1475,6 @@ abstract class NewParser(origin: Origin, tokens: Ls[Stroken -> Loc], newDefs: Bo
     
   }
   }
-
```
LPTK: Needless change to revert.

shared/src/main/scala/mlscript/NewLexer.scala
```diff
@@ -226,7 +280,7 @@ class NewLexer(origin: Origin, raise: Diagnostic => Unit, dbg: Bool) {
     def isQuasiquoteTripleOpening(i: Int): Bool =  matches(i, BracketKind.QuasiquoteTriple.beg, 0)
     def isUnquoteOpening(i: Int): Bool = matches(i, BracketKind.Unquote.beg, 0)
     def isQuasiquoteTripleClosing(i: Int): Bool = matches(i, BracketKind.QuasiquoteTriple.end, 0)
-    // @inline 
+    // inline
```
LPTK: Needless change to revert.

compiler/shared/test/scala/mlscript/compiler/TestIR.scala
```diff
@@ -45,16 +42,30 @@ class IRDiffTestCompiler extends DiffTests {
...
         case err: Exception =>
           output(s"\nIR Processing Failed: ${err.getMessage()}")
-          if (mode.irVerbose) then
-            output("\n" ++ err.getStackTrace().map(_.toString()).mkString("\n"))
+          output("\n" ++ err.getStackTrace().map(_.toString()).mkString("\n"))
         case err: StackOverflowError =>
           output(s"\nIR Processing Failed: ${err.getMessage()}")
-          if (mode.irVerbose) then
-            output("\n" ++ err.getStackTrace().map(_.toString()).mkString("\n"))
+          output("\n" ++ err.getStackTrace().map(_.toString()).mkString("\n"))
```
LPTK: These `.map(_.toString())` calls are redundant.

compiler/shared/test/diff-ir/cpp/Makefile
```diff
@@ -0,0 +1,26 @@
+CXX := clang++
+CFLAGS := $(CFLAGS) -O3 -Wall -Wextra -std=c++20 -I. -I/opt/homebrew/include -Wno-inconsistent-missing-override
+LDFLAGS := $(LDFLAGS) -L/opt/homebrew/lib -lmimalloc -lgmp
```
LPTK: I had trouble with this. I don't have a `/opt/homebrew/lib` directory.

compiler/shared/test/diff-ir/IRTailRec.mls
LPTK, line 1: These files are becoming too big and impossible to read – there's way too much noise in the output.

Suggestions:
- [x] Don't show the default stuff that's inserted in all the test blocks, such as all the `applyn` functions.
- [x] Pretty-print the declarations. Eg instead of having:
    ```
    //│ apply3 -> Def(5, apply3, [x0$2,x1$1,x2$0],
    //│ 1, 
    //│ 0 -- #32
    //│ ),
    ```
    make it:
    ```
    def apply3(x0$2,x1$1,x2$0) = 0 -- #32
    ```
    or something like that.
    You also don't have to show all the metadata (a verbose flag could be used for that).

LPTK (changes requested):
I started to review your changes and already have several comments.

First of all, where are all the tests? This PR introduces many diff-test options but basically does not test them. C++ code-gen is virtually untested. At the very least there should be a couple simple but representative tests that show the generated C++ code as part of their diff output. And this output should not be polluted with all the predefined boilerplate inserted on top of every C++ generated file; it should only have the specifics of the tested code block.

Second, you should _always_ use `:NewDefs` and you should _not_ use `:ParseOnly` in the tests. Just fix the resulting errors. We're not trying to create two separate languages. This code is still MLscript code and there's no reason not to also test the JS backend at the same time and compare that the results are the same/similar. If builtin classes like True and False are needed, just adapt the current builtin stuff accordingly in the type checker and JS backend.

Third, there are no instructions on how to install things. The project seems to require a bunch of stuff like Boost and mimalloc. We need to create a nix configuration to get these in order automatically without the users having to fiddle with this and the C++ tests should actually be compiled and run as part of the DiffTests. This could be done in a separate PR tho, possibly with the help of our resident nix nerd @pca006132.

LPTK:
I think one important thing to do to make testing the IR more practical is to allow later test blocks to depend on earlier test blocks. This way we could define `class True` and `class False` once in text at the beginning of the file and use them in later blocks. And both he IR and the C++ codegen should take these previous defs into consideration.

LPTK:
BTW nice work! The code overall is pretty clean and nice to read.

compiler/shared/main/scala/mlscript/compiler/codegen/CppAst.scala
waterlens: It seems to be a syntax error?

This version is fine:
```scala
  def toDocument = raw(
    this match
    case Extern => "extern"
    case Static => "static"
    case Inline => "inline"
  )
```

LPTK: Right, it needs a more recent version of Scala 3. Let's update the build to the latest one.

compiler/shared/main/scala/mlscript/compiler/codegen/CppAst.scala
waterlens: `virtual` shouldn't be a type qualifier. My fault. For C++, only `const` and `volatile` are type qualifiers.
The code in the link is not used anymore. So I fixed it by just removing it.

LPTK:
FYI I fixed the DiffTests setup on the main branch; you should merge it into this branch. Now the  diff-tests behave like in the main project: if there are unstaged changes, only those files are re-tested.

compiler/shared/main/scala/mlscript/compiler/optimizer/TailRecOpt.scala
```diff
@@ -340,6 +342,9 @@ class TailRecOpt(fnUid: FreshInt, classUid: FreshInt, tag: FreshInt, raise: Diag
                         // If this assignment overwrites the mod cons value, forget it
                         if containingCtors.contains(assignee) then invalidateAndCont(body)
                         else searchOptCalls(body)
+      case LetMethodCall(names, cls, method, args, body) =>
+        // TODO
+        ???
```
CAG2Mark: Unresolved method calls such as virtual calls cannot be analyzed, so we ignore them
```suggestion
        // method call is unresolved, just ignore it
        searchOptCalls(body)(acc, src, scc, start, calledDefn, letCallNode, letCtorNode, containingCtors -- names.toSet) // `containingCtors -- names.toSet` takes care of variable shadowing
```

waterlens:
> I think one important thing to do to make testing the IR more practical is to allow later test blocks to depend on earlier test blocks. This way we could define `class True` and `class False` once in text at the beginning of the file and use them in later blocks. And both he IR and the C++ codegen should take these previous defs into consideration.

Currently, these frequently used classes are defined internally. Are we going to move it back into the source file?

compiler/shared/main/scala/mlscript/compiler/optimizer/TailRecOpt.scala
waterlens: What's the definition of `newMap`?

waterlens:
> I started to review your changes and already have several comments.
> 
> First of all, where are all the tests? This PR introduces many diff-test options but basically does not test them. C++ code-gen is virtually untested. At the very least there should be a couple simple but representative tests that show the generated C++ code as part of their diff output. And this output should not be polluted with all the predefined boilerplate inserted on top of every C++ generated file; it should only have the specifics of the tested code block.
> 
> Second, you should _always_ use `:NewDefs` and you should _not_ use `:ParseOnly` in the tests. Just fix the resulting errors. We're not trying to create two separate languages. This code is still MLscript code and there's no reason not to also test the JS backend at the same time and compare that the results are the same/similar. If builtin classes like True and False are needed, just adapt the current builtin stuff accordingly in the type checker and JS backend.
> 
> Third, there are no instructions on how to install things. The project seems to require a bunch of stuff like Boost and mimalloc. We need to create a nix configuration to get these in order automatically without the users having to fiddle with this and the C++ tests should actually be compiled and run as part of the DiffTests. This could be done in a separate PR tho, possibly with the help of our resident nix nerd @pca006132.

By the way, for class field selection, the IR builder only accepts syntax like `Pair.x(a_pair)`. How to make it compatible with the standard syntax? The syntax above will trigger this error. I guess they only accept `a_pair.x`?
```
//│ ╟── reference of type `forall ?A ?B. (x: A, y: B) -> Pair[?A, ?B]` does not have field 'x'
//│ ║  l.698: 	fun foo(a) = Pair.x(a) + Pair.y(a)
```

LPTK:
> Currently, these frequently used classes are defined internally. Are we going to move it back into the source file?

It might be the simplest way to do it. Because currently in the legacy compiler there's no other easy way of adding new classes visible to all of: the type checker, the JS backend, and the C++ backend. It's probably not a big deal to declare a couple of classes in every test file.

LPTK:
> By the way, for class field selection, the IR builder only accepts syntax like `Pair.x(a_pair)`. How to make it compatible with the standard syntax? The syntax above will trigger this error. I guess they only accept `a_pair.x`?

Ah, right, the type checker would need to be updated to support this style. Ok, let's do this in a different PR, then, and after moving to the new compiler frontend.

waterlens:
> > Currently, these frequently used classes are defined internally. Are we going to move it back into the source file?
> 
> It might be the simplest way to do it. Because currently in the legacy compiler there's no other easy way of adding new classes visible to all of: the type checker, the JS backend, and the C++ backend. It's probably not a big deal to declare a couple of classes in every test file.

The internal definitions have been moved outside. They are hidden to make the output clear.
I also added pretty printers for the IR output. It should look nice now.

waterlens:
> Third, there are no instructions on how to install things. The project seems to require a bunch of stuff like Boost and mimalloc. We need to create a nix configuration to get these in order automatically without the users having to fiddle with this and the C++ tests should actually be compiled and run as part of the DiffTests. This could be done in a separate PR tho, possibly with the help of our resident nix nerd @pca006132.

By the way, John mentioned that it looks funny to test Cpp backend in target `compilerJVM/test`. Shall we add a target for this? He has drafted a nice Nix conf that invokes `sbt` to run the given test.

compiler/shared/main/scala/mlscript/compiler/optimizer/TailRecOpt.scala
CAG2Mark: my bad, typo (fixed)

LPTK:
> By the way, John mentioned that it looks funny to test Cpp backend in target `compilerJVM/test`. Shall we add a target for this?

What? He must be misunderstanding what `compilerJVM/test` means. This just means the MLscript compiler source code is being tested by being run on the JVM. The _output_ code of course has to be run on the intended target. That's already the case when running test output on NodeJS in most tests. The result of running the output code should be summarized into the DiffTest as the testing progresses, so I don't see how we'd separate this into a different target.

> He has drafted a nice Nix conf that invokes `sbt` to run the given test.

Nice! But there's no need to invoke SBT. It's the SBT-ran tests that should invoke the C++ compiler using the toolchains set up by nix. Is that possible? It should be.

waterlens:
> > By the way, John mentioned that it looks funny to test Cpp backend in target `compilerJVM/test`. Shall we add a target for this?
> 
> What? He must be misunderstanding what `compilerJVM/test` means. This just means the MLscript compiler source code is being tested by being run on the JVM. The _output_ code of course has to be run on the intended target. That's already the case when running test output on NodeJS in most tests. The result of running the output code should be summarized into the DiffTest as the testing progresses, so I don't see how we'd separate this into a different target.
> 
> > He has drafted a nice Nix conf that invokes `sbt` to run the given test.
> 
> Nice! But there's no need to invoke SBT. It's the SBT-ran tests that should invoke the C++ compiler using the toolchains set up by nix. Is that possible? It should be.

OK got it. I will modify the conf to make it.

compiler/shared/main/scala/mlscript/compiler/ir/Builder.scala
```diff
@@ -60,64 +97,212 @@ final class Builder(fresh: Fresh, fnUid: FreshInt, classUid: FreshInt, tag: Fres
...
+  private def freeVariables(tu: TypingUnit): Set[Str] =
+    var all_defined = tu.rawEntities.foldLeft(Set.empty[Str]) {
+      case (defined, stmt) => defined ++ freeVariables(stmt)._1
+    }
+    tu.rawEntities.foldLeft(Set.empty[Str]) {
+      case (acc, stmt) => acc ++ (freeVariables(stmt)._2 -- all_defined)
+    }
+
+    // for this fv analysis, we also need to return the variables that are defined in this statement
+  private def freeVariables(stmt: Statement): (Set[Str], Set[Str]) = stmt match
```
LPTK, line 144: Any reason you're not reusing or at least adapting/fixing this one?

https://github.com/hkust-taco/mlscript/blob/4a5a0387a6bd12c93de153a6c510f868f7342aa2/shared/src/main/scala/mlscript/helpers.scala#L895

LPTK: Ok I see your version returns two sets. Is that really necessary?

waterlens: Oops, never noticed it. I will try to see if it could replace the implementation in the IR builder.

LPTK:
If the `compiler` project's tests now need to run only from the nix command (which I don't like), you should change the `build.sbt` so `compilerJVM` is not part of the `.aggregate` root project.

But I don't like this very much. Can't we just run something like "nix install" and then make sure the correct tools are used when calling them from SBT?

waterlens:
> If the `compiler` project's tests now need to run only from the nix command (which I don't like), you should change the `build.sbt` so `compilerJVM` is not part of the `.aggregate` root project.
> 
> But I don't like this very much. Can't we just run something like "nix install" and then make sure the correct tools are used when calling them from SBT?

I think Nix never changes things outside the environment, rather they have overlays where you have your settings and then offer you commands to go inside to use it.

It could be quite slow to invoke `nix` like using `make` in DiffTests, because we need to do that for every single `runCpp` test. If we want that, it would also require us to write the intermediate cpp code into a file in the source tree, otherwise, I don't know how to put it into the nix env ...

pca006132:
You can treat `nix develop` as setting up a docker/vm, and it has the environment variables set inside it.

Technically you can install things globally, but it is not idiomatic and kind of violates the nix philosophy.

LPTK:
@pca006132 Could you please describe the full solution that using `nix develop` implies?
1. What are the commands, if any, that the CI _and local users_ need to run in order to install the tools? (Of course, no global installs!)
2. What are the commands that the SBT process is supposed to shell out in order to run the C++ compiler?

LPTK:
It's probably fine if we have to use the SBT command from some sort of special command like `<setup nix env> sbt test`

pca006132:
CI: `nix develop --command sbt compilerJVM/test`.

Local user:
```sh
# enter shell
nix develop
# inside the shell...
sbt compilerJVM/test
```

I typically use [direnv](https://direnv.net/). If you use it with the file `.envrc` with [`use_flake`](https://direnv.net/man/direnv-stdlib.1.html#codeuse-flake-ltinstallablegtcode) in the directory, you can just enter the directory and just run `sbt compilerJVM/test`, the environment is automatically setup when you enter the directory and destroyed when you exit.

LPTK:
@pca006132 Thanks!

@waterlens Does this work for you?

waterlens:
> @pca006132 Thanks!
> 
> @waterlens Does this work for you?

I think I have used it in the CI. So, what's the next step? Moving `compilerJVM` out of the root target?

compiler/shared/main/scala/mlscript/compiler/ir/Builder.scala
waterlens: I tried this implementation, first, it will recognize `bar` as one of the free variables of `foo.bar`. It will also think the operator `is` of ucs is a free variable. Need to fix that but not sure how to do it properly.

LPTK:
I just checked the CI config. I don't understand why you used `nix develop --command sbt compilerJVM/test` instead of running _everything_ under the nix environment as `nix develop --command sbt test` (replacing the old test command).

waterlens:
I once thought it was only used to set up the environment for testing cpp backend. Anyway, if you accept the idea to test the whole project in a nix environment, I will change it accordingly.

LPTK:
Yeah, there's no reason not to place the whole thing under nix. It would also be useful to fix the nodejs version while we're at it!

waterlens:
As far as I know, nixpkgs only support LTS version NodeJS. Can we upgrade the version from 17 to 18? Otherwise, I may need to install NodeJs from the tarball.

LPTK:
Ok I managed to fix the CI and to make the tests pass on my older local macOS version by switching to g++!

LPTK (approved):

PR #226, opened 2024-10-10
Fix `let` annotation handling

NeilKleistGao:
The annotation for let-binding was not handled correctly:
```ts
// the annotation `forall 'a : 'a -> 'a` is ignored
fun foo(f) =
  let g: forall 'a : 'a -> 'a = x => f(x) in g(123)
//│ fun foo: forall 'a. (123 -> 'a) -> 'a

fun foo(f) =
  let g = (x => f(x)): forall 'a : 'a -> 'a in g(123)
//│ fun foo: (??a -> ??a0) -> 123
```

This simple PR fixed this problem so that two given functions above have the same type:
```ts
fun foo(f) =
  let g = (x => f(x)): forall 'a : 'a -> 'a in g(123)
//│ fun foo: (??a -> ??a0) -> 123

fun foo(f) =
  let g: forall 'a : 'a -> 'a = x => f(x) in g(123)
//│ fun foo: (??a -> ??a0) -> 123
```


LPTK (approved):
Oh! That's a good catch, thanks.

PR #229, opened 2024-11-04
Print javascript stack trace

ychenfo:


LPTK:
Ok it looks like stack traces are very unstable wrt NodeJS version, which is unsurprising. Could you make it so that they only show when there's an _unexpected_ runtime error? (As in: none of `:re`, `:fixme`, and `:todo` is set.)

ychenfo:
I have made the changes. Not sure if this is the best way to do this, but I just add a new parameter `showStackTrace` to the `query` function in ReplHost which controls the printing of stack traces.

LPTK:
Thanks. I see the CI is still mysteriously failing with the `git diff` while showing no patch... Like in the previous PR. I really have no idea why.

ychenfo:
I managed to reproduce the behavior on my machine.

This behavior seems to be caused by how `git diff-files` works; `git diff` seems to behave as desired. Calling `git status` or `git update-index --refresh` before `git diff-file` also makes `git diff-file` to behave as desired on my machine.

Some related posts on stack overflow (thanks to Ruqing who helped me to look into this issue):
https://stackoverflow.com/questions/36367190/git-diff-files-output-changes-after-git-status
https://stackoverflow.com/questions/34807971/why-does-git-diff-index-head-result-change-for-touched-files-after-git-diff-or-g

----------------

Below is my terminal history showing how the problem is reproduced.


```fish
~/c/mlscript (hkmc2-print-stack-trace)> git diff-files --exit-code                                   # exit code 0
~/c/mlscript (hkmc2-print-stack-trace)> git status                                                   # no changes
On branch hkmc2-print-stack-trace
Your branch is up to date with 'anto/hkmc2-print-stack-trace'.

nothing to commit, working tree clean
~/c/mlscript (hkmc2-print-stack-trace)> echo x >> hkmc2/jvm/src/test/scala/hkmc2/ReplHost.scala      # add something
~/c/mlscript (hkmc2-print-stack-trace)> truncate -s-2 hkmc2/jvm/src/test/scala/hkmc2/ReplHost.scala  # undo the previous change
~/c/mlscript (hkmc2-print-stack-trace)> git diff-files --exit-code                                   # now `git diff-files --exit-code` exit with 1, although there is no change in the contents of any file
:100644 100644 d9ccf26c987c5e29e8cda92004077828c62d064e 0000000000000000000000000000000000000000 M      hkmc2/jvm/src/test/scala/hkmc2/ReplHost.scala
~/c/mlscript (hkmc2-print-stack-trace) [1]> git diff-files --exit-code -p                            # `git diff-files --exit-code -p` exit with 1, but do not show any diff
~/c/mlscript (hkmc2-print-stack-trace) [1]> git diff --exit-code -p                                  # `git diff --exit-code -p` exit with 0
~/c/mlscript (hkmc2-print-stack-trace)> git diff-files --exit-code -p                                # `git diff-files --exit-code -p` now also exit with 0
~/c/mlscript (hkmc2-print-stack-trace)> git status                                                   # `git status` shows no changes
On branch hkmc2-print-stack-trace
Your branch is up to date with 'anto/hkmc2-print-stack-trace'.

nothing to commit, working tree clean
~/c/mlscript (hkmc2-print-stack-trace)> echo x >> hkmc2/jvm/src/test/scala/hkmc2/ReplHost.scala      # add something
~/c/mlscript (hkmc2-print-stack-trace)> truncate -s-2 hkmc2/jvm/src/test/scala/hkmc2/ReplHost.scala  # and undo the change again
~/c/mlscript (hkmc2-print-stack-trace)> git status                                                   # execute `git status` first
On branch hkmc2-print-stack-trace
Your branch is up to date with 'anto/hkmc2-print-stack-trace'.

nothing to commit, working tree clean
~/c/mlscript (hkmc2-print-stack-trace)> git diff-files --exit-code -p                                # after `git status`, `git diff-files --exit-code -p` exit with 0
~/c/mlscript (hkmc2-print-stack-trace)> git --version
git version 2.39.5 (Apple Git-154)
```

LPTK:
Wow, thanks for looking into this! I wonder why this was never a problem before. I suppose the new diff-tests is overriding some files even though there are no actual changes.

So what do you suggest to fix it? Fixing the root issue could actually make more sense. Could you commit the fix in this PR?

ychenfo:
Indeed I guess this was not a problem before because previous diff-tests will run `git status` at the directory containing test files so that `git diff-files`, which is executed after `git status`, can work as expected. The new test [`CompileTestRunner`](https://github.com/hkust-taco/mlscript/blob/e1b568c74dfc591f3c56e23a849c52a61b3a1716/hkmc2/jvm/src/test/scala/hkmc2/DiffTestRunner.scala#L117) does not run `git status`, so `git diff-files` report spurious file changes.

For the fix I think although `git diff` works as expected on my machine, I cannot find clear documentation saying that it will never report spurious changes. So maybe it's better to run [`git update-index -q --really-refresh`](https://stackoverflow.com/questions/34807971/why-does-git-diff-index-head-result-change-for-touched-files-after-git-diff-or-g#:~:text=I%20do%20recommend%20to%20use%20git%20update%2Dindex%20%2Dq%20%2D%2Dreally%2Drefresh%20instead%20of%20git%20update%2Dindex%20%2D%2Drefresh%20as%20the%20last%20one%20may%20not%20refresh%20an%20index.%20See%20the%20docs%20for%20details.) before `git diff-files`.

LPTK:
Thanks for the fixes!

PR #230, opened 2024-11-04
Parser for Effect Handlers

CAG2Mark:
Parse terms of the form
```fs
handle h = Effect with
  fun perform(r) = r(foo)
in bar(h)
```


CAG2Mark:
Honestly the code right now is pretty unclean and I would prefer to just parse `handle h = Effect with ...` as a `LetLike`, with `handle` being a `letLike` keyword, and `Effect with ...` as an expression which creates an anonymous instance of some trait.

LPTK:
> Honestly the code right now is pretty unclean and I would prefer to just parse `handle h = Effect with ...` as a `LetLike`, with `handle` being a `letLike` keyword, and `Effect with ...` as an expression which creates an anonymous instance of some trait.

It seems fine for now. We can always refactor it later. And it's not like this is necessarily the final syntax either, anyway. We may well change it at some point.

CAG2Mark:
Merged changes from hkmc2

hkmc2/shared/src/main/scala/hkmc2/syntax/ParseRule.scala
```diff
@@ -189,6 +191,24 @@ object ParseRule:
...
+          ParseRule("'handle' binding head"):
+            Kw(`=`):
+              ParseRule("'handle' binding equals sign"):
+                Expr(
+                  ParseRule("'handle' binding class name"):
+                    typeDeclTemplateThen(
+                      Kw(`in`):
+                        ParseRule(s"'handle' binding `in` clause")(
+                          exprOrBlk(ParseRule(s"'handle' binding body")(End(())))((body, _: Unit) => S(body))
+                      *),
```
LPTK: It seems this line is mis-indented.

hkmc2/shared/src/test/mlscript/parser/Handler.mls
```diff
@@ -0,0 +1,114 @@
...
+  fun f(r) = r(0)
+in
+  foo(h)
+//│ ╔══[ERROR] Unsupported handle binding shape
+//│ ║  l.38: 	handle 1 = 1 with
+//│ ║        	       ^^^^^^^^^^
+//│ ║  l.39: 	  fun f(r) = r(0)
+//│ ╙──      	^^^^^^^^^^^^^^^^^
+
+:elt
```
LPTK: Let's omit this flag from the test file. It generates mostly noise.

hkmc2/shared/src/test/mlscript/parser/Handler.mls
```diff
@@ -0,0 +1,114 @@
...
+handle h = Eff
+//│ ╔══[PARSE ERROR] Expected 'with' keyword after 'handle' binding class name; found end of input instead
+//│ ║  l.16: 	handle h = Eff
+//│ ╙──      	              ^
+
+:pe
+handle h = Eff with
+  fun f(r) = r(0)
+  fun g(a, r) = r(1)
+//│ ╔══[PARSE ERROR] Expected 'in' keyword after type declaration block; found end of input instead
```
LPTK: Why not allow a statement form? Just like `let` – where both are valid:
```fs
fun f() =
  let x = 1 in
    log(x)

// or

fun f() =
  let x = 1
  log(x) 
```

LPTK (approved):
Ok thanks, please address my minor comments, and then I'll make the merge!

hkmc2/shared/src/test/mlscript/parser/Handler.mls
CAG2Mark: Is there a way to run the elaborator without generating all the noise? It seems to do basic name analysis and this is something we would should test.

hkmc2/shared/src/main/scala/hkmc2/syntax/ParseRule.scala
CAG2Mark: oh it seems just the last comma is misindented, will be fixed when I figure out the other change I'm working on

LPTK: No, it's the whole thing. This `*)` corresponds to the `(` in `ParseRule(s"'handle' binding `in` clause")(` and should be indented to the same level.

hkmc2/shared/src/test/mlscript/parser/Handler.mls
LPTK: The elaborator always runs even if you're not printing its result. Just remove the flag.

hkmc2/shared/src/main/scala/hkmc2/syntax/ParseRule.scala
CAG2Mark: Oh then the problem exists in `letLike` as well which I copy-pasted from. I'll fix that as well

CAG2Mark:
All comments addressed, please see the latest changes.

hkmc2/shared/src/main/scala/hkmc2/syntax/ParseRule.scala
LPTK: Probably. Feel free to fix it there as well! 😉

hkmc2/shared/src/test/mlscript/HkScratch.mls
```diff
@@ -4,4 +4,9 @@
 // :elt
 
 
-
+:parseOnly
+(let x = 1) + 1
+x
```
LPTK: This should be removed ~~or moved to another test file. Probably `hkmc2/shared/src/test/mlscript/parser/Let.mls`.~~ EDIT: I see it's already in `codegen/Bindings.mls`.

LPTK: (We try to keep the scratch file empty.)

CAG2Mark: my bad, forgot to remove it before my commit >_>

hkmc2/shared/src/test/mlscript/HkScratch.mls
```diff
@@ -2,6 +2,3 @@
 :sjs
 // :pt
 // :elt
-
-
-
```
LPTK: Please do not commit any unnecessary whitespace changes.

LPTK:
Pro-tip: if you had reviewed your diff first, these repeated rounds of comments on silly things would have been avoided 😉 

https://github.com/hkust-taco/mlscript/blob/mlscript/CONTRIBUTING.md#for-non-draft-prs

CAG2Mark:
Fixed

PR #232, opened 2024-11-07
Support tuple patterns

chengluyu:


hkmc2/shared/src/test/mlscript/codegen/BadIdent.mls
```diff
@@ -0,0 +1,15 @@
+:global
```
LPTK: You no longer need `:global` for the commands block that starts on the first line of the file.

hkmc2/shared/src/test/mlscript/ucs/patterns/SimpleTuple.mls
LPTK, line 1: It would be good to make this file `:js`. And probably most other UCS test files where we want to check the output, too.

hkmc2/shared/src/test/mlscript/codegen/BadIdent.mls
chengluyu: Cool!

hkmc2/shared/src/test/mlscript/ucs/patterns/SimpleTuple.mls
```diff
@@ -1,78 +1,99 @@
...
+//│ = 0
 
 // However, the `Object` type forbids tuples to be used.
 :todo
 working_cases([0, 0])
+//│ = 0
+
 
-:todo
 fun not_working(x) =
```
LPTK: Seems like the name should be updated!

hkmc2/shared/src/test/mlscript/ucs/patterns/RestTuple.mls
```diff
@@ -0,0 +1,135 @@
...
+
+foo([])
+//│ = 0
+
+:re
+foo(["boom"])
+//│ ═══[RUNTIME ERROR] Error: match error
+
+:sjs
+:todo // Better using ... instead of .. for the rest pattern.
```
LPTK: Why?

chengluyu: As we discussed, there will be two “rest patterns”: `..` creates a lazy view, while `...` is the normal one.

chengluyu: So, in this test case we should use `...` instead of `..`. But parsing logic is not finished yet...

hkmc2/shared/src/test/mlscript/ucs/patterns/SimpleTuple.mls
chengluyu: This works right now. The support to simple tuple was limited.

hkmc2/shared/src/test/mlscript/ucs/patterns/RestTuple.mls
LPTK: > So, in this test case we should use ... instead of ...

Why?

hkmc2/shared/src/test/mlscript-compile/List.mjs
LPTK, line 1: There is already `Stack.mls`, which is what this data structure should be called. It's just a functional stack.

hkmc2/shared/src/test/mlscript-compile/String.mls
LPTK, line 1: There is already `Str.mls`.

hkmc2/shared/src/test/mlscript-compile/String.mls
```diff
@@ -0,0 +1,6 @@
+module String with ...
+
+fun (++) concat(lhs, rhs) =
```
LPTK: BTW I'm thinking to reserve `++` for proper collections, which is why `Str.mls` uses `~` (to avoid name clashes later on).

hkmc2/shared/src/test/mlscript-compile/List.mjs
chengluyu: OK. I will remove it.

hkmc2/shared/src/test/mlscript-compile/String.mls
chengluyu: I don't know it because it is introduced after my last rebase. I'll remove it after rebase.

hkmc2/shared/src/test/mlscript-compile/String.mls
chengluyu: OK. I'll update.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -155,6 +157,17 @@ class ClassSymbol(val tree: Tree.TypeDef, val id: Tree.Ident)
...
+  def arity: Int = defn match
+    case S(d) => d.paramsOpt.fold(0)(_.length)
+    case N => tree.paramLists.headOption.fold(0)(_.fields.length)
+  /** Get parameter symbols. Create new symbols if not elaborated. */
+  override lazy val params: Ls[LocalSymbol & NamedSymbol] = defn match
+    case S(d) => d.paramsOpt.fold(Nil)(_.map(_._2))
+    case N => tree.paramLists.headOption.fold(Nil): tup =>
+      tup.fields.iterator.flatMap(_.param).map:
+        case (id, _) => TermSymbol(ParamBind, S(this), id)
+      .toList
```
LPTK: What is the rationale for this complicated logic? If someone somehow calls this while the elaborator is in the process of computing the `defn`, we'll end up with duplicated symbols. It's just a pretty bad/error-prone design.

Why not follow what I suggested? The symbols should be created in `TypeDefImpl`, which does have access to the `ClassSymbol`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -155,6 +155,10 @@ class ClassSymbol(val tree: Tree.TypeDef, val id: Tree.Ident)
   def nme = id.name
   def toLoc: Option[Loc] = id.toLoc // TODO track source tree of classe here
   override def toString: Str = s"class:$nme"
+  /** Compute the arity. */
+  def arity: Int = defn match
+    case S(d) => d.paramsOpt.fold(0)(_.length)
+    case N => tree.paramLists.headOption.fold(0)(_.fields.length)
```
LPTK: Why access `defn` and risk making the result of this function inconsistent? What's the point? You could just use the tree.

hkmc2/shared/src/main/scala/hkmc2/semantics/Desugarer.scala
```diff
@@ -349,6 +360,12 @@ class Desugarer(tl: TraceLogger, elaborator: Elaborator)
       raise(ErrorReport(msg"Unrecognized pattern split." -> tree.toLoc :: Nil))
       _ => _ => Split.default(Term.Error)
 
+  private def tupleSlice: Ctxl[Ctx.Elem] =
+    ctx.get("tupleSlice").getOrElse(lastWords("tupleSlice not found"))
+
+  private def tupleGet: Ctxl[Ctx.Elem] =
+    ctx.get("tupleGet").getOrElse(lastWords("tupleGet not found"))
+
```
LPTK: This logic is very wrong. You're going to pick up _anything_ that's named `tupleGet` in the current context 🤦 

This is why I've been telling you many times (to no avail, apparently) that this part of the logic should use `globalThis`. It should basically be `Sel(Sel(Ref(Ctx.globalThisSymbol), "Predef"), "tupleGet")`.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -255,6 +255,16 @@ class JSBuilder extends CodeBuilder:
         doc" else { #{ ${ returningTerm(el) } #}  # }"
       case N  => doc""
       t :: e :: returningTerm(rest)
+    case Match(scrut, Case.Tup(len, inf) -> trm :: Nil, els, rest) =>
+      val test = doc"Array.isArray(${ result(scrut) }) && ${ result(scrut) }.length ${if inf then ">=" else "==="} ${len}"
```
```suggestion
      val test = doc"globalThis.Array.isArray(${ result(scrut) }) && ${ result(scrut) }.length ${if inf then ">=" else "==="} ${len}"
```

hkmc2/shared/src/test/mlscript/ucs/syntax/TupleRest.mls
```diff
@@ -0,0 +1,127 @@
+:parseOnly
+:pt
```
LPTK: This is not so useful I think. You could at least comment it (so it can be uncommented quickly later, if needed).

hkmc2/shared/src/test/mlscript/codegen/BadIdent.mls
LPTK, line 1: There's already `Primes.mls`. As a general guideline, `BadXXX` tests are supposed to test features used badly/incorrectly, not missing feature implementations.

LPTK (changes requested):
Excellent, thanks! I only have two main change requests.

First, please add a test that pattern-matches tuples within a function `fun foo(tupleGet) = ...`. Realize your foolishness. Then fix the bug.

Second, please comment out the vast majority of the uses of the `:ucs` flag. This produces extremely verbose debugging output that's mostly noise and creates so many diff changes. Especially since it prints symbol UIDs (it probably shouldn't). At most, please only keep **a few small but representative** uses of `:ucs`, which can be used to check if there is a regression in the UCS logic. Currently there is just so much output noise that any change creates tons of diffs and regressions may as well go unnoticed, which completely defeats the purpose of diff-testing.

LPTK (approved):
Awesome contribution, thanks!

PR #233, opened 2024-11-09
Module Methods

FlandiaYingman:
_Module methods_ are the methods defined within a _module_.
_Module parameters_ are function parameters that are prefixed with `module` keyword.
_Module values_ are module literals or applications of functions returning modules.

#### Rules

- [x] Module parameters must have an explicit type.
- [x] Module parameters must have a concrete type.
- [x] Function returning module values must have an explicit return type.
- [x] Function returning module values must have a concrete return type.
- [x] The return type of functions returning module values must be prefixed with `module` keyword.
- [x] Only module arguments (values) may be passed to module parameters.
- [x] Only module methods may return module values.
- [ ] Function returning module values must not have multiple parameter lists.


hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -471,6 +501,7 @@ extends Importer:
           case R(id) =>
             val sym = members.getOrElse(id.name, die)
             val owner = ctx.outer
+            val mod = owner.map(_.isInstanceOf[ModuleSymbol]).getOrElse(false)
```
```suggestion
            val mod = owner.fold(false)(_.isInstanceOf[ModuleSymbol])
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -603,7 +652,23 @@ extends Importer:
...
+      // return S(moduleParam) if t represents a module parameter
+      // return N otherwise
+      def moduleParam(t: Term): Opt[Param] = t match
+        case s: hkmc2.semantics.Term.Sel => s.symbol
+          .flatMap(_.asMod)
+          .map(_ => Param(FldFlags.module, fieldOrVarSym(ParamBind, lhs), S(s)))
+        case hkmc2.semantics.Term.TyApp(s: hkmc2.semantics.Term.Sel, _) => s.symbol
+          .flatMap(_.asMod)
+          .map(_ => Param(FldFlags.module, fieldOrVarSym(ParamBind, lhs), S(s)))
+        case _ => N
```
LPTK: Why not just use `s.symbol` directly?

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -179,12 +180,23 @@ sealed trait Statement extends AutoLocated:
       s"${cls.kind} ${cls.sym.nme}${
         cls.tparams.map(_.showDbg).mkStringOr(", ", "[", "]")}${
         cls.paramsOpt.fold("")(_.map(_.showDbg).mkString("(", ", ", ")"))} ${cls.body}"
+    case mod: ModuleDef =>
```
LPTK: This case is unreachable. `ModuleDef` is a `ClassLikeDef `.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -297,7 +310,12 @@ final case class Param(flags: FldFlags, sym: LocalSymbol & NamedSymbol, sign: Op
   // def showDbg: Str = flags.showDbg + sym.name + ": " + sign.showDbg
   def showDbg: Str = flags.showDbg + sym + sign.fold("")(": " + _.showDbg)
 
-object FldFlags { val empty: FldFlags = FldFlags(false, false, false) }
+object FldFlags { 
+  val empty: FldFlags = FldFlags(false, false, false, false)
+
+  // module parameter / module argument, depending on the context
+  val module: FldFlags = FldFlags(false, false, false, true)
```
LPTK: This should not be useful, please remove it. What you want to do with these things is to add some new flags onto an existing value, as in `flags.copy(mod = true)`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -603,7 +652,23 @@ extends Importer:
     case id: Ident =>
       Param(FldFlags.empty, fieldOrVarSym(ParamBind, id), N) :: Nil
     case InfixApp(lhs: Ident, Keyword.`:`, rhs) =>
-      Param(FldFlags.empty, fieldOrVarSym(ParamBind, lhs), S(term(rhs))) :: Nil
+      // return S(moduleParam) if t represents a module parameter
+      // return N otherwise
```
LPTK: This part of the code is very convoluted. Instead of returning an option that you pattern match immediately and duplicating the `Param`-creation code, why not just make compute the flags alone first then construct the `Param` in only one place?

hkmc2/shared/src/main/scala/hkmc2/utils/utils.scala
```diff
@@ -36,11 +37,16 @@ extension (t: Product)
       case Nil => "Nil"
       case xs: List[_] => "Ls of \n" + xs.iterator.map(aux(_)).mkString("\n").indent("  ")
       case s: String => s.escaped
-      case FldFlags(mut, spec, genGetter) =>
+      case TermDefFlags(mod) =>
+        val flags = Buffer.empty[String]
+        if mod then flags += "module"
+        if flags.isEmpty then "()" else flags.mkString("(", ", ", ")")
```
LPTK: `if` is unnecessary.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -482,11 +513,29 @@ extends Importer:
...
+              // val explicitReturnType = s.isDefined
+              
+              // FIXME: this is magic; we shall determine the return type based on the signature
+              val explicitReturnType = 
+                td.head.isInstanceOf[InfixApp] && td.head.asInstanceOf[InfixApp]._2 == Keyword.`:`
+              
+              if pss.flatMap(_.params).exists(_.flags.mod) && !explicitReturnType then
+                raise:
+                  ErrorReport:
+                    msg"Functions with module parameters must have a explicit return type." ->
```
LPTK: Why is that?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -232,7 +232,33 @@ extends Importer:
...
+        case Term.Tup(fields) => S(fields)
+        case _ => N
+      val argsModFlags = args
+        .map(_.map(_.flags.mod))
+      val paramsModFlags = lt.symbol match
+        case S(sym: BlockMemberSymbol) => sym.defn match
+          case S(defn: TermDefinition) => defn.params.lift(0)
+            .map(_.params.map(_.flags.mod))
+          case _ => argsModFlags.map(_.map(_ => false))
+        case _ => argsModFlags.map(_.map(_ => false))
```
LPTK: This logic is very strange/fishy. In general, I also don't like all that mapping just to consume it in the `for` below. Better zip the sources directly and destructure the corresponding things in the `for`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -482,11 +513,29 @@ extends Importer:
...
+              val tdf = TermDefinition(owner, k, sym, pss, s, b, r, TermDefFlags(mod))
               sym.defn = S(tdf)
+
+              // Restrictions regarding functions receiving module parameters
+              
+              // not working because sometimes the signature is N 
+              // even when the function has a explicit return type
+              // val explicitReturnType = s.isDefined
+              
+              // FIXME: this is magic; we shall determine the return type based on the signature
```
LPTK: So is this FIXME meant to be addressed in this PR?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -482,11 +513,29 @@ extends Importer:
...
               val b = rhs.map(term(_)(using newCtx))
               val r = FlowSymbol(s"‹result of ${sym}›", nextUid)
-              val tdf = TermDefinition(owner, k, sym, pss,
-                td.signature.orElse(newSignatureTrees.get(id.name)).map(term), b, r)
+              val tdf = TermDefinition(owner, k, sym, pss, s, b, r, TermDefFlags(mod))
               sym.defn = S(tdf)
+
+              // Restrictions regarding functions receiving module parameters
+              
+              // not working because sometimes the signature is N
```
LPTK: Currently "signatures" are only picked up from externally-given declarations. Picking up result types when they're given requires a bit more logic to be added, probably in `TypeDefImpl`. It will likely be useful to have in general, along with logic that propagates types to arguments from the signature. But that can be done later.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -179,12 +180,23 @@ sealed trait Statement extends AutoLocated:
...
 
 final case class LetDecl(sym: LocalSymbol) extends Statement
 
 final case class DefineVar(sym: LocalSymbol, rhs: Term) extends Statement
 
+final case class TermDefFlags(mod: Bool):
+  def showDbg: Str = (
+    (if mod then S("module") else N) ::
+    Nil
+  ).flatten.mkString(" ")
```
LPTK: These 2 lines are misindented. Also, why is the intermediate option necessary?

hkmc2/shared/src/main/scala/hkmc2/utils/utils.scala
```diff
@@ -36,11 +37,16 @@ extension (t: Product)
       case Nil => "Nil"
       case xs: List[_] => "Ls of \n" + xs.iterator.map(aux(_)).mkString("\n").indent("  ")
       case s: String => s.escaped
-      case FldFlags(mut, spec, genGetter) =>
+      case TermDefFlags(mod) =>
+        val flags = Buffer.empty[String]
+        if mod then flags += "module"
```
LPTK, line 42: What's the use of this strange mutable buffer?

hkmc2/shared/src/main/scala/hkmc2/utils/utils.scala
```diff
@@ -36,11 +37,16 @@ extension (t: Product)
...
+        if flags.isEmpty then "()" else flags.mkString("(", ", ", ")")
+      case FldFlags(mut, spec, genGetter, mod) =>
         val flags = Buffer.empty[String]
         if mut then flags += "mut"
         if spec then flags += "spec"
         if genGetter then flags += "gen"
+        if mod then flags += "module"
         if flags.isEmpty then "()" else flags.mkString("(", ", ", ")")
```
LPTK: `if` is unnecessary.

hkmc2/shared/src/main/scala/hkmc2/utils/utils.scala
```diff
@@ -36,11 +37,16 @@ extension (t: Product)
...
+      case TermDefFlags(mod) =>
+        val flags = Buffer.empty[String]
+        if mod then flags += "module"
+        if flags.isEmpty then "()" else flags.mkString("(", ", ", ")")
+      case FldFlags(mut, spec, genGetter, mod) =>
         val flags = Buffer.empty[String]
         if mut then flags += "mut"
         if spec then flags += "spec"
         if genGetter then flags += "gen"
+        if mod then flags += "module"
```
LPTK: At this point you'd better just use a `StringBuilder` instead of an intermediate collection.

hkmc2/shared/src/test/mlscript/basics/ModuleMethods.mls
```diff
@@ -0,0 +1,396 @@
+:elt
```
LPTK: No need to commit such noisy output.

LPTK (changes requested):
Ok, so currently we are trying to see if we can get away with _not_ elaborating imported files. So your approach of accessing the `defn` won't work if the function comes from a different file. Based on this, I would suggest adding an explicit `module` modifier for parameters: `fun foo(module m, arg2, arg3)`. I actually like it better than some subtle type-based check.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: IIRC `s.symbol` is only defined on either `s: Term.Sel` or `s: Term.Ref`, so I had to check if the type annotation is really a `Sel` or a `TyApp` that can be unwrapped to a `Sel`. Also I had to check if the type really is a module, so I call `asMod` and see if I can get some module, if so, I create a `Param` with module flag being set; otherwise, I return `N` and fallback to a regular `Param`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
FlandiaYingman: I think I forgot to remove after rebasing and dealing with the merge conflicts, will remove this later

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
FlandiaYingman: So write `FldFlags.empty.copy(mod = true)` instead? Often there are no existing flags for parameters

FlandiaYingman: I think I understand what you mean

hkmc2/shared/src/main/scala/hkmc2/utils/utils.scala
FlandiaYingman: I was thinking that it can be easily extended for future, so I just copied those from below. You don't like it? 🫡

LPTK: I mean the `if` is completely redundant; you can get the same result by:
```suggestion
        flags.mkString("(", ", ", ")")
```

FlandiaYingman: I just copied it from somewhere below... 🥲 Anyway will fix it

hkmc2/shared/src/main/scala/hkmc2/utils/utils.scala
FlandiaYingman: IDK, I just copied it from below. I thought it was some good practice though 😪

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: Not really. Initially I was thinking to just stick to this implementation until we have a better way to get the result type.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: Which one?

hkmc2/shared/src/main/scala/hkmc2/utils/utils.scala
LPTK: Ah yes, this appears to be @chengluyu's [beautiful work](https://github.com/hkust-taco/mlscript/commit/b9190b3630c1979a23953d575fcd78bbeb312dc1#diff-e59b35cf20bc0a11500f04c2828d5134dbee399ce45f1e42c1255261722ab67cR36). Please also fix the original. The idea is that one should strive to leave the code in a better place than they found it, if possible.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: > Functions with module parameters must have a explicit return type.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: This makes me realize we haven't discussed how to mark module returns. Following my latest proposal for module parameters, we should probably make it:

```fs

module fun example(module m: Foo, x, y): Bar = ...
```

WDYT?

LPTK:
So is this ready for review? (BTW pls read https://github.com/hkust-taco/mlscript/blob/mlscript/CONTRIBUTING.md, if you haven't)

Please notify me once you're addressed all our discussions and updated your branch. Also, why not check your TODOs in https://github.com/hkust-taco/mlscript/pull/233#issue-2646388927?

FlandiaYingman:
@LPTK 

Things are getting messy as we come up with new ideas and rules and changing previous ones. I'm trying to refactor the code to ensure everything looks good and concise. That TODO list was intended to track the new rules that have been refactored. 

This PR should be ready for review by tonight. I'll notify you when I'm done. 🫡

LPTK:
Ok sure, no worries!

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -35,10 +36,12 @@ enum Term extends Statement:
   case Ret(result: Term)
   case Try(body: Term, finallyDo: Term)
   case Handle(lhs: LocalSymbol, rhs: Term, defs: ObjBody)
+  case Mod(inner: Term)
```
LPTK: Why do we need a new term form? Why not use the usual `Ref` and `Sel` and check whether something is a module by looking at its symbol?

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -297,7 +320,9 @@ final case class Param(flags: FldFlags, sym: LocalSymbol & NamedSymbol, sign: Op
   // def showDbg: Str = flags.showDbg + sym.name + ": " + sign.showDbg
   def showDbg: Str = flags.showDbg + sym + sign.fold("")(": " + _.showDbg)
 
-object FldFlags { val empty: FldFlags = FldFlags(false, false, false) }
+object FldFlags {
```
```suggestion
object FldFlags:
```

hkmc2/shared/src/test/mlscript/basics/ModuleMethods.mls
```diff
@@ -0,0 +1,67 @@
...
+  fun f5(): M = M
+}
+//│ ╔══[ERROR] The return type of functions returning module values must be prefixed with module keyword.
+//│ ║  l.41: 	  fun f5(): M = M
+//│ ╙──      	      ^^^^^^^
+
+:e
+fun f6(module m: M)
+f6(new C)
+//│ ╔══[ERROR] Only module arguments (values) may be passed to module parameters.
```
LPTK: Any reason for this restriction?

hkmc2/shared/src/test/mlscript/basics/ModuleMethods.mls
```diff
@@ -0,0 +1,67 @@
...
+fun f7(): module M
+//│ ╔══[ERROR] Only module methods may return module values.
+//│ ║  l.55: 	fun f7(): module M
+//│ ╙──      	    ^^^^^^^^^^^^^^
+
+
+fun ok1(module m: M)
+
+module N with {
+  fun ok2(): module M = M
```
LPTK, line 87: Why not add test cases calling this function?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -232,7 +232,32 @@ extends Importer:
...
+
+      // Check if module arguments match module parameters
+      val args = rt match
+        case Term.Tup(fields) => S(fields)
+        case _ => N
+      val params = lt.symbol
+        .filter(_.isInstanceOf[BlockMemberSymbol])
+        .flatMap(_.asInstanceOf[BlockMemberSymbol].trmTree)
+        .filter(_.isInstanceOf[TermDef])
+        .flatMap(_.asInstanceOf[TermDef].paramLists.headOption)
```
LPTK: Methods `collect` + `flatten` can do this but safely. `asInstanceOf` should be avoided if possible.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -271,6 +296,8 @@ extends Importer:
       Term.Ret(term(body))
     case Modified(Keyword.`do`, kwLoc, body) =>
       Term.Blk(term(body) :: Nil, unit)
+    case TypeDef(Mod, head, _, _) =>
+      Term.Mod(term(head))
```
LPTK: Do we have examples of this in the tests? And why is it restricted to modules? (In JS, one can also define and return classes.) Anyway, I don't see a good reason for this. Feels like  term `(module M with { ... })` should be desugared to `{module M with { ... }; M}`. You don't need a new `Mod` term form.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -471,6 +502,7 @@ extends Importer:
           case R(id) =>
             val sym = members.getOrElse(id.name, die)
             val owner = ctx.outer
+            val mod = owner.fold(false)(_.isInstanceOf[ModuleSymbol])
```
```suggestion
            val isModMember = owner.fold(false)(_.isInstanceOf[ModuleSymbol])
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -482,11 +514,40 @@ extends Importer:
...
               val b = rhs.map(term(_)(using newCtx))
               val r = FlowSymbol(s"‹result of ${sym}›", nextUid)
-              val tdf = TermDefinition(owner, k, sym, pss,
-                td.signature.orElse(newSignatureTrees.get(id.name)).map(term), b, r)
+              val tdf = TermDefinition(owner, k, sym, pss, s, b, r, TermDefFlags(mod))
               sym.defn = S(tdf)
+
+              // the return type of the function
+              val result = td.head match
+                case InfixApp(_, Keyword.`:`, rhs) => S(term(rhs)(using newCtx))
```
LPTK: The signature was already elaborated above in `s`. Looks like this will redo the work!

Though I guess the signature might be different from the explicitly-ascribed result type.

Anyway, you shouldn't manually retrieve this type by pattern matching the tree, duplicating logic in `Tree.scala`. There should only be one source of truth about what is the result type.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -672,11 +742,36 @@ extends Importer:
...
+    /** Checks if a term is a reference to a type parameter. */
+    def isTypeParam(t: Term): Bool = t.symbol
+      .filter(_.isInstanceOf[VarSymbol])
+      .flatMap(_.asInstanceOf[VarSymbol].decl)
+      .fold(false)(_.isInstanceOf[TyParam])
+    
+    /** Checks if a term evaluates to a module value. */
+    def evalsToModule(t: Term): Bool = 
+      def returnsModule(t: Tree): Bool = t match
+        case InfixApp(_, Keyword.`:`, rhs) => rhs match
```
LPTK: Same comment as above about the single source of truth.

LPTK (commented):
Did you consider the following scenario?
```fs
module M
fun foo() = M
```
I don't think I saw it in the tests.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
FlandiaYingman: I just reverted that change because there's no need to break this line. Do you want to keep the line break and change `{}` to `:`?

LPTK: Nope!

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
FlandiaYingman: This means a term is modified by a `module` modifier keyword. For example,

```fs
fun f: module M
fun g[T]: module T
```

Here the results get elaborated into

```
Mod(Sel(Ref(M)))
Mod(Ref(T))
```

So I need that to check if a result type is correctly marked as `module`.

Of course, it is not necessary for now that we can always parse the Tree, but I think it'd be useful in future because we will inevitably elaborate some `module` things.

hkmc2/shared/src/test/mlscript/basics/ModuleMethods.mls
FlandiaYingman: Oops, I think I made it in a wrong direction. It should be: "Only module parameters may receive module arguments"

hkmc2/shared/src/test/mlscript/basics/ModuleMethods.mls
FlandiaYingman: The only rule in this PR regarding function call is "only module parameters may receive module arguments", and `ok2` does not involve any module parameters. So I think there's no need to call it.

LPTK: Ah. I would have thought `ok1(N.ok2())` would have qualified. It's perfectly fine to address this in a follow-up PR, but in the meantime it'd be good to add the test case.

FlandiaYingman: Sounds good, I'll add `ok1(N.ok2())`

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
LPTK: Ah ok it's only for module type annotations. Then the name should be more informative, like `ModType`. Though TBH I'm not sure we need a term for that. After elaboration, I suspect no one cares whether the user wrote `: module M` or `: M`. The `module` keyword is just for the elaborator to do its checks.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: I don't really understand your question, but I guess it's related to this comment: https://github.com/hkust-taco/mlscript/pull/233#discussion_r1850475848.

The reason of having this hack is that I can't figure out a good way to parse a module declaration/definition and a module modifier into different `Tree` things, so all of them, including `module` modifiers, get parsed into `TypeDef`. 

I added a case here to elaborate module modifiers into `Term.Mod` so that it can be further processed.

I think this hack is similar to https://github.com/hkust-taco/mlscript/blob/3d84f9cdd7c00fe1d144069d6e89545dfcdcde59/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala#L680

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: What. I thought that's the problem we've just discussed earlier.

In this https://github.com/hkust-taco/mlscript/pull/233#discussion_r1842335992 you said "signatures are only picked up from externally-given declaration"; and at

https://github.com/hkust-taco/mlscript/blob/e3859934e2c38db20f2a6f6998aba55dc49940a9/hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala#L234

it does not always pick up the result type from the parsed tree. That's why I did another parsing to make sure result types are always parsed.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
FlandiaYingman: I think you are probably right that we may not need this. I'll try to remove it and see if anything stops us from doing that.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: Indeed, this was related to my previous assumption that `Mod` meant something different.

> all of them, including `module` modifiers, get parsed into `TypeDef`

BTW there is absolutely no problem with this. Our parsing philosophy is that we parse any tree that makes visual sense (such as module definitions), in any position, and detangle things later in the elaboration pass. This should allow us to have much more specific and helpful error messages while ensuring visual consistency of the syntax.

LPTK: BUT you **must** check that this module `TypeDef` does **not** have any additional things inside, instead of silently discarding them using `_`. Think about the user experience of doing that.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: Ok, let me be completely explicit, so we don't get hung up on details and terminology:

* There should be only one place that determines what is the inline-annotated result type in a definition (if there is any). The logic for finding it by analyzing trees should not be duplicated (as you are doing now), as it means that any update to it must also be duplicated everywhere, which is impractical and error-prone.
  Currently, this is done in `TypeOrTermDef` and the result is confusingly called `signature` (we could change it to `annotatedResultType`).
* In the elaborator, if we see there is no inline-annotated result type, we should use the one given by the separate signature. This could be implemented later, though.
* Such explicitly-annotated result types must not be elaborated several times. Indeed, currently that's not the case, but we should be careful that this does not happen whenever we add the logic to achieve the previous point.
  This is more of a comment for the future and not immediately actionable.

> it does not always pick up the result type from the parsed tree. That's why I did another parsing to make sure result types are always parsed.

If the implementation in `TypeOrTermDef` is missing some cases, you should fix that implementation, instead of leaving it broken and reimplementing it separately!

hkmc2/shared/src/test/mlscript/basics/ModuleMethods.mls
FlandiaYingman: Added in another way: `ok1(M.self())`

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -482,11 +517,41 @@ extends Importer:
...
                     val (qs, newCtx) = params(ps)(using ctx)
                     (pss :+ ParamList(ParamListFlags.empty, qs), newCtx)
+              // * Elaborate signature
+              val s = td.signature.orElse(newSignatureTrees.get(id.name)).map(term)
               val b = rhs.map(term(_)(using newCtx))
               val r = FlowSymbol(s"‹result of ${sym}›", nextUid)
-              val tdf = TermDefinition(owner, k, sym, pss,
-                td.signature.orElse(newSignatureTrees.get(id.name)).map(term), b, r)
+              val tdf = TermDefinition(owner, k, sym, pss, s, b, r, 
+                TermDefFlags.empty.copy(mod = isModMember))
```
LPTK: Better also make the flag name itself more informative! By naming it `isModMember` too.

FlandiaYingman:
@LPTK Everything's done! I

- removed `Term.Mod`
- fixed (maybe) a bug that our parser didn't pick up result types from generic functions (c874fe369e09beaf5a5d39f04e2803a275c47e8a, hope that's not intended)

so now only `Tree.scala` will parse the `Tree` as the single truth

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
FlandiaYingman: removed!

LPTK:
> * fixed (maybe) a bug that our parser didn't pick up result types from generic functions ([c874fe3](https://github.com/hkust-taco/mlscript/commit/c874fe369e09beaf5a5d39f04e2803a275c47e8a), hope that's not intended)

Thanks, this was clearly a typo in the original code.

> @LPTK Everything's done!

Actually you also need to update the branch and fix its conflicts with the main branch. If you choose to rebase, pls make sure that the tests still work on each individual commit: `git rebase -i --exec "sbtn hkmc2JVM/test"`. If you don't rebase and have a dirty history, we can later squash all commits together when merging.

LPTK (approved):
Thanks! I've committed some minor last changes. Notably, there was some dead code that I removed.

PR #234, opened 2024-11-13
Fix missing semicolons

AnsonYeung:
Add semicolons in several places in the codegen

This fixes JS codegen on the IR `Label(lbl, Break(lbl, true), End())`

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
@@ -220,7 +220,7 @@ class JSBuilder extends CodeBuilder:
           case N => doc"$res;${returningTerm(rst)}"
       doc" # ${resJS}"
     case Return(res, true) => doc" # ${result(res)}"
-    case Return(res, false) => doc" # return ${result(res)}"
+    case Return(res, false) => doc" # return ${result(res)};"
     
     // TODO factor out common logic
     case Match(scrut, Case.Lit(syntax.Tree.BoolLit(true)) -> trm :: Nil, els, rest) =>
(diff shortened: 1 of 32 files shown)


LPTK:
Thanks! But there is no need to add semicolons after things like function definitions and classes. `class Foo {} console.log("ok")` is valid JS.

BTW, the CI fails due to flakiness for a reason @chengluyu has already fixed in a separate PR.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -208,19 +208,19 @@ class JSBuilder extends CodeBuilder:
...
               case N =>
                 fun match
-                case S(f) => doc"${f}; # ${sym.nme}.class = ${clsJS}"
+                case S(f) => doc"${f}; # ${sym.nme}.class = ${clsJS};"
                 case N => clsJS
         thisProxy match
-          case S(proxy) => doc" # const $proxy = this; # ${res.stripBreaks};${returningTerm(rst)}"
-          case N => doc"$res;${returningTerm(rst)}"
+          case S(proxy) => doc" # const $proxy = this; # ${res.stripBreaks}${returningTerm(rst)}"
+          case N => doc"$res${returningTerm(rst)}"
```
LPTK, line 220: What's the rationale for removing semicolons in two places and adding them back in five places instead?

AnsonYeung: This is to remove the semicolon for class and functions

LPTK: Ah I see it's just to avoid adding one for the function case.

PR #235, opened 2024-11-13
Port part of nofib benchmarks to hkmc2

ychenfo:


shared/src/test/diff/nofib/lcss.mls
LPTK, line 1: Huh, why is this one in the wrong compiler project?

LPTK:
Cool work!

But what's with all the reverted commits? Do you want to submit these improvements in separate PRs?

Also, can you please add some spacing between things? At least one empty line between each definition – unless, of course, they are mutually recursive. And least two empty lines between the prelude import and the definitions and between the definitions and their final use. Otherwise, the result is barely legible.

LPTK:
Regarding groups of mutually-recursive data structures, please do separate them with empty lines by using the recent `:...` diff-test feature.

Example output:

```fs
:...
//│ ————————————————————————————————————————————————————————————————————————————————

fun foo() = bar()

fun bar() = if false then foo() else 42

foo()

//│ = 42
//│ ————————————————————————————————————————————————————————————————————————————————
```

shared/src/test/diff/nofib/lcss.mls
ychenfo: Ah this is a very early port that I forgot to delete, I will clean up this. Thanks!

ychenfo:
Thanks for the comments! I will make the changes to add empty lines now and also do some further cleanup (delete the `lcss.mls` in the previous compiler project, delete the `0` at the beginning of files, etc.), after that I will turn this draft PR (which I made to see if floating point numbers cause flakiness to the test output) to be ready to review.

The reverted commits are about my changes to difftests so that only changed files will be tested (otherwise [all files are tested](https://github.com/hkust-taco/mlscript/blob/5b294472d57617105ca270e32d97cb2bf455cd00/hkmc2/jvm/src/test/scala/hkmc2/DiffTestRunner.scala#L92), which, if I remember correctly from our meeting, is intended?)

LPTK:
> The reverted commits are about my changes to difftests so that only changed files will be tested (otherwise [all files are tested](https://github.com/hkust-taco/mlscript/blob/5b294472d57617105ca270e32d97cb2bf455cd00/hkmc2/jvm/src/test/scala/hkmc2/DiffTestRunner.scala#L92), which, if I remember correctly from our meeting, is intended?)

It's not specifically intended. It's just a known bug. We haven't bothered to fix it because nowadays we just use the watcher to test things. I hope you do as well. Right?

ychenfo:
> It's not specifically intended. It's just a known bug. We haven't bothered to fix it because nowadays we just use the watcher to test things. I hope you do as well. Right?

Oh I now remember that we mentioned about the watcher during our meeting, I do use the watcher when I make changes to the tests, but it seems that under the watcher mode (`~test`), still all tests are run when `cmd+s` is pressed, as shown in the following gif. 😮 Or did I miss anything?

CleanShot 2024-11-14 at 17 09 08

LPTK:
OMG I can't believe you've done all this work without using the proper testing approach that tests the CURRENT file only when you save. It's just `~hkmc2JVM/Test/run`.

ychenfo:
> OMG I can't believe you've done all this work without using the proper testing approach that tests the CURRENT file only when you save. It's just `~hkmc2JVM/Test/run`.

😵Ah now I see! Thanks a lot!

ychenfo:
I have made further cleanup and added empty lines. One minor thing that I notice is that  adding empty lines between nested functions does not interact well with editor's auto trimming of white spaces. I am not sure whether this should be a concern (`→` below stands for whitespaces):
```f#
fun foo() =
  fun bar() = 1
→ // no problem
  fun baz() = 0
  baz()
```
```f#
fun foo() =
  fun bar() = 1
// without starting white spaces, `foo()` is considered by the compiler to end here
  fun baz() = 0
  baz()
```
Thanks!

LPTK:
> I have made further cleanup and added empty lines. One minor thing that I notice is that adding empty lines between nested functions does not interact well with editor's auto trimming of white spaces.

Please turn auto-trimming off in your editor. I like to keep indentations in place so I don't have to reinsert them all the time. There is no reason to trim it.

> I am not sure whether this should be a concern (`→` below stands for whitespaces):
> 
> ```fsharp
> fun foo() =
>   fun bar() = 1
> → // no problem
>   fun baz() = 0
>   baz()
> ```
> 
> ```fsharp
> fun foo() =
>   fun bar() = 1
> // without starting white spaces, `foo()` is considered by the compiler to end here
>   fun baz() = 0
>   baz()
> ```

Are you saying that your editor ALSO de-indents comments? That would be madness.

In any case, you're welcome to fix this, but please don't use de-indented comments. They look terribly confusing.

LPTK:
Can we merge this PR now? Or is there something else you wanted to fix?

ychenfo:
> Can we merge this PR now? Or is there something else you wanted to fix?

Yes, I think it's ready now. Thanks!

PR #236, opened 2024-11-16
Migrate files for LLIR

waterlens:


CAG2Mark:
FYI in case we port the tailrec mod-cons stuff over, it seems https://github.com/hkust-taco/mlscript/pull/225 broke that so that will need to be revisited.

LPTK:
@CAG2Mark Thanks for the catch. We'll need to reimplement the tail recursion optimization in the new compiler and that will probably have to be at the middle IR level anyway (`Block.scala`).

LPTK:
This PR is superseded by https://github.com/hkust-taco/mlscript/pull/261.

PR #237, opened 2024-11-19
Generate argument number checks for JS code

ychenfo:
I modified `JSBuilder.scala` such that now three places related to function/method codegen call the method `handleFunction` to handle the common logic. The `handleFunction` is overridden in `JSBackendDiffMaker.scala` such that argument number sanity checks are implemented using varargs in javascript.

I checked and also learned that in javascript `arguments.length` can also be used to determine how many arguments are actually passed into a function, but we generate arrow functions which do not have `arguments`.


hkmc2/jvm/src/test/scala/hkmc2/JSBackendDiffMaker.scala
```diff
@@ -4,7 +4,12 @@ import scala.collection.mutable
 
 import mlscript.utils.*, shorthands.*
 import utils.*
-
```
LPTK, line 7: Pls remove unnecessary whitespace change.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -329,7 +329,12 @@ class JSBuilder extends CodeBuilder:
   def body(t: Block)(using Raise, Scope): Document = scope.nest givenIn:
     block(t)
   
-  
+  def handleFunction(ps: List[semantics.Param], b: Block)(using Raise, Scope): (Document, Document) =
```
LPTK: Let's rename this to
```suggestion
  def setupFunction(params: List[semantics.Param], body: Block)(using Raise, Scope): (Document, Document) =
```
As "handle" is a bit too generic.

hkmc2/jvm/src/test/scala/hkmc2/JSBackendDiffMaker.scala
```diff
@@ -13,6 +18,7 @@ abstract class JSBackendDiffMaker extends MLsDiffMaker:
   val sjs = NullaryCommand("sjs")
   val showRepl = NullaryCommand("showRepl")
   val silent = NullaryCommand("silent")
+  val jsSanityCheck = NullaryCommand("jsSanityCheck")
```
LPTK: Pls rename to
```suggestion
  val noSanityChecks = NullaryCommand("noSanityChecks")
```
so that sanity checks are the default.

hkmc2/jvm/src/test/scala/hkmc2/JSBackendDiffMaker.scala
```diff
@@ -116,3 +122,13 @@ abstract class JSBackendDiffMaker extends MLsDiffMaker:
       
       
 
+
+class JSBuilderSanityCheck extends JSBuilder:
```
LPTK: This should be a trait that lives alongside `JSBuilder`. We will want to use it for debug builds outside of the diff-tests.

LPTK: Pls rename to
```suggestion
class JSBuilderSanityChecks extends JSBuilder:
```

hkmc2/jvm/src/test/scala/hkmc2/JSBackendDiffMaker.scala
```diff
@@ -116,3 +122,13 @@ abstract class JSBackendDiffMaker extends MLsDiffMaker:
...
+  override def handleFunction(ps: List[semantics.Param], b: Block)(using Raise, Scope): (Document, Document) =
+    val params = ps.map(p => Scope.scope.allocateName(p.sym))
+    val paramsStr = doc"args_" :: params.mkDocument("_")
```
LPTK: What if this name conflicts with an already-defined name in the source? The proper thing to do here is to use a `TempSymbol` that you make sure to allocate in the scope and that you then `lookup_!`. There does not need to be one such symbol per function. You can create a single one at the top-level of `JSBuilderSanityChecks`.

hkmc2/jvm/src/test/scala/hkmc2/JSBackendDiffMaker.scala
LPTK: It would be best to make this trait take a runtime parameter called `instrument: Bool`, so that whether the instrumentation is done can be configured at runtime (in the difftests or elsewhere).

hkmc2/jvm/src/test/scala/hkmc2/JSBackendDiffMaker.scala
```diff
@@ -42,7 +48,7 @@ abstract class JSBackendDiffMaker extends MLsDiffMaker:
     if js.isSet then
       val low = ltl.givenIn:
         codegen.Lowering()
-      val jsb = codegen.js.JSBuilder()
+      val jsb = if jsSanityCheck.isSet then JSBuilderSanityCheck() else JSBuilder()
```
LPTK: Re suggestions above:
```suggestion
      val jsb = new JSBuilder with JSBuilderSanityCheck(noSanityChecks.isUnset)
```

hkmc2/shared/src/test/mlscript/codegen/SanityCheck.mls
LPTK, line 1: Rename to `SanityChecks.mls`.

hkmc2/shared/src/test/mlscript/codegen/SanityCheck.mls
```diff
@@ -0,0 +1,184 @@
...
+fun f(x, y) = x + y
+f(2)
+//│ JS:
+//│ function f(x, y) { return x + y; } this.f(2)
+//│ = NaN
+
+:sjs
+:re
+:jsSanityCheck
+fun f(x, y) = x + y
```
LPTK: No need to redefine this sanity-checked function. Just use the previous one.

LPTK (changes requested):
Nicely done, thanks! But please address my comments.

ychenfo:
Thanks for the comments! I have made the changes except for the one that makes sanity checks default, because currently several tests have function calls with incorrect number of arguments, but without `:re`. Should I make the change to add `:re` to them and make sanity checks default?

Also previously I use `const x = args[i]` to assign arguments values from varargs, but I just saw https://github.com/hkust-taco/mlscript/pull/238 and realize that arguments can be mutated, so I change to use `let`.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -422,3 +422,18 @@ object JSBuilder:
 end JSBuilder
 
 
+trait JSBuilderSanityChecks(val instrument: Bool) extends JSBuilder:
```
LPTK: No reason to make this a `val`. Doing so will prevent other similar traits (copied from this model) from defining their own `instrument` parameter.

LPTK:
> currently several tests have function calls with incorrect number of arguments, but without `:re`. Should I make the change to add `:re` to them

Well, most are probably actual mistakes! Unless the intent was clearly to test what happens when the wrong number of arguments is passed, please just fix them by passing the correct number of arguments!

ychenfo:
Thanks for the comments! I have removed the `val`, and made the following changes to the tests and made sanity check on by default.

- add an `:re` at line58 in `ClassInFun.mls`
- add missing parameters at around line120 in `IfThenElse.mls`
- add an `:re` at line40 in `Juxtapositions.mls`
- add a `:fixme` at line 170 in `Juxtapositions.mls`

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -417,3 +422,18 @@ object JSBuilder:
...
 
+trait JSBuilderSanityChecks(instrument: Bool) extends JSBuilder:
+  
+  val functionParamVarargSymbol = semantics.TempSymbol(0, N, "args")
+  
+  override def setupFunction(params: List[semantics.Param], body: Block)(using Raise, Scope): (Document, Document) =
+    if instrument then
+      val paramsList = params.map(p => Scope.scope.allocateName(p.sym))
+      val paramsStr = Scope.scope.allocateName(functionParamVarargSymbol)
+      val checkArgsNum = doc"if ($paramsStr.length !== ${params.length}) { throw new globalThis.Error('got ' + $paramsStr.length + ' arguments, expecting ' + ${params.length}) }\n"
```
LPTK: I would change this boilerplate to:
```suggestion
      val checkArgsNum = doc"globalThis.Predef.checkArgs($functionName, $paramsStr.length, ${params.length})"
```
with the appropriate helper method defined in `Predef.mls`.

LPTK (approved):
Great! Just one last touch. And the errors would be more helpful if they had the following  format:

```
Function 'foo' expected 3 arguments but got 2.
```

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
ychenfo: Thanks! I have pushed the commit to improve the error message. As for the helper method `checkArgs` in `Predef.mls`, there seems to be no way to throw errors in `*.mls` files..? Sorry that I vaguely remember that you told me about this in our previous meeting, but cannot recall the detail.

hkmc2/shared/src/test/mlscript/codegen/SanityChecks.mls
```diff
@@ -172,4 +172,4 @@ Cls(1, 2).f(3, 4)(5)
 //│ tmp = this.Cls(1, 2);
 //│ tmp1 = tmp.f(3, 4);
 //│ tmp1(5)
-//│ ═══[RUNTIME ERROR] Error: got 1 arguments, expecting 2
+//│ ═══[RUNTIME ERROR] Error: Function 'arrow function' expected 2 arguments but got 1
```
LPTK: > Function 'arrow function' expected [...]

🤔

LPTK: In this case, just make it

> Function expected [...]

i.e., make the function name optional.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -426,14 +426,14 @@ trait JSBuilderSanityChecks(instrument: Bool) extends JSBuilder:
   
   val functionParamVarargSymbol = semantics.TempSymbol(0, N, "args")
   
-  override def setupFunction(params: List[semantics.Param], body: Block)(using Raise, Scope): (Document, Document) =
+  override def setupFunction(name: Str, params: List[semantics.Param], body: Block)(using Raise, Scope): (Document, Document) =
     if instrument then
       val paramsList = params.map(p => Scope.scope.allocateName(p.sym))
       val paramsStr = Scope.scope.allocateName(functionParamVarargSymbol)
-      val checkArgsNum = doc"if ($paramsStr.length !== ${params.length}) { throw new globalThis.Error('got ' + $paramsStr.length + ' arguments, expecting ' + ${params.length}) }\n"
+      val checkArgsNum = doc"if ($paramsStr.length !== ${params.length}) { throw new globalThis.Error('Function \'$name\' expected ' + ${params.length} + ' arguments but got ' + $paramsStr.length) }\n"
```
LPTK: Please generate strings that use `"` and don't forget to escape `name` for weird characters. You can reuse the logic in `JSBuilder.makeStringLiteral` but extract that logic to a `JSBuilder.escapeStringCharacters` helper function.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
LPTK: Oh, do you mean that we're missing the `throw` keyword? This should be a trivial thing to add (it's already supported internally by the IR), but we can do it in a separate PR.

LPTK: In case you want to do it: I would add `throw` as a symbol in `Elaborator.unaryOps` and handle it specially in `Lowering`.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
ychenfo: Thanks! I have made this and the previous changes to make function name optional.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
ychenfo: Yes, I mean the `throw` keyword is missing, and yes I would like to do this. I just took a look, and it seems that `Elaborator.unaryOps` is only used [here](https://github.com/hkust-taco/mlscript/blob/e3859934e2c38db20f2a6f6998aba55dc49940a9/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala#L105-L107), which assumes all unaryOps must also be binary ops, so maybe there are more changes needed? Maybe I can look into how `return` is handled and do something similar?

ychenfo: I tried to implement the support for `throw` following how `return` is handled and created a new PR https://github.com/hkust-taco/mlscript/pull/239

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -426,11 +429,13 @@ trait JSBuilderSanityChecks(instrument: Bool) extends JSBuilder:
   
   val functionParamVarargSymbol = semantics.TempSymbol(0, N, "args")
   
-  override def setupFunction(name: Str, params: List[semantics.Param], body: Block)(using Raise, Scope): (Document, Document) =
+  override def setupFunction(name: Option[Str], params: List[semantics.Param], body: Block)(using Raise, Scope): (Document, Document) =
     if instrument then
       val paramsList = params.map(p => Scope.scope.allocateName(p.sym))
       val paramsStr = Scope.scope.allocateName(functionParamVarargSymbol)
-      val checkArgsNum = doc"if ($paramsStr.length !== ${params.length}) { throw new globalThis.Error('Function \'$name\' expected ' + ${params.length} + ' arguments but got ' + $paramsStr.length) }\n"
+      val fname = name.map(n => s" '${JSBuilder.escapeStringCharacters(n)}' ").getOrElse(" ")
```
LPTK: This should be a `fold`.

ychenfo: Thanks! I just changed this `map` and `getOrElse` to `fold` and pushed the commit.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
LPTK: Thanks, merged!

ychenfo: Thanks! I have pushed the changes to use `Predef.checkArgs`.

PR #238, opened 2024-11-19
Fix scoping

AnsonYeung:
Fixes issue when setting a variable living in an outer scope.


LPTK (approved):
Nice, thanks for the fix!

PR #239, opened 2024-11-21
Add support for `throw` keyword

ychenfo:


hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -20,6 +20,10 @@ object Ret extends Ret:
 object ImplctRet extends Ret:
   def apply(r: Result): Block = Return(r, implct = true)
 
+abstract class Thrw extends (Result => Block)
```
LPTK: There's no need to add yet another abstract class and `isInstanceOf` check. Just rename the `Ret` abstract class (not the object) to `TailOp` and make `Thrw` extend it.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -11,7 +11,7 @@ import hkmc2.semantics.{Term => st}
 
 import syntax.{Literal, Tree}
 import semantics.*
-import semantics.Term.*
+import semantics.Term.{Throw => _, *}
```
LPTK, line 14: I don't think this change is needed.

LPTK (approved):
Great! Please make the suggested change, and I will merge.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
ychenfo: Oh I made this change because of the name clash between `Term.Throw` and `codegen.Throw`, and `Term.Throw` is only used [here](https://github.com/CrescentonC/mlscript/blob/dbf5b01aa5c6191373bd20b94bd70e81f9abb1e6/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala#L58) by `st.Throw`, while `codegen.Throw` is already used [here](https://github.com/CrescentonC/mlscript/blob/dbf5b01aa5c6191373bd20b94bd70e81f9abb1e6/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala#L264), so I hide the one imported from `Term`.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
ychenfo: Thanks! I have pushed the change.

PR #242, opened 2024-11-28
Naïve compilation of refining patterns

chengluyu:


hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -190,7 +191,7 @@ class JSBuilder(using Elaborator.State) extends CodeBuilder:
                     } + ")""""
                 }; }"""
               } #}  # }"
-            if (clsDefn.kind is syntax.Mod) || (clsDefn.kind is syntax.Obj) then
+            if ((clsDefn.kind is syntax.Mod) || (clsDefn.kind is syntax.Obj)) || (clsDefn.kind is syntax.Pat) then
```
LPTK: The extra parens are not needed.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -109,8 +109,18 @@ extends Importer:
   def mkLetBinding(sym: LocalSymbol, rhs: Term): Ls[Statement] =
     LetDecl(sym) :: DefineVar(sym, rhs) :: Nil
   
-  def resolveField(srcTree: Tree, base: Opt[Symbol], nme: Ident): Opt[Symbol] =
+  def resolveField(srcTree: Tree, base: Opt[Symbol], nme: Ident): Ctxl[Opt[Symbol]] =
     base match
+    // Look up symbols qualified by `globalThis.`.
+    case S(tsym: TopLevelSymbol) =>
```
LPTK: Can you explain why this is needed?

hkmc2/shared/src/main/scala/hkmc2/syntax/ParseRule.scala
```diff
@@ -272,6 +272,20 @@ class ParseRules(using State):
             End(N),
           )
         ) { (lhs, rhs) => TypeDef(Als, lhs, rhs, N) },
+    Kw(`pattern`):
```
LPTK: Couldn't we share the code for type synonyms? Place it in a common shared function.

LPTK (commented):
Nice feature! I just have a few minor comments for now. Please address them first.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
chengluyu: It fixes the issue where we can't get the class symbol when accessing a class in `Predef` through `globalThis`. I need to match terms against `MatchResult`. But I found:
```scala
MatchResult(()) is globalThis.Predef.MatchResult(_)
//│ ╔══[ERROR] Cannot use this selection as an extractor
//│ ║  l.19: 	MatchResult(()) is globalThis.Predef.MatchResult(_)
//│ ╙──      	                   ^^^^^^^^^^
//│ = false
```
I followed the `cls` function and found the `resolveField` function, but it doesn't handle the case where base is `globalThis` (`TopLevelSymbol`).

**But I just tried it again, and I realized the silly part of this code, the `find` function is redundant, and I will fix it.**

LPTK: But what's the justification for writing such code as `is globalThis.Predef.MatchResult(_)` rather than `is Predef.MatchResult(_)`? And if it's not written by the user, you should be able to just add the correct symbol in the section term when constructing it...

chengluyu: > But what's the justification for writing such code as `is globalThis.Predef.MatchResult(_)` rather than `is Predef.MatchResult(_)`?

To prevent `Predef` from being shadowed.

> you should be able to just add the correct symbol in the section term when constructing it...

Maybe this should be the right way. But I haven't found a way to get the exact symbols defined in `Predef`. We access functions via `globalThis.Predef.tupleSlice` but we access classes by `Predef.MatchResult`, which feels very inconsistent.

LPTK: What on Earth are you talking about?

LPTK: Users are NEVER supposed/expected to make use of `globalThis`.

chengluyu: Sorry for the confusion. The code in my initial response is the term I want to generate (I really don't need/expect the user to write it and I also never do that).

In this PR, I need to generate a term similar to `unapply(scrut) is MatchResult(…)` and `MatchResult` is defined in `Predef`. I know I can use `Predef.MatchResult`. But I was told that when generating terms, if I need to use things defined in `Predef`, I should prefix it with `globalThis` to prevent shadowing.

For example, if I need to use `tupleSlice`, the term should be `globalThis.Predef.tupleSlice`. So, to use `MatchResult`, I generate the term `globalThis.Predef.MatchResult`. But the problem is that this term doesn't have a class symbol, therefore it can't be used in pattern matching.

To solve this problem, I modified the `resolveField` function so that it can look up from the context when it encounters the `TopLevelSymbol`.

I apologize for not making it clear. This might not be important at all, as nobody would override `Predef`.

LPTK: Plenty of people might shadow the word "Predef". You can't just say this kind of things isn't important. Correctness is important.

I finally had to check your code myself to find that one place you're talking about because you won't just give me a link or line number. It appears to be this:

```scala
  protected lazy val matchResultClass =
    Sel(Sel(Ident("globalThis"), Ident("Predef")), Ident("MatchResult"))
```

This does not look right. You're not **_generating a term_** here, as you keep writing, you're **_generating a tree_** and then using the elaborator to transform that into a term for you.

If you were truly generating a term, as you're supposed to (trees are what users write, they're generally not meant to be generated, in normal circumstances), you'd be using `Term.Sel` (or rather `Term.SynthSel`), which _allows you to pass the relevant symbol_ at construction. That's why your roundabout approach is not needed and was confusing me.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/DesugaringBase.scala
```diff
@@ -0,0 +1,122 @@
...
+      scrut: => Term.Ref,
+      psym: PatternSymbol,
+      inner: (scrut: TempSymbol) => Split,
+      method: Str = "unapplyStringPrefix"
+  )(fallback: Split): Ctxl[Split] =
+    val matchResultClassTerm = cls(matchResultClass)
+    matchResultClassTerm.symbol match
+    case S(matchResultClassSymbol: ClassSymbol) =>
+      // def makeUnapplyBranch(scrut: => Term.Ref, pat)
+      val resultIdent = Ident("matchResult"): Ident
```
LPTK: This appears to be unused.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/DesugaringBase.scala
```diff
@@ -0,0 +1,122 @@
...
+      val resultSymbol = TempSymbol(N, "matchResult")
+      val globalThis = term(Ident("globalThis"))
```
LPTK: Instead of this hacky approach, you can use `State.globalThisSymbol.ref()`.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/DesugaringBase.scala
```diff
@@ -0,0 +1,122 @@
...
+      case S(_) | N => lastWords("Cannot locate `MatchResult` class in the global scope.")
+
+  /** Make a `Branch` that calls `Pattern` symbols' `unapplyStringPrefix` functions. */
+  def makeUnapplyStringPrefixBranch(
+      scrut: => Term.Ref,
+      psym: PatternSymbol,
+      inner: (scrut: TempSymbol) => Split,
+      method: Str = "unapplyStringPrefix"
+  )(fallback: Split): Ctxl[Split] =
+    val matchResultClassTerm = cls(matchResultClass)
```
LPTK: As I explained in the [other comment](https://github.com/hkust-taco/mlscript/pull/242#discussion_r1863721735), this is not the right approach. But also notice that you're recreating the term every time this function is called, making it doubly wrong.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/DesugaringBase.scala
chengluyu: Thanks! I just learned that I can get this symbol here.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
chengluyu: Thank you for the careful review and your detailed explanation. I'm sorry for not providing you with a link or line number. I'd like to ask, how can I obtain the symbols of `Predef` and the class I need?

For example, now I know I can construct the term in the following way. Where can I get `A` and `B`?

```scala
Term.Sel(
  prefix = Term.Sel(
    prefix = state.globalThisSymbol.ref(),
    nme = Ident("Predef")
  )(S(/* A */)), // <- needs the symbol for `Predef`
  nme = Ident("MatchResult")
)(S(/* B */)) // <- needs the symbol for class `MatchResult`
```

Of course, I know these symbols are all in `Ctx`, but how do I know that what I found is the exact `Predef` imported by the compiler and not the one shadowed by the user?

LPTK: That's a good question, and now I understand your problem better. Thanks for explaining it.

What's for sure is that you don't want to write something like `ctx.get(nme.name).flatMap(_.symbol)` in `resolveField`, as that's just incorrect.

What I would probably do is declare `MatchResult` in `Prelude.mls`, which is used to declare builtins, and then I would use add a line in `Ctx#Builtins`, which can get the corresponding builtin symbols.

However, this will not automatically be the same as the symbol that's eventually going to be created in the `Predef` module, so users won't be able to use `MatchResult` explicitly yet. We can fix this later.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/DesugaringBase.scala
chengluyu: Thanks! I've removed it.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
chengluyu: Thank you for your help! I've already rebased this PR to your commit. Now my implementation does not construct terms and retrieve symbols through elaboration at all.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/DesugaringBase.scala
chengluyu: Thanks for the check. I've already modified this part of the code. Now I don't get symbols or construct terms through `cls` or `term`.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Translator.scala
```diff
@@ -153,31 +144,29 @@ class Translator(tl: TraceLogger, val elaborator: Elaborator)
...
       stringPrefix(scrut, prefix, (captures1, postfixScrut1) =>
         stringPrefix(postfixScrut1, postfix, (captures2, postfixScrut2) =>
           inner(captures2 ++ captures1, postfixScrut2)))
     case ctor @ (_: Ident | _: Sel) =>
-      val clsTrm = elaborator.cls(ctor, inAppPrefix = false)
-      clsTrm.symbol.flatMap(_.asClsLike) match
+      elaborator.cls(ctor, inAppPrefix = false).symbol.flatMap(_.asClsLike) match
       case S(cls: (ClassSymbol | ModuleSymbol)) =>
-        raise(ErrorReport(msg"Cannot match ${ctor.describe} as a string prefix" -> ctor.toLoc :: Nil))
+        val kind = cls match { case _: ClassSymbol => "class" case _ => "module" }
```
LPTK, line 152: This should rather use `cls.tree.k`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -36,13 +36,19 @@ abstract class Symbol(using State) extends Located:
     case mem: BlockMemberSymbol =>
       mem.modTree.map(_.symbol.asInstanceOf[ModuleSymbol])
     case _ => N
+  def asPat: Opt[PatternSymbol] = this match
+    case pat: PatternSymbol => S(pat)
+    case mem: BlockMemberSymbol =>
+      mem.patTree.map(_.symbol.asInstanceOf[PatternSymbol])
```
LPTK: This should probably be
```suggestion
      mem.patTree.flatMap(_.symbol.asPat)
```
(create the `asPat` method if it's not already there)

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/DesugaringBase.scala
```diff
@@ -0,0 +1,116 @@
...
+import mlscript.utils.*, shorthands.*
+import syntax.Tree.*, Elaborator.{Ctxl, ctx}
+
+/** Contains some helpers that makes UCS desugaring easier. */
+trait DesugaringBase(using state: Elaborator.State):
+  val elaborator: Elaborator
+
+  import elaborator.tl.*, state.globalThisSymbol
+
+  protected transparent inline def sel(p: Term, k: Ident): Term.SynthSel = Term.SynthSel(p, k)(N)
```
LPTK: Why are all these functions `transparent inline`? Unless you're doing some type-level magic, they should just be `inline`. In fact, I don't think they should really be `inline`. Just making them `final` should be enough hint that they could be optimized by the JIT if needed.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/DesugaringBase.scala
```diff
@@ -0,0 +1,116 @@
...
+    Split.Let(s, cond, Branch(s.ref(), inner) ~: Split.End)
+
+  /** Make a `Branch` that calls `Pattern` symbols' `unapply` functions. */
+  def makeUnapplyBranch(
+      scrut: => Term.Ref,
+      psym: PatternSymbol,
+      inner: => Split,
+      method: Str = "unapply"
+  )(fallback: Split): Ctxl[Split] =
+    val func = sel(sel(globalThisSymbol.ref(), psym.nme, psym), method)
```
LPTK: I don't understand the logic of this. You've got a symbol pointing to a pattern. And here you just _assume_ that pattern is defined in `globalThis`?! What if it's defined in some module? You should use the pattern TERM here (currently called `clsTrm` at the call site). That's what terms are for.

You should only use `globalThis` when accessing hard-coded things like `Predef` that we KNOW are there. Using it here is nonsensical.

LPTK: Please fix this and also:
* Address other similar incorrect cases in this PR.

LPTK: Please fix this and also:
* Add some tests where the patterns are defined inside some modules (these can be modules defined locally, in the same diff-test file).

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Translator.scala
```diff
@@ -0,0 +1,246 @@
...
+   */
+  private type Scrut = () => Term.Ref
+  
+  private type CaptureMap = Map[Param, Term.Ref]
+  
+  private type Inner = CaptureMap => Split
+  
+  private type PrefixInner = (CaptureMap, Scrut) => Split
+  
+  import Branch as B, Pattern as P, Split as Sp
```
LPTK: Please don't add a renaming import right in the middle of the source file. Is the saving of a few characters worth the absolute confusion that ensures for readers? I'd suggest removing the import entirely and just referring to the things by their original names. They're already quite short.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Translator.scala
```diff
@@ -0,0 +1,246 @@
...
+  private def prefixSuccess(params: Ls[Param]): PrefixInner =
+    val paramIndexMap = params.zipWithIndex.toMap
+    (captures, postfixScrut) => trace(
+      pre = s"prefixSuccess <<< ${params.iterator.map(_.sym).mkString(", ")}", 
+      post = (split: Sp) => s"prefixSuccess >>> ${display(split)}"
+    ):
+      require(captures.forall(_._1 |> paramIndexMap.contains))
+      if captures.size != params.size then
+        // TODO: report uncaptured parameters
+        error(msg"Unmatched number of captures and parameters." -> N)
```
LPTK, line 195: It seems there are no test cases for such errors.

LPTK (changes requested):
Thanks for the nice work! But there are still some things to address.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Translator.scala
```diff
@@ -153,31 +144,29 @@ class Translator(tl: TraceLogger, val elaborator: Elaborator)
...
+    case lit @ StrLit(value) =>
+      plainTest(callStringStartsWith(scrut(), Term.Lit(lit), "startsWith")):
+        tempLet("sliced", callStringDrop(scrut(), value.length, "sliced")): slicedSym =>
           inner(Map.empty, () => slicedSym.ref())
     case App(Ident("~"), Tup(prefix :: postfix :: Nil)) =>
       stringPrefix(scrut, prefix, (captures1, postfixScrut1) =>
         stringPrefix(postfixScrut1, postfix, (captures2, postfixScrut2) =>
           inner(captures2 ++ captures1, postfixScrut2)))
     case ctor @ (_: Ident | _: Sel) =>
-      val clsTrm = elaborator.cls(ctor, inAppPrefix = false)
```
LPTK: This `val clsTrm` pattern is used several times. May be better to keep it here for regularity.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/DesugaringBase.scala
chengluyu: Oops! I really messed up here. I must have been out of my mind at the time.

chengluyu: Also, thanks for the catch. I did not find this during my self review.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
chengluyu: Thanks. I guess I need to modify other methods as well?

https://github.com/chengluyu/mlscript/blob/37e9025c1af976d79b97a40766f8f5a37cef76bd/hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala#L29-L48

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Translator.scala
chengluyu: This is the groundwork for captures/extraction, which has not been fully implemented yet. I have partially implemented it locally and showed it to you. Perhaps I can make a note here and add tests after it's implemented?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Translator.scala
chengluyu: Thanks. There is a small problem, `tree` is a `val` defined on both `ClassSymbol` and `ModuleSymbol` independently (no base class has such a field), so I can't simply access it with `cls.tree`. Do you have any suggestions?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/DesugaringBase.scala
chengluyu: I added tests to test the patterns defined in the module and also tested their cross-references with patterns defined in the file.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
LPTK: Oh, true. Sure.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Translator.scala
LPTK: Yes please.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Translator.scala
LPTK: Oh, I think you can just declare an abstract `val` in `MemberSymbol`.

chengluyu: If that's the case, I need to implement `tree: Tree` in `BlockMemberSymbol`, `TermSymbol`, `TypeAliasSymbol`, and `TopLevelSymbol`. Sorry, I need more guidance.

LPTK: Ok forget it for now.

PR #243, opened 2024-11-29
Sanity Checks for Selection

ychenfo:


LPTK (approved):
Thanks, this is a good first step!

But I'd actually like to refine the scheme a little. We should make the default result `null` rather than `undefined` wherever we can. So MLscript functions that don't return a specific value should return `null`. And it shouldn't be necessary to instrument with `?? null` calls known to be to MLscript functions. You can check that by seeing if the function being called has an associated symbol. This should be done in the Elaborator, which should wrap only unknown calls into an `OrNull(trm)` wrapper that results in `?? null` in JS.

Let's do that in a new PR.

PR #244, opened 2024-11-30
Add `trace` functionality

ychenfo:
~~TODO:~~
- [x] ~~avoid obscure tracing outputs like [here](https://github.com/CrescentonC/mlscript/blob/e30292575acbabb576ece99fd0f7ee5911a530c0/hkmc2/shared/src/test/mlscript/codegen/TraceLog.mls#L217-L230), which is caused by the `toString()` methods, which are called to compose tracing messages.~~
- [x] ~~cleanup~~

The to avoid obscure tracing outputs and non-termination caused by `toString()`, I currently use `util.inspect` to show the argument and return result info in the tracing messages.


hkmc2/shared/src/test/mlscript-compile/Predef.mls
```diff
@@ -31,6 +31,15 @@ fun checkArgs(functionName, expected, isUB, got) =
     //   + " arguments but got " + got)
   else ()
 
+module TraceLogger with
+  mut val indentLvl = 0
+  fun indent() =
+    let prev = indentLvl
+    set indentLvl = indentLvl + 1
```
LPTK: Again, why not use `prev` here?

LPTK:
Obviously you shouldn't pollute every single diff test block with a useless call to `this.Predef.TraceLogger.resetIndent(0)`. Why is that call even necessary, when we do enable tracing?

LPTK:
There is no need to use `:sjs` on so many test cases. Just pick one representative example and leave the others without the command. So you realize that the more useless noise you add to the difftests, the more work it I will be for us to review future changes?

hkmc2/shared/src/test/mlscript-compile/Predef.mls
ychenfo: Thanks! I have pushed the commit to use `prev` here.

ychenfo:
> Obviously you shouldn't pollute every single diff test block with a useless call to `this.Predef.TraceLogger.resetIndent(0)`. Why is that call even necessary, when we do enable tracing?

Thanks, I just realized that I could have done this better without polluting every single diff test block by using `try/finally` in the instrumentation, such that in the `finally` block the indentation level is always reset regardless of whether there is an uncaught error. I will try to make the change now.

I originally wrote this to reset indentation at the beginning of every diff test block because if a previous block throws an error at a higher indent level, blocks after that will start the trace log at a wrong level (this is because functions are instrumented when they are declared in blocks with flag `traceJS`, so the log will be printed no matter whether at call site `traceJS` is on).


> There is no need to use :sjs on so many test cases. Just pick one representative example and leave the others without the command. So you realize that the more useless noise you add to the difftests, the more work it I will be for us to review future changes?

Thanks I see, I will also make the change to remove unnecessary `:sjs`s.

ychenfo:
I have made the changes to cleanup `:sjs` in tests and use `try/finally` to avoid polluting diff test blocks. I also made a minor improvement for multiline trace messages.

hkmc2/shared/src/test/mlscript/codegen/TraceLog.mls
```diff
@@ -0,0 +1,185 @@
...
+//│ > | | calling: toInt(S { x: N {} })
+//│ > | | | calling: toInt(N {})
+//│ > | | | return: 0
+//│ > | | return: 1
+//│ > | return: 2
+//│ > return: 'S(S(N: 0): 1): 2'
+//│ = 'S(S(N: 0): 1): 2'
+
+
+class Cls(a, b, c, d, e, f)
```
LPTK, line 113: I guess the point in having so many fields is to make it break the line. But that makes the output quite noisy, so could you ensure it only prints the thing a couple of times?

LPTK (changes requested):
Thanks, great work.

Some remarks, though:

- [x] I actually preferred the way without `finally`, as it's much less invasive and likely to add much less runtime overhead. Can we move back to it? Before a block that contains tracing, in `JSBackendDiffMaker` you could always just `host.send` a nodeJS command to manually reset the indent, as in `host.send("globalThis.Predef.TraceLogger.indent = 0")`

- [x] I'd prefer it if tracing was disabled in non-tracing blocks, even when the block's execution involves functions that were instrumented. You could simply add a Boolean flag to `Predef.TraceLogger` (or set the indent to -1, tho I don't like sentinel values) which is always `false` unless we're tracing.

- [x] I don't like the "`calling`"/"`return`" pair. Repetitive but not very easy to parse visually. Have you considered using something like the following syntax?

    ```
    f(1,1)(-2)
    //│ > CALL g(-2)
    //│ > | CALL g(-1)
    //│ > | => 42
    //│ > => 42
    //│ = 44
    ```

hkmc2/shared/src/test/mlscript/codegen/TraceLog.mls
ychenfo: Thanks! I have pushed the change to modify this test so that only one multiline trace message is shown.

ychenfo:
> Some remarks, though

Thanks for your suggestions and instructions! I have pushed the changes to avoid `try/finally`, and in `JSBackendDifftest` use `host.execute` to directly reset the indentation level for those diff test blocks that are marked with `traceJS` flag. I added a flag in `Predef.TraceLog` so that all functionalities of tracelog is disabled when `traceJS` is not set for a given block, even if it calls some instrumented functions. And now `CALL/=>` is used.

hkmc2/jvm/src/test/scala/hkmc2/JSBackendDiffMaker.scala
```diff
@@ -110,8 +110,16 @@ abstract class JSBackendDiffMaker extends MLsDiffMaker:
                 source = Diagnostic.Source.Runtime))
         if stderr.nonEmpty then output(s"// Standard Error:\n${stderr}")
       
+      
+      if traceJS.isSet then
+        host.execute("globalThis.Predef.TraceLogger.enabled = true")
+        host.execute("globalThis.Predef.TraceLogger.resetIndent(0)")
```
LPTK: Why not execute both lines in a single call? Splitting them is unnecessarily inefficient.

hkmc2/shared/src/test/mlscript/codegen/TraceLogIndent.mls
```diff
@@ -0,0 +1,92 @@
...
+  x + y + g(z)
+
+:traceJS
+fun h(x) =
+  f(1,2)(x)
+
+:traceJS
+:re
+h(3)
+//│ ═══[RUNTIME ERROR] Error: Function 'plus' expected 2 arguments but got 1
```
LPTK, line 28: Any idea why we're not seeing the trace up to the runtime error?

ychenfo: Yes, because when the ReplHost encounters an error, only the error messages are kept in its replies (https://github.com/hkust-taco/mlscript/blob/99863022a7a9089e162a8d8387c55c17a931daf7/hkmc2/jvm/src/test/scala/hkmc2/ReplHost.scala#L197-L205), maybe I can also make a patch to this in this PR?

LPTK: Sure, that would be great, if you don't mind.

ychenfo: Thanks! I have pushed a commit to address this.

hkmc2/jvm/src/test/scala/hkmc2/JSBackendDiffMaker.scala
```diff
@@ -108,6 +108,10 @@ abstract class JSBackendDiffMaker extends MLsDiffMaker:
               // Otherwise, it is considered a simple runtime error.
               raise(ErrorReport(msg"${message}" -> N :: Nil,
                 source = Diagnostic.Source.Runtime))
+            if traceJS.isSet then
+              output("All outputs:")
```
LPTK: What's with this weird "All outputs" thing? Why would the logging output preceding an error only appear when `traceJS.isSet`? I don't understand how you make this sort of decisions. Is there a relevant rationale, from the point of view of the user of the system?

LPTK: PS: Implementation or historical details are not relevant rationales.

LPTK: I guess you were not thinking about the fact that the output is useful even when no tracing is ongoing. Indeed, that's the output of printing stuff, which is currently being silenced when errors occurs (very annoying).

ychenfo: I previously chose to only show all outputs when `traceJS` is set because I found out that in most of the tests, "all output"s are just the duplicates of the error messages:
image

But indeed I now see that the output is useful even when no tracing is ongoing, I think maybe I can just not print out the lines showing duplicated messages and "undefined"? Or maybe I can another flag which controls whether to print all output when error?

LPTK: > Or maybe I can another flag which controls whether to print all output when error?

No, please! Not another flag. Again, I ask you to think about the _semantics_ of what we're implementing here. How could adding a flag like this that controls low-level details in surprising ways be any good?

Rather, you should figure out why the runtime error appears part of the retrieved output. I believe errors are actually guarded by some invisible unicode characters. Just check how this part of the code works and simply _fix it_. It's not very complicated at all, as far as I remember.

And of course we should also not print the final `undefined`. I'm confident you can figure it out.

ychenfo: Thanks, I have just pushed the changes to filter error messages and final `undefined` in the retrieved output for `ReplHost.Error`. Following the implementation in `ReplHost.scala` I realized that uncaught errors are not guarded by special invisible characters, so I just filter them using the index of substring "Uncaught " ([here](https://github.com/CrescentonC/mlscript/blob/4f4083bd017ee9d3d924db80c6888f4236d30b73/hkmc2/jvm/src/test/scala/hkmc2/ReplHost.scala#L56) and [here](https://github.com/CrescentonC/mlscript/blob/4f4083bd017ee9d3d924db80c6888f4236d30b73/hkmc2/jvm/src/test/scala/hkmc2/ReplHost.scala#L51)); And others are filtered out [by using the invisible character](https://github.com/CrescentonC/mlscript/blob/4f4083bd017ee9d3d924db80c6888f4236d30b73/hkmc2/jvm/src/test/scala/hkmc2/ReplHost.scala#L88)

hkmc2/shared/src/test/mlscript/codegen/TraceLogIndent.mls
LPTK: Great!

One last tweak, though: can we get the runtime error message after the output? Currently, it appears out of logical order:

```
//│ ═══[RUNTIME ERROR] Error: Function 'plus' expected 2 arguments but got 1
//│ CALL h(3)
//│ | CALL g(3)
//│ | | CALL g(2)
//│ | | | CALL g(1)
//│ | | | | CALL g(0)
```

It should be:


```
//│ CALL h(3)
//│ | CALL g(3)
//│ | | CALL g(2)
//│ | | | CALL g(1)
//│ | | | | CALL g(0)
//│ ═══[RUNTIME ERROR] Error: Function 'plus' expected 2 arguments but got 1
```

ychenfo: Thanks, I have pushed the change. I also added the missing ">" at the start of each line of the output, so that now it's:
```
//│ > CALL h(3)
//│ > | CALL g(3)
//│ > | | CALL g(2)
//│ > | | | CALL g(1)
//│ > | | | | CALL g(0)
//│ ═══[RUNTIME ERROR] Error: Function 'plus' expected 2 arguments but got 1
```

hkmc2/jvm/src/test/scala/hkmc2/JSBackendDiffMaker.scala
LPTK: Hmmm... But what if someone prints a message that includes the string "Uncaught "?

ychenfo: Yeah then the test will fail due to a spurious error caused by the message starting with "Uncaught" that is printed... The special invisible characters for caught errors are added by us in [`ReplHost.scala`](https://github.com/hkust-taco/mlscript/blob/76b129a2bb2daac94ddca9bb6180b654d0044841/hkmc2/jvm/src/test/scala/hkmc2/ReplHost.scala#L107) instead of by nodejs, so the detection of uncaught errors currently are entirely dependent on [this two special strings](https://github.com/hkust-taco/mlscript/blob/76b129a2bb2daac94ddca9bb6180b654d0044841/hkmc2/jvm/src/test/scala/hkmc2/ReplHost.scala#L142-L143)

image

LPTK: Well that's nothing short of revolting, but I guess we don't have a better way for now...

PR #245, opened 2024-12-01
Add code generation for bbml

NeilKleistGao:
In this PR:
- Generate code for mutable reference operations in bbml
- Fix some problems with BbML typing
- Map builtin operators and floating operators
- Add predef for BbML
- Add getter code generation and corresponding infrastructures

<del>However, I'm not sure if we still need to make `get` a keyword. There are already some functions using `get` as the name and it seems hard to repick names for them.</del>


LPTK:
> There are already some functions using `get` as the name and it seems hard to repick names for them.

Which functions are you talking about? They should all be in this repo and so the migration should be easy.

Perhaps we could just use a naming scheme. Functions called `get_XXX` generate a getter for `XXX` and similarly for `set_XXX`.

```fs
fun get_x() = _x
fun set_x(value) = set _x = value
```

BTW I noticed that in JS `get` is not actually a keyword. So yet another alternative approach in a similar spirit as JS:

```fs
fun (get) x() = _x
fun (set) x(value) = set _x = value
```

(it looks a bit weird)

Anyway, this should be done in a separate PR.

hkmc2/shared/src/main/scala/hkmc2/bbml/bbML.scala
```diff
@@ -26,7 +26,12 @@ final case class BbCtx(
   env: HashMap[Uid[Symbol], GeneralType]
 ):
   def +=(p: Symbol -> GeneralType): Unit = env += p._1.uid -> p._2
-  def get(sym: Symbol): Option[GeneralType] = env.get(sym.uid) orElse parent.dlof(_.get(sym))(None)
+  def get(sym: Symbol): Option[GeneralType] =
+    if BbCtx.builtinOps(sym.nme) then ctx.get(s"#${sym.nme}") match
```
LPTK: It occurs to me we now have a proper way of retrieving builtin symbols with `getBuiltin` in `Elaborator.Ctx`.

So I think we should rather use that. So the scheme would be: do shadow normal operators like (+) in `bbPrelude.mls`, in elaboration, but on this line, retrieve the correct symbol with `getBuiltin`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -97,6 +111,10 @@ object Elaborator:
         Term.SynthSel(base.ref(Ident(base.nme)),
           new Tree.Ident(nme).withLocOf(id))(symOpt)
       def symbol = symOpt
+    final case class GetElem(val base: Elem) extends Elem:
+      def nme: Str = base.nme
+      def ref(id: Tree.Ident): Term = base.ref(id)
```
LPTK: The whole point of adding `GetElem` was to make this function return the application term. Why did you basically add a useless forwarder to a base `Elem` that doesn't do anything? Don't you think that makes no sense?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -43,17 +43,30 @@ object Elaborator:
...
+              case S(TermDefinition(_, Fun, _, Nil, _, _, _, _)) => nme -> Ctx.GetElem(elem)
+              case _ => nme -> elem
       )
     
     def nest(outer: Opt[InnerSymbol]): Ctx = Ctx(outer, Some(this), Map.empty)
     
-    def get(name: Str): Opt[Ctx.Elem] =
-      env.get(name).orElse(parent.flatMap(_.get(name)))
+    private def getImpl(name: Str): Opt[Ctx.Elem] = env.get(name).orElse(parent.flatMap(_.getImpl(name)))
+    def isGetter(name: Str): Bool = getImpl(name) match
```
LPTK: This is a hack that should be removed (see other comment).

LPTK:
What's the rationale for adding a competing Predef for BbML? Can't we use the existing one, and add all functionality needed by BbML in some separate BbPredef module?

NeilKleistGao:
> What's the rationale for adding a competing Predef for BbML? Can't we use the existing one, and add all functionality needed by BbML in some separate BbPredef module?

So far only `checkArgs` in the predef is required. I added a new implementation that can be type checked in Bbml (even though the type-checking for importing is still not done).  Or do we plan to use the untyped predef everywhere?

LPTK:
Ah you mean that the current Predef doesn't type check. Then I guess this design is fine for now.

hkmc2/shared/src/main/scala/hkmc2/bbml/bbML.scala
NeilKleistGao: The typer can always find the correct symbol if we shadow normal operators in the prelude. The problem is that the JS builder needs to know the symbol should be builtin.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -354,12 +350,21 @@ extends Importer:
...
       Term.SynthSel(preTrm, nme)(sym)
-    case Sel(pre, nme) =>
+    case t @ Sel(pre, nme) =>
       val preTrm = term(pre)
       val sym = resolveField(nme, preTrm.symbol, nme)
-      if inAppPrefix
+      val res = if inAppPrefix
       then Term.SynthSel(preTrm, nme)(sym)
       else Term.Sel(preTrm, nme)(sym)
+      val isGetter = ctx.get(nme.name) match // TODO: create a function for this logic
```
LPTK: This `ctx.get(nme.name)` makes zero sense.

hkmc2/shared/src/main/scala/hkmc2/bbml/bbML.scala
```diff
@@ -482,7 +482,14 @@ class BBTyper(using elState: Elaborator.State, tl: TL):
...
           case N => 
             (error(msg"Not a valid class: ${cls.describe}" -> cls.toLoc :: Nil), Bot)
-      case Term.App(lhs, Term.Tup(rhs)) =>
+      case Term.App(lhs @ Term.SynthSel(_, Ident(nme)), Term.Tup(Nil)) if ctx.ctx.get(nme).map(e => e match {
+        case _: Elaborator.Ctx.GetElem => true
+        case _ => false
+      }).getOrElse(lhs.sym.map(_.defn match {
+        case S(TermDefinition(_, Fun, _, Nil, _, S(_), _, _)) => true
+        case _ => false
+      }).getOrElse(false)) => typeCheck(lhs) // * Gettters
```
LPTK: OMG this code is _cursed_. Even the comment is 😅

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -354,12 +350,21 @@ extends Importer:
...
+        case S(_: Ctx.GetElem) => true
+        case _ => sym.flatMap(_.defn) match
```
LPTK: The code should not have to access `defn` to work. The relevant info could be stored in the symbol itself I think.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
NeilKleistGao: I've fixed this problem (and also the following ones).

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -355,12 +358,7 @@ extends Importer:
       val res = if inAppPrefix
       then Term.SynthSel(preTrm, nme)(sym)
       else Term.Sel(preTrm, nme)(sym)
-      val isGetter = ctx.get(nme.name) match // TODO: create a function for this logic
-        case S(_: Ctx.GetElem) => true
-        case _ => sym.flatMap(_.defn) match
-          case S(TermDefinition(_, Fun, _, Nil, _, S(_), _, _)) => true
-          case _ => false
-      if isGetter then
+      if sym.map(_.isGetter).getOrElse(false) then
```
```suggestion
      if sym.exists(_.isGetter) then
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -170,6 +170,9 @@ extends Importer:
           N
       case N =>
         N
+    case S(_: TopLevelSymbol) => ctx.get(nme.name).flatMap(_.symbol) match
+      case S(sym: FieldSymbol) => S(sym)
```
LPTK: I think this logic is incorrect. A top-level member may very well be shadowed by a local one in `ctx`.

You may want to have a read of https://github.com/hkust-taco/mlscript/pull/242#discussion_r1861679152

LPTK (changes requested):
Nice work, but please see my comment.

Also, we should really now look into using `Scope` to generate proper TV names without a counter. Currently, the counter makes diffs like this one very noisy.

LPTK:
I just realized we must forbid getters from being defined in non-module and non-function scopes. Indeed, they cannot be used correctly if we don't know their symbol.

```fs
fun foo(f) =
  f.oops

class Foo(x) with
  fun oops = x

foo(Foo(42))
//│ = [Function: oops]

foo(Foo(42))()
//│ ═══[RUNTIME ERROR] TypeError: Cannot read properties of undefined (reading 'x')
//│     at oops (REPL25:1:176)
//│     at REPL27:1:80
//│     at ContextifyScript.runInThisContext (node:vm:136:12)
//│     at REPLServer.defaultEval (node:repl:598:22)
//│     at bound (node:domain:432:15)
//│     at REPLServer.runBound [as eval] (node:domain:443:12)
//│     at REPLServer.onLine (node:repl:927:10)
//│     at REPLServer.emit (node:events:520:28)
//│     at REPLServer.emit (node:domain:488:12)
//│     at [_onLine] [as _onLine] (node:internal/readline/interface:416:12)
```

Note that a property of module methods is that we always know their symbol when calling them.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -170,9 +170,15 @@ extends Importer:
           N
       case N =>
         N
-    case S(_: TopLevelSymbol) => ctx.get(nme.name).flatMap(_.symbol) match
-      case S(sym: FieldSymbol) => S(sym)
-      case _ => N
+    case S(_: TopLevelSymbol) =>
+      def rec(ctx: Ctx): Opt[Ctx] = // * To avoid shadowing if users write `globalThis.somefield`
```
LPTK: Users are not really expected to write code like `globalThis.somefield`.

`globalThis` is a low-level implementation trick to access the global JS scope, mostly used in generated code. Normal users will instead just use the normal module system with imports.

So what is this for? What part of the code needs this?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -642,9 +642,16 @@ extends Importer:
               val s = st.map(term(_)(using newCtx))
               val b = rhs.map(term(_)(using newCtx))
               val r = FlowSymbol(s"‹result of ${sym}›")
+              val isGetter = k === syntax.Fun && td.paramLists.isEmpty && b.nonEmpty
```
LPTK: Why `b.nonEmpty`?

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -122,6 +122,9 @@ end BlockMemberSymbol
 sealed abstract class MemberSymbol[Defn <: Definition](using State) extends Symbol:
   def nme: Str
   var defn: Opt[Defn] = N
+  lazy val isGetter: Bool = defn match
+    case S(TermDefinition(_, Fun, _, Nil, _, S(_), _, _)) => true
+    case _ => false
```
LPTK: Did you miss my previous review comment? Regarding the use of `defn`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -71,7 +72,9 @@ enum Term extends Statement:
     case Region(name, body) => "region expression"
     case RegRef(reg, value) => "reference creation"
     case Assgn(lhs, rhs) => "assignment"
+    case SetRef(ref, value) => "set"
```
```suggestion
    case SetRef(ref, value) => "mutable reference assignment"
```

hkmc2/shared/src/test/mlscript/codegen/FunctionsThis.mls
```diff
@@ -35,4 +35,3 @@ new Test
 //│ > 3
 //│ = Test {}
 
-
```
LPTK: ?

hkmc2/shared/src/test/mlscript/bbml/bbBasics.mls
```diff
@@ -257,16 +262,34 @@ class Foo[A](x: A -> A)
...
+:e
+throw 42
+//│ ╔══[ERROR] Type error in throw
+//│ ║  l.283: 	throw 42
+//│ ║         	      ^^
+//│ ╙── because: cannot constrain  Int  <:  Error
+//│ Type: ⊥
 
 
+fun not = case
```
LPTK: This is already tested in [bbCodeGen.mls](https://github.com/hkust-taco/mlscript/pull/245/files#diff-32c10f136d369b0f19058873e261ca0fcc6a825ff7a05c78e579377f7abaae9b)

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -279,15 +279,13 @@ class Lowering(using TL, Raise, Elaborator.State):
           if usesResTmp then k(Value.Ref(l))
           else k(Value.Lit(syntax.Tree.UnitLit(true))) // * it seems this currently never happens
         )
-
```
LPTK: ?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -304,7 +302,22 @@ class Lowering(using TL, Raise, Elaborator.State):
         term(finallyDo)(_ => End()),
         k(Value.Ref(l))
       )
-
```
LPTK, line 307: Let's keep these BbML-specific cases grouped up with empty lines surrounding them, and ideally a comment on top.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -92,7 +92,12 @@ class JSBuilder(using Elaborator.State, Elaborator.Ctx) extends CodeBuilder:
       else err(msg"Cannot call non-unary builtin symbol '${l.nme}'")
     case Call(Value.Ref(l: BuiltinSymbol), args) =>
       err(msg"Illeal arity for builtin symbol '${l.nme}'")
-
```
LPTK, line 95: ?

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -179,6 +184,10 @@ class JSBuilder(using Elaborator.State, Elaborator.Ctx) extends CodeBuilder:
                     doc" # ${td.sym.nme}($params) { #{  # ${
                       bodyDoc
                     } #}  # }"
+                  case td @ FunDefn(_, Nil, bod) =>
+                    doc" # ${td.sym.nme}() { #{  # ${
```
LPTK: Argh, I now just realized that in fact, for classes, getters don't need any special treatment, since JS already supports them! They should be of the form `get ${td.sym.nme}() { ... }`. Then there is nothing special to do for them in this context and we only need to take care of those getters that are NOT in object/class scopes. There is no need for the module restriction.

Note: IIRC you can check if something is or isn't in a class/object scope by checking whether the symbol has an owner.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -43,17 +46,24 @@ object Elaborator:
...
+            case N => sym: Ctx.Elem
+          sym match
+            case sym: BlockMemberSymbol =>
+              sym.trees.headOption match // TODO: Support module overloading
+                case S(td @ TermDef(Fun, _, _)) if td.rhs.isDefined && td.paramLists.isEmpty =>
+                  nme -> Ctx.GetElem(elem)
+                case _ => nme -> elem
+            case _ => sym.defn match
+              case S(TermDefinition(_, Fun, _, Nil, _, S(_), _, _)) => nme -> Ctx.GetElem(elem)
+              case _ => nme -> elem
```
LPTK: Why isn't `isGetter` used here? This logic seems pretty brittle.

LPTK (changes requested):
Argh, sorry that the getter implementation is unexpectedly blocking this PR. It would have been better to make it in a separate PR. Anyway, let's fix it now.

There are also still problems with your hack around `globalThis`. I still don't know what it achieves/what it's for.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
NeilKleistGao: So the following cases except the normal `t` term are not considered:
```js
:sjs
fun t = 42
//│ JS:
//│ function t() { return 42; } null

t
//│ = 42

globalThis.t
//│ = [Function: t]

this.t
//│ = [Function: t]
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
NeilKleistGao: Because function declarations also have empty parameter lists:
```js
fun a() = 1
// paramLists = List(Tup(List()))

fun b = 1
// paramLists = List()

fun c: Int -> Int
// paramLists = List()
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
NeilKleistGao: Maybe I misunderstood your suggestion. Did you suggest we pass another parameter to `MemberSymbol` to indicate if this symbol is a getter, instead of checking `defn` every time?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: We should probably forbid uses of `this` at the top level, like here. So I think this is fine.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: Right, this is another good reason why we should _not_ define what is or isn't a getter based on this criterion, and rather use a special syntax. Anyway, that can be done later. I guess this is fine for now.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
LPTK: Yes except not in `MemberSymbol`, where it doesn't belong, more like in `TermSymbol`.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
NeilKleistGao: I've fixed all related implementations.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -116,12 +116,16 @@ class BlockMemberSymbol(val nme: Str, val trees: Ls[Tree])(using State)
   override def toString: Str =
     s"member:$nme${State.dbgUid(uid)}"
 
+  override val isGetter: Bool = // TODO: should be based on another syntax & kind
```
LPTK: Can you clarify? I don't understand what this comment means.

hkmc2/shared/src/test/mlscript/codegen/Getters.mls
```diff
@@ -0,0 +1,256 @@
+:js
+
+:import PredefJS.mls
```
LPTK: Oh, this is outdated. I will remove this file later. There should be no need to import it.

LPTK (approved):
Great changes, thanks! Please fix the few remaining issues and I'll make the merge.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
NeilKleistGao: > should not define what is or isn't a getter based on this criterion, and rather use a special syntax.

Oh, I just left a TODO here for the future getter syntax.

PR #246, opened 2024-12-03
Support the use of `do` as a connective in UCS

chengluyu:
Main features:

- Keyword `do` can appear in place of `then`, and the default value of its UCS expression will be replaced with `UnitLit`.
- Keyword `do` can appear in a split of UCS to introduce expressions whose results are not needed between branches.

Relevant changes:

- Keyword `do` has the same left and right precedence as `then` and can be parsed as an infix operator.
- Add two `Set`s to `Desugarer` tracking the usage of `do` and `then`. If they are all used in the same UCS expression, an error will be reported.
- Add test file `ucs/syntax/Do.mls`, which shows the semantic difference between `do` and `then`.

Not so relevant changes:

- Add a unified `apply` to `Desugarer`. It should be called no matter we want to desugar `case`, `if`, or UCS shorthands.
- Add `kwLoc: Opt[Loc]` field to `IfLike` and `Case` for better error reporting.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/syntax/Keyword.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/syntax/Keyword.scala
@@ -65,8 +65,8 @@ object Keyword:
   val `while` = Keyword("while", N, curPrec)
   
   val thenPrec = nextPrec
-  val `then` = Keyword("then", thenPrec, curPrec)
-  val `do` = Keyword("do", thenPrec, curPrec)
+  val `then` = Keyword("then", thenPrec, thenPrec)
+  val `do` = Keyword("do", thenPrec, thenPrec)
   
   val `else` = Keyword("else", nextPrec, curPrec)
   val `case` = Keyword("case", N, N)
--- a/hkmc2/shared/src/main/scala/hkmc2/syntax/Keyword.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/syntax/Keyword.scala
@@ -63,7 +63,11 @@ object Keyword:
   
   val `if` = Keyword("if", N, nextPrec)
   val `while` = Keyword("while", N, curPrec)
-  val `then` = Keyword("then", nextPrec, curPrec)
+  
+  val thenPrec = nextPrec
+  val `then` = Keyword("then", thenPrec, curPrec)
+  val `do` = Keyword("do", thenPrec, curPrec)
+  
   val `else` = Keyword("else", nextPrec, curPrec)
   val `case` = Keyword("case", N, N)
   val `fun` = Keyword("fun", N, N)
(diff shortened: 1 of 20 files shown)


hkmc2/shared/src/main/scala/hkmc2/syntax/Keyword.scala
```diff
@@ -63,7 +63,11 @@ object Keyword:
   
   val `if` = Keyword("if", N, nextPrec)
   val `while` = Keyword("while", N, curPrec)
-  val `then` = Keyword("then", nextPrec, curPrec)
+  
+  val thenPrec = nextPrec
+  val `then` = Keyword("then", thenPrec, curPrec)
+  val `do` = Keyword("do", thenPrec, curPrec)
```
LPTK: It seems this is equivalent but clearer?

```suggestion
  val thenPrec = nextPrec
  val `then` = Keyword("then", thenPrec, thenPrec)
  val `do` = Keyword("do", thenPrec, thenPrec)
```

LPTK:
Nice change! There are a few places to update that currently use `while` imperatively but with `then` because we didn't have `do` at the time.

hkmc2/shared/src/test/mlscript/codegen/Do.mls
```diff
@@ -26,15 +26,32 @@ val f = case
   do console.log("non-null")
```
chengluyu, line 26: I just realized that I didn't support `do` as a way to interleave computations.

chengluyu: It's supported in 006ce44! 🎉

chengluyu:
Can you please squash this PR when you merge?

hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala
```diff
@@ -900,6 +900,12 @@ abstract class Parser(
       case (NEWLINE, _) :: (KEYWORD(kw), _) :: _
       if kw.canStartInfixOnNewLine && kw.leftPrecOrMin > prec
       && infixRules.kwAlts.contains(kw.name)
+      && kw != Keyword.`do` // This is to avoid the following case:
```
LPTK: `!=` should be avoided. 
```suggestion
      && kw isnt Keyword.`do` // This is to avoid the following case:
```

LPTK (approved):
All right, let's merge this.

Though I am not quite sure about the `do` situation. I don't like that it's currently overloaded, making things a bit difficult to parse sometimes, as in:

```fs
while
  do log(i)
  i < 3 do set i += 1
```
or
```fs
while
  do print(ls)
  ls is Cons(h, tl) do set ls = tl
```

Here the two uses of `do` serve quite different purposes, making things read a bit awkward.

PR #247, opened 2024-12-04
Support for annotations

chengluyu:
I'm trying to support annotations. As we discussed, recursive pattern normalization/compilation should be an opt-in feature and can be enabled by an annotation. For example,

```
fun foo(xs) = if xs is
  @compile EmailList(values) then ...
  ...
```

I have only completed parsing as the first step. I am still not sure how it should be handled in the `Elaborator`, so I put this PR in draft status.


LPTK:
For reference, the PR where this was implemented in the previous version of the compiler by @CAG2Mark: https://github.com/hkust-taco/mlscript/pull/210

hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala
```diff
@@ -256,14 +264,14 @@ abstract class Parser(
           yeetSpaces match
           case (tok @ BRACKETS(Indent | Curly, toks), loc) :: _ if subRule.blkAlt.isEmpty =>
             consume
-            val blk = rec(toks, S(tok.innerLoc), tok.describe).concludeWith(_.blockOf(subRule, allowNewlines)) // FIXME allowNewlines?
+            val blk = rec(toks, S(tok.innerLoc), tok.describe).concludeWith(_.blockOf(subRule, Nil, allowNewlines)) // FIXME allowNewlines?
             if blk.isEmpty then
               err((msg"Expected ${subRule.whatComesAfter} ${subRule.mkAfterStr}; found end of block instead" -> S(loc) :: Nil))
               errExpr
-            blk ::: blockContOf(rule)
+            blk ::: blockContOf(rule) // TODO: apply headAnnotations
```
LPTK: Not a fan of this. We're going to be silently dropping some information, which is always the worst to debug, if/when we eventually look into why this case is not working.

Do you have a test case that reproduces this path?

chengluyu:
### Summary

- Add case `Annotated(qualifier, target)` to both `Tree` and `Terms`.
- Only annotations for declarations are stored in a new field `annotations` added to `LetDecl`, `TermDefinition`, and `TypeLikeDef`. 
- Annotating expressions and missing annotation targets will raise a warning. In such cases, annotation qualifiers will be discarded and not elaborated.
- Annotation qualifiers must be a simple expression of selections or identifiers (or applications to them).

chengluyu:
Please squash and merge after review. Thanks.

hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala
chengluyu: Thanks. This is fixed.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -467,16 +467,16 @@ extends Importer:
...
-        msg"The annotation is not used" -> prefix.toLoc ::
-        msg"Because annotations cannot be applied to ${receiver.describe}" -> receiver.toLoc :: Nil))
-      val ann = prefix match
-      case App(_: Ident | _: SynthSel | _: Sel, _) | _: Ident | _: SynthSel | _: Sel => term(prefix)
-      case _ =>
-        raise(ErrorReport(msg"Unsupported annotation expression." -> prefix.toLoc :: Nil))
-        Term.Error
-      Term.Annotated(ann, term(receiver))
+        msg"This annotation qualifier is not applied" -> lhs.toLoc ::
+        msg"Because annotations are not supported for ${rhs.describe}" -> rhs.toLoc :: Nil))
```
```suggestion
        msg"This annotation has no effect." -> lhs.toLoc ::
        msg"Annotations are not supported on ${rhs.describe} terms." -> rhs.toLoc :: Nil))
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -532,25 +532,25 @@ extends Importer:
...
-          raise(WarningReport(
-            msg"The annotation is not used" -> (annotations.foldLeft[Opt[Loc]](N):
-              case (acc, ann) => acc match
-                case N => ann.toLoc
-                case S(loc) => S(loc ++ ann.toLoc)) ::
-            (sts.headOption match
-              case N => msg"Because nothing follows" -> blk.toLoc.map(_.right)
-              case S(head) => msg"Because annotations cannot be applied to ${head.describe}" -> head.toLoc
-            ) :: Nil))
+    def go(sts: Ls[Tree], qualifiers: Ls[Term], acc: Ls[Statement]): Ctxl[(Term.Blk, Ctx)] =
```
LPTK: Why rename to annotations to qualifiers? The former is much clearer and more specific.correct. There are many forms of qualifiers, but annotations are not usually understood as qualifiers.

hkmc2/shared/src/test/mlscript/syntax/annotations/Declarations.mls
```diff
@@ -33,13 +33,46 @@ module Foo with
...
+let
+  @inline success = 0
+  failure = 1
+//│ failure = 1
+//│ success = 0
+
+// Multiple annotations
+// ====================
+
+// :pt
```
LPTK: Remove the comments? (Below as well.)

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -309,7 +309,7 @@ class Lowering(using TL, Raise, Elaborator.State):
         term(finallyDo)(_ => End()),
         k(Value.Ref(l))
       )
-
+
```
LPTK: Weird spaces.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -337,6 +337,10 @@ class Lowering(using TL, Raise, Elaborator.State):
         subTerm(rhs): value =>
           AssignField(ref, Tree.Ident("value"), value, k(value))(N)
 
+    case Annotated(prefix, receiver) => 
+      // TODO: handle annotations
```
LPTK: It seems that the warning about useless annotations should be reported here rather than in the elaborator.

hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala
```diff
@@ -256,14 +264,14 @@ abstract class Parser(
           yeetSpaces match
           case (tok @ BRACKETS(Indent | Curly, toks), loc) :: _ if subRule.blkAlt.isEmpty =>
             consume
-            val blk = rec(toks, S(tok.innerLoc), tok.describe).concludeWith(_.blockOf(subRule, allowNewlines)) // FIXME allowNewlines?
+            val blk = rec(toks, S(tok.innerLoc), tok.describe).concludeWith(_.blockOf(subRule, Nil, allowNewlines)) // FIXME allowNewlines?
             if blk.isEmpty then
               err((msg"Expected ${subRule.whatComesAfter} ${subRule.mkAfterStr}; found end of block instead" -> S(loc) :: Nil))
               errExpr
-            blk ::: blockContOf(rule)
+            blk.map(qualifiers.annotate) ::: blockContOf(rule) // TODO: apply qualifiers
```
LPTK: What is the TODO for?

hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala
```diff
@@ -273,26 +281,27 @@ abstract class Parser(
             yeetSpaces match
             case (tok @ BRACKETS(Indent | Curly, toks), loc) :: _ /* if subRule.blkAlt.isEmpty */ =>
               consume
+              if qualifiers.nonEmpty then
+                err((msg"Blocks cannot be annotated" -> S(loc) :: Nil))
```
LPTK: This error should be diff-tested.

LPTK (changes requested):
Thanks, it looks mostly good, but:
* You should probably warn about useless annotations in Lowering rather than Elaborator – indeed, the lowering will later make use of some annotations, like `@tailrec`.
* Similarly, in Lowering, you should warn against useless annotations even on definitions (like class and value definitions).

chengluyu:
It seems that there are still some warnings about unused annotations that need to be addressed in the elaborator, because by the time they reach the lowering phase, these trees are no longer present, such as `open`. Therefore, I kept the warnings in the elaborator when processing blocks.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -532,25 +529,25 @@ extends Importer:
...
-          msg"The annotation qualifier is not applied" -> (qualifiers.foldLeft[Opt[Loc]](N):
+          msg"This annotation has no effect" -> (annotations.foldLeft[Opt[Loc]](N):
             case (acc, ann) => acc match
               case N => ann.toLoc
               case S(loc) => S(loc ++ ann.toLoc)
           ) :: (sts.headOption match
-            case N => msg"Because the target term is missing" -> blk.toLoc.map(_.right)
-            case S(head) => msg"Because annotations are not supported for ${head.describe}" -> head.toLoc
+            case N => msg"A target term is expected at the end of block" -> blk.toLoc.map(_.right)
+            case S(head) => msg"Annotations are not supported on ${head.describe}" -> head.toLoc
```
```suggestion
            case S(head) => msg"Annotations are not supported on ${head.describe} terms." -> head.toLoc
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -378,6 +383,17 @@ class Lowering(using TL, Raise, Elaborator.State):
...
+  def reportAnnotations(target: Statement, annotations: Ls[Term]): Unit = if annotations.nonEmpty then
+    raise(WarningReport(
+      (msg"This annotation has no effect." -> annotations.foldLeft[Opt[Loc]](N):
+        case (acc, term) => acc match
+          case N => term.toLoc
+          case S(loc) => S(loc ++ term.toLoc)) ::
+      msg"Annotations are not supported on this ${target match
+            case _: LetDecl => "let declaration"
+            case td: TermDefinition => td.k.desc
+            case cls: ClassLikeDef => cls.kind.desc }." -> target.toLoc :: Nil))
```
```suggestion
      Nil))
```

hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala
chengluyu: I added a test, but this case is tricky because the condition under which it appears is: the foremost token is an identifier; this identifier is also keyword; this keyword has a rule starting with it in the `prefixRules1, but there is no rule starting with this keyword in the current `rule`.

chengluyu: I tried adding `???` to understand where it's used by seeing where it crashes, but I found that all of these tests are marked as todo or fixme.

LPTK: Why are you not simply passing the annotations to the `p.blockOf` call below instead of `Nil`? The error as it is now makes little sense.

LPTK (approved):
LGTM, thanks.

I'll merge after you the last conversation is resolved.

hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala
chengluyu: That's a great idea! I overlooked the fact that it called `blockOf`.

hkmc2/shared/src/test/mlscript/syntax/annotations/Unsupported.mls
```diff
@@ -90,13 +90,3 @@ id(@Log 5)
...
-
-:todo
-class Foo with
-  @debug
-  constructor
-    Bar(qax: Str)
-//│ ╔══[PARSE ERROR] Blocks cannot be annotated
-//│ ║  l.98: 	    Bar(qax: Str)
-//│ ╙──      	^^^^
-//│ /!!!\ Uncaught error: scala.NotImplementedError: List() (of class Nil$)
```
LPTK: Why remove the test case?

chengluyu: The code that produced the error "Blocks cannot be annotated" has been removed, and this code actually couldn't work; it was rewritten from a test block of `ADTs.mls`. It's no longer useful.

LPTK: Why couldn't this code work? The code looks fine to me.

Clearly, this code was triggering a path in the compiler that was previously untested. So it's important that we keep the test around! Why would you remove it and revert to having that path untested?

chengluyu: The code path is tested. There are around 11 test blocks which reveals this code path IIRC, for example, in `ADTs.mls`. Also, the `UnimplementedError` is not caused by any code introduced in this commit and can be found in all other test blocks revealing this code path.

chengluyu: Here is a test block that reveals the same code path: https://github.com/hkust-taco/mlscript/blob/e9e1354cd539ca8c0d40cdc201b18d6068444d95/hkmc2/shared/src/test/mlscript/basics/ADTs.mls#L14-L18

LPTK: I meant that there is no test that tries to annotate a split statement like this. In fact, the test I'd like to see is:
```fs
@debug
val
  x = 1
  y = 2
```

chengluyu: Do you mean a test like this: https://github.com/hkust-taco/mlscript/blob/e907ad68197d93d63b00bb53fadac7dbed0c0522/hkmc2/shared/src/test/mlscript/syntax/annotations/Declarations.mls#L117-L122

chengluyu: Or do you want me to specifically introduce a test for `val`?

LPTK: Ok that one seems to be enough 👍

PR #248, opened 2024-12-04
Return `null` by default and remove unnecessary `?? null`s

ychenfo:
Current changes made by this PR:
1. Functions that do not have a specific return value return `null` by default. This is done by flipping what `UnitLit(true)` and `UnitLit(false)` represent respectively (i.e. previously `UnitLit(true)` represents `undefined`, now it represents `null`). 
2. Detect if the function is a mls function conservatively in `Lowering.scala` by [using the symbol of term](https://github.com/CrescentonC/mlscript/blob/b9bb849addc1b9e3fa3ed9751b9222181df38d03/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala#L71-L76) being called and the associated tree of that symbol. This information is stored [in a new field in `codegen.Call`](https://github.com/CrescentonC/mlscript/blob/b9bb849addc1b9e3fa3ed9751b9222181df38d03/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala#L124). 
3. If `noSanityCheck` flag is unset, insert `?? null` to calls of unknown functions in `JSBuilder.scala`.

----

This PR is currently marked as a draft because of the following problem: when a function defined in a non-sanity-checked block or imported `mls` files is called, that function will be considered as a mls function so that no `?? null` will be inserted for calls to that function. However, an imported `mls` may not be compiled with `?? null` inserted, therefore the function being called may still return `undefined`, and this `undefined` value will leak into the current program which is not supposed to produce any `undefined` value under sanity check flag.

Maybe one way to address this is to make the following two things default behaviors, no matter whether the sanity check flag is set or not, but I so far am not sure whether there is a better design...
   - Mls functions always return `null` when there is no specific return value
   - Calls to unknown functions are always guarded by `?? null`, even in non-sanity-checked blocks or compiled files like `Predef.mls`


LPTK:
Thanks for the changes!

> Maybe one way to address this is to make the following two things default behaviors, no matter whether the sanity check flag is set or not, but I so far am not sure whether there is a better design...

Yes, that's absolutely the correct approach. I think you are confused because you think preventing `undefined` is a sanity-checking thing. It is not. It's part of the official semantics of MLscript, and it should _always_ be done. By contrast, all a sanity check can do is crash a program when some conditions are not meant – sanity checks are not supposed to change the semantics of programs that run successfully.

Now, the sanity check that this feature _enables_ (in a sound and parametric way) is the one that makes sure things like field selections never result in `undefined`.

ychenfo:
Thanks! I have pushed the commit to make `?? null` default behavior instead of being related to sanity check. 

Upon reviewing this PR, I found that although I think [my checks for mls functions](https://github.com/hkust-taco/mlscript/pull/248/files#diff-b8abb35f8c1d9f1708d239d7c71bdda2c4be29154d84ba6df912e6b5a26d387bR71-R75) are already conservative (only those lhs that are lambdas, or builtins, class constructors or functions are considered to be mls functions for sure), some program using mutations may still break:

```
:sjs
class A(x)
set A = console.log
A(2)
//│ JS:
//│ let selRes, tmp;
//│ this.A = function A(...args) { return new A.class(...args); };
//│ this.A.class = class A {
//│   constructor(x) {
//│     this.x = x;
//│     
//│   }
//│   toString() { return "A(" + this.x + ")"; }
//│ };
//│ selRes = this.console.log;
//│ if (selRes === undefined) {
//│   throw new this.Error("Access to required field 'log' yielded 'undefined'");
//│ } else {
//│   tmp = selRes;
//│ }
//│ this.A = tmp;
//│ this.A(2)
//│ > 2
```

In the above program, `A` is resigned to be `console.log` so `A(2)` still returns `undefined`. But I guess maybe this will not be considered as a valid mls program and therefore some sort of UB?

LPTK:
> In the above program, `A` is resigned to be `console.log` so `A(2)` still returns `undefined`. But I guess maybe this will not be considered as a valid mls program and therefore some sort of UB?

OMG I didn't know you could reassign classes in the current compiler 😄 

Yeah, it's not a valid program, and it would still not be even if it was using a `mut val`. That's because `console.log` is ill-formed – due to wacky JS semantics it fails to capture `this`. It should rather be `console.log(_)`, ie `arg => console.log(arg)`, which should never return `undefined`.

LPTK (approved):
LGTM

ychenfo:
> That's because `console.log` is ill-formed – due to wacky JS semantics it fails to capture `this`. It should rather be `console.log(_)`, ie `arg => console.log(arg)`, which should never return `undefined`.

Thanks for the explanation! Yeah when a lambda is created and `console.log` is explicitly called, `?? null` will be inserted to that call so that the lambda will never return `undefined`.

PR #249, opened 2024-12-11
Remove unnecessary Alloc type in BbML

NeilKleistGao:
Both the implementation and the paper no longer require this `Alloc` type.


LPTK (approved):
LGTM

PR #250, opened 2024-12-14
Improve type variable names in Bbml

NeilKleistGao:
- Give temporary type variable meaningful names (e.g., `"ctx"`, `"scrut"`) instead of using α everywhere
- <del>Adopt `allocateName` approach from `Scope` for tv name clash</del> Reuse `Scope` for tv name clash, avoiding using global uid everywhere
- Move `Scope` to `hkmc2.utils`
- Add `show` & `showDbg` for Bbml types


LPTK:
That's better but not really what I wanted.

The reason I suggested reusing `Scope` is because we want name generation NOT to be global. Adding a definition at the top should not affect the names generated for later definitions – otherwise very small changes in early definitions can create **hundreds** of cascading name changes which pollute the diffs. We can avoid that because these names are _scoped_. This logic is already implemented in `Scope`. Why not reuse it?

NeilKleistGao:
I improved the implementation by reusing `Scope`.

LPTK (changes requested):
Sorry but I think you misunderstood how I wanted you to use `Scope`. It's only meant to be used _when pretty-printing types_. It has nothing to do inside the type checker's context. In particular, it's not meant to be used to generate name hints! The name hints are just plain names from the source code that a re supposed to be used later during type pretty-printing (at the time we will allocate a name in the corresponding scope). Look at what the legacy compiler is doing with `ShowCtx`. You should do something similar but less global (follow the scoping structure of the type).

In other words, all you need to do is to make the type pretty printing functions take an implicit `(using Scope)` and use that scope to show type variable names.

hkmc2/shared/src/main/scala/hkmc2/bbml/bbML.scala
```diff
@@ -13,6 +13,7 @@ import semantics.*, semantics.Term.*
 import Elaborator.Ctx
 import syntax.*
 import Tree.*
+import codegen.js.Scope
```
LPTK: Also don't forget to move `Scope` out of the `hkmc2.codegen.js` package and into `hkmc2.utils`.

NeilKleistGao:
I think this version should work. All pretty-printer-related methods take an implicit parameter `using Scope`. The `nest` method is called when entering forall bodies.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -79,6 +79,13 @@ class TempSymbol(val trm: Opt[Term], dbgNme: Str = "tmp")(using State) extends B
   override def toLoc: Option[Loc] = trm.flatMap(_.toLoc)
   override def toString: Str = s"$$${super.toString}"
 
+
+class RefSymbol(val origin: Symbol)(using State) extends LocalSymbol:
```
LPTK: What's the purpose of this class? Either document that purpose with a small comment or remove the class.

NeilKleistGao: Sorry, this is missing in the previous commit. I've fixed it in the following one.

LPTK (approved):
Great, it finally LGTM!

PR #251, opened 2024-12-16
nofib Benchmarks

ychenfo:
* remove `not` and some other unnecessary functions from `NofibPrelude.mls`
* import `NofibPrelude.mls` after the top command block
* use `::` for lists, and arrays for tuples
* fix missing laziness in `knights.mls`


hkmc2/shared/src/test/mlscript/nofib/NofibPrelude.mls
```diff
@@ -45,32 +49,28 @@ abstract class LzList[out T]: LzCons[T] | LzNil
...
-class Tup6[out T1, T2, T3, T4, T5, T6](fst: T1, snd: T2, thd: T3, fth: T4, ffth: T5, sxth: T6)
-class Tup7[out T1, T2, T3, T4, T5, T6, T7](fst: T1, snd: T2, thd: T3, fth: T4, ffth: T5, sxth: T6, svth: T7)
-class Tup8[out T1, T2, T3, T4, T5, T6, T7, T8](fst: T1, snd: T2, thd: T3, fth: T4, ffth: T5, sxth: T6, svth: T7, eghth: T8)
+fun eqTuup2(t1, t2) = if t1 is [a, b] and t2 is [c, d] then a == c and b == d
+type Tup3[out T1, T2, T3]
+type Tup4[out T1, T2, T3, T4]
+type Tup5[out T1, T2, T3, T4, T5]
+type Tup6[out T1, T2, T3, T4, T5, T6]
+type Tup7[out T1, T2, T3, T4, T5, T6, T7]
+type Tup8[out T1, T2, T3, T4, T5, T6, T7, T8]
```
LPTK: What are these type synonym for?

hkmc2/shared/src/test/mlscript/nofib/NofibPrelude.mls
```diff
@@ -1,12 +1,15 @@
 
-fun nofibPrint(x) = globalThis.process.stdout.write(globalThis.JSON.stringify(x) + "\n")
-fun nofibError(e) = globalThis.eval("throw new Error('" + e + "')")
-fun nofibBigInt(i) = globalThis.BigInt(i)
+declare type Array
+set Array.prototype.toString = globalThis.eval("(function() { return '(' + Array.prototype.join.call(this, ',') + ')'; })")
```
LPTK, line 3: WOW. Is this really necessary?

hkmc2/shared/src/test/mlscript/nofib/mate.mls
```diff
@@ -0,0 +1,652 @@
...
+// rookmoves :: Colour -> Square -> Board -> [Move]
+// rookmoves c sq bd =
+//   ( moveLine bd c sq (\xy -> case xy of {(x,y) -> (x-1,y)}) $ (moveLine bd c sq (\xy -> case xy of {(x,y) -> (x+1,y)}) $ (moveLine bd c sq (\xy -> case xy of {(x,y) -> (x,y-1)}) $ (moveLine bd c sq (\xy -> case xy of {(x,y) -> (x,y+1)}) (\x -> x))))) []
+fun rookmoves(c, sq, bd) =
+  moveLine(
+    bd,
+    c,
+    sq,
+    xy => if xy is [x, y] then [x - 1, y],
+    moveLine(
```
LPTK: Why not use `of` to avoid the parenthesis (and achieve something similar to the original's `$`)?

For the record, this works:

```fs
fun f(a)(b) = [a, b]
fun g(a, b) = [a, b]

(f of
    g of
      1
      2
) of 3
//│ = [ [ 1, 2 ], 3 ]
```

hkmc2/shared/src/test/mlscript/nofib/mate.mls
```diff
@@ -0,0 +1,652 @@
...
+
+// rookmoves :: Colour -> Square -> Board -> [Move]
+// rookmoves c sq bd =
+//   ( moveLine bd c sq (\xy -> case xy of {(x,y) -> (x-1,y)}) $ (moveLine bd c sq (\xy -> case xy of {(x,y) -> (x+1,y)}) $ (moveLine bd c sq (\xy -> case xy of {(x,y) -> (x,y-1)}) $ (moveLine bd c sq (\xy -> case xy of {(x,y) -> (x,y+1)}) (\x -> x))))) []
+fun rookmoves(c, sq, bd) =
+  moveLine(
+    bd,
+    c,
+    sq,
+    xy => if xy is [x, y] then [x - 1, y],
```
LPTK:
```suggestion
    case [x, y] then [x - 1, y],
```

hkmc2/shared/src/test/mlscript/nofib/mate.mls
```diff
@@ -0,0 +1,652 @@
...
+// fun polyLt(a, b) = a < b
+
+fun tab(n) = if n <= 0 then Nil else " " :: tab(n - 1)
+
+:...
+//│ ————————————————————————————————————————————————————————————————————————————————
+fun showReplies(rs, n) = if rs is
+  Nil then Nil
+  [mifs, s] :: rs then
+    append(
```
LPTK: I suggest using `of` again here.

LPTK: And similarly in other places of the file.

hkmc2/shared/src/test/mlscript/nofib/minimax.mls
```diff
@@ -0,0 +1,387 @@
...
+    Nil then Nil
+    x :: xs then
+      fun lscomp2(ls) = if ls is
+        Nil then lscomp1(xs)
+        y :: ys then [x, y] :: lscomp2(ys)
+      lscomp2(1 :: 2 :: 3 :: Nil)
+  concat(map(
+    pos => placePiece(piece, board, pos),
+    lscomp1(1 :: 2 :: 3 :: Nil)
+  ))
```
LPTK:
```suggestion
  concat of map of
    pos => placePiece(piece, board, pos)
    lscomp1(1 :: 2 :: 3 :: Nil)
```

hkmc2/shared/src/test/mlscript/nofib/treejoin.mls
```diff
@@ -0,0 +1,81 @@
...
...3484|26304 3973 21085|16318 10974 25171|12535 15786 14623|1432 6269 12461|15964 652 1353|11183 17378 14039|18294 1435 17411|4732 1436 11900|9200 9981 15368|2899 25483 9363|17591 13978 7506|26549 7672 26717|7441 12180 21633|15002 20578 7231|22 18606 23947|22953 1028 3900|5448 26168 24120|21355 3083 18819|4167 8986 6703|2925 7856 944|9780 13556 9284|25439 1450 4082|9758 1451 16878|9953 6452 1852|1453 5696 20024|14918 22638 14667|18983 9527 6327|4160 8589 7317|281 21460 16308|25634 8138 20767|14078 9086 11835|21673 11809 21001|13981 24424 16496|9451 4955 11118|20175 8618 5455|22736 2405 8989|22321 26300 17889|23914 23426 24575|20598 12774 3742|5065 23257 18201|1165 16760 21224|17779 1470 10915|8714 25079 6458|20365 1213 6760|948 3641 12012|22247 22367 450|14659 20038 2574|25116 12433 23121|18648 11232 9688|24926 14275 12934|12919 22375 10439|11064 9760 12096|7708 19396 545|18506 6575 14578|17838 14414 3211|22129 21844 17169|13408 24120 20021|6843 10166 2475|9087 6338 15679|8381 8597 1080|13596 14105 25076|26247 1490 7626|2190 12494 19427|15505 6025 13724|19909 8480 5416|7403 1219 20294|6570 1759 4474|5368 11080 15808|12281 1497 26401|20895 18287 7647|2403 16566 6934|17393 1500 14217")
```
LPTK: Are these huge string literals given in the source like this in nofib?

LPTK:
Really impressive work, thanks!

Looking at the failed test's stack trace, it looks like a dead lock. Weird.

```
Reporter completed abruptly with an exception after receiving event: TestFailed(Ordinal(0, 122),The test did not complete within the specified 60 seconds time limit.,AllTests$$anon$2,hkmc2.AllTests$$anon$2,Some(hkmc2.AllTests$$anon$2),mlscript/nofib/mate,mlscript/nofib/mate,Vector(),Vector(),Some(org.scalatest.exceptions.TestFailedDueToTimeoutException: The test did not complete within the specified 60 seconds time limit.),Some(106404),Some(IndentedText(- mlscript/nofib/mate,mlscript/nofib/mate,0)),Some(SeeStackDepthException),Some(hkmc2.AllTests$$anon$2),None,pool-5-thread-1-ScalaTest-running-AllTests$$anon$2,1734323933595).
java.lang.InterruptedException
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireInterruptibly(AbstractQueuedSynchronizer.java:1220)
	at java.util.concurrent.locks.ReentrantLock.lockInterruptibly(ReentrantLock.java:335)
	at java.util.concurrent.LinkedBlockingQueue.put(LinkedBlockingQueue.java:339)
	at org.scalatest.LinkedBlockingQueue.put(JavaClassesWrappers.scala:43)
	at org.scalatest.DispatchReporter.apply(DispatchReporter.scala:300)
	at org.scalatest.tools.SuiteSortingReporter.handleTestEvents(SuiteSortingReporter.scala:161)
	at org.scalatest.tools.SuiteSortingReporter.doApply(SuiteSortingReporter.scala:80)
	at org.scalatest.CatchReporter.apply(CatchReporter.scala:36)
	at org.scalatest.CatchReporter.apply$(CatchReporter.scala:30)
	at org.scalatest.tools.SuiteSortingReporter.apply(SuiteSortingReporter.scala:26)
	at org.scalatest.tools.Framework$SbtReporter.apply(Framework.scala:1187)
	at org.scalatest.WrapperCatchReporter.doApply(CatchReporter.scala:70)
	at org.scalatest.CatchReporter.apply(CatchReporter.scala:36)
	at org.scalatest.CatchReporter.apply$(CatchReporter.scala:30)
	at org.scalatest.WrapperCatchReporter.apply(CatchReporter.scala:63)
	at org.scalatest.Suite$.reportTestFailed(Suite.scala:1644)
	at org.scalatest.SuperEngine.runTestImpl(Engine.scala:347)
	at org.scalatest.funsuite.AnyFunSuiteLike.runTest(AnyFunSuiteLike.scala:216)
	at org.scalatest.funsuite.AnyFunSuiteLike.runTest$(AnyFunSuiteLike.scala:47)
	at hkmc2.DiffTestRunner.org$scalatest$OneInstancePerTest$$super$runTest(DiffTestRunner.scala:69)
	[...]
```

ychenfo:
Thanks for the suggestions! I will address them.

> Looking at the failed test's stack trace, it looks like a dead lock. Weird.

Yeah the tests can pass on my machine, but I found that indeed that one failing test (`mate.mls`) takes abnormally longer to run on my machine too, and I just realized that it turns out to be another laziness that I missed😢 I have just pushed a fix to this, and I will also fix the things you mentioned above. Thanks!

hkmc2/shared/src/test/mlscript/nofib/treejoin.mls
ychenfo: Not really, they are read from files. Maybe I can write some utility function in prelude to read files and use them to get these big string literals?

hkmc2/shared/src/test/mlscript/nofib/minimax.mls
ychenfo: Thanks! I have fixed this in the latest commit.

hkmc2/shared/src/test/mlscript/nofib/mate.mls
ychenfo: Thanks for the suggestion! For all the list appension operations in the latest commit I now use an infix operator `+:`, maybe that seems cleaner.

hkmc2/shared/src/test/mlscript/nofib/mate.mls
ychenfo: It seems that the `case` shorthand requires a new line after it:
```fsharp
:fixme
case [x] then x
//│ ╔══[PARSE ERROR] Expected block after `case` keyword; found square bracket section instead
//│ ║  l.5: 	case [x] then x
//│ ╙──     	     ^^^
//│ /!!!\ Uncaught error: scala.MatchError: InfixApp(App(Error(),TyTup(List(Ident(x)))),keyword 'then',Ident(x)) (of class hkmc2.syntax.Tree$InfixApp)

:sjs
case
  [x] then x
//│ JS (unsanitized):
//│ (caseScrut) => {
//│   let first0, x;
//│   if (globalThis.Array.isArray(caseScrut) && caseScrut.length === 1) {
//│     first0 = caseScrut[0];
//│     x = first0;
//│     return x;
//│   } else {
//│     throw new this.Error("match error");
//│   }
//│ }
//│ = [Function (anonymous)]
```

hkmc2/shared/src/test/mlscript/nofib/mate.mls
ychenfo: Thanks for the tip! I have pushed the commit to use `of` to make the code cleaner.

hkmc2/shared/src/test/mlscript/nofib/mate.mls
LPTK: Oh you're right, currently it needs to be `case { [x] then x }`. We should fix that at some point.

hkmc2/shared/src/test/mlscript/nofib/NofibPrelude.mls
ychenfo: I missed these during cleanup, this is fixed in the latest commit.

hkmc2/shared/src/test/mlscript/nofib/mate.mls
LPTK: Indeed much cleaner!

hkmc2/shared/src/test/mlscript/nofib/treejoin.mls
LPTK: It should be really easy to read things from a file using the NodeJS API. There's an example of writing `Accounting.mls`. So I don't think it really needs helper functions (but if it does please don't put them in `Predef`; put them in an `IO` module).

hkmc2/shared/src/test/mlscript/nofib/NofibPrelude.mls
ychenfo: Yeah I used this gross hack to override `toString` for Arrays because otherwise, some outputs of list of tuples become weird... for example in [this diff for `mandel.mls`](https://github.com/hkust-taco/mlscript/pull/251/commits/198c05da1f053ad8d6fecef96ab2ec59a416cb72#diff-06e58e019cdd717b37cc60596be892ea083b144b7d4fd21870d09eb129612e73L74), the list of tuples (`[(1,2),(2,3)]`) will be displayed as `[1,2,2,3]` without this..

hkmc2/shared/src/test/mlscript/nofib/treejoin.mls
ychenfo: Thanks! I will refer to that to read the input from files.

hkmc2/shared/src/test/mlscript/nofib/mate.mls
ychenfo: Thanks I see, I will use curly brackets for now.

hkmc2/shared/src/test/mlscript/nofib/treejoin.mls
ychenfo: Now `mate.mls`, `rsa.mls`, `sorting.mls` and `treejoin.mls` read from files. Indeed no helper function is needed😃

hkmc2/shared/src/test/mlscript/nofib/NofibPrelude.mls
LPTK: Wow that's _bad_. I can't believe people in JS and TS live with arrays/tuples that do not have a useful `toString` implementation when nested...

ychenfo: Yes I just checked and confirmed that the default JS array `toString()` method gives outputs [that look like flattened...](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toString#:~:text=Array.prototype.toString%20recursively%20converts%20each%20element%2C%20including%20other%20arrays%2C%20to%20strings.%20Because%20the%20string%20returned%20by%20Array.prototype.toString%20does%20not%20have%20delimiters%2C%20nested%20arrays%20look%20like%20they%20are%20flattened.)

```javascript
> [[2,3], [1]].toString()
'2,3,1'
```

Maybe I can also use `util.inspect`, which gives more reasonable output shown below. But I think in `NofibPrelude.mls`, if I want to override `Array.prototype.toString`, I still have to use `eval` because `this` is not available to arrow functions generated from mls..
```javascript
> util.inspect([[2,3], [1]])
'[ [ 2, 3 ], [ 1 ] ]'
```

EDIT:
In JS it seems the following works:
```js
> function toStr() { return "(" + this.join(",") + ")" }
undefined
> Array.prototype.toString = toStr
[Function: toStr]
> [1,2,3].toString()
'(1,2,3)'
```

I tried to do the similar thing in mls, but found that this cannot be a workaround because `this` in `toStr` will be become `globalThis`

```fsharp
:sjs
fun toStr() = "(" + this.join(",") + ")"
set Array.prototype.toString = toStr
//│ JS (unsanitized):
//│ function toStr() {
//│   let tmp, tmp1;
//│   tmp = globalThis.join(",") ?? null;
//│   tmp1 = "(" + tmp;
//│   return tmp1 + ")";
//│ }
//│ this.Array.prototype.toString = this.toStr;
//│ null

:re
[1,2,[3,4]].toString()
//│ ═══[RUNTIME ERROR] TypeError: globalThis.join is not a function
```

LPTK: Yeah we don't have the crazy `this` semantics of JS. We'll need to add a syntax that explicitly binds `this` when needed, in cases like this.

For now it's fine to use your hack, but this is of course not a real solution, as other parts of the program and libraries written in JS might rely on the default `toString` behavior.

We could use a wrapper class that properly implements `toString`. Then something like `[1, 2, 3]` would be compiled to `new Tuple(1, 2, 3)`:

```js
> class Tuple extends Array { constructor(...xs) { super(...xs) } }
undefined
> new Tuple(1, 2, 3)
Tuple(3) [ 1, 2, 3 ]
```

But this could come in a separate PR.

ychenfo: Thanks I see. So it seems that to enable this we need to have support for `super` and for `extends` in the js backend.

LPTK (approved):
Great, thanks for the hard work!

As a side note, I would use more varargs to make the code more readable.

For example, instead of:
```fs
  let yes = nofibStringToList("You seem quite positive.") ::
    nofibStringToList("Are you Sure?") ::
    nofibStringToList("I see.") ::
    nofibStringToList("I understand.") ::
    Nil
```
I would make it:
```fs
  let yes = List of
    nofibStringToList("You seem quite positive.")
    nofibStringToList("Are you Sure?")
    nofibStringToList("I see.")
    nofibStringToList("I understand.")
```
But that's a minor change that can come later.

PR #222, opened 2024-05-17
Defunctionalization with simple-sub Control-flow Analysis

HarrisL2:
This PR adds an implementation of defunctionalization for selection on objects that uses simple-sub for control flow analysis.

PR #252, opened 2024-12-17
User guided specialisation stage

Oli-Ar:
Draft PR for implementing a user-guided specialisation pass in the mlscript compiler, very much still WIP

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
@@ -307,6 +307,7 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
           case Elaborator.ctx.builtins.Num => doc"typeof $sd === 'number'"
           case Elaborator.ctx.builtins.Bool => doc"typeof $sd === 'boolean'"
           case Elaborator.ctx.builtins.Int => doc"globalThis.Number.isInteger($sd)"
+          case Elaborator.ctx.builtins.Function => doc"typeof $sd === 'function'"
           case _ => doc"$sd instanceof ${result(pth)}"
         case Case.Tup(len, inf) => doc"globalThis.Array.isArray($sd) && $sd.length ${if inf then ">=" else "==="} ${len}"
       val h = doc" # if (${ cond(hd._1) }) ${ braced(returningTerm(hd._2, endSemi = false)) }"
--- a/hkmc2/shared/src/main/scala/hkmc2/semantics/Specialiser.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/semantics/Specialiser.scala
@@ -242,13 +242,11 @@ class SimpleSub(val tl: TraceLogger):
     fields: Map[Str, SimpleType] = Map.empty
   ):
     override def toString: String = 
-      val superStr = if supers.isEmpty then "" else 
-        s" extends ${supers.mkString(" with ")}"
-      val fieldsStr = if fields.isEmpty then "" else 
-        fields.map { case (n, t) => s"val $n: $t" }.mkString(", ")
+      val superStr = if supers.isEmpty then "" else s" extends ${supers.mkString(" with ")}"
+      val fieldsStr = if fields.isEmpty then "" else fields.map { case (n, t) => s"val $n: $t" }.mkString(", ")
       val membersStr = members.map { case (n, t) => s"$n: $t" }.mkString(", ")
       val contentStr = (if fieldsStr.nonEmpty && membersStr.nonEmpty then s"$fieldsStr; $membersStr" 
-                       else fieldsStr + membersStr)
+                        else fieldsStr + membersStr)
       s"class ${sym.nme}$superStr { $contentStr }"
 
   enum SimpleType:
@@ -280,7 +278,7 @@ class SimpleSub(val tl: TraceLogger):
   
   class TypeContext(val mapping: Map[Symbol, SimpleType] = Map.empty):
     def get(sym: Symbol): Option[SimpleType] = mapping.get(sym)
-    def getOrFresh(sym: Symbol): SimpleType = mapping.getOrElse(sym, freshVar())
+    def getOrFresh(sym: Symbol): SimpleType = mapping.getOrElse(sym, freshVar)
     def +(pair: (Symbol, SimpleType)): TypeContext = TypeContext(mapping + pair)
     def ++(pairs: Iterable[(Symbol, SimpleType)]): TypeContext = TypeContext(mapping ++ pairs)
     override def toString: String = mapping.map { case (sym, ty) => s"$sym: $ty" }.mkString(", ")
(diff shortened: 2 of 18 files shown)


hkmc2/shared/src/test/mlscript/specialisation/Basic.mls
```diff
@@ -0,0 +1,87 @@
+
+
+:spt
```
LPTK: It would be helpful to also show the JS code, which is much easier to read.

hkmc2/shared/src/main/scala/hkmc2/semantics/Specialiser.scala
```diff
@@ -0,0 +1,753 @@
...
+
+class SimpleSub(val tl: TraceLogger):
+  import tl.*
+
+  enum SimpleType:
+    case Variable(state: VariableState)
+    case Primitive(name: Str)
+    case Function(lhs: SimpleType, rhs: SimpleType)
+    case Record(fields: Ls[(Str, SimpleType)])
+    case ClassType(sym: ClassSymbol, supers: Ls[SimpleType], members: Map[Str, SimpleType])
```
LPTK: This is more like `ClassInfo` (which is _not_ a `SimpleType`). The types of actual class instances should be a different constructor (which points back to the `ClassInfo`).

LPTK:
TODO: Write specification/architecture of algorithm first.

You need to attach the flows of normal parameters to the flows of specialized ones.

Example:

```
fun foo(spec x: ?a, p1: ?p1, p2: ?p2) = ...

fun bar(x: ?b, arg: ?arg) =
  foo(x, arg, 123)
// At this point, we know we'll need to specialize this call to `foo` for each new concrete LB coming into ?b
// Each of these specializations will need to create the flows `?arg <: ?p1` and `123 <: ?p2`.
```

Another way of thinking about this: specialized functions are polymorphic, but the polymorphic instances are indexed not by the individual call sites but by the list of concrete shapes flowing into their specialized parameters.

```
fun foo(spec x, spec y, p1, p2) = ... constraint ?p1 <: ?p2 -> ?result
foo: (x, y) => forall p1, p2, result where { p1 <: p2 -> result }. T
```

In this approach, we remember whatever flows into `(p1, p2)` by attaching it (indexing it) to what flows into `(x, y)`. Only when the `(x, y)` flow becomes concrete do we look up the specialization table: then, if there is not yet an entry, we instantiate the polymorphic type scheme; afterwards, we connect the stashes `(p1, p2)` flows with the corresponding instantiated signature.

Note: Similarly, `ClassInfo`s themselves should be indexed on their spec params.

LPTK:
Another interesting thing: the calls to specialized functions happening in non-specialized contexts may have to be turned into pattern matches.

Reusing the previous example:

```fs
fun foo(spec x, p1, p2) = ...

fun bar(x, arg) =
  foo(x, arg, 123)

bar(Int, 0)
bar(Str, true)
```

This should basically specialize into:

```fs
fun foo_Int(x, p1, p2) = ...
fun foo_Str(x, p1, p2) = ...

fun bar(x, arg) =
  if x is
    Int then foo_Int(x, arg, 123)
    Str then foo_Str(x, arg, 123)

bar(Int, 0)
bar(Str, true)
```

In fact, at least conceptually, we should probably always generate pattern matching by default and possibly optimize it later. So even a direct call like

```fs
foo(42, true, 0)
```

should be turned into

```fs
let $arg = 42
if $arg is
  Int then foo_Int($arg, true, 0)
```

which will be optimized the the expected

```fs
let $arg = 42
foo_Int($arg, true, 0)
```

So this looks pretty similar to the earlier [defunctionalization/demethodization project](https://github.com/hkust-taco/mlscript/pull/222)!

Oli-Ar:
The approach I took to specialisation works by taking the SimpleType that a parameter would be and then wrapping it in a SpecialisableType SimpleType, which stores the underlying SimpleType, along with a SpecPoint instance, which stores more details about the specialisation.

Specifically I store any known concrete types, any variables that flow into it, and any flows out of it into other SpecialisableTypes. With this, I can do one pass of the program, and constrain the types, and while I constrain I fill in the rest of the information about the specialised parameter. Then with all of the SpecialisableTypes I can figure out what each function should be specialised to. Then on the next pass I case use that table to actually specialise the functions and the applications of the functions.

The reason I did it this way over the linked list was just that I was struggling to get the linked list method working, and I had this idea so tried it quickly and it worked better than anything I'd managed with the first approach so I decided to keep at it for a bit longer.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -307,6 +307,7 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
           case Elaborator.ctx.builtins.Num => doc"typeof $sd === 'number'"
           case Elaborator.ctx.builtins.Bool => doc"typeof $sd === 'boolean'"
           case Elaborator.ctx.builtins.Int => doc"globalThis.Number.isInteger($sd)"
+          case Elaborator.ctx.builtins.Function => doc"typeof $sd === 'function'"
```
Oli-Ar, line 310: I added this as I had a test where I specialise a parameter which I pass a lambda to, so when I generate the pattern matching for the application I needed to be able to match the lambda, but I'd just like to check that this is okay to do since I'm not very familiar with the codegen side of mlscript.

hkmc2/shared/src/test/mlscript/specialiser/Basic.mls
```diff
@@ -200,86 +289,40 @@ foo("Str")
...
+//│ if (globalThis.Number.isInteger(id_arg05)) {
+//│   runtime.safeCall(id_Int3(id_arg05))
 //│ } else {
 //│   throw new this.Error("match error");
 //│ }
 //│ = 1
 
-
-:sjs
-fun foo(x, y) =
```
LPTK: Any reason to remove this test?

hkmc2DiffTests/src/test/scala/hkmc2/MLsDiffMaker.scala
```diff
@@ -245,8 +245,8 @@ abstract class MLsDiffMaker extends DiffMaker:
       output(s"Elaborated tree:")
       output(e.showAsTree(using post))
 
-    val checker = new semantics.SimpleSub(curCtx, stl)
-    val spt = checker.analyzeTermTypes(e).asInstanceOf[semantics.Term.Blk]
+    val checker = new semantics.Specialiser(curCtx, stl)
+    val spt = checker.topLevel(e).asInstanceOf[semantics.Term.Blk]
```
LPTK, line 249: Why is the cast needed here but not in `MLsCompiler`? And if we really need a `Blk`, couldn't you just make `topLevel` take and return one?

hkmc2/shared/src/test/mlscript/specialiser/Basic.mls
Oli-Ar: When debugging, I found it gave the same error as the test with `add`, or the test with `Pair`, but was harder to work with than either of them, so I removed it. I've added it back for consistency in commit c08a954.

LPTK:
This is unlikely to progress further, so I marked it to revisit later and will close it for now.

PR #253, opened 2024-12-19
Add new IR forms for handlers

AnsonYeung:
This PR adds 2 new IR forms: `HandlerBlock` and `HandlerBlockReturn`.

Semantics:

`HandleBlock` binds a list of handlers specified in `handlers` during the execution of `body`. This will
- create a object stored in `lhs`, with the type `cls` (via instantiation of a internal class inheriting `cls`)
- `handlers` will be new method of `lhs` that allows the body to raise effects by calling these methods.
- the result of the block will be stored in `res` after the block finishes.
Note: binding a list of handlers has different semantics from binding handlers one by one. In the later case, if the outer handler does not resume, the inner handler will not continue. In the former case, each handler, once started, cannot be stopped by other handler in the same list.

`HandleBlockReturn` is the way for a handler block to assign the value back to `res` after its completion. The current implementation uses a function to wrap the handle block in order to reduce code duplication so this will be translated to an actual return.

Other changes:
- Removed some additional semicolons and added `_` as valid JS identifier name character
- Fixed Makefile not working on linux gcc due to link ordering
- Added a flag to enable the handler translation
- Added class inheritance support and fixed some spacing issue in codegen


hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -470,3 +478,28 @@ trait LoweringTraceLog
...
+            N
+          case Some(bod) =>
+            val (paramLists, bodyBlock) = setupFunctionDef(td.params, bod, S(td.sym.nme))      
+            S(Handler(td.sym, resumeSym, paramLists, bodyBlock))
+      }.collect{ case Some(v) => v }
+      val resSym = TempSymbol(S(t))
+      subTerm(rhs): cls =>
+        HandleBlock(lhs, resSym, cls, handlers, term(st.Blk(stmts, res))(HandleBlockReturn(_)), k(Value.Ref(resSym)))
+    case _ => super.term(t)(k)
+  override def topLevel(t: st): Block =
```
LPTK: It looks like this override does nothing useful.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -87,18 +87,21 @@ class JSBuilder(using Elaborator.State, Elaborator.Ctx) extends CodeBuilder:
...
       if l.unary then
         val res = doc"${l.nme} ${operand(rhs)}"
         if needsParens(l.nme) then doc"(${res})" else res
       else err(msg"Cannot call non-unary builtin symbol '${l.nme}'")
     case Call(Value.Ref(l: BuiltinSymbol), args) =>
-      err(msg"Illeal arity for builtin symbol '${l.nme}'")
+      if l.anyarity then
+        val argsDoc = args.map(argument).mkDocument(", ")
+        doc"${l.nme}(${argsDoc})"
+      else err(msg"Illeal arity for builtin symbol '${l.nme}'")
```
```suggestion
      else err(msg"Illegal arity for builtin symbol '${l.nme}'")
```

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
LPTK, line 1: Nice simplifications to the JSBuilder code!

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -89,10 +93,13 @@ case class TryBlock(sub: Block, finallyDo: Block, rest: Block) extends Block wit
 case class Assign(lhs: Local, rhs: Result, rest: Block) extends Block with ProductWithTail
 // case class Assign(lhs: Path, rhs: Result, rest: Block) extends Block with ProductWithTail
 
-case class AssignField(lhs: Path, nme: Tree.Ident, rhs: Result, rest: Block)(symbol: Opt[FieldSymbol]) extends Block with ProductWithTail
+case class AssignField(lhs: Path, nme: Tree.Ident, rhs: Result, rest: Block)(val symbol: Opt[FieldSymbol]) extends Block with ProductWithTail
 
 case class Define(defn: Defn, rest: Block) extends Block with ProductWithTail
 
+case class HandleBlock(lhs: Local, res: Local, cls: Path, handlers: Ls[Handler], body: Block, rest: Block) extends Block with ProductWithTail
```
LPTK, line 100: So the fact that the result has to be stored in `res` seems to imply that no calls inside `body` are considered tail calls, right? But we might want to optimize such tail calls by not instrumenting them (when `rest` is empty, that is), so it would be more helpful if they were apparent and not hidden behind `Assign` nodes.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -94,7 +94,7 @@ class VarSymbol(val id: Ident)(using State) extends BlockLocalSymbol(id.name) wi
   // override def toString: Str = s"$name@$uid"
 
 class BuiltinSymbol
-    (val nme: Str, val binary: Bool, val unary: Bool, val nullary: Bool)(using State)
+    (val nme: Str, val binary: Bool, val unary: Bool, val nullary: Bool, val anyarity: Bool)(using State)
```
```suggestion
    (val nme: Str, val binary: Bool, val unary: Bool, val nullary: Bool, val functionLike: Bool)(using State)
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -308,6 +309,13 @@ class Lowering(using TL, Raise, Elaborator.State):
         term(finallyDo)(_ => End()),
         k(Value.Ref(l))
       )
+
+    case Handle(lhs, rhs, defs) =>
+      raise(ErrorReport(
+        msg"Effect handlers are not enabled" ->
+        t.toLoc :: Nil,
+        source = Diagnostic.Source.Compilation))
+      Assign(lhs, Value.Lit(syntax.Tree.UnitLit(true)), k(Value.Lit(syntax.Tree.UnitLit(true))))
```
```suggestion
      End("error")
```

LPTK (commented):
Nice changes! Although it's a bit uncomfortable that the class inheritance code is untested. But I can live with it for now, since it will be tested soon by implementing handlers code gen.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
AnsonYeung: This also fixes multi-arms Match which was apparently not implemented

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
AnsonYeung: Optimizating tail call inside handle block would require verifying all handlers are tail-resumptive or else the tail call may not be a tail call in the result. The handle block is responsible for resuming the non tail-resumptive handlers.

LPTK: Ah, that's right. But that's an easy check to make, and it's probably a relatively common case. So it would still make sense if we could optimize it.

Would it be difficult to express the handled code as a block with possible tail calls?

LPTK: Ok actually I guess the optimization wouldn't be sound because the resuming code might use the effect of the handler in question, so we must register the fact that this part of the code has this handler in the continuation object. Is that handled correctly in your current implementation of the instrumentation?

AnsonYeung: The tail position in the handle block are marked with HandleBlockReturn currently.

Actually I think even if all handlers are tail calls, we still need to check whether the returned value is a continuation so I think it can't be a tail call. But the common scenario should not make a handler for each recursive call anyway, and the recursive function should not contain any handle block

AnsonYeung: The current transformation inserts code to the handle block before adding continuation checks. So it should be fine because continuation check is only skipped when it is tail after adding the handle block code

LPTK: Yeah, that's a good point.

LPTK:
Don't forget to update the branch so we can merge it.

LPTK (approved):
Thanks, LGTM.

PR #255, opened 2024-12-20
Add ticks for type variable pretty-printing in BbML

NeilKleistGao:


LPTK:
I think skolems can be shown as simply `A` (like, again, in the legacy compiler) instead of the very strange-looking `'<A>`.

hkmc2/shared/src/test/mlscript/bbml/bbCheck.mls
```diff
@@ -92,11 +92,11 @@ high(x => x + 1)
 //│ ╔══[ERROR] Type error in reference with expected type Int
 //│ ║  l.91: 	high(x => x + 1)
 //│ ║        	          ^
-//│ ╙── because: cannot constrain  '<A>  <:  Int
-//│ ╔══[ERROR] Type error in application with expected type '<A>
+//│ ╙── because: cannot constrain  'A  <:  Int
```
LPTK: It's not right. It has a tick...

NeilKleistGao: Previously I was worried that using the same name for skolems and variables is confusing. But I just realized it is actually fine. 😅

LPTK (approved):
LGTM

PR #256, opened 2024-12-22
Type Classes

FlandiaYingman:
This (draft) PR introduces type classes by a minimal implementation of contextual parameters, including:

### Parser rules for parsing `use` and `using` keywords.

```fs
abstract class Foo with
  fun foo(): Int

class FooImpl extends Foo with
  fun foo(): Int = 40

use Int as i = 2

use Foo = new FooImpl()
```

```fs
module M with
  fun f(using someInt: Int, someFoo: Foo): Int = someInt + someFoo.foo()
```

### Elaborator rules for elaborating `use` and `using` keywords.

A `use` definition is elaborated into a `TermDefinition` of kind instance (`Ins`). If there is no explicitly specified identifier with the instance, a synthesized identifier is used. 

If any parameter in the parameter list is modified by the `using` keyword, the whole parameter list becomes an "implicit parameter list". 

### Further elaboration on module methods

Module method applications (or implicit applications) are further elaborated with the type information. 

In particular, if there is any missing contextual argument lists, the elaborator will *unify* the argument lists with the parameter lists, with some placeholder `Elem`. These `Elem`s are populated later in the new implicit resolution pass.

```fs
module M with
  fun foo(using someInt: Int, someFoo: Foo): Int = someInt + someFoo.foo()
  fun bar()(using someInt: Int) = someInt

M.foo
// => M.foo(using Int ‹unpopulated›, using Foo ‹unpopulated›)

M.bar
// => elaboration error: mismatch

M.bar()
// => M.foo()(using Int ‹unpopulated›)
```

### New pass: Implicit Resolution

A new pass called `ImplicitResolution` is introduced to resolve implicit arguments after elaboration. 

It traverses the elaboration tree, find:

- Instance definitions: add them into the *context* with their type signature as the key.
- Contextual argument placeholder: resolve the contextual argument with the corresponding instance definition in the context.

**TODO**: Move rules of module methods to this pass.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
@@ -186,6 +186,14 @@ class Lowering(lowerHandlers: Bool, stackLimit: Option[Int])(using TL, Raise, St
             case sem.Fld(flags, value, asc) =>
               TODO("Other argument forms")
             case spd: Spd => true -> spd.term
+            case ca: sem.CtxArg => ca.term match
+              case S(t) => 
+                false -> t
+              case N => 
+                // All contextual arguments should have been
+                // populated by implicit resolution before lowering.
+                // Fail silently.
+                false -> Term.Error
           val l = new TempSymbol(S(t))
             def rec(as: Ls[Bool -> st], asr: Ls[Arg]): Block = as match
               case Nil => k(Call(fr, asr.reverse)(isMlsFun, true))
@@ -231,6 +239,7 @@ class Lowering(lowerHandlers: Bool, stackLimit: Option[Int])(using TL, Raise, St
             term_nonTail(st.Blk(stmts, res))(HandleBlockReturn(_)),
             k(Value.Ref(resSym)))
       
+    case st.TyApp(lhs, _) => term(lhs)(k)
     case st.Blk(Nil, res) => term(res)(k)
     case st.Blk((t: sem.Term) :: stats, res) =>
       subTerm(t, inStmtPos = true)(r => term_nonTail(st.Blk(stats, res))(k))
@@ -254,6 +263,12 @@ class Lowering(lowerHandlers: Bool, stackLimit: Option[Int])(using TL, Raise, St
             val (paramLists, bodyBlock) = setupFunctionOrByNameDef(td.params, bod, S(td.sym.nme))
             Define(FunDefn(td.owner, td.sym, paramLists, bodyBlock),
               term_nonTail(st.Blk(stats, res))(k))
+          case syntax.Ins =>
+            // Implicit instances are not parameterized for now.
+            assert(td.params.isEmpty)
+            subTerm(bod)(r =>
+              Define(ValDefn(td.owner, syntax.ImmutVal, td.sym, r),
+                term_nonTail(st.Blk(stats, res))(k)))
       case cls: ClassLikeDef if cls.sym.defn.exists(_.isDeclare.isDefined) =>
         // * Declarations have no lowering
         term(st.Blk(stats, res))(k)
--- a/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
@@ -1047,23 +1047,28 @@ extends Importer:
   
   def params(t: Tree): Ctxl[(ParamList, Ctx)] = t match
     case Tup(ps) =>
-      val plf = ParamListFlags.empty
-      def go(ps: Ls[Tree], acc: Ls[Param], ctx: Ctx): (ParamList, Ctx) =
+      def go(ps: Ls[Tree], acc: Ls[Param], ctx: Ctx, flags: ParamListFlags): (ParamList, Ctx) =
         ps match
-        case Nil => (ParamList(plf, acc.reverse, N), ctx)
+        case Nil => (ParamList(flags, acc.reverse, N), ctx)
         case hd :: tl =>
           param(hd)(using ctx) match
           case S((isSpd, p)) =>
+            val isCtx = hd match
+              case Modified(Keyword.`using`, _, _) => true
+              case _ => false
             val newCtx = ctx + (p.sym.name -> p.sym)
+            val newFlags = if isCtx then flags.copy(ctx = true) else flags
+            if isCtx && acc.nonEmpty then
+              raise(ErrorReport(msg"Keyword `using` must occur before all parameters." -> hd.toLoc :: Nil))
             isSpd match
             case S(spdKnd) =>
               if tl.nonEmpty then
                 raise(ErrorReport(msg"Spread parameters must be the last in the parameter list." -> hd.toLoc :: Nil))
-              (ParamList(plf, acc.reverse, S(p)), newCtx)
-            case N => go(tl, p :: acc, newCtx)
+              (ParamList(flags, acc.reverse, S(p)), newCtx)
+            case N => go(tl, p :: acc, newCtx, newFlags)
           case N =>
             ???
-      go(ps, Nil, ctx)
+      go(ps, Nil, ctx, ParamListFlags.empty)
   
   def typeParams(t: Tree): Ctxl[(Ls[Param], Ctx)] = t match
     case TyTup(ps) =>
--- a/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
@@ -38,7 +38,8 @@ object Elaborator:
 
   val reservedNames = binaryOps.toSet ++ aliasOps.keySet + "NaN" + "Infinity"
   
-  case class Ctx(outer: Opt[InnerSymbol], parent: Opt[Ctx], env: Map[Str, Ctx.Elem]):
+  case class Ctx(outer: Opt[InnerSymbol], parent: Opt[Ctx], env: Map[Str, Ctx.Elem], 
+    mode: Mode):
     
     def +(local: Str -> Symbol): Ctx = copy(outer, env = env + local.mapSecond(Ctx.RefElem(_)))
     def ++(locals: IterableOnce[Str -> Symbol]): Ctx =
@@ -55,7 +56,7 @@ object Elaborator:
           nme -> elem
       )
     
-    def nest(outer: Opt[InnerSymbol]): Ctx = Ctx(outer, Some(this), Map.empty)
+    def nest(outer: Opt[InnerSymbol]): Ctx = Ctx(outer, Some(this), Map.empty, mode)
     
     def get(name: Str): Opt[Ctx.Elem] =
       env.get(name).orElse(parent.flatMap(_.get(name)))
@@ -126,7 +127,11 @@ object Elaborator:
           new Tree.Ident(nme).withLocOf(id))(symOpt)
       def symbol = symOpt
     given Conversion[Symbol, Elem] = RefElem(_)
-    val empty: Ctx = Ctx(N, N, Map.empty)
+    val empty: Ctx = Ctx(N, N, Map.empty, Mode.Full)
+    
+  enum Mode:
+    case Full
+    case Light
   
   type Ctxl[A] = Ctx ?=> A
   
@@ -163,7 +168,7 @@ end Elaborator
 import Elaborator.*
 
 
-class Elaborator(val tl: TraceLogger, val wd: os.Path)
+class Elaborator(val tl: TraceLogger, val wd: os.Path, val prelude: Ctx)
 (using val raise: Raise, val state: State)
 extends Importer:
   import tl.*
@@ -894,7 +899,9 @@ extends Importer:
               // * Elaborate signature
               val st = td.annotatedResultType.orElse(newSignatureTrees.get(id.name))
               val s = st.map(term(_)(using newCtx))
-              val b = rhs.map(term(_)(using newCtx))
+              val b = if ctx.mode != Mode.Light
+                then rhs.map(term(_)(using newCtx))
+                else S(Term.Missing)
               val r = FlowSymbol(s"‹result of ${sym}›")
               val tdf = TermDefinition(owner, k, sym, pss, tps, s, b, r, 
                 TermDefFlags.empty.copy(isModMember = isModMember), annotations)
@@ -1142,7 +1149,6 @@ extends Importer:
     val (res, newCtx) = block(sts, hasResult = false)
     // TODO handle name clashes
     (res, newCtx)
-  
   
   def topLevel(sts: Tree.Block)(using c: Ctx): (Term.Blk, Ctx) =
     val (res, ctx) = block(sts, hasResult = false)
--- a/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
@@ -218,10 +218,11 @@ extends Importer:
         case _ => ()
         S(Annot.Trm(trm))
   
-  def term(tree: Tree, inAppPrefix: Bool = false): Ctxl[Term] =
+  def term(tree: Tree, inAppPrefix: Bool = false, inTyAppPrefix: Bool = false): Ctxl[Term] =
   trace[Term](s"Elab term ${tree.showDbg}", r => s"~> $r"):
-    def maybeModuleMethodApp(t: Term): Ctxl[Term] =
-      if !inAppPrefix then moduleMethodApp(t)
+    def maybeApp(t: Term): Ctxl[Term] =
+      if !inAppPrefix && !inTyAppPrefix // to ensure that nested App/TyApp are only wrapped once.
+      then maybeModuleMethodApp(t)
       else t
     tree.desugared match
     case unt @ Unt() => unit.withLocOf(unt)
@@ -279,10 +280,10 @@ extends Importer:
       case trm => raise(WarningReport(msg"Terms in handler block do nothing" -> trm.toLoc :: Nil))
       
       val tds = elabed.stats.map {
-          case td @ TermDefinition(owner, Fun, sym, params, sign, body, resSym, flags, annotations) =>
+          case td @ TermDefinition(owner, Fun, sym, params, tparams, sign, body, resSym, flags, annotations) =>
             params.reverse match
               case ParamList(_, value :: Nil, _) :: newParams =>
-                val newTd = TermDefinition(owner, Fun, sym, newParams.reverse, sign, body, resSym, flags, annotations)
+                val newTd = TermDefinition(owner, Fun, sym, newParams.reverse, tparams, sign, body, resSym, flags, annotations)
                 S(HandlerTermDefinition(value.sym, newTd))
               case _ => 
                 raise(ErrorReport(msg"Handler function is missing resumption parameter" -> td.toLoc :: Nil))
@@ -324,8 +325,8 @@ extends Importer:
         case N =>
           raise(ErrorReport(msg"Name not found: $name" -> tree.toLoc :: Nil))
           Term.Error
-    case TyApp(lhs, targs) =>
-      Term.TyApp(term(lhs), targs.map {
+    case TyApp(lhs, targs) => maybeApp:
+      Term.TyApp(term(lhs, inTyAppPrefix = true), targs.map {
         case Modified(Keyword.`in`, inLoc, arg) => Term.WildcardTy(S(term(arg)), N)
         case Modified(Keyword.`out`, outLoc, arg) => Term.WildcardTy(N, S(term(arg)))
         case Tup(Modified(Keyword.`in`, inLoc, arg1) :: Modified(Keyword.`out`, outLoc, arg2) :: Nil) =>
@@ -467,11 +468,13 @@ extends Importer:
                 msg"Only module parameters may receive module arguments (values)." -> 
                 arg.toLoc :: Nil
       
-      maybeModuleMethodApp(Term.App(lt, rt)(tree, sym))
+      maybeApp:
+        Term.App(lt, rt)(tree, sym)
     case SynthSel(pre, nme) =>
       val preTrm = term(pre)
       val sym = resolveField(nme, preTrm.symbol, nme)
-      maybeModuleMethodApp(Term.SynthSel(preTrm, nme)(sym))
+      maybeApp:
+        Term.SynthSel(preTrm, nme)(sym)
     case Sel(pre, nme) =>
       val preTrm = term(pre)
       val sym = resolveField(nme, preTrm.symbol, nme)
@@ -483,7 +486,8 @@ extends Importer:
           ErrorReport(
             msg"[debinding error] Method '${nme.name}' cannot be accessed without being called." -> nme.toLoc :: Nil)
       case S(_) | N => ()
-      maybeModuleMethodApp(Term.Sel(preTrm, nme)(sym))
+      maybeApp:
+        Term.Sel(preTrm, nme)(sym)
     case MemberProj(ct, nme) =>
       val c = cls(ct, inAppPrefix = false)
       val f = c.symbol.flatMap(_.asCls) match
@@ -639,48 +643,53 @@ extends Importer:
     //   ???
   
   /** Module method applications that require further elaboration with type information. */
-  def moduleMethodApp(t: Term): Ctxl[Term] =
-  trace[Term](s"Elab module method ${t.showDbg}", r => s"~> $r"):
-    /** Returns the module method definition of the innermost Sel wrapped by some App and TyApp. */
-    def defn(t: Term, argLists: Ls[Term]): (Opt[Definition], Term, Ls[Term]) = t match
-      case Term.App(f, r) => defn(f, r :: argLists)
-      case Term.TyApp(f, _) => defn(f, argLists)
-      case Term.SynthSel(pre, nme) if ModuleChecker.evalsToModule(pre) =>
-        (t.symbol.flatMap(_.asBlkMember).flatMap(_.defn), t, argLists)
-      case Term.Sel(pre, nme) if ModuleChecker.evalsToModule(pre) =>
-        (t.symbol.flatMap(_.asBlkMember).flatMap(_.defn), t, argLists)
-      case _ => (N, t, argLists)
-    defn(t, Nil) match
-      case (S(defn: TermDefinition), inner, argLists) =>
-        log(s"Elab module method definition w/ type information ${defn}.")
-        val emptyTreeApp = new Tree.App(Tree.Empty(), Tree.Empty())
-        /**
-         * Zips a module method application term along with its parameter lists,
-         * inserting any missing contextual argument lists.
-         * 
-         * M.foo -> M.foo(<using> ...)
-         * M.foo(a, b) -> M.foo(<using> ...)(a, b)(<using> ...)
-         * 
-         * Note: This *doesn't* handle explicit contextual arguments.
-         */
-        def zip(t: Term, paramLists: Ls[ParamList]): Term = (t, paramLists) match
-            case (_, params :: pRest) if params.flags.ctx =>
-              log(s"Insert a missing contextual argument list for ${params}")
-              val args = Term.Tup(params.params.map(CtxArgImpl(_)))(Tree.Tup(Nil))
-              Term.App(zip(t, pRest), args)(emptyTreeApp, FlowSymbol("‹app-res›"))
-            case (t @ Term.App(lhs, rhs), params :: pRest) =>
-              // Match the outermost App with the next non-contextual parameter list.
-              Term.App(zip(lhs, pRest), rhs)(t.tree, t.resSym)
-            case (t, params :: pRest) =>
-              // LHS is not a app but it still expects more param lists - a partial application.
-              // Just suppose it is legal and don't fail here. 
-              // TODO: Check in the implicit resolver.
-              t
-            case (_, Nil) => t
-        val newTerm = zip(t, defn.params.reverse)
-        log(s"Zip module method application: ${newTerm}")
-        newTerm
-      case _ => 
+  def maybeModuleMethodApp(t: Term): Ctxl[Term] =
+    // * Some function definitions might not be fully elaborated yet.
+    // * We need to do some very lightweight tree parsing here.
+    case class Param(ctx: Bool)(tree: Tree)
+    case class ParamList(ps: Ls[Param], ctx: Bool)(tree: Tree)
+    def param(tree: Tree): Param = tree match
+      case Tree.Modified(Keyword.`using`, _, tree) => Param(true)(tree)
+      case _ => Param(false)(tree)
+    def paramList(tree: Tree.Tup): ParamList =
+      val ps = tree.fields.map(param)
+      ParamList(ps, ps.exists(_.ctx))(tree)
+      
+    /**
+     * Zips a module method application term along with its parameter lists,
+     * inserting any missing contextual argument lists.
+     * 
+     * M.foo -> M.foo(<using> ...)
+     * M.foo(a, b) -> M.foo(<using> ...)(a, b)(<using> ...)
+     * 
+     * Note: This *doesn't* handle explicit contextual arguments.
+     */
+    def zip(t: Term, paramLists: Ls[ParamList]): Term = (t, paramLists) match
+      case (t, ps :: pss) if ps.ctx =>
+        val appTree = new Tree.App(Tree.Empty(), Tree.Empty())
+        val tupTree = new Tree.Tup(Nil)
+        val args = Term.Tup(ps.ps.map(_ => CtxArgImpl()))(tupTree)
+        Term.App(zip(t, pss), args)(appTree, FlowSymbol("‹app-res›"))
+      case (t @ Term.App(lhs, rhs), ps :: pss) =>
+        Term.App(zip(lhs, pss), rhs)(t.tree, t.resSym)
+      case (t, params :: pRest) =>
+        // LHS is not a app but it still expects more param lists - a partial application.
+        // Just suppose it is legal and don't fail here. 
+        // TODO: Check in the implicit resolver.
+        t
+      case (_, Nil) => t
+    
+    t match
+      // M.f[T](foo)(bar)
+      case semantics.Apps(Term.TyApp(ModuleChecker.MethodTreeDef(tree), _), argss) =>
+        trace[Term](s"Elab module method application ${t.showDbg}", r => s"~> $r"):
+          zip(t, tree.paramLists.map(paramList).reverse)
+      // M.f(foo)(bar)
+      case semantics.Apps(ModuleChecker.MethodTreeDef(tree), argss) =>
+        trace[Term](s"Elab module method application ${t.showDbg}", r => s"~> $r"):
+          zip(t, tree.paramLists.map(paramList).reverse)
+      // Not a module method application.
+      case _ =>
         t
   
   def fld(tree: Tree): Ctxl[Elem] = tree match
@@ -872,7 +881,9 @@ extends Importer:
             val tdf = ctx.nest(N).givenIn:
               // * Add type parameters to context
               val (tps, newCtx1) = td.typeParams match
-                case S(t) => typeParams(t)
+                case S(t) => 
+                  val (tps, ctx) = typeParams(t)
+                  (S(tps), ctx)
                 case N => (N, ctx)
               // * Add parameters to context
               var newCtx = newCtx1
@@ -885,7 +896,7 @@ extends Importer:
               val s = st.map(term(_)(using newCtx))
               val b = rhs.map(term(_)(using newCtx))
               val r = FlowSymbol(s"‹result of ${sym}›")
-              val tdf = TermDefinition(owner, k, sym, pss, s, b, r, 
+              val tdf = TermDefinition(owner, k, sym, pss, tps, s, b, r, 
                 TermDefFlags.empty.copy(isModMember = isModMember), annotations)
               sym.defn = S(tdf)
               
@@ -1141,7 +1152,7 @@ extends Importer:
   def computeVariances(s: Statement): Unit =
     val trav = VarianceTraverser()
     def go(s: Statement): Unit = s match
-      case TermDefinition(_, k, sym, pss, sign, body, r, _, _) =>
+      case TermDefinition(_, k, sym, pss, _, sign, body, r, _, _) =>
         pss.foreach(ps => ps.params.foreach(trav.traverseType(S(false))))
         sign.foreach(trav.traverseType(S(true)))
         body match
@@ -1159,31 +1170,6 @@ extends Importer:
     while trav.changed do
       trav.changed = false
       go(s)
-  
-  object ModuleChecker:
-    
-    /** Checks if a term is a reference to a type parameter. */
-    def isTypeParam(t: Term): Bool = t.symbol
-      .filter(_.isInstanceOf[VarSymbol])
-      .flatMap(_.asInstanceOf[VarSymbol].decl)
-      .exists(_.isInstanceOf[TyParam])
-    
-    /** Checks if a term evaluates to a module value. */
-    def evalsToModule(t: Term): Bool = 
-      def isModule(t: Tree): Bool = t match
-        case TypeDef(Mod, _, _, _) => true
-        case _ => false
-      def returnsModule(t: TermDef): Bool = t.annotatedResultType match
-        case S(TypeDef(Mod, _, N, N)) => true
-        case _ => false
-      t match
-        case Term.Blk(_, res) => evalsToModule(res)
-        case Term.App(lhs, rhs) => lhs.symbol match
-          case S(sym: BlockMemberSymbol) => sym.trmTree.exists(returnsModule)
-          case _ => false
-        case t => t.symbol match
-          case S(sym: BlockMemberSymbol) => sym.modTree.exists(isModule)
-          case _ => false
   
   class VarianceTraverser(var changed: Bool = true) extends Traverser:
     override def traverseType(pol: Pol)(trm: Term): Unit = trm match
--- a/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
@@ -220,6 +220,9 @@ extends Importer:
   
   def term(tree: Tree, inAppPrefix: Bool = false): Ctxl[Term] =
   trace[Term](s"Elab term ${tree.showDbg}", r => s"~> $r"):
+    def maybeModuleMethodApp(t: Term): Ctxl[Term] =
+      if !inAppPrefix then moduleMethodApp(t)
+      else t
     tree.desugared match
     case unt @ Unt() => unit.withLocOf(unt)
     case Bra(k, e) =>
@@ -464,11 +467,11 @@ extends Importer:
                 msg"Only module parameters may receive module arguments (values)." -> 
                 arg.toLoc :: Nil
       
-      Term.App(lt, rt)(tree, sym)
+      maybeModuleMethodApp(Term.App(lt, rt)(tree, sym))
     case SynthSel(pre, nme) =>
       val preTrm = term(pre)
       val sym = resolveField(nme, preTrm.symbol, nme)
-      Term.SynthSel(preTrm, nme)(sym)
+      maybeModuleMethodApp(Term.SynthSel(preTrm, nme)(sym))
     case Sel(pre, nme) =>
       val preTrm = term(pre)
       val sym = resolveField(nme, preTrm.symbol, nme)
@@ -480,7 +483,7 @@ extends Importer:
           ErrorReport(
             msg"[debinding error] Method '${nme.name}' cannot be accessed without being called." -> nme.toLoc :: Nil)
       case S(_) | N => ()
-      Term.Sel(preTrm, nme)(sym)
+      maybeModuleMethodApp(Term.Sel(preTrm, nme)(sym))
     case MemberProj(ct, nme) =>
       val c = cls(ct, inAppPrefix = false)
       val f = c.symbol.flatMap(_.asCls) match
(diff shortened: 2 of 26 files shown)


FlandiaYingman:
Wow it works like magic 🫨🤓

```fs
// Monoid Example

abstract class Monoid[T] extends Semigroup[T] with
  fun combine(a: T, b: T): T
  fun empty: T

object IntAddMonoid extends Monoid[Int] with
  fun combine(a: Int, b: Int): Int = a + b
  fun empty: Int = 0

object IntMulMonoid extends Monoid[Int] with
  fun combine(a: Int, b: Int): Int = a * b
  fun empty: Int = 1

module M with
  fun foldInt(x1: Int, x2: Int, x3: Int)(using m: Monoid[Int]): Int =
    m.combine(x1, m.combine(x2, m.combine(x3, m.empty)))
  fun fold[T](x1: T, x2: T, x3: T)(using m: Monoid[T]): T =
    m.combine(x1, m.combine(x2, m.combine(x3, m.empty)))
    
use Monoid[Int] = IntAddMonoid
M.foldInt(2, 3, 4)
//│ = 9

use Monoid[Int] = IntMulMonoid
M.foldInt(2, 3, 4)
//│ = 24

use Monoid[Int] = IntAddMonoid
M.fold(1, 2, 3)
//│ ═══[ERROR] Missing instance for context argument Param(‹›,m,Some(TyApp(SynthSel(Ref(globalThis:block#18),Ident(Monoid)),List(Ref(T)))))
```

FlandiaYingman:
**TODO**: Generate JS

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -178,6 +178,9 @@ extends Importer:
   
   def term(tree: Tree, inAppPrefix: Bool = false): Ctxl[Term] =
   trace[Term](s"Elab term ${tree.showDbg}", r => s"~> $r"):
+    def maybeModuleMethod(t: Term): Ctxl[Term] =
+      if !inAppPrefix && ModuleChecker.isModuleMethodApp(t) then moduleMethod(t)
```
LPTK: Why `!inAppPrefix`?

Admittedly we should document what the purpose of `inAppPrefix` does: it's used to avoid generating `undefined`-checks when a field being accessed is actually called like a method, because in this case `undefined` will crash as needed already.

I don't think this has anything to do with module methods.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -469,6 +473,39 @@ extends Importer:
     // case _ =>
     //   ???
   
+  /** Module method applications that requrie further elaboration. */
+  def moduleMethod(t: Term): Ctxl[Term] = 
+  trace[Term](s"Elab (further) module method ${t.showDbg}", r => s"~> $r"):
+    def defn(t: Term, argLists: Ls[Term] = Nil): (Opt[Definition], Term, Ls[Term]) = t match
```
LPTK: Default argument here is a code smell.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -820,23 +857,26 @@ extends Importer:
...
-        case Nil => (ParamList(plf, acc.reverse, N), ctx)
+        case Nil => (ParamList(flags, acc.reverse, N), ctx)
         case hd :: tl =>
           param(hd)(using ctx) match
           case S((isSpd, p)) =>
+            val isCtx = hd match
+              case Modified(Keyword.`using`, _, _) => true
+              case _ => false
             val newCtx = ctx + (p.sym.name -> p.sym)
+            val newFlags = if isCtx then flags.copy(ctx = true) else flags
```
LPTK, line 1126: Does this mean that in `(x: Int, using y: Str)`, both `x` and `y` are contextual? What's the point of allowing this confusing syntax?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -924,6 +964,13 @@ extends Importer:
           case S(sym: BlockMemberSymbol) => sym.modTree.exists(isModule)
           case _ => false
   
+    def isModuleMethodApp(t: Term): Bool = t match
```
LPTK: Better follow the "parse, don't validate" methodology (https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/).

Instead of returning a Boolean, just return an option of the relevant information so that info doesn't have to be retrieved separately in a way that might not be perfectly in sync, creating an unnecessary source of bugs.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -346,6 +347,15 @@ object PlainFld:
...
 
+/** 
+ * Contextual Arguments: to be filled in by the implicit resolver.
+ */
+abstract class CtxArg extends Elem:
+  def term: Opt[Term]
+  def showDbg: Str = s"using ${term.fold("‹unpopulated›")(_.showDbg)}"
+
+object CtxArg:
+  def unapply(arg: CtxArg): Opt[Term] = arg.term
```
LPTK: What is this extractor for? It relies on the object's state, which could be confusing/a source of bugs.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -469,6 +473,39 @@ extends Importer:
...
+  trace[Term](s"Elab (further) module method ${t.showDbg}", r => s"~> $r"):
+    def defn(t: Term, argLists: Ls[Term] = Nil): (Opt[Definition], Term, Ls[Term]) = t match
+      case Term.App(f, r) => defn(f, r :: argLists)
+      case Term.TyApp(f, _) => defn(f, argLists)
+      case _ => (t.symbol.flatMap(_.asBlkMember).flatMap(_.defn), t, argLists)
+    defn(t) match
+      case (S(defn: TermDefinition), inner, argLists) =>
+        log(s"Elab module method definition w/ type information ${defn}.")
+        val emptyTreeApp = new Tree.App(Tree.Empty(), Tree.Empty())
+        def unify(t: Term, paramLists: Ls[ParamList]): Term =
```
LPTK: I'm not sure "unify" is the right term. In PL, that term already has a heavy proof-theoretic connotation. Why nit something like `zip`?

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -0,0 +1,108 @@
...
+import hkmc2.utils.TraceLogger
+import hkmc2.Raise
+import hkmc2.semantics.Elaborator.State
+import hkmc2.syntax.Ins
+import hkmc2.ErrorReport
+import hkmc2.Message.MessageContext
+import hkmc2.syntax.Tree.TermDef
+import hkmc2.syntax.Fun
+
+case class CtxArgImpl(private val param: Param) extends CtxArg:
```
LPTK: This should probably not be a case class because it's basically a state wrapper, not a plain data holder.

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -0,0 +1,108 @@
...
+    
+    def get(sign: Term): Opt[ICtx.Instance] = env.get(sign)
+  
+  object ICtx:
+    final case class Instance(sym: Symbol)
+      
+    val empty: ICtx = ICtx(N, Map.empty)
+
+class ImplicitResolver(val tl: TraceLogger)
+(using val raise: Raise, val state: State):
```
LPTK: Does any of these really need to be a `val`

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -0,0 +1,108 @@
...
+          case N =>
+            raise(ErrorReport(msg"Missing instance for context argument ${param.toString}" -> N :: Nil))
+          case S(ICtx.Instance(sym)) =>
+            log(s"Resolved context argument: $param -> $sym")
+            arg.populate(sym.ref())
+      case _ => 
+        elem.subTerms.foreach(resolve(_))
+  
+  def resolveBlk(blk: Term.Blk)(using ictx: ICtx): ICtx = 
+  trace[ICtx](s"Resolving block: $blk", * => s"~> $blk"):
```
LPTK: Wow what's that cursed lambda? And it should be
```suggestion
  trace[ICtx](s"Resolving block: $blk", * => s"~> ${*}"):
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -0,0 +1,108 @@
...
+            log(s"Resolving instance definition $t")
+            sign match
+              case N =>
+                raise(ErrorReport(msg"Missing signature for instance ${t.toString}" -> N :: Nil))
+                ictx
+              case Some(sign) =>
+                ictx + (sign, sym)
+          case t @ TermDefinition(_, Fun, _, pss, _, _, _, _) =>
+            given newICtx: ICtx = pss
+            .filter(_.flags.ctx)
```
LPTK: An indentation is warranted here.

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -0,0 +1,108 @@
...
+    def go(rest: Ls[Statement])(using ictx: ICtx): ICtx = rest match
+      case Nil => ictx
+      case stmt :: rest => 
+        log(s"Resolving statement: $stmt")
+        given newICtx: ICtx = stmt match
+          case t @ TermDefinition(_, Ins, sym, _, sign, _, _, _) => 
+            log(s"Resolving instance definition $t")
+            sign match
+              case N =>
+                raise(ErrorReport(msg"Missing signature for instance ${t.toString}" -> N :: Nil))
```
LPTK: Ideally all the errors raised in your PR need to have a test reproducing them. Usually in a file with a name like `BadTypeClasses.mls`.

LPTK (commented):
Cool, nice draft changes!

> **TODO**: Instance identifiers should not be able to be referred directly in the code.
> 
> **Edit**: Wait. Do we really want to do that? I thought in Scala implicit instances are in another name scope, but seems that's not the case 🤓

Indeed, there need be no such restriction.

> * Instance definitions: add them into the context with their type signature as the key.

The key should be the type constructor of the type class being instantiated. It can be concrete or abstract (a type parameter or type member). As in `fun foo[A](someA) = { use A = someA; ... }`

BTW eventually we'll like to define things like:

```fs
module Foo[A](using Bar[A]) with
  ...
    [baz, Foo.baz] // equivalent to `[this.baz, Foo(using Bar[A].baz]`
    use Bar[Int] // possible to change the context here
    Foo.baz // equivalent to `Foo(using Bar[Int]).baz`
  ...
```

But for this we first need to add support for parameterized modules.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: As its name suggests and by looking at the implementation of `term`, I think this flag is `true` only if the elaborator is evaluating the LHS of an App. Without this flag check, if the elaborator sees some App like `f()()()`, it'll "further elaborate" the term 3 times, namely `f()`, `f()()`, `f()()()`, while it is expected to be further elaborated only once with `f()()()`!

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: To me `zip` sounds like it's doing the same as the `zip` in Scala stdlib. How about `normalize`?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: Yes this is what the code currently does. I'm going to fix it (and some other strange behaviors 🫨) later when I convert the draft PR into a PR.

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
FlandiaYingman: Because terms are mutable in this resolution stage, the functions work by side effects to the `blk`. So here I print the `blk` at beginning and print it again after the whole function is done. The param `*` here is an `ICtx`, which I thought it didn't need to be printed? I forgot why did I choose `*` as the name, but now I no more agree with myself 😶‍🌫️

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
FlandiaYingman: Theoretically the error is never raised because the elaboration promised that an instance must have a type signature. I guess later we can have some term other than `TermDefinition` to describe an instance definition? I found it didn't fit our use case here.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: Actually I chose *unify* because it is similar to what we did in the prolog terms unification, no? 🙈

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: Ah, sure, if we want to elaborate calls with multiple argument lists at once. Does this work well in your current implementation approach? What if only some of the argument lists are provided (ie it's partially applied)?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: I'm not sure. Would be helpful if you could document, in the code, what this does. It's not obvious.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: I couldn't think of a way to have module methods partially applied. If we do that, the result may not be recognized as a module method and then cannot be further elaborated with type information. I remember we discussed this, and we ended up with to reject partial application on module methods. (not implemented yet)

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
LPTK: Ok, sure. And normally we'd just use `_` here. I didn't even know `*` could be bound as a parameter name like this.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: I'll add some documentation. In brief, this piece of code takes a list of lists of parameters and a list of lists of arguments; comparing them one by one, and trying to insert some contextual argument placeholders if some are missing.

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
LPTK: Either it's "impossible" and you should make this report an internal error (e.g. by calling `lastWords`, although we do need a better way of reporting internal errors, without necessarily terminating compilation) or it's possible but an error will have already been reported about this, in which case the correct behavior is to silently return some error term (with a code comment explaining why that's fine).

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: If the omitted parameter lists don't have any special feature in them (no `using`, etc.) then there's no reason to forbid it, as it's functionally equivalent to a module method returning a lambda. There would be no reason to reject that.

> I remember we discussed this, and we ended up with to reject partial application on module methods.

IIRC I was just suggesting to have this arbitrary implementation restriction to ease progress on the first PR.

FlandiaYingman: Alright. Let's formally restrict partial application if the resulting function does require further elaboration with type information. Otherwise, let it behaves just like normal methods.

LPTK: > if the resulting function does require further elaboration with type information

Note that we should also check that the result type also does not need module-method treatment. E.g., it shouldn't return a module!

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: Yeah that's zipping alright.

FlandiaYingman:
> The key should be the type constructor of the type class being instantiated. It can be concrete or abstract (a type parameter or type member). As in `fun foo[A](someA) = { use A = someA; ... }`

If only the type constructors are specified as keys, wouldn't the following application fail?

```fs
use Monoid[Int] = IntAddMonoid
use Monoid[Str] = StrConcatMonoid

M.foldInt(2, 3, 4)
// resolves to M.foldInt(2, 3, 4)(using StrConcatMonoid)
```

LPTK:
It does fail, indeed. I guess we could make the rules a wee bit more fine-grained: we could look for the innermost instance in scope that could match the type query, where only those types explicitly specified (like the `Int` of `Ord[Int]`) are taken into account – so, as expected, `Ord[A]` for some unspecified `A` will unconditionally use `StrConcatMonoid`.

In any case, the map should _still_ use keys that are type constructor symbols and not signatures. You can't just use syntactic types as keys as several type expressions could be equivalent and represent the same semantic type.

FlandiaYingman:
Most comments addressed, except:

### Symbols as Context Keys

I tried, but it seems that it is not possible to use pure `Symbol` as the keys, because by doing so I couldn't get a previously defined `Ord[Int]` after defining a `Ord[Str]` (or maybe I didn't get your point 🙋‍♂️). Currently, a enum `Type` is used as the keys:

```scala
enum Type:
  case Any
  case Sym(sym: TypeSymbol)
  case App(t: Sym, typeArgs: Ls[Type])
```

`Any` is a placeholder for type parameters that are not explicitly provided. It could unconditionally match the innermost instance.

`Map` is not working here because many "query" keys can map to the same value, so instead a `List` is used, and it performs linear search to find an instance that matches the query.

I'm also aware that a `TypeSymbol` might be a `TypeAliasSymbol`. Should we resolve a type alias' definition (possibly perform a "lightweight" elaboration if absent) and substitute?

### Partial Application Restrictions

It is currently not enforced because I find it easier to implement it when we move all module methods related checks to the new pass. I also find this PR is becoming large so let's have the checks in another PR

FlandiaYingman:
All tests should pass now, except for this one

https://github.com/hkust-taco/mlscript/blob/e37d8f8d4018d0c5fbc7d49d47ac0255dee604d7/hkmc2/shared/src/test/mlscript/bbml/bbBorrowing.mls#L12

The ident `use` is now a keyword, so the parsing fails. I have no idea on bbml. How could I fix it 🫨

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -36,7 +36,8 @@ object Elaborator:
 
   val reservedNames = binaryOps.toSet ++ aliasOps.keySet + "NaN" + "Infinity"
   
-  case class Ctx(outer: Opt[InnerSymbol], parent: Opt[Ctx], env: Map[Str, Ctx.Elem]):
+  case class Ctx(outer: Opt[InnerSymbol], parent: Opt[Ctx], env: Map[Str, Ctx.Elem], 
+  mode: Mode = Mode.Full):
```
```suggestion
      mode: Mode = Mode.Full):
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -474,37 +480,63 @@ extends Importer:
     //   ???
   
   /** Module method applications that requrie further elaboration. */
```
```suggestion
  /** Module method applications that require further elaboration. */
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -851,20 +891,28 @@ extends Importer:
...
           param(hd)(using ctx) match
           case S((isSpd, p)) =>
             val isCtx = hd match
-              case Modified(Keyword.`using`, _, _) => true
+              case Modified(Keyword.`using`, _, _) => 
+                if first then
+                  true
+                else
+                  raise(ErrorReport(msg"Unexpected `using` keyword." -> hd.toLoc :: Nil))
+                  false
```
```suggestion
                if !first then
                  raise(ErrorReport(msg"Unexpected `using` keyword." -> hd.toLoc :: Nil))
                first
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
LPTK, line 1: Plz remove all the junk imports of `hkmc2` stuff and just declare the package as
```
package hkmc2
package semantics
```
instead.

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -27,19 +30,43 @@ object ImplicitResolver:
...
+        case _ if a == Any || b == Any => true
+        case (Sym(a), Sym(b)) => a == b
+        case (App(aSym, aArgs), App(bSym, bArgs)) =>
+          compare(aSym, bSym) && 
+          aArgs.length == bArgs.length && 
+          aArgs.zip(bArgs).forall((a, b) => compare(a, b))
+        case _ => false
+    
+    enum Type:
+      case Any
```
LPTK:
```suggestion
      case Unspecified
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -51,53 +78,71 @@ class ImplicitResolver(val tl: TraceLogger)
...
+                        -> p.toLoc :: Nil
+                      ))
+                      ictx
+                    case S(sign) => resolveSign(sign) match
+                      case N => ictx
+                      case S(tpe) => ictx + (tpe, p.sym)
             t.subTerms.foreach(resolve(_))
             ictx
-          case _ => 
+          // Default Case. Simply resovle all subterms.
```
LPTK, line 223:
```suggestion
          // Default Case. Simply resolve all subterms.
```

hkmc2/shared/src/test/mlscript/basics/TypeClasses.mls
```diff
@@ -1,44 +1,126 @@
...
-use Foo = new FooImpl()
+use Foo[Int] = new IntFoo()
+
+use Foo[Str] = new StrFoo()
+
+// should resolve to foo(2)
+M.foo
+//│ = 42
+
+// should resolve to strFoo(2, new IntFoo())
```
LPTK:
```suggestion
// should resolve to intFoo(2, new IntFoo())
```

LPTK (commented):
Cool, looks very promising!

Though the light elaboration logic seems inappropriate. It appears that you're re-elaborating term definitions every time they are accessed, and in the wrong context! What you should do instead is to elaborate (in light mode) things that are imported from other files. This should be done in `Importer.scala`. Notice the old comment:

https://github.com/hkust-taco/mlscript/blob/8aed18709266cd0433f146f4e476ebd54b093998/hkmc2/shared/src/main/scala/hkmc2/semantics/Importer.scala#L63-L71

(Forget about the laziness for now!)

Then, if we still find that a definition is missing, it means either there was an error while elaborating the thing (and trying to re-elaborate it won't help) or there is a bug in the compiler.

> I tried, but it seems that it is not possible to use pure `Symbol` as the keys, because by doing so I couldn't get a previously defined `Ord[Int]` after defining a `Ord[Str]` (or maybe I didn't get your point 🙋‍♂️).

Instead of a list of all instances and plain linear searches, we should at least index these instances on the type symbol. Thus "using symbols as keys".

> I'm also aware that a `TypeSymbol` might be a `TypeAliasSymbol`. Should we resolve a type alias' definition and substitute?

Yes, if the type alias is not abstract, this should be done.

> (possibly perform a "lightweight" elaboration if absent)

No (for the reason mentioned above).

> [Partial Application Restrictions] currently not enforced because I find it easier to implement it when we move all module methods related checks to the new pass. I also find this PR is becoming large so let's have the checks in another PR

Sure, that's fine. Let's get small PRs merged regularly rather than big ones including many loosely related changes.

---

Don't forget to update the branch and mark the PR ready when it's ready for a thorough review.

LPTK:
PS:

> The ident `use` is now a keyword, so the parsing fails. I have no idea on bbml. How could I fix it 🫨

Just rename the `use` function to something else! E.g., `use_`.

FlandiaYingman:
This PR is ready for review. The history is also clean. Main changes since last review:

### Light Elaboration

I fixed the light elaboration according to the suggestion. There is a problem that there are two kinds of imports - one in the `Importer.scala` and the other in the `MLsDiffMaker.scala`. All sources will implicitly import `Predef.mls`. But this would not succeed without importing `Prelude.mls` because we will now *elaborate* all definitions and that needs the symbols from `Prelude.mls`.

I added a field `prelude: Ctx` to `Elaborator` so that instead of empty `Ctx`, the `Importer` will elaborate the imported files with `prelude` to bypass the problem. I think it might be not perfect, but I couldn't think of a cleaner way to do it.

FlandiaYingman:
Oh wait I forgot to resolve the conflicts

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -147,7 +152,7 @@ end Elaborator
 import Elaborator.*
 
 
-class Elaborator(val tl: TraceLogger, val wd: os.Path)
+class Elaborator(val tl: TraceLogger, val wd: os.Path, val prelude: Ctx = Ctx.empty)
```
LPTK: Using a default argument here is an anti-pattern and will just cause needless bugs.

LPTK:
Thanks, I will try to review it soon. But I already notice some problems.

First, I just realized that I wanted to comment on this last time but somehow forgot: It seems you didn't understand the reason why we need to do resolution in a pass that's separate from elaboration. Indeed, what you implemented is not a separate pass: you do resolution on the fly, as elaboration progresses. So examples like this don't work:

```fs
:r
fun main() =
  use Bar = new Bar
  Test.test()
class Bar
module Test with
  fun test()(using bar: Bar): Bar = bar
//│ Resolved: { ‹› fun member:main() = { ‹› use member:instance$Ident_Bar_: globalThis:block#11#666(.)Bar‹member:Bar› = new SynthSel(Ref(globalThis:block#11),Ident(Bar))(); globalThis:block#11#666(.)Test‹member:Test›.test‹member:test›() }; Cls Bar { }; Mod Test { ‹module› fun member:test()ctx (Param(‹›,bar,Some(SynthSel(Ref(globalThis:block#11),Ident(Bar)))), ): globalThis:block#11#666(.)Bar‹member:Bar› = bar#666; }; }

main()
//│ = [Function (anonymous)]
```

Notice that the `Test.test()` call is not resolved (because we haven't even elaborated `Bar` or `Test` yet, at this point) and thus `main()` returns an incorrect result.

This test also exemplifies a critical flaw in your way of writing code: instead of reporting errors when things fail to elaborate, you silently generate wrong code! This sort of behavior is really to be avoided at all costs. Again, the default should be rejection, and you should only accept the code when it's 100% sure to be good. Even with a compiler crash is by far preferable to a miscompilation.

Notice that the following works and returns a different result. We don't what that changing the order of module and method definitions cause this sort of semantic discrepancies!

```fs
class Bar
module Test with
  fun test()(using bar: Bar): Bar = bar
fun main() =
  use Bar = new Bar
  Test.test()

main()
//│ = Bar {}
```

FlandiaYingman:
Thanks for the comments! You are right that this doesn't work, but I think the reason is not really related to the resolution pass itself. Prior to resolution, the elaborator identifies all module method applications and "marks" replaces any missing contextual parameters as some placeholders. What I was thinking is that it can be performed during elaboration to speed up the process, but it turns out that it indeed requires all symbols to be resolved, so I'll move it to the resolution pass.

LPTK:
Ah you're right, I read your diff too fast. You indeed have a properly separated phase, which is not the same as the `moduleMethodApp` method.

FlandiaYingman:
I tried to move the process of identifying module method applications later to the resolution stage, but it turned out to be a little bit difficult to do that directly. This is because during elaboration, we don't know whether some term represents a module method application or not. E.g., a `Sel` like `M.f` can become a module method application if `f` takes some implicit parameters!

If we want to completely move the process to the resolution stage, while keeping our performance goal (make arguments mutable instead of re-building the tree again), we will have to either change all `Sel`, `App` or `TyApp` to mutable, or create a mutable term wrapper that wraps over all `Sel`, `App` and `TyApp` for further modification by resolution. I think either is problematic because the former is error-prone and the latter breaks the semantic structure of our `Term` tree! I tried to implement in (60ad410716a66acb55260b821011cb241d026307), but *many* tests fail because of the wrapper, as I have to add one more case to all pattern matchings related to `Sel`, `App` and `TyApp`!

Now I'd suggest (and implemented in the latest push) to keep the current " identifying all module method applications" process during elaboration, but in addition we would have to "elaborate“ (in a very lightweight way) the module methods' parameter lists (because they might not be available at the time), so that we may construct a correct shape of nested `Term.App` to be populated later in the resolution pass.

LPTK:
> we will have to either change all `Sel`, `App` or `TyApp` to mutable, or [...]

> I think either is problematic because the former is error-prone

I don't follow. Why is this error-prone? I don't see a problem with it and this is more or less how I envisioned the feature. BTW you will also have to do this for `Ref` (I noticed that your PR currently mishandles references that should elaborate into applications to implicit arguments). All these ctors can be made to extend an additional `ArgumentTaking` trait that stores the actual state.

So, to be more specific: there is only one place where the trait is stored (the trait) and there can be only one place where this data is used (in lowering) to introduce the missing applications into the lowered tree. We just need to remember to do this change in lowering.

> Now I'd suggest (and implemented in the latest push) to keep the current " identifying all module method applications" process during elaboration, but in addition we would have to "elaborate“ (in a very lightweight way) the module methods' parameter lists

This seems unnecessarily inefficient.

FlandiaYingman:
> I don't follow. Why is this error-prone? 

IMO a better design may only allow to modify things that really need to be modified. Notice that almost all `Sel`s, `TyApp`s and `App`s are not eventually evaluated to implicit argument applications, only a few of them are. In contrast, the current implementation only allows modifying contextual arguments themselves, which, sounds more natural to me.

> All these ctors can be made to extend an additional `ArgumentTaking` trait that stores the actual state. 
> So, to be more specific: there is only one place where the trait is stored (the trait) and there can be only one place where this data is used (in lowering) to introduce the missing applications into the lowered tree. We just need to remember to do this change in lowering.

I am not sure if I exactly understand this, but this sounds way better than what I was originally thinking 🥲. 

Do you mean to introduce a trait that stores only the potential *contextual* arguments? So that other parts of the compiler may follow the original flow if they are not interested in the contextual arguments (or they may handle the arguments implicitly via `subterms`). E.g.,

```diff
  M.foo[T](using ...)()(using ...)  // has the following tree
  //│ App:
  //│   lhs = TyApp:
  //│     lhs = Sel:
  //│       prefix = SynthSel{member:M}:
  //│         prefix = Ref of globalThis:block#7
  //│         nme = Ident of "M"
  //│       nme = Ident of "foo"
  //│     targs = Nil
+ //|     ctxArgs = Tup of A :: Nil
  //│   rhs = Tup of Nil
+ //|   ctxArgs = Tup of B :: Nil
  // where ctxArgs are stored in the trait
```

> BTW you will also have to do this for `Ref`

Isn't that only module method applications are resolved? And since module methods must somehow be `Sel`ed from some modules instances, I only did the resolution for `Sel`.

> This seems unnecessarily inefficient.

Maybe we can do some caching? 🙈

LPTK:
Just a heads-up @FlandiaYingman: I've merged some pretty significant changes on the main branch, which affects code generation for local definitions (including the top-level ones of test worksheets). You may have to fix a few merge conflicts and possibly some bugs revealed by the change.

LPTK:
Oh, and I also think local functions should be allowed the same affordances as module methods (regarding type classes etc.) since we also always know their symbol when calling them.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: Please add a test reproducing these cases with appropriate TODOs.

LPTK (commented):
In private conversation you told me you want to merge these changes as is. Which is a wise choice as I intend to make further changes to the elaboration system that would cause conflicts if this is not merged. So please merge with the main branch, address the current comments, and make sure this is ready for my final review.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -217,8 +222,13 @@ extends Importer:
         case _ => ()
         S(Annot.Trm(trm))
   
-  def term(tree: Tree, inAppPrefix: Bool = false): Ctxl[Term] =
+  def term(tree: Tree, inAppPrefix: Bool = false, inTyAppPrefix: Bool = false): Ctxl[Term] =
   trace[Term](s"Elab term ${tree.showDbg}", r => s"~> $r"):
+    def maybeApp(t: Term): Ctxl[Term] =
+      // !inAppPrefix && !inTyAppPrefix is to ensure that nested App/TyApp are only wrapped once.
+      if !inAppPrefix && !inTyAppPrefix
```
```suggestion
      if !inAppPrefix && !inTyAppPrefix // to ensure that nested App/TyApp are only wrapped once.
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -0,0 +1,336 @@
...
+          case (ps, Term.Tup(args)) =>
+            // * TODO: Check the arity.
+            // if ps.paramCountUB
+            // then if ps.paramCountLB != args.length then
+            //   raise(ErrorReport(msg"Expected ${ps.paramCountLB.toString()} arguments, " +
+            //     msg"got ${args.length.toString()}" -> base.toLoc :: Nil))
+            // else if ps.paramCountLB > args.length then
+            //   raise(ErrorReport(msg"Expected at least ${ps.paramCountLB.toString()} arguments, " +
+            //     msg"got ${args.length.toString()}" -> base.toLoc :: Nil))
+            (ps.params zip args).foreach((p, arg) => resolveArg(p, arg)(base))
```
LPTK, line 154: At least make the compiler crash instead of doing the wrong thing silently.
```suggestion
            assert(ps.params.sizeCompare(args) === 0)
            (ps.params zip args).foreach((p, arg) => resolveArg(p, arg)(base))
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -0,0 +1,336 @@
...
+  /** Resolve the type signature of a term. */
+  def resolveType(t: Term): Opt[ICtx.Type.Concrete] = t match
+      // If the term is a type application, e.g., T[A, ...],
+      // resolve the type constructor and arguments respectively.
+      case Term.TyApp(con, args) => 
+        (resolveType(con), args.map(resolveType(_)).sequence) match
+          case (S(sym: ICtx.Type.Sym), S(typeArgs)) =>
+            S(ICtx.Type.App(sym, typeArgs))
+          case _ =>
+            // Either the type constructor or the arguments are not resolved.
```
LPTK:
```suggestion
            // Either the type constructor or the arguments is not resolved.
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Importer.scala
```diff
@@ -60,15 +60,9 @@ class Importer:
...
-          // * Note: we don't even need to elaborate the block!
-          // * Though doing so may be needed later for type checking,
-          // * so we should probably do it lazily in the future.
-          /* 
-          given Elaborator.State = new Elaborator.State
-          given Elaborator.Ctx = Elaborator.Ctx.init.nest(N)
-          val elab = Elaborator(tl, file / os.up)
-          val (blk, newCtx) = elab.importFrom(resBlk)
-          */
+          given Elaborator.Ctx = self.prelude.copy(mode = Mode.Light).nest(N)
```
LPTK: Why use `self.prelude` here and just `prelude` below?
```suggestion
          given Elaborator.Ctx = prelude.copy(mode = Mode.Light).nest(N)
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -31,6 +31,7 @@ enum Annot extends AutoLocated:
 enum Term extends Statement:
   case Error
   case UnitVal()
+  case Missing
```
```suggestion
  case Missing // Placeholder terms that were not elaborated due to the "lightweight" elaboration mode `Mode.Light`
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -459,6 +465,15 @@ object PlainFld:
 final case class Spd(eager: Bool, term: Term) extends Elem:
   def showDbg: Str = (if eager then "..." else "..") + term.showDbg
 
+/** 
+ * Context arguments. 
+ * 
+ * Placeholders represents a nil term by default;
```
LPTK: What do you mean by `nil term` here? That's not standard terminology.

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
```diff
@@ -202,6 +202,12 @@ enum Tree extends AutoLocated:
     case Spread(Keyword.`...`, _, S(und: Under)) => S(S(true), new Ident("_").withLocOf(und), N)
     case InfixApp(lhs: Ident, Keyword.`:`, rhs) => S(N, lhs, S(rhs))
     case TermDef(ImmutVal, inner, _) => inner.asParam
+    case Modified(Keyword.`using`, _, inner) => inner match
+      // Param of form (using ..., name: Type). Parse it as usual.
+      case inner: InfixApp => inner.asParam
+      // Param of form (using ..., Type). Synthesize an identifier for it.
+      // TODO: Synthesize a better name.
+      case _ => S(N, Ident(inner.showDbg.replaceAll("\\W", "_")), S(inner))
```
LPTK: That's _cursed_. Why do you need a weird name for this? Just use something like
```suggestion
      case _ => S(N, Ident("$ev")), S(inner))
```
Or maybe just an empty string, which could be useful later on to check whether there was an actual name given.
The name doesn't need to be unique.

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
```diff
@@ -296,10 +303,25 @@ trait TypeOrTermDef:
...
       t match
       
+      // use Foo as foo = ...
+      case InfixApp(typ, Keyword.`as`, id: Ident) if k == Ins =>
+        (S(R(id)), R(id), Nil, N, S(typ))
+      
+      // use Foo = ...
+      case typ if k == Ins =>
+        // TODO: Synthesize a better name.
+        val name = typ.showDbg.replaceAll("\\W", "_")
```
LPTK: Cursed again...

hkmc2/shared/src/test/mlscript-compile/Stack.mls
```diff
@@ -6,7 +6,7 @@ open Predef
 // TODO
 // type Stack[A] = Stack.Cons[A] | Stack.Nil
 
-module Stack with ...
+module Stack[T] with ...
```
LPTK, line 9: What is this strange type parameter for?

hkmc2/shared/src/test/mlscript/basics/BadTypeClasses.mls
```diff
@@ -0,0 +1,59 @@
...
+//│ ╔══[ERROR] Keyword `using` must occur before all parameters.
+//│ ║  l.4: 	  fun f(foo: Int, using bar: Int)
+//│ ╙──     	                        ^^^^^^^^
+
+module M with
+  fun f(using Int)
+
+:e
+M.f
+//│ ╔══[ERROR] Missing instance for contextual parameter Ident_Int_: member:Int#666
```
LPTK: `member:Int#666` should never leak into a user-facing error. Neither names like `Ident_Int_` that are not part of the program source. For now you could just print the name of the type symbol ("Int" here).

hkmc2/shared/src/test/mlscript/basics/BadTypeClasses.mls
```diff
@@ -0,0 +1,59 @@
...
+use 42 = 42
+//│ ╔══[ERROR] Expected a type symbol, got 42
+//│ ║  l.23: 	use 42 = 42
+//│ ╙──      	    ^^
+
+val someInt = 42
+
+:e
+use someInt = 42
+//│ ╔══[ERROR] Expected a type, got member:someInt#666
```
LPTK: Use the `describe` method of `Term` instead?

hkmc2/shared/src/test/mlscript/ucs/normalization/OverlapOfPrimitives.mls
```diff
@@ -1,6 +1,242 @@
 :js
 :ucs normalized
```
LPTK: No idea what happened here. We should obviously not commit that garbage.

You can make it
```suggestion

:global
:ucs normalized
```

hkmc2/shared/src/test/mlscript/ucs/patterns/Refinement.mls
```diff
@@ -1,31 +1,507 @@
 :todo
 :ucs normalized
```
LPTK: Same remark as above.

hkmc2DiffTests/src/test/scala/hkmc2/MLsDiffMaker.scala
```diff
@@ -69,13 +75,21 @@ abstract class MLsDiffMaker extends DiffMaker:
...
+    override def emitDbg(str: String): Unit = output(str)
   
   var curCtx = Elaborator.State.init
+  var curICtx = ImplicitResolver.ICtx.empty
   
+  var prelude = Elaborator.Ctx.empty
   
   override def run(): Unit =
-    if file =/= preludeFile then importFile(preludeFile, verbose = false)
+    // if file =/= preludeFile then
```
LPTK: Any idea why this is no longer needed? 🤔

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -254,6 +263,12 @@ class Lowering(lowerHandlers: Bool, stackLimit: Option[Int])(using TL, Raise, St
             val (paramLists, bodyBlock) = setupFunctionOrByNameDef(td.params, bod, S(td.sym.nme))
             Define(FunDefn(td.owner, td.sym, paramLists, bodyBlock),
               term_nonTail(st.Blk(stats, res))(k))
+          case syntax.Ins =>
+            // Implciit instances are not parameterized for now.
```
```suggestion
            // Implicit instances are not parameterized for now.
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -826,9 +891,11 @@ extends Importer:
               // * Elaborate signature
               val st = td.annotatedResultType.orElse(newSignatureTrees.get(id.name))
               val s = st.map(term(_)(using newCtx))
-              val b = rhs.map(term(_)(using newCtx))
+              val b = if ctx.mode != Mode.Light
```
LPTK, line 902: It looks like this is the only place `ctx.mode` is used. Are you sure there are no other places where it'd be appropriate?

LPTK (approved):
Thanks! I only have minor comments, but overall the changes look good enough. I'll make the merge once you resolve everything.

hkmc2/shared/src/test/mlscript/ucs/normalization/OverlapOfPrimitives.mls
LPTK: Oh I guess that's because it now elaborates `Predef.mls`. But why are we seeing the UCS stuff here? It's supposed to be `Light` elaboration mode, so no UCS term should be processed...

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
FlandiaYingman: Changed to

```scala
/** 
 * Context arguments. 
 * 
 * Placeholders represents a nonexistent term that is *expected*
 * to be populated (resolved) by the implicit resolver later to represent some concrete term.
 */
```

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
FlandiaYingman: A long time ago when I created the PR, this didn't work because JS complained about the identifier containing illegal characters. Now it seems to work fine. I'll remove the `replaceAll` part.

But I don't understand why doesn't the name need to be unique. If the name is not unique, we won't be able to define multiple `use`s in one block, e.g.,

```fs
use Bar[Int, Str] = new IntStrBar()
use Bar[Str, Int] = new StrIntBar()
//│ ╔══[ERROR] Multiple definitions of symbol 'instance'
//│ ╟── defined here
//│ ║  l.99: 	use Bar[Int, Str] = new IntStrBar()
//│ ║        	    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//│ ╟── defined here
//│ ║  l.100: 	use Bar[Str, Int] = new StrIntBar()
//│ ╙──       	    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

which might be a common use case.

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
FlandiaYingman: A long time ago when I created the PR, this didn't work because JS complained about the identifier containing illegal characters. Now it seems to work fine. I'll remove the `replaceAll` part.

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
FlandiaYingman: I think we cannot just assert if the argument lengths are equal because there are spread parameters.

I don't remember why I commented out the check, but seems it works fine. I just uncommented it and resolved any unit test problems in a8871fb1bee9f7b11669ee737daa3ef05157e6aa.

hkmc2/shared/src/test/mlscript-compile/Stack.mls
FlandiaYingman: I guess `Stack` should receive a type param? Otherwise `GeneratorStack.mls` will fail

```fs
class Generator with
  fun produce(result: Stack[Int]): ()
//│ ╔══[ERROR] Wrong number of type arguments
//│ ║  l.8: 	  fun produce(result: Stack[Int]): ()
//│ ╙──     	                      ^^^^^^^^^
```

Or maybe I shall modify `GeneratorStack.mls` instead.

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
LPTK: > I think we cannot just assert if the argument lengths are equal because there are spread parameters.

That was exactly the point: when you leave some checks commented and assume that something hold, you should back the assumption by a runtime check, otherwise you're going to do the wrong thing silently...

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -459,6 +465,15 @@ object PlainFld:
 final case class Spd(eager: Bool, term: Term) extends Elem:
   def showDbg: Str = (if eager then "..." else "..") + term.showDbg
 
+/** 
+ * Context arguments. 
+ * 
+ * Placeholders represents a nonexistent term that is *expected*
```
```suggestion
 * Placeholders represent nonexistent terms that are *expected*
```

hkmc2/shared/src/test/mlscript-compile/Stack.mls
LPTK: Ah yeah that's because `GeneratorStack.mls` expects that `Stack` is already overloaded with the corresponding type (`type Stack[T] = Nil | Cons[T]`). But we don't support overloading yet. It's fine to keep the type parameter here for now.

hkmc2/shared/src/test/mlscript/basics/StrTest.mls
```diff
@@ -17,35 +17,42 @@ concat of
...
 // TODO (~) should be sanitized
 
+:e
 (~)("a", "b", "c")
+//│ ╔══[ERROR] Expected 2 arguments, got 3
+//│ ║  l.21: 	(~)("a", "b", "c")
+//│ ╙──      	 ^
 //│ = "ab"
 
-
```
LPTK, line 23: Please avoid changing the amount of whitespace between each test block.

hkmc2DiffTests/src/test/scala/hkmc2/MLsDiffMaker.scala
FlandiaYingman: Ah, this is a workaround that I now realize it is useless. I just reverted it.

LPTK: On a side note, it's interesting that this doesn't cause a problem even running diff-tests on `Prelude.mls`.

LPTK:
I see you force-push to clean the history, which is laudable. But if you want me to merge this without squashing, make sure you run `git rebase -i --exec "sbt hkmc2AllTests/test" hkmc2` first, to make sure the commits are not broken by rebasing.

hkmc2/shared/src/test/mlscript/ucs/normalization/OverlapOfPrimitives.mls
FlandiaYingman: It turns out that I forgot to copy the mode in `Ctx.nest` so actually all the imported trees were fully elaborated instead. 🤦I set mode to have default argument `Mode.Full` at the constructor and was hoping all copy procedure was done by the case class `copy` method. I just removed the default argument.

I used to wonder why do you so hate default arguments and now I know why 🤦

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: I checked again and AFAIK that's the only place where we don't elaborate if the mode is light. Given that the TermDef body terms are the only terms that we want to omit.

hkmc2/shared/src/test/mlscript/ucs/normalization/OverlapOfPrimitives.mls
LPTK: Yeah default arguments for important logic like this really are just asking for trouble as they will silence any unintentional omission. IMO default arguments are only acceptable in external-facing APIs for ease of use, not in internal logic, except in rare cases (like `inAppPrefix`).

LPTK:
PS: Next time please pick a more informative branch name 😉

PR #257, opened 2024-12-26
Function instrumentation and effect handlers transformation

CAG2Mark:
# Instrumentation of functions
Allows functions executions to be suspended by creating an associated continuation class with each function / lambda body which can be resumed. The continuation class are intended to be chained together with other functions' continuation classes to create a "heapified" structured stack frame.

The main idea is to treat functions as a state machine, with state transitions occurring at all possibly effectful function calls. More state transitions may be inserted to support various kinds of control flow, such as if/while.

At each possibly effectful function call, functions will check if the result of that call is a continuation. If so, it will capture its current stack frame into a continuation, including the current state in the state machine, and append it onto the other continuation.

# Effect handlers
Use the above instrumentation of functions to compile effect handlers such as:
```fs
handle h = Effect with
  fun perform(arg)(k) = k(arg); arg
in foo(h)
```
Most of the additional infrastructure needed to compile handlers are in the Predef. Nested handlers, and handlers within other handlers, are also supported. This part was mostly done by @AnsonYeung 

# TODO
- [x] Verify our implementation's semantics against other implementations


hkmc2/shared/src/test/mlscript/handlers/RecursiveHandlers.mls
```diff
@@ -0,0 +1,123 @@
...
+//│ > B hola amigos
+//│ = '‹,‹,‹,‹,hola,amigos,›,adios,›,friend,›,bye,›'
+
+
+:handler 
+let str = ""
+if true do
+  handle h1 = Effect with
+    fun perform(arg)(k) =
+      set str = str + "A"
```
LPTK:
```suggestion
      set str += "A"
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -52,14 +54,147 @@ sealed abstract class Block extends Product with AutoLocated:
...
+    }
+  
+  def mapSyms(f: Local => Local)(using State): Block = 
+    def g(t: Local) = t match 
+      case t: TermSymbol =>
+        TermSymbol(t.k, t.owner.map(symMap), t.id)
+  
+      case _ => f(t)
+    def symMap[T <: Local : ClassTag](l: T): T = g(l) match
+      case r : T => r
```
LPTK: The use of `ClassTag` for pattern matching is a completely unsound misfeature of Scala 2. Doesn't this yield a warning by now?

The proper way in Scala 3 is through [TypeTest](https://docs.scala-lang.org/scala3/reference/other-new-features/type-test.html).

Though I'm not convinced `symMap` is a very justified function to have. It looks like it's a hack fixing a symptom of another problem. I'm pretty sure the design of the current `Symbol` class hierarchy has problems and could be improved and would rather try to understand what the root problem being addressed here is.

hkmc2/shared/src/test/mlscript/handlers/Generators.mls
```diff
@@ -0,0 +1,57 @@
+:js
+:handler
+import "../../mlscript-compile/Stack.mls"
```
LPTK: This isn't used.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -0,0 +1,609 @@
...
+    val newBlk = defns.foldLeft(blk)((acc, defn) => Define(defn, acc))
+    newBlk.mapSyms(symMap)
+  
+  private def translateFun(f: FunDefn): FunDefn =
+    FunDefn(f.sym, f.params, translateBlock(f.body, functionHandlerCtx))
+  
+  private def translateCls(cls: ClsLikeDefn): ClsLikeDefn =
+    cls.copy(methods = cls.methods.map(translateFun), ctor = translateBlock(cls.ctor, functionHandlerCtx))
+
+  // to ensure the fun and class reference sin the continuation class is properly scoped,
```
```suggestion
  // to ensure the fun and class references in the continuation class are properly scoped,
```

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -348,11 +350,106 @@ class HandlerLowering(using TL, Raise, Elaborator.State):
...
+  
   private def translateFun(f: FunDefn): FunDefn =
     FunDefn(f.sym, f.params, translateBlock(f.body, functionHandlerCtx))
   
   private def translateCls(cls: ClsLikeDefn): ClsLikeDefn =
     cls.copy(methods = cls.methods.map(translateFun), ctor = translateBlock(cls.ctor, functionHandlerCtx))
+
+  // to ensure the fun and class reference sin the continuation class is properly scoped,
+  // we move all function defns to the top level of the handler block
+  def popDefns(b: Block, acc: List[Defn]): (Block, List[Defn]) =
```
LPTK: Minor: the conventional name for this operation is "float out", not "pop".

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -348,11 +350,106 @@ class HandlerLowering(using TL, Raise, Elaborator.State):
...
+      case sym => sym -> sym
+    }.toMap
+
+    val clsNmeMap = newSyms.collect {
+      case (c1: ClassSymbol, c2: ClassSymbol) => c1.id.name -> c2.id.name
+    }.toMap
+
+    def symMap(l: Local) = newSyms.get(l) match
+      case None => l match
+        case b: BlockMemberSymbol => clsNmeMap.get(b.nme) match
```
LPTK: Could you document the point of this `clsNmeMap`?

In principle, after the name resolution stage (done by `Elaborator`), there should be zero string-based resolution happening anywhere in the compiler. That's the purpose of symbols. Why do we need it here?

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
CAG2Mark: The root problem is that the input of `mapSyms` *must* be a function that maps symbols of the same subtype to the same subtype because many blocks' parameters require specific types of symbols, and these types must be preserved after being mapped. But, I'm not sure how to express such a function using the type system other than this kind of hack.

A similar problem exists for `Result`. In that case I can make a separate function for each kind of `Result` but it's not ideal.

The only "solution" I can think of is to pass in different functions for each specific symbol type, but that's very clunky and does not solve the problem when a union or intersection type of symbols is expected.

Also no, using `ClasssTag` does not yield a warning.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
CAG2Mark: The problem is the combination of the fact that `BlockMemberSymbol` is string-based, is used for class constructors, and is completely separated from `ClassSymbol`.

Every class has its own `ClassSymbol`. But they also have a separate `BlockMemberSymbol` for their constructor, which is currently purely string-based. Therefore, if we wish to change the name of a class, we are forced to update the name of the associated `BlockMemberSymbol` which involves changing the string.

The reason I need to change the names in the first place is to make them unique. This stems from the need to move function and class definitions to the top level of each instrumented function so that they can be accessed by the continuation. For example,
```fs
fun f() =
  if something then
    fun g() =
      fun h() = 2
      h()
    g()
  else
    fun g() = 3
    g()
```
will be changed to something like
```fs
fun f() =
  fun g1() =
    fun h() = 2
    h()
  fun g2() = 3
  if something then
    g1()
  else
    g2()
```
If two symbols are different but have the same name, they will be still emitted with the same name in the JS builder, so I need to rename the functions after moving them to the top level. If this is not the intended behaviour of the JS builder, then we can fix that instead and I can get rid of basically all the hacky stuff I've done.

hkmc2/shared/src/test/mlscript/handlers/RecursiveHandlers.mls
CAG2Mark: Oh I didn't realize we had this operator, nice

LPTK: It works for all symbolic operators ;^)

BTW there are many places to update, not just the one I marked here.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
LPTK: Here is how I suggest to fix this:
* You define a `SymbolSubst` class that contains one Map per type of symbol to substitute.
* You declare a generic `subst(using SymbolSubst): Symbol` method in `Symbol` and give implementations in the Symbol subclasses  that _refine_ this signature, e.g. `subst(using BlockMemberSymbol): BlockMemberSymbol` in `BlockMemberSymbol`.

Note that you don't need to define one such Map + refining implementation per concrete leaf subclass of Symbol. For some symbol sub-hierarchies, it will be sufficient to define a single one.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
LPTK: > The problem is the combination of the fact that `BlockMemberSymbol` is string-based

I guess by this you mean that it's "nominal", i.e., its name is visible/accessible to surface-language users and must be code-gen'd as is (without allocating a fresh name).

> is used for class constructors, and is completely separated from `ClassSymbol`.

This is by design, yes.

The issue is that so far, assuming nominality for that `BlockMemberSymbol` was fine. But since we now need to synthesize fresh `BlockMemberSymbol`, these should get fresh names allocated for them. Note that this fresh name allocation should happen in `JSBuilder` through the `Scope` construct. Manually dealing with creating fresh names in your transformation pass is unnecessary and suboptimal.

Let me see what's the best way of changing the current definitions to accommodate this.

LPTK:
The CI fails because some auto-generated number is different. So it looks like there's some nondeterminism in how you generate these numbers. Any idea what's the source of it? Ideally you'd want to use a fresh state for each individual compilation unit. Like we do with `Elaborator.State` (which you could also use).

CAG2Mark:
We use the same state that is used in the Elaborator to generate symbols, so that should not be the cause. We also use a private counter to generate some other IDs, but this is deterministic.

However, we use sets and maps in a few places. Maybe in some of those, the items/keys being hashed don't have a properly overridden hashcode function, causing some non-determinism in the order in which some IDs are generated.

AnsonYeung:
The change in number is probably caused by the merge done by the CI.
On the other hand, the number is generated using `State.suid.nextUid`

LPTK:
Ah, good catch!

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -670,7 +670,7 @@ extends Importer:
         val res: Term.Blk = ctx.nest(N).givenIn:
           val sym = fieldOrVarSym(HandlerBind, id)
           log(s"Processing `handle` statement $id (${sym}) ${ctx.outer}")
-
```
LPTK: Please avoid needless whitespace changes. Disable automatic space trimming if needed.

CAG2Mark:
Updated to use `SymbolSubst`. I've resolved the original comment https://github.com/hkust-taco/mlscript/pull/257#discussion_r1898571379 since this addresses the main problem there.

However, there are still some other problems. The smaller ones are labelled `HACK:` in the diff.

A bigger problem is more related to https://github.com/hkust-taco/mlscript/pull/257#discussion_r1898577646. I realized my original method of mapping names and replacing symbols using the general `mapSyms` function does not work for classes, such as in this example:
```fs
fun foo(h) =
  h.perform()
  if false then
    class A() with
      fun f() = 2
    A().f()
  else
    class A() with
      fun f() = 3
    A().f()
```
I'll try to see if there's a way to fix this that is not too tedious.

BTW, I recall you mentioned you were going to update the symbol system in some way to improve the handling of nominal names -- is that the case?

CAG2Mark:
OK I managed to fix all the issues mentioned above.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -279,11 +280,10 @@ class HandlerLowering(using TL, Raise, Elaborator.State, Elaborator.Ctx):
   
   private def translateBlock(b: Block, h: HandlerCtx): Block =
     given HandlerCtx = h
+
```
LPTK: It's weird to have an empty lie here in the middle of `translateBlock`'s body but none between `translateBlock` and `firstPass`.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -316,45 +316,75 @@ class HandlerLowering(using TL, Raise, Elaborator.State, Elaborator.Ctx):
     // to ensure the fun and class references in the continuation class are properly scoped,
     // we move all function defns to the top level of the handler block
     val (blk, defns) = b.floatOutDefns
-    val syms = defns.collect {
+    val clsDefns = defns.collect {
```
```suggestion
    val clsDefns = defns.collect:
```

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -316,45 +316,75 @@ class HandlerLowering(using TL, Raise, Elaborator.State, Elaborator.Ctx):
...
       case ClsLikeDefn(sym, k, parentPath, methods, privateFields, publicFields, preCtor, ctor) => sym
+    }
+    val funDefns = defns.collect {
```
```suggestion
    val funDefns = defns.collect:
```

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -316,45 +316,75 @@ class HandlerLowering(using TL, Raise, Elaborator.State, Elaborator.Ctx):
...
+    val bmsMap = (fnBmsMap ++ clsBmsMap).toMap
+
+    val clsMap = clsBmsMap.map {
+      case b1 -> b2 => b1.asCls match
+        case Some(value) => 
+          val newSym = ClassSymbol(value.tree, Tree.Ident(b2.nme))
+          newSym.defn = value.defn
+          S(value -> newSym)
+        case None => None
+    }.collect{ case Some(x) => x }.toMap
```
```suggestion
    val clsMap = clsBmsMap.map:
        case b1 -> b2 => b1.asCls match
          case Some(value) => 
            val newSym = ClassSymbol(value.tree, Tree.Ident(b2.nme))
            newSym.defn = value.defn
            S(value -> newSym)
          case None => None
      .collect{ case Some(x) => x }.toMap 
```

LPTK: Same suggestion below.

hkmc2/shared/src/test/mlscript/handlers/Effects.mls
```diff
@@ -230,7 +230,7 @@ fun f(perform) =
   handle h = Effect with
     fun perform(arg)(k) = arg
   perform()
-f(() => 3)
+f(() => 3)
```
LPTK: Weird space.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -65,19 +68,29 @@ class FlowSymbol(label: Str)(using State) extends Symbol:
...
   def id: Ident
+  override def subst(using s: SymbolSubst): NamedSymbol
 
 abstract class BlockLocalSymbol(name: Str)(using State) extends FlowSymbol(name) with LocalSymbol:
   var decl: Opt[Declaration] = N
+  // HACK: for some reason, if there is no concrete implementation, Scala overrides this subst with the one in
+  // FlowSymbol and then gives a type error because FlowSymbol is not a subtype of BlockLocalSymbol...
+  // Any way to force it to be abstract?
+  // For now, you should be careful when extending BlockLocalSymbol.
+  override def subst(using SymbolSubst): BlockLocalSymbol = lastWords("tried to subst BlockLocalSymbol")
```
LPTK: Yeah, Scala has some annoying limitation like this. This one is due to the requirement that `extends`/`with` clauses need to be checked for validity locally (contrary to self-type annotations). I fixed this by turning the `with LocalSymbol` clause into a self-type annotation.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -51,15 +53,142 @@ sealed abstract class Block extends Product with AutoLocated:
...
+    this match
+      case Return(res, implct) => Return(resMap(res), implct)
+      case Throw(exc: Path) => Throw(pathMap(exc))
+      case Assign(l, r, rst) => Assign(l.subst, resMap(r), rst.mapSyms)
+      case blk @ AssignField(l, n, r, rst) => AssignField(pathMap(l), n, _resMap(r), rst.mapSyms)(blk.symbol)
+      case Match(scrut, arms, dflt, rst) => 
+        val newArms = arms.map((cse, blk) =>
+          val newCse = cse match
+            case Case.Lit(lit) => cse
+            // HACK: cls.subst gives something other than ClassSymbol | ModuleSymbol and it doesn't type check...
```
LPTK: Another limitation of Scala, though it is not too severe. I wouldn't call this a hack, just a mostly benign workaround. This could be fixed by introducing yet another parent trait and using that instead of `ClassSymbol | ModuleSymbol`.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -0,0 +1,552 @@
...
+  def partitionBlock(blk: Block, labelIds: Map[Symbol, (StateId, StateId)] = Map.empty): Ls[BlockState] = 
+    // for readability :)
+    case class PartRet(head: Block, states: Ls[BlockState])
+
+    // returns (truncated input block, child block states)
+    // TODO: don't split within Match, Begin and Labels when not needed, ideally keep it intact. Need careful analysis for this
+    // blk: The block to transform
+    // labelIds: maps label IDs to the state at the start of the label and the state after the label
+    // jumpTo: what state End should jump to, if at all 
+    // freshState: uid generator
```
LPTK: Just a small suggestion to make the comment easier to parse.
```suggestion
    // * returns (truncated input block, child block states)
    // * blk: The block to transform
    // * labelIds: maps label IDs to the state at the start of the label and the state after the label
    // * jumpTo: what state End should jump to, if at all 
    // * freshState: uid generator
    // TODO: don't split within Match, Begin and Labels when not needed, ideally keep it intact. Need careful analysis for this.
```

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -0,0 +1,552 @@
...
+        sym -> BlockMemberSymbol(sym.nme + "$" + thirdPassFresh(), sym.trees)
+    }.toMap
+
+    val clsBmsMap = toConvert.map(b =>
+      b -> BlockMemberSymbol(b.nme + "$" + thirdPassFresh(), b.trees)  
+    ).toMap
+
+    val bmsMap = (fnBmsMap ++ clsBmsMap).toMap
+
+    val clsMap = clsBmsMap.map {
```
```suggestion
    val clsMap = clsBmsMap.map:
```

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -152,12 +152,14 @@ class HandlerLowering(using TL, Raise, Elaborator.State, Elaborator.Ctx):
...
 
-    // returns (truncated input block, child block states)
-    // TODO: don't split within Match, Begin and Labels when not needed, ideally keep it intact. Need careful analysis for this
-    // blk: The block to transform
-    // labelIds: maps label IDs to the state at the start of the label and the state after the label
-    // jumpTo: what state End should jump to, if at all 
-    // freshState: uid generator
+    /* 
+    * returns (truncated input block, child block states)
+    * TODO: don't split within Match, Begin and Labels when not needed, ideally keep it intact. Need careful analysis for this
```
LPTK: Part of my comment was meant to distinguish between the TODO and the parameter enumeration. This is why I added ticks in front of each parameter and a period at the end of the TODO. Your change misses out on both things.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -345,34 +346,33 @@ class HandlerLowering(using TL, Raise, Elaborator.State, Elaborator.Ctx):
       else None
     ).collect { case Some(b) => b }
 
-    val fnBmsMap = funDefns.map {
-      case sym: BlockMemberSymbol => 
-        sym -> BlockMemberSymbol(sym.nme + "$" + thirdPassFresh(), sym.trees)
-    }.toMap
+    val fnBmsMap = funDefns.map(b =>
+      b -> BlockMemberSymbol(b.nme + "$" + thirdPassFresh(), b.trees)
+    ).toMap
```
LPTK: The RHS of such definitions should be indented (same remark for the other similar places in this file).

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -345,34 +346,33 @@ class HandlerLowering(using TL, Raise, Elaborator.State, Elaborator.Ctx):
...
-    val clsMap = clsBmsMap.map {
+    val clsMap = clsBmsMap.map:
       case b1 -> b2 => b1.asCls match
         case Some(value) => 
           val newSym = ClassSymbol(value.tree, Tree.Ident(b2.nme))
           newSym.defn = value.defn
           S(value -> newSym)
         case None => None
-    }.collect{ case Some(x) => x }.toMap 
+    .collect{ case Some(x) => x }.toMap
```
LPTK: Please indent this, following my previous suggestions (and similarly below).

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -195,7 +198,7 @@ sealed trait InnerSymbol extends Symbol:
   def subst(using SymbolSubst): InnerSymbol
 
 class ClassSymbol(val tree: Tree.TypeDef, val id: Tree.Ident)(using State)
-    extends MemberSymbol[ClassDef] with CtorSymbol with InnerSymbol:
+    extends ClassLikeSymbol with CtorSymbol with InnerSymbol:
```
LPTK: `ClassSymbol` is a `MemberSymbol[ClassDef]`, not a `MemberSymbol[ClassDef | ModuleDef]`. So `ClassLikeSymbol` should probably either take a type parameter or be a trait. Not sure if this difference is problematic for your use case. Let me know if it causes problems.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -358,7 +355,7 @@ case class End(msg: Str = "") extends BlockTail with ProductWithTail
 
 enum Case:
   case Lit(lit: Literal)
-  case Cls(cls: ClassSymbol | ModuleSymbol, path: Path)
+  case Cls(cls: ClassLikeSymbol | ModuleSymbol, path: Path)
```
LPTK: Shouldn't this be
```suggestion
  case Cls(cls: ClassLikeSymbol, path: Path)
```
?

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
CAG2Mark: It seems to cause problems if I rewrite it to `ClassSymbol[T <: Definition]`. The type system can't prove that `ClassLikeSymbol[ClassDef] <: ClassLikeSymbol[ClassDef | ModuleDefn]`. If I make the type parameter `T` covariant, I also have to make the type parameter `Defn` of `MemberSymbol` covariant, but that's not possible due to `var defn: Opt[Defn]` in `MemberSymbol`.

LPTK: > The type system can't prove that `ClassLikeSymbol[ClassDef] <: ClassLikeSymbol[ClassDef | ModuleDefn]`.

Well, that's because it's not the case. Where is that needed?

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
CAG2Mark: So you mean like
```fs
    val fnBmsMap = funDefns.map(b =>
        b -> BlockMemberSymbol(b.nme + "$" + thirdPassFresh(), b.trees)
      ).toMap
```
?

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
CAG2Mark: OK, when I experimented with it, I made `ClassLikeSymbol` take a type parameter `T <: Definition` and updated `ClassSymbol` and `ModuleSymbol` accordingly.

then, I tried to change
```
case Cls(cls: ClassLikeSymbol, path: Path)
```
into
```
case Cls(cls: ClassLikeSymbol[ClassDef | ModuleDef], path: Path)
```
which is why I need `ClassSymbol <: ClassLikeSymbol[ClassDef] <: ClassLikeSymbol[ClassDef | ModuleDef]` in some places.

If I instead changed it to
```
case Cls(cls: ClassLikeSymbol[ClassDef] | ClassLikeSymbol[ModuleDef], path: Path)
```
then I need to use the original workaround here https://github.com/hkust-taco/mlscript/pull/257#discussion_r1914145665 again, which defeats the whole point of this.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
LPTK: Yes, or
```scala
    val fnBmsMap = funDefns.map: b =>
        b -> BlockMemberSymbol(b.nme + "$" + thirdPassFresh(), b.trees)
      .toMap
```
if you want.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
LPTK: I fixed it for you. It was not very hard. Note that the self-type I added mostly for documentation (it's not currently necessary).

BTW you don't need to use `override` when you're not actually overriding any implementation. Better leave it out.

CAG2Mark: Thanks! I wasn't too familiar with self-types before. I'll keep it in mind for the future.

CAG2Mark:
Now ready for review.

LPTK:
I see the build fails with `mlscript/nofib/eliza`. It's a JVM stack overflow (due to the expression being quite large). It happened sometimes on my machine but seems nondeterministic (heisenbug). We should eventually look into making the lowering `term` function tail-recursive...

LPTK:
Ok my change seems to fix it. I rolled the handler lowering into `term` and made most `term` recursive calls tail-recursive, which should lower its stack usage a lot.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -51,15 +53,139 @@ sealed abstract class Block extends Product with AutoLocated:
...
+        body.mapSyms
+      )
+      case Value.Arr(elems) => Value.Arr(elems.map(a => Arg(a.spread, pathMap(a.value))))
+
+    def _resMap(r: Result) = r match
+      case p: Path => pathMap(p)
+      case c @ Call(fun, args) => Call(pathMap(fun), args.map(a => Arg(a.spread, pathMap(a.value))))(c.isMlsFun)
+      case Instantiate(cls, args) => Instantiate(pathMap(cls), args.map(pathMap))
+
+    def resMap[T <: Result : ClassTag](l: T): T = _resMap(l) match
```
LPTK: This is not good, as already explained. Please remove it.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -12,6 +12,8 @@ import hkmc2.semantics.{Term => st}
 import syntax.{Literal, Tree}
 import semantics.*
 import semantics.Term.*
+import scala.reflect.ClassTag
```
```suggestion
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -80,11 +206,69 @@ sealed abstract class Block extends Product with AutoLocated:
...
-    case HandleBlock(lhs, res, cls, handlers, body, rest) =>
-      (body.freeVars - lhs) ++ rest.freeVars ++ handlers.flatMap(_.freeVars)
+    case HandleBlock(lhs, res, par, cls, hdr, bod, rst) =>
+      (bod.freeVars - lhs) ++ rst.freeVars ++ hdr.flatMap(_.freeVars)
     case HandleBlockReturn(res) => res.freeVars
     case End(msg) => Set.empty
 
+  def floatOutDefns =
+    def rec(b: Block, acc: List[Defn]): (Block, List[Defn]) =
+      b match
```
LPTK: Pro-tip: you don't need to indent after `match`. For nested matches (like here), this can save a lot on of needless indentation.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -0,0 +1,561 @@
...
+import hkmc2.Message.MessageContext
+
+import semantics.Elaborator.State
+
+import syntax.{Literal, Tree, ParamBind}
+import semantics.*
+import scala.annotation.tailrec
+import hkmc2.semantics.Elaborator.ctx
+
+import hkmc2.utils.SymbolSubst
```
LPTK: I like to order imports:
* Scala level imports
* Project level imports
* Module level imports

```suggestion
import scala.annotation.tailrec

import mlscript.utils.*, shorthands.*
import hkmc2.utils.*
import hkmc2.utils.SymbolSubst
import hkmc2.Message.MessageContext

import syntax.{Literal, Tree, ParamBind}
import semantics.*
import semantics.Elaborator.ctx
import semantics.Elaborator.State
```

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -0,0 +1,561 @@
...
+    def chain(other: Block => Block): Block => Block = b => k(other(b))
+    def rest(b: Block): Block = k(b)
+    def transform(f: (Block => Block) => (Block => Block)) = f(k)
+    
+    def assign(l: Local, r: Result) = k.chain(Assign(l, r, _))
+    def assignFieldN(lhs: Path, nme: Tree.Ident, rhs: Result) = k.chain(AssignField(lhs, nme, rhs, _)(N))
+    def break(l: Local): Block = k.rest(Break(l))
+    def continue(l: Local): Block = k.rest(Continue(l))
+    def define(defn: Defn) = k.chain(Define(defn, _))
+    def end() = k.rest(End())
```
```suggestion
    def end = k.rest(End())
```

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -0,0 +1,561 @@
...
+
+import hkmc2.utils.SymbolSubst
+
+object HandlerLowering:
+
+  private val pcIdent: Tree.Ident = Tree.Ident("pc")
+  private val nextIdent: Tree.Ident = Tree.Ident("next")
+  private val tailIdent: Tree.Ident = Tree.Ident("tail")
+
+  extension (k: Block => Block)
```
LPTK: These look like pretty general helpers. Maybe they belong in a more general file, perhaps even `Block.scala`.

LPTK: Same remark for `extension (p: Path)`.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -0,0 +1,561 @@
...
+
+      case Match(scrut, arms, dflt, rest) => 
+        val restParts = go(rest)
+        val restId: StateId = restParts.head match
+          case StateTransition(uid) => uid
+          case _ => freshId()
+        
+        val armsParts = arms.map((cse, blkk) => (cse, go(blkk)(afterEnd = S(restId))))
+        val dfltParts = dflt.map(blkk => go(blkk)(afterEnd = S(restId)))
+
```
```suggestion
```

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -0,0 +1,561 @@
...
+            )
+          case _ =>
+            PartRet(
+              StateTransition(startId), 
+              BlockState(startId, bodyNew, N) :: BlockState(endId, restNew, N) :: parts ::: restParts
+            )
+
+      case Break(label) =>
+        val (start, end) = labelIds.get(label) match
+          case N => raise(ErrorReport(
```
LPTK: This looks likes it should never happen unless there's a compiler bug, isn't it? In this case you should just crash the compiler.

```suggestion
          case N => raise(InternalError(
```

`raise` is for actionable _user_ errors with nice diagnostic messages.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -0,0 +1,561 @@
...
+
+    val result = go(blk)(using labelIds, N)
+    BlockState(headId, result.head, N) :: result.states
+  
+  /**
+   * The actual translation:
+   * 1. add call markers, transform class, function, lambda and sub handler blocks
+   * 2.
+   * a) generate continuation class
+   * b) generate normal function body
```
```suggestion
   *   a) generate continuation class
   *   b) generate normal function body
```

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -0,0 +1,561 @@
...
+        case b => b.mapResult:
+          case r @ Call(Value.Ref(_: BuiltinSymbol), _) => N
+          case c: Call =>
+            val res = freshTmp("res")
+            S(k => CallPlaceholder(res, freshId(), false, c, k(Value.Ref(res))))
+          case r => N
+      } match
+        case Define(f: FunDefn, rst) => Define(translateFun(f), rst)
+        case Define(c: ClsLikeDefn, rst) => Define(translateCls(c), rst)
+        case b => b
```
```suggestion
    b.map(firstPass) match
    case b: HandleBlock => translateHandleBlock(b)
    case b => b.mapValue:
        case Value.Lam(params, body) => Value.Lam(params, translateBlock(body, functionHandlerCtx))
        case v => v
      .match
        case Return(c: Call, implct) if handlerCtx.isHandleFree => Return(c, implct)
        case b => b.mapResult:
          case r @ Call(Value.Ref(_: BuiltinSymbol), _) => N
          case c: Call =>
            val res = freshTmp("res")
            S(k => CallPlaceholder(res, freshId(), false, c, k(Value.Ref(res))))
          case r => N
      .match
        case Define(f: FunDefn, rst) => Define(translateFun(f), rst)
        case Define(c: ClsLikeDefn, rst) => Define(translateCls(c), rst)
        case b => b
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -653,36 +653,44 @@ extends Importer:
...
-              
-          case st => 
-            raise(ErrorReport(msg"Only function definitions are allowed in handler blocks" -> st.toLoc :: Nil))
-            None
-        }.collect { case Some(x) => x }
+        val res: Term.Blk = ctx.nest(N).givenIn:
+          val sym = fieldOrVarSym(HandlerBind, id)
+          log(s"Processing `handle` statement $id (${sym}) ${ctx.outer}")
+          
+          // TODO: shouldn't need uid here
```
LPTK, line 696: Indeed, it shouldn't be needed. We'll be able to remove it when I change my symbols change.

hkmc2/shared/src/main/scala/hkmc2/utils/SymbolSubst.scala
```diff
@@ -0,0 +1,18 @@
...
+  def mapVarSym(s: VarSymbol): VarSymbol = s
+  def mapInstSym(s: InstSymbol): InstSymbol = s
+  def mapBuiltInSym(s: BuiltinSymbol): BuiltinSymbol = s
+  def mapTermSym(s: TermSymbol): TermSymbol = s
+  def mapCtorSym(s: CtorSymbol): CtorSymbol = s
+  def mapClsSym(s: ClassSymbol): ClassSymbol = s
+  def mapModuleSym(s: ModuleSymbol): ModuleSymbol = s
+  def mapTypeAliasSym(s: TypeAliasSymbol): TypeAliasSymbol = s
+  def mapPatSym(s: PatternSymbol): PatternSymbol = s
+  def mapTopLevelSym(s: TopLevelSymbol): TopLevelSymbol = s
```
LPTK: Always end files on an empty line.

```suggestion
  def mapTopLevelSym(s: TopLevelSymbol): TopLevelSymbol = s

```

LPTK (commented):
The changes look reasonable on a first skim. But code could use more documentation, given the tricky logic. (Yes, I know, the rest of the code is also lacking in doc.)

I'll need to have a deeper look tomorrow.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
CAG2Mark: Yes I agree, I actually made the exact comment on the stack-safety branch. I'll do that once I address some other things.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
CAG2Mark: > `raise` is for actionable user errors with nice diagnostic messages.

I suppose that's unless an `InternalError` is raised, as suggested above?

CAG2Mark:
I think that documenting the logic is a tricky task in itself, as we probably need a lot of examples and perhaps diagrams to illustrate it. A proper formalization will of course take even more time.

Perhaps we can at least do the former in our progress report so we have more time to polish it.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -390,3 +390,28 @@ enum Value extends Path:
...
+  def staticif(b: Boolean, f: (Block => Block) => (Block => Block)) = if b then k.transform(f) else k
+
+def blockBuilder: Block => Block = identity
+
+extension (p: Path)
+  def selN(id: Tree.Ident) = Select(p, id)(N)
+  def asArg = Arg(false, p)
+
+extension (l: Local)
+  def asPath: Path = Value.Ref(l)
```
AnsonYeung: could this be methods in the class directly instead of extension method

LPTK: Agreed.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
LPTK: Ah yeah I that's what I meant.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
CAG2Mark: Good catch, thanks

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -80,11 +201,68 @@ sealed abstract class Block extends Product with AutoLocated:
...
     case AssignField(lhs, nme, rhs, rest) => lhs.freeVars ++ rhs.freeVars ++ rest.freeVars
     case Define(defn, rest) => defn.freeVars ++ rest.freeVars
-    case HandleBlock(lhs, res, cls, handlers, body, rest) =>
-      (body.freeVars - lhs) ++ rest.freeVars ++ handlers.flatMap(_.freeVars)
+    case HandleBlock(lhs, res, par, cls, hdr, bod, rst) =>
+      (bod.freeVars - lhs) ++ rst.freeVars ++ hdr.flatMap(_.freeVars)
     case HandleBlockReturn(res) => res.freeVars
     case End(msg) => Set.empty
 
+  def floatOutDefns =
```
LPTK, line 90: Please document what this function does.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -51,15 +52,135 @@ sealed abstract class Block extends Product with AutoLocated:
     case Define(_, rst) => 1 + rst.size
     case TryBlock(sub, fin, rst) => 1 + sub.size + fin.size + rst.size
     case Label(_, bod, rst) => 1 + bod.size + rst.size
+    case HandleBlock(lhs, res, par, cls, handlers, bdy, rst) => 1 + handlers.map(_.body.size).sum + bdy.size + rst.size
+  
+  // ignoring blocks inside functions and handle block
```
LPTK: I'm not a fan of this menagerie of `map`-like functions, especially with ones like this that are very specialized (not touching handlers).

There's a good way of organizing such code modularly in a way that won't require lots of work every time we change the constructors of the IR. Take a look at `Traverser2`:

https://github.com/hkust-taco/mlscript/blob/dedfc2d57698c9d509cd8fdf7b35648fd5246dc7/shared/src/main/scala/mlscript/TyperHelpers.scala#L1276C9-L1348

and how it uses different `applyX` functions for each kind `X` of tree.

It would be a good improvement to add a `BlockTransformer` (in a separate file) with `mapBlock`, `mapResult`, etc. where the default implementation just maps to the same thing. Then, one can selectively override the things they want to map or the ones they don't want to be mapped (like `HandleBlock`, here). 

A further optimization is to avoid reconstructing the trees when they haven't actually changed. That is, instead of 
```scala
case Begin(sub, rst) => Begin(f(sub), f(rst))
```
you make it
```scala
case b @ Begin(sub, rst) =>
  val sub2 = f(sub)
  val rst2 = f(rst)
  if (sub2 is sub) && (rst2 is rst) then b else Begin(f(sub), f(rst))
```

Since you're doing lots of mapping, including for things like symbols where the majority probably won't change, this can reduce the number of allocations by a lot and thus make the code much faster.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -51,15 +52,135 @@ sealed abstract class Block extends Product with AutoLocated:
...
+  
+  def mapPath(f: Path => Path): Block = this match
+    case Return(res: Path, implct) => Return(f(res), implct)
+    case Throw(exc: Path) => Throw(f(exc))
+    case Assign(l, r: Path, rst) => Assign(l, f(r), rst)
+    case b @ AssignField(l, n, r: Path, rst) => AssignField(l, n, f(r), rst)(b.symbol)
+    case Match(scrut: Path, arms, dflt, rst) => Match(f(scrut), arms, dflt, rst)
+    case Define(ValDefn(owner, k, sym, rhs: Path), rst) => Define(ValDefn(owner, k, sym, f(rhs)), rst)
+    case HandleBlockReturn(res: Path) => HandleBlockReturn(f(res))
+    case _: Return | _: Throw | _: Assign | _: AssignField | _: End | _: Break | _: Continue | _: Match | _: Label | _: Begin | _: TryBlock | _: Define | _: HandleBlock | _: HandleBlockReturn => this.mapResult {
```
LPTK: Please try not to exceed 100-character lines.

Also braces are lame:
```suggestion
    case _: Return | _: Throw | _: Assign | _: AssignField | _: End | _: Break | _: Continue | _: Match | _: Label | _: Begin | _: TryBlock | _: Define | _: HandleBlock | _: HandleBlockReturn => this.mapResult:
```

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -0,0 +1,533 @@
...
+        case Return(res, implct) =>
+          // In the case of res is effectful, it will be handled in translateBlock
+          Assign(tmp, res, Return(Instantiate(retClsPath, tmp.asPath :: Nil), implct))
+        case HandleBlockReturn(res) =>
+          Return(res, false)
+        case b => b
+      go(b)
+    val handlerBody = translateBlock(prepareBody(h.body), HandlerCtx(false, false, state => blockBuilder
+      .assignFieldN(state.res.tail, nextIdent, Instantiate(state.cls, Value.Lit(Tree.IntLit(state.uid)) :: Nil))
+      // .assignFieldN(state.res, tailIdent, state.res.tail.next)
```
LPTK: Any reason to leave this commented code in?

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -0,0 +1,533 @@
...
+      .ret(SimpleCall(handleBlockImplPath, state.res :: h.lhs.asPath :: Nil))))
+    
+    val handlers = h.handlers.map: handler =>
+      val lam = Value.Lam(PlainParamList(Param(FldFlags.empty, handler.resumeSym, N) :: Nil), translateBlock(handler.body, functionHandlerCtx))
+      val tmp = freshTmp()
+      FunDefn(handler.sym, handler.params, Return(SimpleCall(mkEffectPath, h.lhs.asPath :: lam :: Nil), false))
+    
+    // TODO: it seems that our current syntax didn't know how to call super, calling it with empty param list now
```
LPTK, line 449: Can you elaborate? Maybe open a GitHub issue with a minimal example.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -0,0 +1,533 @@
...
+    val result = Define(defn, CallPlaceholder(h.res, freshId(), true, Call(sym.asPath, Nil)(true), h.rest))
+    result
+  
+  private def genContClass(b: Block)(using HandlerCtx): Opt[ClsLikeDefn] =
+    val sym = ClassSymbol(
+      Tree.TypeDef(syntax.Cls, Tree.Error(), N, N),
+      Tree.Ident("Cont$" + State.suid.nextUid)
+    )
+    val pcVar = VarSymbol(Tree.Ident("pc"))
+    sym.defn = S(ClassDef(N, syntax.Cls, sym, Nil, S(PlainParamList(Param(FldFlags.empty, pcVar, N) :: Nil)), ObjBody(Term.Blk(Nil, Term.Lit(Tree.UnitLit(true)))), List()))
```
LPTK: Line too long.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -0,0 +1,533 @@
...
+    
+    val resumeSym = BlockMemberSymbol("resume", List())
+    val resumeFnDef = FunDefn(
+      resumeSym,
+      List(PlainParamList(List(Param(FldFlags.empty, resumedVal, N)))),
+      resumeBody
+    )
+    
+    S(ClsLikeDefn(sym, syntax.Cls, S(contClsPath), resumeFnDef :: Nil, Nil, Nil,
+      Assign(freshTmp(), SimpleCall(Value.Ref(State.builtinOpsMap("super")), Value.Lit(Tree.UnitLit(true)) :: Value.Lit(Tree.UnitLit(true)) :: Nil), End()), End()))
```
LPTK: Line too long.

LPTK (changes requested):
Ok, please make the requested change with `BlockTransformer` and then we can merge this.

I didn't review the transformation's logic as it's quite complex and undocumented. After you write your FYP report you can import some of the explanations from it as code comments and also link to the document for other people to see.

LPTK:
I found some minor problems. I think we should
1. Forbid handlers in constructors.
2. For now, crash when an effect is not handled inside a constructor. A proper workaround implementation (such as using a separate factory function to run the ctor code) could be added later.

hkmc2/shared/src/test/mlscript/handlers/HandlerBlockBidnings.mls
```diff
@@ -0,0 +1,34 @@
...
+handle h = Effect with
+  fun perform(arg)(k) =
+    print(arg)
+let x = 123
+
+:fixme
+x
+//│ ╔══[ERROR] Name not found: x
+//│ ║  l.15: 	x
+//│ ╙──      	^
```
AnsonYeung: the rationale of this is that `x` may not be initialized if the handler is non-resumptive

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
AnsonYeung: It is the syntax `handle h = Foo(123)` as right now it is just calling it without any parameters. Should this syntax be also implemented inside this PR?

hkmc2/shared/src/test/mlscript/handlers/HandlerBlockBidnings.mls
LPTK: Ah, good point... For now it's fine like this, then. But later when we have optional bindings, we can mark it as optional.

AnsonYeung:
For class constructors, the old example code was due to the reliance of implicitly returning this in the constructor. I've modified it to explicitly return the correct `this` in the continuation class.

TODO:
- [x] Check `Instantiate` for effects

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
LPTK: Not necessarily. Better do it in a follow-up.

LPTK: But the comment could be clarified.

hkmc2/shared/src/test/mlscript/handlers/HandlersInClasses.mls
```diff
@@ -24,28 +25,12 @@ class Lol() with
   print(h.perform("k"))
   val x = 123
 
-//  FIXME should either return 'b' or crash, not return a partially-initialized instance of Lol
```
LPTK: Again, I think this should be a runtime crash. Or we should forbid handlers in ctors. It's obviously wrong to return an object instance that's only partly initialized.

That said, maybe we could catch this in the type system (checking that the handler is resumptive).

AnsonYeung: So the current implementation interprets this as same as
```
class Lol() with
  "b"
```
because handlers are handled in the function. Changes to elaboration is required to fix this if the intended behaviour is to break out of the constructor as consider the following currently equivalent code:
```
class Lol() with
  handle h = Effect with
    fun perform(arg)(k) =
      print(arg)
      "b"
  in
    print(h.perform("k"))
    val x = 123
  val y = 1234 // the constructor continues here
```

My point is that `x` is never supposed to be a field here even if the handler is resumptive. Having a field that may or may not exist depending on control flow seems weird (it could be something like a optional field that always exist instead).

btw, the problem is actually not related to handlers inherently. For example:
```
module M with
  if true then
    val a = 1
    ()
  else
    val b = 2
    ()
```

Now M.a always exist but M.b does not (in the generated JS code). Note that `true` can be replaced by anything. The exact problem occur in classes as well.

We should give the same treatment to if and handlers whatever semantics is desired. If we need to reject handlers in constructors we need to reject ifs in constructor for the same reason too and I don't think there is actually any difference here.

LPTK: > My point is that `x` is never supposed to be a field here even if the handler is resumptive.

I disagree. Because it is a `val` binding syntactically in the body of the class, it is a field of the class.

> Having a field that may or may not exist depending on control flow seems weird

This should never be the case.

> btw, the problem is actually not related to handlers inherently. For example:
> 
> ```
> module M with
>   if true then
>     val a = 1
>     ()
>   else
>     val b = 2
>     ()
> ```
> 
> Now M.a always exist but M.b does not (in the generated JS code). Note that `true` can be replaced by anything. The exact problem occur in classes as well.

Here `a` and `b` clearly aren't fields of `M`. They're just local bindings inside local blocks of the constructor. Avoiding this kind of confusion is actually one reason I've been considering disallowing `val` in local blocks.

Notice the error here:

```
object Foo with
  if true then
    val x = 1
    ()

Foo.x
//│ ╔══[ERROR] Object 'Foo' does not contain member 'x'
//│ ║  l.37: 	Foo.x
//│ ╙──      	   ^^
//│ = 1
```

The fact that it still compiles to a field and returns `1` at the end is an elaborator bug.

> we need to reject ifs in constructor for the same reason

Absolutely not!

To be clear, what I'm proposing is to reject `handle` statements (ie without an `in` body) in ctors.

LPTK:
Do you guys @CAG2Mark and @AnsonYeung intend to address the remaining issues quickly so we can finally merge this? I don't think you want to wait until after I merge the symbol changes (it would give you more work to fix things).

LPTK (approved):
All right, I think we can merge now. There are still unresolved things to address in future PRs, though. I'll open a tracking issue for them.

PR #259, opened 2025-01-07
Stack safe recursion

CAG2Mark:
Uses effect handlers to implement stack safe recursion.

Depends on https://github.com/hkust-taco/mlscript/pull/257, which should be merged first.

TODO:
- [x] Write more tests and check for bugs
- [x] `__Return` is currently leaked, which should be fixed
- [x] Don't rewrite trivial functions


LPTK:
### Notes from today's meeting

Example that does not work:

```fs
:stackSafe 100
:handler
fun foo(f) = print(f(f)) // or just `f(f)`
foo(foo)
//│     at foo (REPL11:1:60)
//│     at foo (REPL11:1:3053)
//│     at foo (REPL11:1:3053)
//│     at foo (REPL11:1:3053)
//│     at foo (REPL11:1:3053)
//│     at foo (REPL11:1:3053)
//│     at foo (REPL11:1:3053)
//│     at foo (REPL11:1:3053)
//│     at foo (REPL11:1:3053)
//│     at foo (REPL11:1:3053)
```

To preserve tail calls, we want to update the stack depth global variable on function calls:

```fs
let tmp = __stackDepth
set __stackDepth = tmp + 1
tailCall()
```

```fs
let tmp = __stackDepth
set __stackDepth = tmp + 1
nonTailCall()
__stackDepth = tmp
```

However, the `__stackDepth = tmp` will interact poorly with effect resumptions. How to deal with this?

CAG2Mark:
Note: seems I made some mistakes in the logic, will fix

CAG2Mark:
OK, ready for review.

I am considering rewriting `floutOutDefns` using by overriding `BlockTransformer` with a function which collects definitions using a mutable list or vector. Would mutability be acceptable here if it makes the logic much simpler?

CAG2Mark:
Need quite a lot of refactoring after the merge, re-marked as a draft. Hopefully will be ready by tomorrow.

CAG2Mark:
OK, took a bit of wrestling with git, for some reason it didn't want to accept that I had actually merged from hkmc2. But all looks good now so it should be ready for review.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -88,75 +88,74 @@ sealed abstract class Block extends Product with AutoLocated:
...
+    case Match(_, arms, dflt, rest) => arms.map(_._2) ++ dflt.toList :+ rest
+    case Begin(sub, rest) => sub :: rest :: Nil
+    case TryBlock(sub, finallyDo, rest) => sub :: finallyDo :: rest :: Nil
+    case Assign(_, rhs, rest) => rest :: Nil
+    case AssignField(_, _, rhs, rest) => rest :: Nil
+    case Define(_, rest) => rest :: Nil
+    case HandleBlock(_, _, _, _, handlers, body, rest) => handlers.map(_.body) :+ body :+ rest
+    case _: Return | _: Throw | _: Label | _: Break | _: Continue | _: End | _: HandleBlockReturn => Nil
+  
+  // Moves definitions in a block to the top. Only scans one definition deep, i.e. definitions inside other definitions
```
LPTK: I wouldn't call this "one definition deep" (which would imply you look into at most one definition).

More like

```suggestion
  // Moves definitions in a block to the top. Only scans the top-level definitions of the block; i.e., definitions inside other definitions
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -88,75 +88,74 @@ sealed abstract class Block extends Product with AutoLocated:
...
+  // are not moved out. Definitions inside `if` and `while` statements are moved out.
+  def floatOutDefns =
```
LPTK, line 108: Actually couldn't this be ported as a `BlockTransformer` with a `ListBuffer`? Seems that it would be much shorter (and more efficient!) that way.

hkmc2/shared/src/main/scala/hkmc2/codegen/StackSafeTransform.scala
```diff
@@ -1,14 +1,11 @@
 package hkmc2
 
 import mlscript.utils.*, shorthands.*
-import utils.*
```
LPTK: Let's not move this one. I like to order imports from most general to most specific, with each level of generality separated by an empty line.

Please disable automatic import reordering, if any.

hkmc2/shared/src/main/scala/hkmc2/codegen/StackSafeTransform.scala
```diff
@@ -120,10 +151,11 @@ class StackSafeTransform(depthLimit: Int)(using State):
     defn match
     case d: FunDefn => rewriteFn(d)
     case _: ValDefn => defn
-    case ClsLikeDefn(sym, k, parentPath, methods, privateFields, publicFields, preCtor, ctor) =>
+    case ClsLikeDefn(owner, isym, sym, k, paramsOpt, 
+      parentPath, methods, privateFields, publicFields, preCtor, ctor) =>
       ClsLikeDefn(
-        sym, k, parentPath, methods.map(rewriteFn), privateFields, publicFields,
-        rewriteBlk(preCtor), rewriteBlk(ctor) // TODO: do we need to rewrite the preCtor?
```
LPTK: > // TODO: do we need to rewrite the preCtor?

So I assume the answer was yes? Or was the comment lost in the merge?

hkmc2/shared/src/main/scala/hkmc2/codegen/StackSafeTransform.scala
```diff
@@ -0,0 +1,189 @@
...
+  private val stackOffsetPath: Path = State.globalThisSymbol.asPath.selN(Tree.Ident("Predef")).selN(STACK_OFFSET_IDENT)
+  private val stackHandlerPath: Path = State.globalThisSymbol.asPath.selN(Tree.Ident("Predef")).selN(STACK_HANDLER_IDENT)
+  private val predefPath: Path = State.globalThisSymbol.asPath.selN(Tree.Ident("Predef"))
+
+  private def intLit(n: BigInt) = Value.Lit(Tree.IntLit(n))
+  
+  private def op(op: String, a: Path, b: Path) =
+    Call(State.builtinOpsMap(op).asPath, List(a.asArg, b.asArg))(true)
+
+  // TODO: this code is copied from HandlerLowering and is quite useful. Maybe refactor it into a utils file
```
LPTK: Yes. Let's not add duplicated code, which will need twice the maintenance burden and may diverge.

LPTK (changes requested):
Very cool, thanks!

But I found some problems (added as a test file):

```fs
:js
:handler
:stackSafe 1000


fun selfApp(f) = f(f)

fun mkrec(g) =
  selfApp of self =>
    g of y => self(self)(y)

let fact = mkrec of self => x =>
  if x == 0 then 1 else self(x - 1) * x
//│ fact = [Function (anonymous)]

fact(3)
//│ = 6

fact(10)
//│ = 3628800

:fixme
fact(5000)


// * Interestingly, this variation works:

fun mkrec(g) =
  selfApp of self =>
    g of y => selfApp(self)(y)

let fact = mkrec of self => x =>
  if x == 0 then 1 else self(x - 1) * x
//│ fact = [Function (anonymous)]

// * Without `:stackSafe`, gives `RangeError: Maximum call stack size exceeded`
fact(5000)
//│ = Infinity


// * FIXME: changing the instrumentation's stack depth shouldn't change the result!

:stackSafe 100
set selfApp = f => f(f)

fact(5000)
//│ = 9.33262154439441e+157


:stackSafe 10
set selfApp = f => f(f)

fact(5000)
//│ = 3628800
```

CAG2Mark:
It seems that the last remaining problem is a bug with the instrumentation itself.

Since fixing that issue may take some time, should we just merge this anyway once the other comments are dealt with? This implementation of stack safety is orthogonal to the actual instrumentation anyway.

AnsonYeung:
The problem in the instrumentation is dealt with and tests look fine now.

hkmc2/shared/src/test/mlscript/handlers/ZCombinator.mls
```diff
@@ -18,40 +18,56 @@ fact(3)
...
+fact(10000)
 //│ = Infinity
 
+:re
+:stackSafe off
+fact(10000)
+//│ ═══[RUNTIME ERROR] RangeError: Maximum call stack size exceeded
+
 
 // * FIXME: changing the instrumentation's stack depth shouldn't change the result!
```
LPTK: Remove

hkmc2/shared/src/main/scala/hkmc2/codegen/StackSafeTransform.scala
CAG2Mark: Assuming the preCtor can have arbitrary code, yes

CAG2Mark:
Should be good to go now

- [x] EDIT: actually there's one small change I want to make to make the tests a bit more consistent

hkmc2/shared/src/main/scala/hkmc2/MLsCompiler.scala
```diff
@@ -74,7 +74,7 @@ class MLsCompiler(preludeFile: os.Path):
       val parsed = mainParse.resultBlk
       val (blk, newCtx) = elab.importFrom(parsed)
       val low = ltl.givenIn:
-        codegen.Lowering(lowerHandlers = false)
+        codegen.Lowering(lowerHandlers = false, None) // TODO: properly hook up stack limit
```
LPTK: When passing in some generic constant, it's better to name it.
```suggestion
        codegen.Lowering(lowerHandlers = false, stackLimit = None) // TODO: properly hook up stack limit
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -99,63 +100,22 @@ sealed abstract class Block extends Product with AutoLocated:
     case HandleBlock(_, _, _, _, handlers, body, rest) => handlers.map(_.body) :+ body :+ rest
     case _: Return | _: Throw | _: Label | _: Break | _: Continue | _: End | _: HandleBlockReturn => Nil
   
+  // Moves definitions in a block to the top. Only scans the top-level definitions of the block;
+  // i.e, definitions inside other definitions are not moved out. Definitions inside `if` and
```
```suggestion
  // i.e, definitions inside other definitions are not moved out. Definitions inside `match`/`if` and 
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -99,63 +100,22 @@ sealed abstract class Block extends Product with AutoLocated:
...
+    val transformer = new BlockTransformerShallow(SymbolSubst()):
+      override def applyBlock(b: Block): Block = b match
+        case Define(defn, rest) => defn match
+          case v: ValDefn => super.applyBlock(b)
+          case _ =>
+            defns.addOne(defn)
+            applyBlock(rest)
+        case _ => super.applyBlock(b)
+    
+    (transformer.applyBlock(this), defns.reverse.toList)
```
LPTK: If you're going to reverse it at the end, there's no need to use `ListBuffer`. Just build the reversed list directly!
```suggestion
    var defns: List[Defn] = Nil
    val transformer = new BlockTransformerShallow(SymbolSubst()):
      override def applyBlock(b: Block): Block = b match
        case Define(defn, rest) => defn match
          case v: ValDefn => super.applyBlock(b)
          case _ =>
            defns ::= defn
            applyBlock(rest)
        case _ => super.applyBlock(b)
    
    (transformer.applyBlock(this), defns.toList)
```

LPTK: I do wonder why you reverse the list in the first place, though. A comment explaining why could help.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -288,24 +288,24 @@ class HandlerLowering(using TL, Raise, Elaborator.State, Elaborator.Ctx):
...
           val rest = applyBlock(b.rest)
           translateHandleBlock(b.copy(rest = rest))
-        case Return(c: Call, implct) if handlerCtx.isHandleFree =>
-          val fun2 = applyPath(c.fun)
-          val args2 = c.args.map(applyArg)
-          val c2 = if (fun2 is c.fun) && (args2 zip c.args).forall(_ is _) then c else Call(fun2, args2)(c.isMlsFun)
+        case Return(c @ Call(fun, args), implct) if handlerCtx.isHandleFree && !handlerCtx.isTopLevel =>
+          val fun2 = applyPath(fun)
+          val args2 = args.map(applyArg)
+          val c2 = if (fun2 is fun) && (args2 zip args).forall(_ is _) then c else Call(fun2, args2)(c.isMlsFun)
```
```suggestion
          val args2 = args.mapConserve(applyArg)
          val c2 = if (fun2 is fun) && (args2 is args) then c else Call(fun2, args2)(c.isMlsFun)
```

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -288,24 +288,24 @@ class HandlerLowering(using TL, Raise, Elaborator.State, Elaborator.Ctx):
         case b: HandleBlock =>
           val rest = applyBlock(b.rest)
           translateHandleBlock(b.copy(rest = rest))
-        case Return(c: Call, implct) if handlerCtx.isHandleFree =>
-          val fun2 = applyPath(c.fun)
-          val args2 = c.args.map(applyArg)
-          val c2 = if (fun2 is c.fun) && (args2 zip c.args).forall(_ is _) then c else Call(fun2, args2)(c.isMlsFun)
+        case Return(c @ Call(fun, args), implct) if handlerCtx.isHandleFree && !handlerCtx.isTopLevel =>
```
LPTK: Why do we special-case `!handlerCtx.isTopLevel`? This needs a comment.

I'd also like some doc, in the definition of `HandlerCtx`, on what each parameter of  mean, especially (but not only)  `isHandleFree` and `isTopLevel`.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -288,24 +288,24 @@ class HandlerLowering(using TL, Raise, Elaborator.State, Elaborator.Ctx):
...
+          val c2 = if (fun2 is fun) && (args2 zip args).forall(_ is _) then c else Call(fun2, args2)(c.isMlsFun)
           if c2 is c then b else Return(c2, implct)
         case _ => super.applyBlock(b)
       override def applyResult2(r: Result)(k: Result => Block): Block = r match
         case r @ Call(Value.Ref(_: BuiltinSymbol), _) => super.applyResult2(r)(k)
         case c @ Call(fun, args) =>
           val res = freshTmp("res")
           val fun2 = applyPath(fun)
-          val args2 = c.args.map(applyArg)
+          val args2 = args.map(applyArg)
```
LPTK: Please apply the same change.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -288,24 +288,24 @@ class HandlerLowering(using TL, Raise, Elaborator.State, Elaborator.Ctx):
...
           val c2 = if (fun2 is fun) && (args2 zip args).forall(_ is _) then c else Call(fun2, args2)(c.isMlsFun)
           ResultPlaceholder(res, freshId(), false, c2, k(Value.Ref(res)))
         case c @ Instantiate(cls, args) =>
           val res = freshTmp("res")
           val cls2 = applyPath(cls)
-          val args2 = c.args.map(applyPath)
+          val args2 = args.map(applyPath)
```
LPTK: Please apply the same change.

hkmc2/shared/src/main/scala/hkmc2/codegen/StackSafeTransform.scala
```diff
@@ -0,0 +1,169 @@
...
+  private val stackDelayClsPath: Path = predefPath.selN(Tree.Ident("__StackDelay")).selN(Tree.Ident("class"))
+  private val stackLimitPath: Path = predefPath.selN(STACK_LIMIT_IDENT)
+  private val stackDepthPath: Path = predefPath.selN(STACK_DEPTH_IDENT)
+  private val stackOffsetPath: Path = predefPath.selN(STACK_OFFSET_IDENT)
+  private val stackHandlerPath: Path = predefPath.selN(STACK_HANDLER_IDENT)
+
+  private def intLit(n: BigInt) = Value.Lit(Tree.IntLit(n))
+  
+  private def op(op: String, a: Path, b: Path) =
+    Call(State.builtinOpsMap(op).asPath, List(a.asArg, b.asArg))(true)
```
LPTK: I normally try to avoid `List.apply` and similar vararg-taking constructor functions, as calling them will always allocate an extra array at runtime. It's better to use `Nil` (or `List.empty`) and `::`.

hkmc2/shared/src/main/scala/hkmc2/codegen/StackSafeTransform.scala
```diff
@@ -0,0 +1,169 @@
...
+  def rewriteBlk(blk: Block) =
+    val newBody = transform(blk)
+
+    if isTrivial(blk) then 
+      newBody
+    else
+      val diffSym = TempSymbol(None, "diff")
+      val scrut1Sym = TempSymbol(None, "scrut1")
+      val scrut2Sym = TempSymbol(None, "scrut2")
+      val scrutSym = TempSymbol(None, "scrut")
```
LPTK: How about you give these slightly more informative names? "scrut" is inappropriate as this code is not really doing pattern matching.

hkmc2/shared/src/main/scala/hkmc2/codegen/StackSafeTransform.scala
```diff
@@ -0,0 +1,169 @@
...
+      val scrutVal = op("&&", scrut1Sym.asPath, scrut2Sym.asPath)
+      blockBuilder
+        .assign(diffSym, diff)        // diff = stackDepth - stackOffset
+        .assign(scrut1Sym, scrut1)    // diff >= depthLimit
+        .assign(scrut2Sym, scrut2)    // stackHandler !== null
+        .assign(scrutSym, scrutVal)   // diff >= depthLimit && stackHandler !== null
+        .ifthen(
+          scrutSym.asPath, Case.Lit(Tree.BoolLit(true)), 
+          blockBuilder.assign( // tmp = perform(undefined)
+            TempSymbol(None, "tmp"),
```
LPTK: Better name?

LPTK (approved):
LGTM! Please address the comments and I'll make the merge!

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
AnsonYeung: actually more accurately, it should have matched implct = False here. it is currently inconsistent that we throw Error("unhandled effect") on all calls on the top level except the last one which is implicitly returned.

AnsonYeung: I've added doc for HandlerCtx

PR #260, opened 2025-01-08
Add disjointness reasoning for BbML

NeilKleistGao:
- Track outer regions for disjointness reasoning
- Support `outer` variables for <del>functions</del> polymorphic types
- Fix some pretty-print bugs
- Fix negative type elaboration
- Refactor type variable collection logic
- Fix some missing cases

<del>_This PR is still being prepared, e.g., containing redundant code. Such problems will be addressed after the main problems are discussed in the meeting._</del>


hkmc2/shared/src/test/mlscript/bbml/bbDisjoint.mls
```diff
@@ -113,29 +113,76 @@ region x in
...
 
 
-fun annohelper: Region[outer] ->{outer} @outer Pair[out Ref[Int, out outer], out Ref[Int, out Any]]
+fun annohelper: [outer, T extends outer] -> Region[T] ->{T} Pair[out Ref[Int, out T], out Ref[Int, out Any]]
 fun annohelper(r1) =
   region r2 in
     fork((_ => r1.ref 1), (_ => r2.ref 2))
 annohelper
-//│ Type: (Region[envouter]) ->{envouter} Pair[out Ref[Int, out envouter], out Ref[Int, ?]]
+//│ Type: forall 'T: (Region['T]) ->{'T} Pair[out Ref[Int, out 'T], out Ref[Int, ?]]
```
LPTK: The `outer` binding is missing from this signature.

hkmc2/shared/src/test/mlscript/bbml/bbDisjoint.mls
```diff
@@ -113,29 +113,76 @@ region x in
...
-//│ ╔══[ERROR] Type error in integer literal with expected type ⊥
-//│ ║  l.121: 	42: @outer Int
-//│ ║         	^^
-//│ ╙── because: cannot constrain  Int  <:  ⊥
-//│ Type: ⊥
+//│ Type: ⊤
 
 
-fun annohelper: Region[outer] ->{outer} @outer Pair[out Ref[Int, out outer], out Ref[Int, out Any]]
+fun annohelper: [outer, T extends outer] -> Region[T] ->{T} Pair[out Ref[Int, out T], out Ref[Int, out Any]]
```
LPTK: A more natural signature could be
```suggestion
fun annohelper: [outer, T] -> Region[T] ->{T | outer} Pair[out Ref[Int, out T], out Ref[Int, out outer]]
```
Does it work? If it does, it would be good to add it to the test file (in addition to the existing one).

LPTK: Why do we have `Any` in `Ref[Int, out Any]`?

hkmc2/shared/src/test/mlscript/bbml/bbRef.mls
```diff
@@ -36,7 +36,7 @@ let r = region x in x.ref 42
 
 fun mkRef() = region x in x.ref 42
 mkRef
-//│ Type: () -> Ref[Int, out ¬envmkRef]
+//│ Type: forall : () -> Ref[Int, out ¬env@mkRef]
```
LPTK: `forall : ` 👀

hkmc2/shared/src/test/mlscript/bbml/bbDisjoint.mls
NeilKleistGao: No, it doesn't. We can only track the evidence that `r2 <: ~outer`, but not `r2 <: ~T`. So this code will be rejected.

LPTK:
BTW, since you are modifying the way forall types are printed. Why not print them like in the source? Eg `[A, B, outer C] -> ...`

hkmc2/shared/src/test/mlscript/bbml/bbDisjoint.mls
NeilKleistGao: Because the second reference returned by `fork` is located in `r2`. It is extruded after we leave the scope of `r2`. But it can be written as `out ~outer`.

LPTK: Yes, `~outer` is what I meant in my other comment.

LPTK: Why would we need `r2 <: ~T` here?

NeilKleistGao: The disjointness requirement is `(T | outer) & r2 <: bot`, i.e., `(T & r2) | (outer & r2) <: bot`. Then only the second subterm `(outer & r2) <: bot` can be solved. The equivalent one should be `T & outer` instead.

LPTK: Right, right, ofc.

hkmc2/shared/src/test/mlscript/bbml/bbBasics.mls
```diff
@@ -20,7 +20,7 @@ false
 
 fun id(x) = x
 id
-//│ Type: forall outer, 'x: 'x -> 'x
+//│ Type: [outer, 'x] -> 'x -> 'x
```
LPTK: Can't the `outer` type be omitted here?

NeilKleistGao: Good suggestion. I also fixed the problem with simplification of higher-ranked polymorphic function types.

hkmc2/shared/src/test/mlscript/bbml/bbCheck.mls
```diff
@@ -187,3 +187,42 @@ foo(42, 1)
...
+
+
+fun f: ([A] -> ([B] -> A | B -> B) -> A) -> ([A] -> ([B] -> A | B -> B) -> A)
+f
+//│ Type: ((['B] -> ('B) ->{⊥} 'B) ->{⊥} ⊥) ->{⊥} (['B] -> ('B) ->{⊥} 'B) ->{⊥} ⊥
+
+
+fun f: ([A] -> ([B] -> A | B -> B | A) -> A) -> ([A] -> ([B] -> A | B -> B | A) -> A)
+f
+//│ Type: (['A] -> (['B] -> ('A ∨ 'B) ->{⊥} 'B ∨ 'A) ->{⊥} 'A) ->{⊥} ['A] -> (['B] -> ('A ∨ 'B) ->{⊥} 'B ∨ 'A) ->{⊥} 'A
```
LPTK, line 218: Can we please use `|` and `&` rather than `∨` and `∧`?

hkmc2/shared/src/main/scala/hkmc2/bbml/bbML.scala
```diff
@@ -57,27 +67,38 @@ object BbCtx:
...
+  private def freshReg(sym: Symbol)(using ctx: BbCtx) =
+    val state = new VarState()
+    state.upperBounds = ctx.getRegEnv.! :: Nil
+    InfVar(ctx.lvl + 1, infVarState.nextUid, state, true)(sym, "")
+  private def freshOuter(sym: Symbol)(using ctx: BbCtx): InfVar =
+    InfVar(ctx.lvl + 1, infVarState.nextUid, new VarState(), true)(sym, "")
+  private def freshEnv(sym: Symbol)(using ctx: BbCtx): InfVar =
+    val state = new VarState()
+    state.upperBounds = ctx.getRegEnv :: Nil
+    state.lowerBounds = ctx.getRegEnv :: Nil
```
LPTK, line 100: I wonder what's the point of creating a new TV that's equivalent to an underlying type `ctx.getRegEnv`. I see `freshEnv` is only used in `instantiate`. Why can't `instantiate` substitute simple types for variables? Instead of just variables for variables?

hkmc2/shared/src/test/mlscript/bbml/bbCheck.mls
LPTK: This can be done in a future PR, though.

LPTK (approved):
LGTM, thanks!

I've left some comments that can be addressed later. Let's merge this now!

PR #262, opened 2025-01-10
Preliminary implementation of refining patterns

chengluyu:
This PR implements the feature of defining recursive pattern synonyms and compiling them into UCS splits at the use site.  The syntax for defining patterns and type aliases is similar except that patterns can be used in pattern matching.

```
pattern Bit = 0 | 1
pattern BitSeq = null | Pair(Bit, BitSeq)
```

The declarations above defines a sequence made of either 0 or 1 using `Pair`, with the empty sequence represented by `null`. Users can check whether a value has the defined shape simply by matching against the synonym.

```
if Pair(0, Pair(1, Pair(0, null))) is BitSeq then "yes" else "no" // => "yes"
```


hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1069,4 +1086,3 @@ end Elaborator
 
 type Pol = Opt[Bool]
 extension (p: Pol) def ! : Pol = p.map(!_)
-
```
LPTK: ?

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -192,11 +194,15 @@ class TypeAliasSymbol(val id: Tree.Ident)(using State) extends MemberSymbol[Type
...
   override def toString: Str = s"module:${id.name}${State.dbgUid(uid)}"
 
-class PatternSymbol(val id: Tree.Ident)(using State)
+class PatternSymbol(val id: Tree.Ident, val params: Opt[Tree.Tup], val body: Tree)(using State)
     extends MemberSymbol[PatternDef] with CtorSymbol with InnerSymbol:
   def nme = id.name
   def toLoc: Option[Loc] = id.toLoc // TODO track source tree of pattern here
   override def toString: Str = s"pattern:${id.name}"
+  /** Compute the arity. */
+  def arity: Int = params.fold(0)(_.fields.length)
```
LPTK: Looks like this does not take spreads into account. You shouldn't just take the length of `fields`. If there's a spread and you don't support it, at least crash the compiler.

hkmc2/shared/src/test/mlscript/rp/recursion/BitTree.mls
```diff
@@ -0,0 +1,49 @@
...
+//│ = true
+:todo
+
+Pair(0, null) is BinTree
+//│ = true
+:todo
+
+Pair(0, Pair(0, null)) is BinTree
+//│ = true
+:todo
```
LPTK: ?

hkmc2/shared/src/test/mlscript/rp/parametric/HighOrderPattern.mls
LPTK, line 1: It's high**er** order pattern.

hkmc2/shared/src/test/mlscript/rp/recursion/BitTree.mls
LPTK: Looks like my old comment was on a test line that was removed, so it can be resolved.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
chengluyu: Thanks for pointing this out. I also realized that computing arity is not that straightforward as this. I no longer use this field in the latest implementation. I will consider the case of `Spread` next up.

chengluyu:
### What's Missing?

From a high-level perspective, the PR still lacks the following features.

- Extractions. Extract values from the sub-patterns in the pattern body and bind them to the parameters of the pattern.
- Case analysis and computation. Use `case` expressions as the pattern body and perform certain computations in each case. 
- Detection of non-determinism. Specifically, reject and report merges on the same constructor coming from different pattern expansions.

The following features have not been completed yet.

- The specialization of string joining is not implemented in the pattern compiler, and the whole pattern compiler currently does not support string matching at all.
- The pattern translator does not yet support pattern parameters and lacks related tests.
- The specialization for tuple patterns has not been implemented in both the compiler and the translator.

LPTK:
Great! Don't forget to mark the PR ready after merging with the main branch. We don't have to have all features in this PR (but it would be good to fail gracefully on missing ones).

LPTK:
It looks like you forgot to rerun the tests.

LPTK:
You forgot to rerun the tests again 🌚

chengluyu:
Sorry, I merged it on my phone with the web editor. I haven't had the chance to pull it down and run a test yet.

LPTK (approved):
Ok, I haven't reviewed the entire change as there's a lot of code and little time, but let's merge now. Thanks!

PR #265, opened 2025-01-26
Handler and Stack Safety Fixes

CAG2Mark:
# Handler Fixes
When a handler calls `resume`, the handler is prepended to tail list. Prior to this fix, if the handler calls other effect first, this prepend is not done correctly. This is the root cause of the test (https://github.com/hkust-taco/mlscript/pull/259/commits/ea340666c7f56a74ade878314ebda52e90c0eb92).

Handler may result in the following 3 states when it returns:
- Finished (this means no effect is raised and it is not resuming anything)
- Tail call into either resume or some other effect raising function.
- Normal call into either resume or some other effect raising function.

Right now normal call into effect raising function, and resuming afterwards will cause a bug due to append code is hardcoded in the continuation and the fixup code after `cur.handleFun` in `Predef.__handleEffect` is not present for continuation. The fix implements another function `appendInCont` which decide whether it should prepend or append depending on the type of tail.

# Stack Safety Fixes
The variable `__stackOffset` was not being properly reset in many cases which has now been fixed. Also added a test to demonstrate the original problem and the fix.

Handlers are now:
```fs
handle h = __StackDelay 
  fun perform()(resume) =
    // no longer needed: let curOffset = __stackOffset
    __stackOffset = __stackDepth
    let ret = resume()
    // incorrect line: set __stackOffset = curOffset
    ret // prevent tail-resumptive call optimization later
```

Non-tail-calls are now:
```fs
let prevDepth = __stackDepth
set __stackDepth += 1
nonTailCall()
set __stackDepth = prevDepth
// correct way to reset __stackOffset
if prevDepth < __stackOffset then
  set __stackOffset = prevDepth
```


hkmc2/shared/src/main/scala/hkmc2/codegen/StackSafeTransform.scala
```diff
@@ -35,20 +35,26 @@ class StackSafeTransform(depthLimit: Int)(using State):
...
         .assign(prevDepth, stackDepthPath)
         .assignFieldN(predefPath, STACK_DEPTH_IDENT, op("+", stackDepthPath, intLit(1)))
         .assign(tmp, res)
         .assignFieldN(predefPath, STACK_DEPTH_IDENT, prevDepth.asPath)
+        .assign(offsetGtDepth, op("<", prevDepth.asPath, stackOffsetPath))
+        .ifthen(
+          offsetGtDepth.asPath, 
+          Case.Lit(Tree.BoolLit(true)), 
+          blockBuilder.assignFieldN(predefPath, STACK_OFFSET_IDENT, prevDepth.asPath).end
+          )
```
```suggestion
        )
```

LPTK (changes requested):
Thanks for the fixes! And sorry that I still cannot check the soundness of your changes since your logic is still unexplained/undocumented (as expected; the explanation and doc will come with the FYP report).

I'm a bit worried about your pseudocode here:
```fs
handle h = __StackDelay 
  fun perform()(resume) =
    // no longer needed: let curOffset = __stackOffset
    __stackOffset = __stackDepth
    let ret = resume()
    // incorrect line: set __stackOffset = curOffset
    ret // prevent tail-resumptive call optimization later
```
This does not look like a robust thing to do. Eta expansion and contraction of let bindings should not affect the semantics, and one would hope that the compiler would be free to turn any `let x = f() in x` into a tail call. If you require tail calls in handlers to be handled specially, you should do that properly. Or just add a call to some dummy `nonTail` identity function (defined in `Predef`) to block the tail call. We'd later mark this as `@noinline`.

Or did I misunderstand what you meant in the comments?

CAG2Mark:
Maybe I worded it badly. I don't want this effect to be tail-resumptive so
that it doesn't get optimized into just a virtual function call later.

LPTK:
Yeah, well, I think we should introduce a `@nonTailResumptive` annotation and use that on `fun perform`. Then later, when we implement the optimizations, we can make sure the annotation is respected. I updated the tracking issue so this can be merged now.

LPTK (approved):

PR #266, opened 2025-01-27
Class Lifter

CAG2Mark:
Lifts classes and nested functions to the top-level.

## Method
General idea: when functions have to capture local variables, we need to rewrite their parameter lists. For example:
```fs
fun f(arg) =
  let unused = 0
  fun g(arg2) = set arg1 = 1
  let foo = g
  foo(unused)
  arg
```
is transformed into
```fs
fun g(f_capture)(arg2) =
  set f_capture.arg1 = 1 
class f$capture(arg1)
fun f(arg) =
  let capture = f$capture(arg)
  let foo = g(capture)
  foo(unused)
  arg
```
Variables which are not mutated will be passed as a parameter directly.
## TODO
- [x] Lift functions.
- [x] Lift classes:
  - [x] `Instantiate` does not work at all
  - [ ] Functions within functions within classes are lifted out of the class. This is technically fine, but it's better to have them as class methods. However, we need private methods for this, which is not yet implemented. 
  - [ ] Private variables can't be accessed by nested classes/functions after being lifted out.
  - [x] ~~Modules are not lifted properly~~ For now we don't lift them at all
- [x] Only pass mutated variables through the capture class. Move other variables by passing them as an argument directly. Note that there are some subtle cases where this causes problems, such as
  ```fs
  fun f() =
    let x = g
    let y = 2
    fun g() = y
    x()
  ```
  so for now, we only do this when the function is not passed around as a first-class object. More thorough analysis could be done in another PR.
  - [x] More advanced analysis is work-in-progress.
- [x] Replace calls with a single call where we can. For example
  ```fs
  fun f(arg) =
    let unused = 0
    fun g(arg2) = set arg1 = 1
    g(unused)
    arg
  ```
  could be rewritten as
  ```fs
  fun g(f_capture, arg2) =
    set f_capture.arg1 = 1
  fun g$(f_capture)(arg2) = g(f_capture, arg2)
  class f$capture(arg1)
  fun f(arg) =
    let capture = f$capture(arg)
    g(capture, unused)
    arg
  ```
  where the call `g(unused)` is *symbolically* replaced.
  ("lifting" curried functions is much easier than the lifting in general, so it can be done later)
- [ ] https://github.com/hkust-taco/mlscript/pull/266#issuecomment-2639361716

## Auxiliary Parameters
Classes can now have auxiliary parameters (IR only currently) with curried constructors.


LPTK:
What's the intended scheme regarding mutable local variables?

FWIW, we should distinguish mutable from immutable variables. I intend to let `mut` be used to annotate pattern variables, like in Rust, and to reject re-assignments to non-`mut` variables. Note that this would still leave `let x; if b do x = 1` well-formed.

When there are only immutable variables being captured, the scheme can be simpler.

CAG2Mark:
> What's the intended scheme regarding mutable local variables?
> 
> FWIW, we should distinguish mutable from immutable variables. I intend to let `mut` be used to annotate pattern variables, like in Rust, and to reject re-assignments to non-`mut` variables. Note that this would still leave `let x; if b do x = 1` well-formed.
> 
> When there are only immutable variables being captured, the scheme can be simpler.

I don't distinguish between local variables that are mutated / not mutated for now. I intend to move every local variable that's referenced inside a nested class / function into a closure.

Would it be better to add the immutable local variables to each class/function directly?

P.S. seems I accidentally marked this PR as ready earlier, but it's definitely not. It's now marked as a draft.

CAG2Mark:
Added much more into to the the first comment.

CAG2Mark:
A lot of progress. See the top comment

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -314,11 +314,12 @@ class Lifter(using State):
     val reqdCaptures: List[BlockMemberSymbol],
     val reqdVars: List[Local],
     val reqdInnerSyms: List[InnerSymbol],
-    val fakeCtorBms: Option[BlockMemberSymbol] // only for classes
+    val fakeCtorBms: Option[BlockMemberSymbol], // only for classes
+    val singleCallBms: BlockMemberSymbol // optimization
```
```suggestion
    val singleCallBms: BlockMemberSymbol, // optimization
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -105,7 +131,7 @@ sealed abstract class Block extends Product with AutoLocated:
     case Throw(r) => r.subBlocks
     
     case _: Return | _: Throw | _: Label | _: Break | _: Continue | _: End | _: HandleBlockReturn => Nil
-  
+
```
LPTK: Let's not add unnecessary space changes.

hkmc2/jvm/src/test/scala/hkmc2/CompileTestRunner.scala
```diff
@@ -54,4 +54,3 @@ class CompileTestRunner
       
 end CompileTestRunner
 
-
```
LPTK: Let's not add unnecessary space changes.

LPTK:
As discussed in the meeting, it would be nice to emit a warning if lifted classes are used in instance tests or as first-class values, which is not supported yet.

And please add this test to the PR:
```fs
fun foo1(x, n) =
  class C()
  print(x is C)
  if n > 0 do foo1(C(), n - 1)

:lift
fun foo2(x, n) =
  class C()
  print(x is C)
  if n > 0 do foo2(C(), n - 1)

foo1(0, 1)
//│ > false
//│ > false

foo2(0, 1)
//│ > false
//│ > true
```

hkmc2/shared/src/test/mlscript/lifter/ClassInFun.mls
```diff
@@ -115,6 +115,21 @@ fun f(used1, unused1) =
...
+  fun g(g_arg) =
+    let used3 = 2
+    fun h = used3
+    used1 + h
+  let unused2 = 2
+  class Test with
+    fun get() = used1
+  new Test
+f(1, 2).get()
+//│ ═══[WARNING] Cannot yet lift the class `Test` as it is used as a higher-order class.
```
LPTK: Is it? 🤔

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1259 @@
...
+          r
+        case Instantiate(Select(Value.Ref(_: BlockMemberSymbol), Tree.Ident("class")), args) =>
+          args.map(applyPath)
+          r
+        case _ => super.applyResult(r)
+
+      override def applyValue(v: Value): Value = v match
+        case RefOfBms(l) if clsSyms.contains(l) && !l.isModule =>
+          raise(WarningReport(
+            msg"Cannot yet lift the class `${l.nme}` as it is used as a higher-order class." -> N :: Nil,
```
LPTK: I guess you mean
```suggestion
            msg"Cannot yet lift the class `${l.nme}` as it is used as a first-class class." -> N :: Nil,
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -99,7 +103,16 @@ object Lifter:
...
+  
+  object InstSel:
+    def unapply(p: Path) = p match
+      case Value.Ref(l: BlockMemberSymbol) => S(l)
+      case s @ Select(Value.Ref(l: BlockMemberSymbol), Tree.Ident("class")) => S(l)
+      case _ => N
 
+  def modOrObj(d: Defn) = d match
+    case c: ClsLikeDefn => (c.k is syntax.Mod) || (c.k is syntax.Obj)
+    case _ => false
```
LPTK, line 113: You're missing an empty line:
```suggestion
    case _ => false

```

hkmc2/shared/src/test/mlscript/codegen/BadNew.mls
```diff
@@ -11,14 +11,14 @@ new
 new 2
 //│ ═══[RUNTIME ERROR] TypeError: 2 is not a constructor
 
+// TODO: somehow desugaring lambdas fixed this, is this intended?
```
LPTK: The problem was that `new (...args) => ...` in JS is not valid syntax. But `new lambda` is fine. The fix would be to add parentheses around such new-ed arguments. But since lambdas should no longer be values, it probably doesn't matter much.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -312,56 +311,77 @@ class Lifter(using State, Raise):
...
           r
         case Instantiate(InstSel(_), args) =>
           args.map(applyPath)
           r
-        
+
         case _ => super.applyResult(r)
 
-      // don't search within `extends`
+      // don't search within `extends`, otherwise it'll think it's used as a first-class class
```
LPTK: But it could be used as a genuine first-class class in an `extends` clause.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -312,56 +311,77 @@ class Lifter(using State, Raise):
...
+            case Some(value) if !liftable.contains(value) =>
               raise(WarningReport(
                 msg"Cannot yet lift the class/module `${value.nme}` as it is used in an instance check." -> N :: Nil,
                 N, Diagnostic.Source.Compilation
               ))
               unliftable += value
+            case _ => ()
           case _ => ()
         cse
-
```
LPTK: Let's not

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: Ah never mind I guess extends clauses in the IR are paths.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -885,27 +959,68 @@ class Lifter(using State, Raise):
         super.applyBlock(ret)
     transformer.applyBlock(b)
 
+  // lift things out of top-level modules. this is easy
```
LPTK: Is this really necessary, though?

CAG2Mark: no it's not :P removed

hkmc2/shared/src/test/mlscript/codegen/ReboundLet.mls
```diff
@@ -15,7 +15,7 @@ foo()
 
 let x = 1
 let f = () => x
-//│ f = [function f]
+//│ f = [function lambda]
```
CAG2Mark: Hm. Did this actually rewrite it as a proper function before? Or is this just due to the way it's printed?

CAG2Mark:
OK, I think it's ready for review.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -307,16 +353,28 @@ sealed abstract class Result:
     case Value.Lam(params, body) => body :: Nil
     case Value.Arr(elems) => elems.flatMap(_.value.subBlocks)
     case _ => Nil
-
```
LPTK, line 354: 👎

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockTransformer.scala
```diff
@@ -238,3 +244,28 @@ class BlockTransformerShallow(subst: SymbolSubst) extends BlockTransformer(subst
         then b else HandleBlock(l2, res2, par2, args2, cls2, hdr2, bod, rst2)
     case _ => super.applyBlock(b)
 
+// does not traverse into any other block
```
LPTK: The semantics of this transformer seems broken wrt `Begin`. If you have `Begin{foo()}; bar()` you're not going to change `foo` even though you would have for the equivalent `foo(); bar()`.

hkmc2/shared/src/test/mlscript/codegen/Getters.mls
```diff
@@ -276,7 +277,9 @@ class Foo(x) with
   fun oops = x
 //│ JS (unsanitized):
 //│ let Foo1;
-//│ Foo1 = function Foo(x1) { return new Foo.class(x1); };
+//│ Foo1 = function Foo(x1) {
+//│   return new Foo.class(x1);
+//│ };
```
LPTK, line 282: Any idea why there are no longer on one line?

hkmc2/shared/src/test/mlscript/codegen/ReboundLet.mls
LPTK: What do you mean? The name is taken from the name of the binding, because we generate things like `f = function f ...`.

hkmc2/shared/src/main/scala/hkmc2/Config.scala
```diff
@@ -24,6 +25,7 @@ object Config:
     sanityChecks = N, // TODO make the default S
     // sanityChecks = S(SanityChecks(light = true)),
     effectHandlers = N,
+    liftDefns = N
```
```suggestion
    liftDefns = N,
```

hkmc2/shared/src/test/mlscript/basics/BadMemberProjections.mls
```diff
@@ -10,27 +10,32 @@
...
+//│   globalThis.Predef.checkArgs("lambda", 1, false, args1.length);
 //│   let self = args1[0];
 //│   let args = globalThis.Predef.tupleSlice(args1, 1, 0);
 //│   return runtime.safeCall(self.x(...args))
-//│ })());
+//│ };
+//│ block$res2 = runtime.checkCall(lambda1());
 //│ undefined
-//│ ═══[RUNTIME ERROR] Error: Function expected at least 1 argument but got 0
+//│ ═══[RUNTIME ERROR] Error: Function 'lambda' expected at least 1 argument but got 0
```
LPTK: It would be good if we could avoid making every lambda get the name "lambda", at least in such error messages. Ideally they'd be unnamed and we would still have the old message here.

LPTK:
> OK, I think it's ready for review.

Thanks, I'll try to properly review it tomorrow. In the meantime please fix the conflicts and comments.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockTransformer.scala
CAG2Mark: The job of this transformer is just to modify the paths and results in a block which do not require traversing another block to reach -- it's the job of the person using this transformer to traverse the sub-blocks, probably used in combination with one of the other block transformers.

As for `Begin`, both of its sub-blocks are expected to be traversed by whoever is using this transformer, not the transformer itself.

hkmc2/shared/src/test/mlscript/codegen/Getters.mls
CAG2Mark: When there's parameter lists, we need to curry the ctor using lambdas, and it's quite ugly to have it all on one line. I decided it was not worth special-casing non-curried ctors to be on one line such it's purely a superficial change.

LPTK: Oh I see, I didn't realize the line was previously missing the breaks.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockTransformer.scala
LPTK: Ok, then please pick a better name for this class and document in the code what you've just written here. The current comment is inadequate as documentation.

LPTK: Also it feels like a better architecture might be to use  a method called `applySubBlock` in `BlockTransformer` whenever a sub-block (not a top-level block) is traversed, which would mean you only have to override that here to return input unchanged, instead of all this boilerplate.

CAG2Mark: Good point, I think that will be a better solution.

hkmc2/shared/src/test/mlscript/codegen/Getters.mls
CAG2Mark: Can this be marked as resolved?

hkmc2/shared/src/test/mlscript/basics/BadMemberProjections.mls
CAG2Mark: Currently it seems `BlockMemberSymbol`s must have names. Maybe we could add an flag which disables the name from being printed in error messages -- I don't know if it makes sense for BlockMemberSymbols to be unnamed.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -91,12 +92,37 @@ sealed abstract class Block extends Product with AutoLocated:
     case TryBlock(sub, finallyDo, rest) => sub.freeVars ++ finallyDo.freeVars ++ rest.freeVars
     case Assign(lhs, rhs, rest) => Set(lhs) ++ rhs.freeVars ++ rest.freeVars
     case AssignField(lhs, nme, rhs, rest) => lhs.freeVars ++ rhs.freeVars ++ rest.freeVars
+    case AssignDynField(lhs, fld, arrayIdx, rhs, rest) => lhs.freeVarsLLIR ++ fld.freeVarsLLIR ++ rhs.freeVarsLLIR ++ rest.freeVarsLLIR
```
LPTK: Is this really meant to use `freeVarsLLIR`? 🤔

CAG2Mark: Thanks, good catch

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockTransformer.scala
LPTK: Any intention of solving this in the near future? I'd like to merge the PR, it's been open long enough.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -122,15 +148,18 @@ sealed abstract class Block extends Product with AutoLocated:
   // Note that this returns the definitions in reverse order, with the bottommost definiton appearing
   // last. This is so that using defns.foldLeft later to add the definitions to the front of a block, 
   // we don't need to reverse the list again to preserve the order of the definitions.
-  def floatOutDefns =
+  def floatOutDefns(
+    ignore: Defn => Bool = _ => false, 
+    preserve: Defn => Bool = _ => false) =
```
LPTK: Better to distinguish parameters from body using indentation:
```suggestion
      ignore: Defn => Bool = _ => false,
      preserve: Defn => Bool = _ => false,
  ) =
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+import hkmc2.Message.*
+import hkmc2.semantics.Elaborator.State
+import hkmc2.syntax.Tree
+import hkmc2.codegen.llir.FreshInt
+
+import scala.collection.mutable.ListBuffer
+import scala.collection.mutable.LinkedHashSet
+import scala.collection.mutable.LinkedHashMap
+import scala.collection.mutable.Map as MutMap
+import scala.collection.mutable.Set as MutSet
```
LPTK: Several imports are unused.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+    * @param vars The free variables that are accessed by nested classes/functions.
+    * @param reqCapture The free variables that must be captured using a heap-allocated object.
+    */
+  case class FreeVars(vars: Set[Local], reqCapture: Set[Local]):
+    def ++(that: FreeVars) = FreeVars(vars ++ that.vars, reqCapture ++ that.reqCapture)
+  object FreeVars:
+    val empty = FreeVars(Set.empty, Set.empty)
+
+  /**
+    * Describes the free variables of a function that have been accessed by nested definitions.
```
LPTK: Comment seems copy-pasted and not updated.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+    * @param mp The map from functions' `BlockMemberSymbol`s to their accessed variables.
+    */
+  class UsedLocalsMap(val mp: Map[BlockMemberSymbol, FreeVars]):
+    def apply(f: BlockMemberSymbol) = mp(f)
+    private lazy val inverse = mp.flatMap:
+      case fn -> vars => vars.vars.map(v => v -> fn)
+    // gets the function to which a local belongs
+    def lookup(l: Local) = inverse.get(l)
+
+  object AccessInfo:
```
LPTK: Why not follow the usual pattern and move this after the `AccessInfo` class?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+    * Describes previously defined locals and definitions which could possibly be accessed or mutated by a definition.
+    *
+    * @param accessed Previously defined locals which could possibly be accessed or mutated.
+    * @param mutated Such locals which could also be mutated by this definition.
+    * @param refdDefns Previously defined definitions which could possibly be used by this definition.
+    */
+  case class AccessInfo(
+    accessed: Set[Local], 
+    mutated: Set[Local], 
+    refdDefns: Set[BlockMemberSymbol]):
```
```suggestion
      accessed: Set[Local],
      mutated: Set[Local],
      refdDefns: Set[BlockMemberSymbol],
  ):
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+        refdDefns -- locals
+      )
+    def intersectBms(locals: Set[BlockMemberSymbol]) = AccessInfo(
+        accessed,
+        mutated,
+        refdDefns.intersect(locals)
+      )
+    def addAccess(l: Local) = this.copy(accessed = accessed + l)
+    def addMutated(l: Local) = this.copy(accessed = accessed + l, mutated = mutated + l)
+    def addRefdDefn(l: BlockMemberSymbol) = this.copy(refdDefns = refdDefns + l)
```
```suggestion
    def addAccess(l: Local) = copy(accessed = accessed + l)
    def addMutated(l: Local) = copy(accessed = accessed + l, mutated = mutated + l)
    def addRefdDefn(l: BlockMemberSymbol) = copy(refdDefns = refdDefns + l)
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+      (f.body.definedVars ++ f.params.flatMap(_.paramSyms)).collect:
+        case s: FlowSymbol if !(s is state.runtimeSymbol) => s
+    case c: ClsLikeDefn =>
+      (c.preCtor.definedVars ++ c.ctor.definedVars).collect:
+        case s: FlowSymbol if !(s is state.runtimeSymbol) => s
+    case _ => Set.empty
+
+  def getVarsBlk(b: Block)(using state: State): Set[Local] =
+    b.definedVars.collect:
+      case s: FlowSymbol if !(s is state.runtimeSymbol) => s
```
LPTK, line 95: These two require `State`. Any reason not to move them to the `Lifter` class?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+
+    val clsSyms = clsSymToBms.values.toSet
+    val walker2 = new BlockTransformer(SymbolSubst()):
+      override def applyCase(cse: Case): Case =
+        cse match
+          case Case.Cls(cls, path) =>
+            clsSymToBms.get(cls) match
+            case Some(value) if !ignored.contains(value) => // don't generate a warning if it's already ignored
+              raise(WarningReport(
+                msg"Cannot yet lift the class/module `${value.nme}` as it is used in an instance check." -> N :: Nil,
```
```suggestion
                msg"Cannot yet lift class/module `${value.nme}` as it is used in an instance check." -> N :: Nil,
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+              (privateFields2 is privateFields) &&
+              (publicFields2 is publicFields) &&
+              (preCtor2 is preCtor) && (ctor2 is ctor)
+            then defn else ClsLikeDefn(own2, isym2, sym2, k, paramsOpt2,
+              auxParams2, parentPath, methods2, privateFields2, publicFields2, preCtor2, ctor2)
+
+      override def applyValue(v: Value): Value = v match
+        case RefOfBms(l) if clsSyms.contains(l) && !modOrObj(ctx.defns(l)) =>
+          raise(WarningReport(
+            msg"Cannot yet lift the class `${l.nme}` as it is used as a first-class class." -> N :: Nil,
```
```suggestion
            msg"Cannot yet lift class `${l.nme}` as it is used as a first-class class." -> N :: Nil,
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+
+/**
+  * Lifts classes and functions to the top-level. Also automatically rewrites lambdas.
+  * Assumes the input block does not have any `HandleBlock`s.
+  */
+class Lifter(using State, Raise):
+  import Lifter.*
+
+  /**
+    * The context of the class lifter. One can create an empty context using `Lifter.empty`.
```
```suggestion
    * The context of the class lifter. One can create an empty context using `LifterCtx.empty`.
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+    * @param prevClsDefns Class definitions that have already been traversed, excluding modules
+    * @param curModules Modules that that we are currently nested in (cleared if we are lifted out)
+    * @param capturePaths The path to access a particular function's capture in the local scope
+    * @param bmsReqdInfo The (mutable) captures and (immutable) local variables each function requires
+    * @param ignoredBmsPaths The path to access a particular BlockMemberSymbol (for definitions which could not be lifted)
+    * @param localPaths The path to access a particular local (possibly belonging to a previous function) in the current scope
+    * @param iSymPaths The path to access a particular `innerSymbol` (possibly belonging to a previous class) in the current scope
+    * @param replDefns Ignored definitions whose definitions we should replace in the block itself
+    */
+  case class LifterCtx(
```
LPTK: This class is not meant to be instantiated directly except in `empty`. Doing so would be dangerous because of all the default arguments. So make it
```suggestion
  case class LifterCtx private(
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+    private lazy val inverse = mp.flatMap:
+      case fn -> vars => vars.vars.map(v => v -> fn)
+    // gets the function to which a local belongs
+    def lookup(l: Local) = inverse.get(l)
+
+  object AccessInfo:
+    val empty = AccessInfo(Set.empty, Set.empty, Set.empty)
+
+  /**
+    * Describes previously defined locals and definitions which could possibly be accessed or mutated by a definition.
```
LPTK: What does "previously defined" mean, here and in other doc comments? This wording makes it _look_ like it's alluding to a definition order, but I think it's not. Clearly, these things should be independent of definition order.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+    * @param localCaptureSyms The symbols in a capture corresponding to a particular local
+    * @param prevFnLocals Locals belonging to function definitions that have already been traversed
+    * @param prevClsDefns Class definitions that have already been traversed, excluding modules
+    * @param curModules Modules that that we are currently nested in (cleared if we are lifted out)
+    * @param capturePaths The path to access a particular function's capture in the local scope
+    * @param bmsReqdInfo The (mutable) captures and (immutable) local variables each function requires
+    * @param ignoredBmsPaths The path to access a particular BlockMemberSymbol (for definitions which could not be lifted)
+    * @param localPaths The path to access a particular local (possibly belonging to a previous function) in the current scope
+    * @param iSymPaths The path to access a particular `innerSymbol` (possibly belonging to a previous class) in the current scope
+    * @param replDefns Ignored definitions whose definitions we should replace in the block itself
```
LPTK: What does "replace in the block itself" here?

Also, does the "repl" part of the name refer to diff-test style worksheets?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+    * @param ctx The lifter context. Determines which variables will be captured.
+    * @return The triple (defn, varsMap, varsList), where `defn` is the capture class's definition,
+    * `varsMap` maps the function's locals to the correpsonding `VarSymbol` in the class, and
+    * `varsList` specifies the order of these variables in the class's constructor. 
+    */
+  def createCaptureCls(f: FunDefn, ctx: LifterCtx) =
+    val nme = f.sym.nme + "$capture"
+
+    val clsSym = ClassSymbol(
+      Tree.TypeDef(syntax.Cls, Tree.Error(), N, N),
```
```suggestion
      Tree.DummyTypeDef(syntax.Cls),
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+    )
+
+    val FreeVars(_, cap) = ctx.usedLocals(f.sym)
+
+    val fresh = FreshInt()
+
+    val varsMap: Map[Local, TermSymbol] = cap.map: s =>
+      val id = fresh.make
+      s -> TermSymbol(syntax.ParamBind, S(clsSym), Tree.Ident(s.nme + id + "$"))
+    .toMap
```
```suggestion
        val id = fresh.make
        s -> TermSymbol(syntax.ParamBind, S(clsSym), Tree.Ident(s.nme + id + "$"))
      .toMap
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+  /**
+    * Gets the inner symbols referenced within a class (including those within a member symbol).
+    * @param c The class from which to get the inner symbols.
+    * @return The inner symbols reference within a class.
+    */
+  def getInnerSymbols(c: Defn) =
+    val sym = c match
+      case f: FunDefn => f.sym
+      case c: ClsLikeDefn => c.isym
+      case _ => c.sym // unreachable
```
```suggestion
      case _ => wat("unreachable", c.sym)
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+
+    innerSymCache.get(sym) match
```
LPTK: You could use `getOrElseUpdate`.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+    * @param liftDefn The definition being lifted.
+    * @return Whether the function needs to be captured.
+    */
+  private def needsCapture(captureFn: FunDefn, liftDefn: Defn, ctx: LifterCtx) =
+    val candVars = liftDefn.freeVars
+    val captureFnVars = ctx.usedLocals(captureFn.sym).reqCapture.toSet
+    !candVars.intersect(captureFnVars).isEmpty
+  
+  /**
+    * Gets the immutable local variables of a function that need to captured by a definition being lifted.
```
```suggestion
    * Gets the immutable local variables of a function that need to be captured by a definition being lifted.
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+    val walker = new BlockTransformer(SymbolSubst()):
+      override def applyDefn(defn: Defn): Defn =
+        if defn === d then super.applyDefn(defn)
+        else defn match
+          case c: ClsLikeDefn =>
+            clsSymToBms += c.isym -> c.sym
+            
+            if c.k is syntax.Mod then
+              raise(WarningReport(
+                msg"Modules are not yet properly lifted and will break." -> N :: Nil,
```
LPTK: Please document what's currently wrong and what's the plan to fix it in https://github.com/hkust-taco/mlscript/issues/285

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+    var modules: List[ClsLikeDefn] = Nil
+    var objects: List[ClsLikeDefn] = Nil
+
+    d match
+      case c @ ClsLikeDefn(k = syntax.Mod) => modules +:= c
+      case c @ ClsLikeDefn(k = syntax.Obj) => objects +:= c
+      case _ => ()
+
+    // search for modules
+    val walker = new BlockTransformer(SymbolSubst()):
```
LPTK: I think it'd be good to define a `BlockTraverser` that doesn't need to perform so many checks to see if anything changed. That's what they do in the Scala compiler.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+              raise(WarningReport(
+                msg"Modules are not yet properly lifted and will break." -> N :: Nil,
+                N, Diagnostic.Source.Compilation
+              ))
+              modules +:= c
+            else if c.k is syntax.Obj then
+              objects +:= c
+          case _ => ()
+        super.applyDefn(defn)
+    walker.applyDefn(d)
```
```suggestion
    new BlockTransformer(SymbolSubst()):
      applyDefn(d)
      override def applyDefn(defn: Defn): Defn =
        if defn === d then super.applyDefn(defn)
        else defn match
          case c: ClsLikeDefn =>
            clsSymToBms += c.isym -> c.sym
            
            if c.k is syntax.Mod then
              raise(WarningReport(
                msg"Modules are not yet properly lifted and will break." -> N :: Nil,
                N, Diagnostic.Source.Compilation
              ))
              modules +:= c
            else if c.k is syntax.Obj then
              objects +:= c
          case _ => ()
        super.applyDefn(defn)
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+        case Call(Value.Ref(_: BlockMemberSymbol), args) =>
+          args.map(applyArg)
+          r
+        case Instantiate(InstSel(_), args) =>
+          args.map(applyPath)
+          r
+
+        case _ => super.applyResult(r)
+
+      // don't search within `extends`, otherwise it'll think it's used as a first-class class
```
LPTK: What could go wrong? 😛 
```fs
:lift
fun foo() =
  class C
  class D extends C
  C
//│ ═══[WARNING] Cannot yet lift the class `C` as it is used as a first-class class.
//│ /!!!\ Uncaught error: hkmc2.InternalError: Not in scope: member:C‹392› (class hkmc2.semantics.BlockMemberSymbol)
```

LPTK: I suggest you special-case the valid scenarios where extension clauses do not cause a problem for lifting instead of blanket-assuming no extends clause is problematic.

LPTK: I just noticed clause `extends id(C)` is currently miscompiled (regardless of lifting), but you have to also account for the fact that things like this could happen:

```fs
fun foo() =
  class C
  class D extends id(C)
```
Tho in the future we'll require this to be written
```fs
fun foo() =
  let class C
  let class D extends id(C)
```
and it'll be clear they won't be lifted.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+        val (rewriten, syms) = rewriteBms(b, ctx)
+        val pre = syms.foldLeft(blockBuilder):
+          case (blk, (bms, local)) =>
+            val initial = blk.assign(local, createCall(bms, ctx))
+            ctx.defns(bms) match
+              case c: ClsLikeDefn => initial.assignFieldN(local.asPath, Tree.Ident("class"), bms.asPath)
+              case _ => initial
+        pre.rest(super.applyBlock(rewriten))
+
+    b |> transformer1.applyBlock |> transformer2.applyBlock
```
LPTK: Do these really need to happen in sequence? Couldn't they be combined into a single pass? Cf the "miniphase" paper – this can be done modularly by using class extension.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+        // from the capture; otherwise, we see if that local is passed directly as a parameter to this defn.
+        case Value.Ref(l) => ctx.getLocalCaptureSym(l) match
+          case Some(captureSym) => 
+            Select(ctx.getLocalClosPath(l).get, captureSym.id)(N)
+          case None => ctx.getLocalPath(l) match
+            case Some(value) => Value.Ref(value)
+            case None => super.applyPath(p)
+        case _ => super.applyPath(p)
+
+  def rewriteBlk(b: Block, ctorCls: Opt[ClsLikeDefn], ctx: LifterCtx): Block =
```
LPTK: Does it really make sense to have this as `rewriteBlk` a function containing many anonymous block transformers instead of making `rewriteBlk` itself a class that extends `BlockTransformerNoRec`?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+    val defns = ctx.nestedDefns(f.sym)
+    defns.flatMap(createLiftInfoCont(_, N, ctx.addFnLocals(ctx.usedLocals(f.sym)))).toMap
+
+  def createLiftInfoCls(c: ClsLikeDefn, ctx: LifterCtx): Map[BlockMemberSymbol, LiftedInfo] =
+    val defns = c.preCtor.floatOut(ctx)._2 ++ c.ctor.floatOut(ctx)._2
+    val newCtx = if c.k is syntax.Mod then ctx else ctx.addClsDefn(c)
+    defns.flatMap(f => createLiftInfoCont(f, S(c), newCtx)).toMap
+      ++ c.methods.flatMap(f => createLiftInfoFn(f, newCtx))
+
+  def refRewriter(ctx: LifterCtx, ctorCls: Opt[ClsLikeDefn]) =
```
LPTK: Same remark as for `rewriteBlk`. Just food for thought – maybe some of these transformations could be defined more simply instead of being nested in definitions.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+        (sym, createSym(sym.nme + "$instance"))
+
+      val bmsSymbols = reqdBms.map: sym =>
+        (sym, createSym(sym.nme + "$member"))
+
+      val extraParamsCaptures = capturesSymbols.map: // parameter list
+        case (d, sym) => Param(FldFlags.empty, sym, None)
+      val newCapturePaths = capturesSymbols.map: // mapping from sym to param symbol
+        case (d, sym) => d -> sym.asPath
+      .toMap
```
LPTK: Indentation 😭

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+          val newDef = FunDefn(
+            base.owner, f.sym, PlainParamList(extraParams) :: f.params, f.body
+          )
+          val Lifted(lifted, extras) = liftDefnsInFn(newDef, newCtx)
+
+          val args1 = extraParamsCpy.map(p => p.sym.asPath.asArg)
+          val args2 = headPlistCopy.params.map(p => p.sym.asPath.asArg)
+
+          val bdy = blockBuilder
+            .ret(Call(singleCallBms.asPath, args1 ++ args2)(true, false)) // TODO: restParams not considered
```
LPTK, line 762: Can you add a test where this TODO causes a failure? And please add an item to https://github.com/hkust-taco/mlscript/issues/285

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+    val ctxx = if c.k is syntax.Mod then ctx.inModule(c) else ctx
+    
+    val (preCtor, preCtorDefns) = c.preCtor.floatOut(ctxx)
+    val (ctor, ctorDefns) = c.ctor.floatOut(ctxx)
+
+    val allCtorDefns = preCtorDefns ++ ctorDefns
+    val (ctorIgnored, ctorIncluded) = allCtorDefns.partition(d => ctxx.ignored(d.sym))
+
+    // if this is a module, add a method to reference the classes lifted out
+    val extraMethods = if c.k is syntax.Mod then ctorIncluded.collect:
```
LPTK: Why is this only done if it's a module?

Is that related?
```fs
:lift
class C with
  class D()

let c = new C
//│ c = C

c.D
//│ ═══[RUNTIME ERROR] Error: Access to required field 'D' yielded 'undefined'
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+          case (sym, Value.Lam(params, body)) =>
+            FunDefn(None, sym, params :: Nil, body)
+        val ret = lambdaDefns.foldLeft(newBlk):
+          case (acc, defn) => Define(defn, acc)
+        super.applyBlock(ret)
+    transformer.applyBlock(b)
+
+  // top-level
+  def transform(b: Block) =
+    // this is already done once in the lowering, but the handler lowering adds lambdas currently
```
LPTK, line 976: Then shouldn't the code to do this be written only once and shared?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+      .withNestedDefns(analyzer.nestedDefns)
+      .withAccesses(analyzer.accessMap)
+      .withInScopes(analyzer.inScopeDefns)
+
+    var unliftable: Set[BlockMemberSymbol] = Set.empty
+    var modules: List[ClsLikeDefn] = List.empty
+    var objects: List[ClsLikeDefn] = List.empty
+
+
+    val walker = new BlockTransformerShallow(SymbolSubst()):
```
LPTK: Same remarks as before regarding possibly merging walkers and also the fact it's unnecessary to name this instance.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+        case _ => N
+    .collect:
+      case S(v) => v
+    .toMap
+
+    val ctxx = ctx
+      .addIgnored(unliftable)
+      .withModLocals(modLocals)
+
+    val walker1 = new BlockTransformerShallow(SymbolSubst()):
```
LPTK, line 988: Same remarks as before regarding possibly merging walkers and also the fact it's unnecessary to name this instance.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+        case _ => super.applyBlock(b)
+    walker1.applyBlock(blk)
+/**
+  * Analyzes which variables have been used and mutated by which functions.
+  * Also finds which variables can be passed to a capture class without a heap
+  * allocation (during class lifting) despite being mutable.
+  *
+  * Assumes the input trees have no lambdas.
+  */
+class UsedVarAnalyzer(b: Block)(using State):
```
LPTK: This looks like it should go into its own file.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1504 @@
...
+      walker.applyBlock(b)
+
+      CaptureInfo(reqCapture, hasReader, hasMutator)
+
+    val reqCapture = go(f.body, Set.empty, Set.empty, Set.empty).reqCapture
+    val usedVars = defns.flatMap(_.freeVars.intersect(thisVars)).toSet
+    (usedVars, reqCapture)
+
+  // the current problem is that we need extra code to find which variables were really defined by a function
+  // this may be resolved in the future when the IR gets explicit variable declarations
```
LPTK: This comment is duplicated. Is that intentional?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: In fact I just noticed the exact same code is also added in `Lowering`, which looks like a mistake.

hkmc2/shared/src/main/scala/hkmc2/codegen/llir/Builder.scala
```diff
@@ -304,7 +306,10 @@ final class LlirBuilder(tl: TraceLogger)(fresh: Fresh, fnUid: FreshInt, clsUid:
   
   def registerClasses(b: Block)(using ctx: Ctx)(using Raise, Scope): Ctx =
     b match
-    case Define(cd @ ClsLikeDefn(_own, isym, sym, kind, _paramsOpt, parentSym, methods, privateFields, publicFields, preCtor, ctor), rest) =>
+    case Define(cd @ ClsLikeDefn(_own, isym, sym, kind, _paramsOpt, auxParams,
+      parentSym, methods, privateFields, publicFields, preCtor, ctor), rest) =>
```
LPTK:
```suggestion
        parentSym, methods, privateFields, publicFields, preCtor, ctor), rest) =>
```

hkmc2/shared/src/test/mlscript/backlog/Lifter.mls
```diff
@@ -0,0 +1,149 @@
+:js
+:lift
+:todo
+
+// We use the optional `symbol` parameter of `Select` to detect references to the
+// BlockMemberSymbol. But when this symbol is not present, there is no way to properly
+// detect it. What do to?
```
LPTK: This could be lifted as, essentially:
```fsharp
class B$(y)(a) with
  fun getB() = a.x + y
class A(x) with
  fun B(y) = B$(y)(this)
  set B.class = B // so that `new a.B(n)` still works
  fun getA() = this.B(2).getB()
A(1).getA()
```
and we'd require that the class `B` be final, so it can't be used in `extends` clauses (that check can be done by the type checker later).

It can be done in a future PR. But please update the comment and document this in https://github.com/hkust-taco/mlscript/issues/285

hkmc2/shared/src/test/mlscript/backlog/Lifter.mls
```diff
@@ -0,0 +1,149 @@
...
+
+// This is due to the order of classes after lifting
+class Test
+fun hello() =
+  class Test2 extends Test
+  2
+//│ ═══[RUNTIME ERROR] TypeError: Class extends value undefined is not a constructor or null
+
+// This is due to subclasses not calling `super` with the required locals. The analysis to
+// determine which locals are required is in-place yet.
```
LPTK: What do you mean by "is in-place yet"? This?
```suggestion
// determine which locals are required is not in place yet.
```

hkmc2/shared/src/test/mlscript/basics/BadMemberProjections.mls
LPTK: I'd say that if the symbol was not named by the user, it really shouldn't get a name like this. I checked and using `""` works, here (but then all lambdas are bound to `tmp`-like names, which is a pity).

We could also add a flag that says the name is just a temporary variable name hint and not provided by the user, with the same effect as the flag you proposed.

hkmc2/shared/src/test/mlscript/handlers/Effects.mls
```diff
@@ -367,6 +372,7 @@ handle h = Eff with
 foo(h) 
 //│ = 123
 
+:sjs
```
LPTK: Is this intended? It adds lots of output.

LPTK (approved):
Thanks for your patience; I finally have time to review this PR now (though I'm a bit disappointed that in the meantime you still haven't addressed some of the previous comments, including https://github.com/hkust-taco/mlscript/pull/266#issuecomment-2639361716).

Impressive work, thanks! This was much more difficult than we thought, but your PR seems like a good first step in the right direction.

PS: Before marking my comments as "resolved", make sure to add the test cases implied by the comment, if any.

CAG2Mark:
> though I'm a bit disappointed that in the meantime you still haven't addressed some of the previous comments, including https://github.com/hkust-taco/mlscript/pull/266#issuecomment-2639361716

I thought I did? The warnings are in the backlog file.

hkmc2/shared/src/test/mlscript/codegen/ReboundLet.mls
CAG2Mark: Any suggestions on how this should be changed? To be honest I think it makes more sense here to print `[function anonymous]` or something, but I don't know

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: They're used by `UsedVarAnalyzer`, which is a different class. I could pass `Lifter` into `UsedVarAnalyzer` and move the functions into `Lifter`, but that would make `UsedVarAnalyzer` depend on `Lifter`. I think we should avoid that, as `UsedVarAnalyzer` might be useful for other things in the future -- who knows.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: I've updated the comment to clarify, let me know if it's satisfactory.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: I've updated the name and comment. Please let me know if it's clearer.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: Would doing it using class extensions improve performance? The function calls to `applyX` are re-ordered, but the number of such function calls remains the same (maybe up to a constant).

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -0,0 +1,1442 @@
...
+  def modOrObj(d: Defn) = d match
+    case c: ClsLikeDefn => (c.k is syntax.Mod) || (c.k is syntax.Obj)
+    case _ => false
+
+  // TODO: this is very unclean
+  def isContClassPth(p: Path)(using s: State) = p match
+    case Select(Select(Select(Value.Ref(s.globalThisSymbol), Tree.Ident("Predef")), 
+      Tree.Ident("__Cont")), Tree.Ident("class")) => true
+    case _ => false
+
```
AnsonYeung: Expose the path in HandlerLowering and access it here, the path already changed in the handler runtime PR so we won't forget to update the path

hkmc2/shared/src/test/mlscript/codegen/ReboundLet.mls
AnsonYeung: In JS, you can remove the name in 2 ways:
```js
let x = (0, () => {}); // This makes the lambda to occur inside an expression, which disable the automatic name detection)
let y = () => {};
// set the "name" field to empty. Note that y.name = ""; does not work, as the field is not writable
// but we can still write to not writable field using the following:
Object.defineProperty(y, "name", { value: "", writable: false, enumerable: false, configurable: false});
```

AnsonYeung: btw, both of the ways work exactly the same for proper anon functions (`function () {}`) and anon arrow functions (`() => {}`)

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: You can also just merge the other PR into this one. The other PR will most likely be merged first.

hkmc2/shared/src/test/mlscript/codegen/ReboundLet.mls
LPTK: Seems like a much better idea to use the first approach. Messing with writable fields in lamdas is exactly the kind of things that destroy JIT performance (as it's likely to make it throw its hands in the air and say "A lambda with overridden fields? All right this is some JS nonsense, let's not even try to optimize it...")

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: Actually it's merged now.

LPTK:
> I thought I did? The warnings are in the backlog file.

I was talking about the part that says "And please add this test to the PR [...]". I didn't find this test (but may have missed it).

Don't forget to assign me (or request my review again) when you're done addressing everything.

CAG2Mark:
> > I thought I did? The warnings are in the backlog file.
> 
> I was talking about the part that says "And please add this test to the PR [...]". I didn't find this test (but may have missed it).
> 
> Don't forget to assign me (or request my review again) when you're done addressing everything.

Ah, I see. I recall adding a modified version of the test that returns the output rather than printing it so I can use :expect. I think that's where the confusion is coming from.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: > What could go wrong? 😛
> 
> ```fsharp
> :lift
> fun foo() =
>   class C
>   class D extends C
>   C
> //│ ═══[WARNING] Cannot yet lift the class `C` as it is used as a first-class class.
> //│ FAILURE: Unexpected exception
> //│ /!!!\ Uncaught error: hkmc2.InternalError: Not in scope: member:C‹392› (class hkmc2.semantics.BlockMemberSymbol)
> ```

This case is already in the backlog test file.

CAG2Mark: > I suggest you special-case the valid scenarios where extension clauses do not cause a problem for lifting instead of blanket-assuming no extends clause is problematic.

I don't know if this is a good idea. In general, it's not possible to tell if an extends clause is directly extending a class or something else -- for example:
```fs
object Foo with
  class A
  val class = A
class Bar extends Foo.class
```
Here, the path after `extends` is indistinguishable from
```fs
class Foo(x)
class Bar extends Foo
```
since the latter will be converted to `Select(Foo, "class")` during lowering.

I think it's better to just have `let class` handle this stuff properly.

EDIT: Actually nevermind, the above doesn't even parse because `class` is a keyword

EDIT 2: Right now, it also seems that such a check breaks lifting continuation classes and such, because the `Select`s generated by the handler lowering does not have the `BlockMemberSymbol` field symbol. It's not possible to tell that it's selecting a class directly.

CAG2Mark: I've added the code in the lifter class to perform this check, but it is currently disabled as it breaks things the above mentioned things.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: You're rewriting immutable things here. Doing two rewritings at the same time will produce fewer intermediate data structures allocations.

Besides, cache locality will be better if you only traverse something once and do two things on the same elements together rather than twice, doing them separately.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: > I don't know if this is a good idea. In general, it's not possible to tell if an extends clause is directly extending a class or something else

I don't understand your line of reasoning. Are you suggesting that when the compiler can't know something, it's free to assume that the thing holds and do a transformation that may turn out to be unsound? We're not compiling C++ code, here!

> the `Select`s generated by the handler lowering does not have the `BlockMemberSymbol` field symbol. It's not possible to tell that it's selecting a class directly.

Clearly, handler lowering should be putting the right symbol in there in the first place. That's the point of having symbols.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: I later decided to not lift classes inside modules.

Also, I don't know if this can be fixed with the current IR, at least without eta-expanding the constructor or making it a getter that allocates a lambda on each call.

CAG2Mark: Nevermind, it seems you addressed this in a later comment https://github.com/hkust-taco/mlscript/pull/266#discussion_r1966445502

hkmc2/shared/src/test/mlscript/backlog/Lifter.mls
```diff
@@ -2,9 +2,28 @@
...
+
+// Lifting functions with spread arguments is broken.
+:expect [1]
+fun f(x) =
+  fun g(...rest) = 
+    print(x)
+    rest
+  let a = g
+  a(1)
+f(2)
```
LPTK, line 17: That test case is already in `FunInFun.mls`.

We _really_ don't want to duplicate test cases! De-duplication of test cases is horribly laborious and no one will want to do it, so we'll just end up with many redundant tests.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: Is there a way to get the BlockMemberSymbol of a class defined in the runtime?

hkmc2/shared/src/test/mlscript/backlog/Lifter.mls
CAG2Mark: Oh? All the tests in the backlog are copied directly from the main tests. I thought the point is that the backlog tests would be removed as they are fixed.

CAG2Mark:
@LPTK OK, I've addressed everything. There's some minor things that I would like your opinion on, which are all in the unresolved comments.

hkmc2/shared/src/test/mlscript/backlog/Lifter.mls
LPTK: The backlog is just for broken stuff to fix later. Either you add the broken thing to the backlog and when it's fixed we'll move it where it belongs (you can specify that in a comment if needed). Or you leave the test here it belongs with a `:todo` or `:fixme` and ideally also update the issue so we don't lose track of it.

LPTK:
Thanks, I'll take a look. Maybe next week tho. In the meantime can you pls deduplicate all tests?

hkmc2/shared/src/test/mlscript/backlog/Lifter.mls
CAG2Mark: Got it

PR #268, opened 2025-01-31
Handler and stack safety changes

AnsonYeung:
Changes:
- Support for syntax `handle h = Cls(args) ...`
- Refactor stack safety code into Predef functions
- Optimized some usage of tempoararies away
- Added an extra effectful parameter in `Call` to distinguish calls that cannot yield any effect


hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -80,7 +80,7 @@ class HandlerLowering(using TL, Raise, Elaborator.State, Elaborator.Ctx):
   )
   
   object SimpleCall:
-    def apply(fun: Path, args: List[Path]) = Call(fun, args.map(Arg(false, _)))(true)
+    def apply(fun: Path, args: List[Path]) = Call(fun, args.map(Arg(false, _)))(true, false)
```
LPTK, line 83: This change looks dangerous. It seems to me that `SimpleCall` is a fairly generic helper, notably because of its name (in fact, we could probably move it to a place where more code could use it). Assuming such calls will always be effect-free may cause bugs to future users of the helper. It would be better to pass `isEffectful` as an argument to `apply`.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -591,7 +596,7 @@ trait LoweringTraceLog
       case ((sym, res), acc) => Assign(sym, res, acc)
   
   private def call(fn: Path, args: Ls[Arg]): Call =
-    Call(fn, args)(true)
+    Call(fn, args)(true, false)
```
LPTK, line 599: Same remark as above. You could rename this method to `pureCall`.

hkmc2/shared/src/test/mlscript/decls/Prelude.mls
```diff
@@ -36,6 +36,8 @@ declare module Predef with
   class MatchFailure(errors)
   class __EffectSig(next, tail, handleBlockList, resumed, handler, handlerFun)
   class __Return(value)
+  abstract class __StackDelay with
```
LPTK, line 47: BTW how come this was not needed before?

LPTK (commented):
Thanks, this looks reasonable to me. But can we get a review from @CAG2Mark before we merge?

hkmc2/shared/src/test/mlscript/decls/Prelude.mls
AnsonYeung: The code does not rely on the symbol `__StackDelay` being present. But I added it because I want to manually create instance of this class during my benchmark (due to how stack safety works, it only instantiate the effect handler at the top level which it doesn't have access to during the instrumentation of modules)

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
AnsonYeung: So what should probably be documented for `isEffectful` is that currently it is `true` only when it is both an effectful call and the check is not added for the call (i.e. for tail call, it can be converted to isEffectful=false without any instrumentation as done currently in this PR)

But I don't know whether this is actually a good idea, because the handler lowering should only be ever done once (due to generation of a lot of additional class) and the parameter become a bit misleading if there were any passes after handler lowering that wish to know what calls can be effectful. I'm debating whether to change `isEffectful` to mean it is an effectful call (ignoring whether or not checks are added for the function)

I did find a bug when changing the name tho. There is now a usage of `super` which can be effectful and not instrumented (which actually exist already before this PR).

CAG2Mark (approved):
Are we allowing arbitrary expressions in `handle h = <expr> with`? Technically it seems fine but I'm not sure if it'll cause problems later on.

Otherwise, LGTM.

AnsonYeung:
> Are we allowing arbitrary expressions in `handle h = <expr> with`? Technically it seems fine but I'm not sure if it'll cause problems later on.
> 
> Otherwise, LGTM.

The code is adapted from new expressions. So that part can be anything that evaluate to a JS class currently afaik

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
LPTK: So you're saying you want to add a flag to the `Call` IR node although that flag only makes sense in the context of one specific transformation? Is there not a cleaner way?

At the very least it should have a much more specific and targeted name. Or, as you say, if that's not too much work you could generalize its meaning so it's not so misleading and could be used in other passes.

> I did find a bug when changing the name tho. There is now a usage of `super` which can be effectful and not instrumented (which actually exist already before this PR).

Curious what bug you are referring to. And is it fixed now?

hkmc2/shared/src/test/mlscript/decls/Prelude.mls
LPTK: Then yo should say something to that effect in a comment (that this declaration is not really useful except for personal experimentation).

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
AnsonYeung: I added a new test that is broken now. This is a pretty weird use of the handler feature as an effect is raised during the construction of the handler. The most natural semantics (following what currently normal constructor do) is that the handler is not yet constructed but after resumption it will be constructed and executed normally.

It is not fixed. I think fixing it require some change to JSBuilder, if super yields some effect then those `this.x = x;` together with the real constructor code will be paused and resumed later (the super call is not finished yet so you cannot access `this`, and also it should suspend even if it can access it anyway). Implementing this will cause those code to (still) be generated in the final code because JSBuilder is unaware that I already included them elsewhere.

LPTK: Wow ok, thanks for the explanation. That *is* a weird case.  We may want to just forbid it and throw an exception, though supporting the "obvious" semantics would be better. This can be done in a follow-up PR.

LPTK (changes requested):
Could you please address this part of my comments:

> At the very least it should have a much more specific and targeted name. Or, as you say, if that's not too much work you could generalize its meaning so it's not so misleading and could be used in other passes.

AnsonYeung:
Right, I've changed it to mean it is an effectful call (regardless of whether the check is added yet). This should make more sense to later passes but passes after handlerlowering should not create new effectful call anyway as it won't be instrumented.

LPTK:
> Right, I've changed it to mean it is an effectful call (regardless of whether the check is added yet). This should make more sense to later passes but passes after handlerlowering should not create new effectful call anyway as it won't be instrumented.

You need to realize that everything you just wrote here needs to be documented in the code. How else do you think another project member will know what this field means and how to use it correctly?

LPTK:
Also please be careful about the terminology. "Effectful" in most compilers normally mean that the code may have side effects, which is a distinct thing from raising algebraic effects. We need to establish clear terminology. If you want to indicate the code may raise effects, please call this `mayRaiseEffects`.

AnsonYeung:
I've addressed these issues just now

LPTK (approved):
LGTM!

PR #269, opened 2025-02-01
Add missing reserved words in `JSBuilder`

NeilKleistGao:
Add the following reserved words in strict mode:
- implements
- package
- protected
- static

and identifiers with special meanings:
- arguments
- eval

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words


LPTK (approved):
LGTM, thanks!

PR #271, opened 2025-02-01
Better handle overlapping patterns in normalization (fix #270)

### LINKED ISSUE #270: [UCS] Patterns `null` and `undefined` are treated as being covered by `Object`
```fs
:sjs
fun f(arg) = if
  arg is
    undefined then "undefined"
    null      then "null"
    Object    then String(arg)
//│ JS (unsanitized):
//│ let f;
//│ f = function f(arg) {
//│   if (arg === undefined) {
//│     return "undefined";
//│   } else {
//│     if (arg === null) {
//│       return "null";
//│     } else {
//│       if (arg instanceof globalThis.Object) {
//│         return globalThis.String(arg);
//│       } else {
//│         throw new globalThis.Error("match error");
//│       }
//│     }
//│   }
//│ };
//│ null

:sjs
fun f(arg) = if
  arg is
    Object    then String(arg)
    undefined then "undefined"
    null      then "null"
//│ JS (unsanitized):
//│ let f1;
//│ f1 = function f(arg) {
//│   if (arg instanceof globalThis.Object) {
//│     return globalThis.String(arg);
//│   } else {
//│     throw new globalThis.Error("match error");
//│   }
//│ };
//│ null
```

chengluyu:
I have solved the issue in #270. Additionally,

1. Replace hard-coded class names with symbols assumed in `Elaborator.ctx.Builtins`.
2. Place built-in classes that do not use `instanceof` tests in `JSBuilder` into a `Set`, avoiding processing those classes during lowering (otherwise, the classes cannot be found at runtime).
3. Branches with `Object` as the pattern will overshadow all branches with classes and modules as patterns.
4. Fix the normalization of Boolean conditions. To be specific, the temporary scrutinee created for`if f(x) then ... else ...` should not be specialized, but the treatment was mistakenly applied to all matches against `true`.


hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -562,7 +567,7 @@ trait LoweringSelSanityChecks
         val selRes = TempSymbol(N, "selRes")
         val split = Split.Cons(
             Branch(
-              selRes.ref(),
+            selRes.ref(),
```
LPTK: ?

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -97,6 +97,9 @@ class TempSymbol(val trm: Opt[Term], dbgNme: Str = "tmp")(using State) extends B
   override def toString: Str = s"$$${super.toString}"
   override def subst(using s: SymbolSubst): TempSymbol = s.mapTempSym(this)
 
+/** A special `TempSymbol` created to cache the result of Boolean condition
+ *  terms. We need to distinguish such symbols in normalization. */
+class TestSymbol(dbgNme: Str)(using State) extends TempSymbol(N, dbgNme)
```
LPTK: There used to be 2 lines separating `TempSymbol` from `InstSymbol` but you broke that separation.
```suggestion


```

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -180,9 +175,7 @@ class Normalization(elaborator: Elaborator)(using raise: Raise, ctx: Ctx):
       case split @ Split.Cons(head, tail) =>
         log(s"CASE Cons ${head.showDbg}")
         head match
-          case Branch(thatScrutineeVar, Pattern.Var(alias), continuation) =>
-            Split.Let(alias, thatScrutineeVar, rec(continuation))
-          case Branch(test, Pattern.Lit(Tree.BoolLit(true)), continuation) =>
+          case Branch(Term.Ref(_: TestSymbol), Pattern.Lit(Tree.BoolLit(true)), continuation) =>
```
LPTK: This logic looks quite weird and brittle. Why do you need to assume that some other part of the UCS put a `TestSymbol` there which is matched specifically against `true`? Why not represent things more cleanly as two different cases in internal UCS the representation itself?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -40,8 +40,6 @@ class Normalization(elaborator: Elaborator)(using raise: Raise, ctx: Ctx):
         case Split.Cons(head, tail) => Split.Cons(head, tail ++ those)
         case Split.Let(name, term, tail) => Split.Let(name, term, tail ++ those)
         case Split.Else(_) /* impossible */ | Split.End => those)
-
```
LPTK, line 43: Is this line removal on purpose?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -388,6 +388,11 @@ class Lowering(lowerHandlers: Bool, stackLimit: Option[Int])(using TL, Raise, St
               )
             pat match
               case Pattern.Lit(lit) => mkMatch(Case.Lit(lit) -> go(tail, topLevel = false))
+              case Pattern.ClassLike(cls: ClassSymbol, _trm, _args0, _refined)
+                  // Do not elaborate `_trm` when the `cls` is virtual.
+                  if Elaborator.ctx.Builtins.virtualClasses contains cls =>
+                // `Value.Arr(Nil)` is a dummy result and will not be used.
```
LPTK: Why pick this weird value as the dummy? Why will it not be used? What's doing on exactly, here? A concrete example could help.

LPTK (commented):
Thanks, but I have a few questions.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
chengluyu: Some classes (e.g. `Int`) from `Prelude` does not exist at runtime. If we do lowering on `trm`, then `JSBuilder` will generate selection that crashes.

`Case.Cls` requires `path: Path` as the second parameter. Since the selection is not available, I just passed a dummy value here.

It's not used because if `JSBuilder` found the class symbol does not exist at runtime, then `JSBuilder` does not use the second parameter.

LPTK: Then make the dummy value something like `globalThis.Predef.unreachable` (and make that a getter that throws an internal error). And add the explanation you commented here as documentation in the actual code. How else do you expect someone will understand this subtle non-local logic? Moreover, in the doc, please include the string `[invariant:0]` and add the same `[invariant:0]` string in a comment on the corresponding `JSBuilder` part that discards the path. Because you're introducing a non-local invariant here that we need to be able to track! (Not to mention other backends will need to be careful about this as well.)

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -391,8 +394,13 @@ class Lowering(lowerHandlers: Bool, stackLimit: Option[Int])(using TL, Raise, St
               case Pattern.ClassLike(cls: ClassSymbol, _trm, _args0, _refined)
                   // Do not elaborate `_trm` when the `cls` is virtual.
                   if Elaborator.ctx.Builtins.virtualClasses contains cls =>
-                // `Value.Arr(Nil)` is a dummy result and will not be used.
-                mkMatch(Case.Cls(cls, Value.Arr(Nil)) -> go(tail, topLevel = false))
+                // [invariant:0] Some classes (e.g., `Int`) from `Prelude` does
```
```suggestion
                // [invariant:0] Some classes (e.g., `Int`) from `Prelude` do
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -391,8 +394,13 @@ class Lowering(lowerHandlers: Bool, stackLimit: Option[Int])(using TL, Raise, St
...
+                // not exist at runtime. If we do lowering on `trm`, backends
+                // (e.g., `JSBuilder`) will generate selection that crashes.
```
```suggestion
                // (e.g., `JSBuilder`) will not be able to handle the corresponding selections.
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -391,8 +394,13 @@ class Lowering(lowerHandlers: Bool, stackLimit: Option[Int])(using TL, Raise, St
...
+                // The second parameter of `Case.Cls` is a dummy value that
+                // leads the program into an unreachable state and will be
+                // discarded in the backend.
```
```suggestion
                // In this case the second parameter of `Case.Cls` will not be used.
                // So we make it `Predef.unreachable` here.
```

hkmc2/shared/src/test/mlscript/decls/Prelude.mls
```diff
@@ -43,7 +43,8 @@ declare module Predef with
   class MatchFailure(errors)
   class __EffectSig(next, tail, handleBlockList, resumed, handler, handlerFun)
   class __Return(value)
-
+  
+  fun unreachable
```
LPTK: Is this needed? If not, please remove it.

chengluyu: Removed.

LPTK (approved):
LGTM!

chengluyu:
Thanks for your review!

PR #273, opened 2025-02-07
Merge Match Arms

ychenfo:
```
if x is
 A then 1
 B then 2
```

Prevously this is compiled to

```
Match:
  scrut = x
  arms = List of (case = A, body = 1)
  dflt =
    Match:
      scrut = x
      arms = List of (case = B, body = 2)
      dflt = Throw error
      rest = End
  rest = End("")
```

Now this is:
```
Match:
  scrut = x
  arms = List of
    (case = A, body = 1)
    (case = B, body = 2)
  dflt = Throw error
  rest = End("")
```

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
@@ -133,29 +133,45 @@ sealed abstract class Block extends Product with AutoLocated:
         case _ => super.applyBlock(b)
     
     (transformer.applyBlock(this), defns)
-  
-  def append(b: Block): Block = this match
-    case Match(scrut, arms, dflt, rest) => Match(scrut, arms, dflt, rest.append(b))
-    case Label(label, body, rest) => Label(label, body, rest.append(b))
-    case Begin(sub, rest) => ??? // appending on a Begin Block should never happen
-    case TryBlock(sub, finallyDo, rest) => TryBlock(sub, finallyDo, rest.append(b))
-    case AssignDynField(lhs, fld, arrayIdx, rhs, rest) => AssignDynField(lhs, fld, arrayIdx, rhs, rest.append(b))
-    case Assign(lhs, rhs, rest) => Assign(lhs, rhs, rest.append(b))
-    case a@AssignField(lhs, nme, rhs, rest) => AssignField(lhs, nme, rhs, rest.append(b))(a.symbol)
-    case Define(defn, rest) => Define(defn, rest.append(b))
-    case HandleBlock(lhs, res, par, args, cls, handlers, body, rest) => HandleBlock(lhs, res, par, args, cls, handlers, body, rest.append(b))
-    case End(msg) => b
-    case _: BlockTail => this
-  
-  object FlattenTransformer extends BlockTransformer(new SymbolSubst()):
-    // no transformation for nested sub-definitions
-    override def applyDefn(defn: Defn): Defn = defn
-    override def applyBlock(b: Block): Block = super.applyBlock(b) match
-      case Begin(sub, rest) => sub.append(rest)
-      case b => b
-  
-  lazy val flatten: Block =
-    FlattenTransformer.applyBlock(this)
+    
+  lazy val flatten: Block = 
+    // traverses a Block like a list, flatten `Begin`s using an accumulator
+    // returns the flattend but reversed Block (with the dummy tail `End("for flatten only")`) and the actual tail of the Block
+    def getReversedFlattenAndTrueTail(b: Block, acc: Block): (Block, BlockTail) = b match
+      case Match(scrut, arms, dflt, rest) => getReversedFlattenAndTrueTail(rest, Match(scrut, arms, dflt, acc))
+      case Label(label, body, rest) => getReversedFlattenAndTrueTail(rest, Label(label, body, acc))
+      case Begin(sub, rest) =>
+        val (firstBlockRev, firstTail) = getReversedFlattenAndTrueTail(sub, acc)
+        firstTail match
+          case _: End => getReversedFlattenAndTrueTail(rest, firstBlockRev)
+          // if the tail of `sub` is not `End`, ignore the `rest` of this `Begin`
+          case _ => firstBlockRev -> firstTail
+      case TryBlock(sub, finallyDo, rest) => getReversedFlattenAndTrueTail(rest, TryBlock(sub, finallyDo, acc))
+      case Assign(lhs, rhs, rest) => getReversedFlattenAndTrueTail(rest, Assign(lhs, rhs, acc))
+      case a@AssignField(lhs, nme, rhs, rest) => getReversedFlattenAndTrueTail(rest, AssignField(lhs, nme, rhs, acc)(a.symbol))
+      case AssignDynField(lhs, fld, arrayIdx, rhs, rest) => getReversedFlattenAndTrueTail(rest, AssignDynField(lhs, fld, arrayIdx, rhs, acc))
+      case Define(defn, rest) => getReversedFlattenAndTrueTail(rest, Define(defn, acc))
+      case HandleBlock(lhs, res, par, args, cls, handlers, body, rest) => getReversedFlattenAndTrueTail(rest, HandleBlock(lhs, res, par, args, cls, handlers, body, acc))
+      case t: BlockTail => acc -> t
+    
+    // reverse the Block returnned from the previous function,
+    // which does not contain `Begin` (except for the nested ones),
+    // and whose tail must be the dummy `End("for flatten only")`
+    def rev(b: Block, t: Block): Block = b match
+      case Match(scrut, arms, dflt, rest) => rev(rest, Match(scrut, arms, dflt, t))
+      case Label(label, body, rest) => rev(rest, Label(label, body, t))
+      case TryBlock(sub, finallyDo, rest) => rev(rest, TryBlock(sub, finallyDo, t))
+      case Assign(lhs, rhs, rest) => rev(rest, Assign(lhs, rhs, t))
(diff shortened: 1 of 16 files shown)


hkmc2/shared/src/test/mlscript/codegen/MergeMatchArms.mls
```diff
@@ -0,0 +1,146 @@
...
+//│   }
+//│   tmp = Predef.print("done");
+//│ }
+//│ x = tmp;
+//│ Predef.print(x)
+//│ > 1
+//│ x = 1
+
+
+// NOTE: these two branches are not merged...
```
LPTK: That's expected because in general we try to preserve the order of `let` bindings. But in trivial cases like these we could actually reorder them.

hkmc2/shared/src/test/mlscript/codegen/MergeMatchArms.mls
```diff
@@ -0,0 +1,146 @@
...
+//│ = 2
+
+
+:sjs
+let x = if a is
+  A then 1
+  B then 2
+  C then 3
+  else
+    if a is
```
LPTK, line 92: Any idea why this one doesn't seem to be merged?

LPTK (approved):
Nice simplifications!

I think @chengluyu had some more advanced post-processing transformations in mind, notably commuting some matches on interleaved scrutinees when possible, but these can come in later PRs.

hkmc2/shared/src/test/mlscript/codegen/MergeMatchArms.mls
ychenfo: Ah sorry I forgot to add comments for this test. Here it's not merged because `if a is D ...` is not in a tail position ([the `rest` field of its `Match` block is not `End`](https://github.com/hkust-taco/mlscript/pull/273/commits/7c3a1be8303b72631bffc61faf62d48d0c9e2f82#:~:text=scrutRewritten%2C%20armsRewritten%2C%20dfltRewritten%2C-,_%3A%20End,-%29%20if%20scrutRewritten)).

One way I can currently think of is to duplicate the `rest` into each arm body, but this seems to cause too much duplication when there are many branches.

hkmc2/shared/src/test/mlscript/codegen/MergeMatchArms.mls
ychenfo: Thanks for the explanation! Now I see.

LPTK:
```suggestion
// NOTE: These two branches are not merged because there is an interleaved `let` binding.
//   For trivial bindings like these, we could move them out of the way and still merge the branches.
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -729,3 +731,21 @@ trait LoweringTraceLog
     )
 
 
+object MergeMatchArmTransformer extends BlockTransformer(new SymbolSubst()):
+  override def applyBlock(b: Block): Block = b match
```
LPTK: Instead of all the boilerplate in this case, why don't you rewrite this match to
```suggestion
  override def applyBlock(b: Block): Block = super.applyBlock(b) match
```
so that you don't need to manually apply the transformation to all the extracted components yourself?

hkmc2/shared/src/test/mlscript/codegen/MergeMatchArms.mls
LPTK: For sure, you should never blindly duplicate arbitrary tails in general, as that could lead to an exponential code blowup. But here you could match the tail as `SimpleTail(k, rest)` where `SimpleTail` is an extractor that matches tails of small size that are cheap to duplicate, such as single calls (like here) as well as trivial bindings that can be moved out of the way, like `x = 42`. `SimpleTail` would extract such a binding into `k = newTail => Assign("x", 42, newTail)` and the call as `rest = Return(...)`.

ychenfo: Thanks for the suggestion! I was trying to implement the `SimpleTail` things, and found that what I said ("the `rest` field of its `Match` block is not `End`") was not actually true😰. The real reason for it not being merged is that in this test case, the `dflt` block is not a `Match`, but a `Begin` block (that starts with a `Match`, though). So my block transformer did not treat this as transformable.

At the same time, I also notice your comment (`// TODO: remove this form?`) in the definition of `Begin`. If I understand correctly, `Begin` is just for concatenating two blocks, like appending two lists; and for the `SimpleTail`s, I also need to append them to the end of each arm blocks.

So maybe I can first write an `append` function for blocks, and then replace the `Begin` with that `append` function, and then finish the implementations to append `SimpleTails` to the arms then merge them?

And for the details of `SimpleTail`, I understand that it should match small blocks, but I couldn't understand why `k` and `rest` needs to be extracted. I thought just appending the small blocks after each arms would be fine, like below...

```
let x = if a is
  A then 1
  B then 2
  C then 3
  else
    if a is
      D then
        if a is
          A then 1
          B then 2
      E then 5
    print("done")
```

would be transformed to

```
let x = if a is
  A then 1
  B then 2
  C then 3
  else
    if a is
      D then
        if a is
          A then 1
          B then 2
        print("done")
      E then
        5
        print("done")
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
ychenfo: Thanks for the suggestion! Indeed this is much cleaner, I will make the change.

hkmc2/shared/src/test/mlscript/codegen/MergeMatchArms.mls
LPTK: Concatenating in constant time might be important in some cases, so I'm not sure about removing it. Instead, we could have a `lazy val` field called `flattened` representing the flattened block (without any `Begin` in it except nested in sub-definitions). And you'd match on that in your analysis. This would mean that no matter in which order we concatenate things, we only linearize once (in linear time).

LPTK: > And for the details of `SimpleTail`, I understand that it should match small blocks, but I couldn't understand why `k` and `rest` needs to be extracted.

I was thinking in case there are intervening trivial bindings, but I realize I mixed things up. I was thinking of:

```
if a is
  A then 1
  else
    let x = 42
    if a is
      B then x

```

Here, we still want to merge into

```
let x = 42
if a is
  A then 1
  B then x
```

You could do that by adding a `TrivialStatements(k, rest)` extractor, allowing you to inspect `rest` and if needed rebuild the trivial context captured by `k`.

ychenfo: Thanks for the suggestions! I have tried to make the changes implementing the `TrivialStatements` for extracting the trivial contexts that can be moved out of the way to help merging match arms.

For the small `SimpleTail` extractor, I notice that there is a `size` lazy val for `Block`s, so I currently just use that and the number of branches as the [criteria](https://github.com/hkust-taco/mlscript/blob/d2a5f7646346e474c66065af8a9d630bee3fa17d/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala#L774) to determine whether the `rest` field of `Match` is too big to be duplicated into arm bodies.

hkmc2/shared/src/test/mlscript/codegen/MergeMatchArms.mls
ychenfo: Thanks for the suggested changes. Now this trivial interleaved `let` binding is moved out of the way by using `TrivialStatements`, and I have updated the comment further to clarify this test.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -133,6 +134,29 @@ sealed abstract class Block extends Product with AutoLocated:
...
+    case HandleBlock(lhs, res, par, args, cls, handlers, body, rest) => HandleBlock(lhs, res, par, args, cls, handlers, body, rest.append(b))
+    case End(msg) => b
+    case _: BlockTail => this
+  
+  object FlattenTransformer extends BlockTransformer(new SymbolSubst()):
+    // no transformation for nested sub-definitions
+    override def applyDefn(defn: Defn): Defn = defn
+    override def applyBlock(b: Block): Block = super.applyBlock(b) match
+      case Begin(sub, rest) => sub.append(rest)
+      case b => b
```
LPTK: This has the wrong complexity. If your block has a left-leaning structure, you will keep re-traversing the flattened prefix over and over. I think this can be fixed by changing the `append` function and giving it an accumulator.

ychenfo: Thanks for the suggestion! I have pushed an updated `flatten`, which I believe now only traverses the Block twice (once for building a reversed accumulated result, and once for reversing it again). I couldn't think of a way to make the code more concise using `BlockTransformer` so currently I just manually write the recursive functions on Block...

LPTK (approved):
Well... the way you treat the Block construct like a linked list by building it in the wrong order and then using an ad-hoc reversal function (when you could have just accumulated a function) is extraordinarily inelegant, but let's merge this for now as it seems to work 😅

ychenfo:
> Well... the way you treat the Block construct like a linked list by building it in the wrong order and then using an ad-hoc reversal function (when you could have just accumulated a function) is extraordinarily inelegant, but let's merge this for now as it seems to work 😅

😅Ah now I see, thanks! I think my new PR should improve this.

PR #275, opened 2025-02-11
Fix syntax of `(using ..., Type)`

### LINKED ISSUE #274: Syntax `def f(using Type)` works but `def f(using Type1, Type2)` doesn't work.

FlandiaYingman:
Fix #274.

This PR fixes

```fs
module M with
  fun f(using Int, Int) = 42
```

which used to parse the `using Int` part only but failed to parse the (second) `Int` part.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
@@ -198,6 +198,7 @@ enum Tree extends AutoLocated:
    */
   def asParam(inUsing: Bool): Opt[(Opt[Bool], Ident, Opt[Tree])] = this match
     case und: Under => S(N, new Ident("_").withLocOf(und), N)
+    // * In `using` clauses, identifiers are understood as type names for unnamed contextual parameters:
     case id: Ident if inUsing => S(N, Ident(""), S(id))
     case id: Ident => S(N, id, N)
     case Spread(Keyword.`..`, _, S(id: Ident)) => S(S(false), id, N)
--- a/hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
@@ -193,8 +193,8 @@ enum Tree extends AutoLocated:
     case _ => this
   
   /** 
-   * S(true) means eager spread, S(false) means lazy spread, N means no spread.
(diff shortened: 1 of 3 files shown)


hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
```diff
@@ -192,20 +192,24 @@ enum Tree extends AutoLocated:
       LetLike(letLike, id, S(App(Ident(nme.init), Tup(id :: r :: Nil))), bodo).withLocOf(this).desugared
     case _ => this
   
-  /** S(true) means eager spread, S(false) means lazy spread, N means no spread. */
-  def asParam: Opt[(Opt[Bool], Ident, Opt[Tree])] = this match
+  /** 
+   * S(true) means eager spread, S(false) means lazy spread, N means no spread.
+   * ctx means the the param list is modified by `using`.
```
LPTK:
```suggestion
   * Parameter `inUsing` means the param list is modified by `using`.
   * In the first result, `S(true)` means eager spread, `S(false)` means lazy spread, and `N` means no spread.
```

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
```diff
@@ -192,20 +192,24 @@ enum Tree extends AutoLocated:
...
+   */
+  def asParam(inUsing: Bool): Opt[(Opt[Bool], Ident, Opt[Tree])] = this match
     case und: Under => S(N, new Ident("_").withLocOf(und), N)
+    case id: Ident if inUsing => S(N, Ident(""), S(id))
```
LPTK, line 202: Would be good to document this.
```suggestion
    // * In `using` clauses, identifiers are understood as type names for unnamed contextual parameters:
    case id: Ident if inUsing => S(N, Ident(""), S(id))
```

LPTK (commented):
Does the following syntaxes work? Do we have tests for them?

```fs
fun foo(using Int, arg: Str)

fun foo(using arg: Str, Int)
```

FlandiaYingman:
> Does the following syntaxes work? Do we have tests for them?
> 
> ```fsharp
> fun foo(using Int, arg: Str)
> 
> fun foo(using arg: Str, Int)
> ```

They work fine and I just added them to the tests.

LPTK (approved):
LGTM!

PR #276, opened 2025-02-11
Better Flatten

ychenfo:
This PR improves the flattening of Blocks to remove `Begin`s. Now the I believe the internal function `flattenConcat` traverses everything once to build the flattened Block. The second parameter is by-name [so that in the case of `Begin`, `rest` is not traversed if `sub` ends with something other than `End`](https://github.com/CrescentonC/mlscript/commit/87fffa970c75e61591aaa2d6a9cf5ad7bb6e39b7#diff-72bbb753659dd44a5e47f3cd0b3954e4ee95fda127ca491a2c267461779c60fbR142).

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
@@ -135,22 +135,47 @@ sealed abstract class Block extends Product with AutoLocated:
     
     (transformer.applyBlock(this), defns)
     
+  private lazy val fAndRest: (Block => Block, BlockTail) = this match
+    case Match(scrut, arms, dflt, rest) =>
+      val (f, r) = rest.fAndRest
+      ((k: Block) => Match(scrut, arms, dflt, f(k))) -> r
+    case Label(label, body, rest) =>
+      val (f, r) = rest.fAndRest
+      ((k: Block) => Label(label, body, f(k))) -> r
+    case Begin(sub, rest) =>
+      val (f, r) = sub.fAndRest
+      r match
+        case End(_) =>
+          val (f1, r1) = rest.fAndRest
+          ((k: Block) => f(f1(k))) -> r1
+        case _ => f -> r
+    case TryBlock(sub, finallyDo, rest) =>
+      val (f, r) = rest.fAndRest
+      ((k: Block) => TryBlock(sub, finallyDo, f(k))) -> r
+    case Assign(lhs, rhs, rest) =>
+      val (f, r) = rest.fAndRest
+      ((k: Block) => Assign(lhs, rhs, f(k))) -> r
+    case a@AssignField(lhs, nme, rhs, rest) =>
+      val (f, r) = rest.fAndRest
+      ((k: Block) => AssignField(lhs, nme, rhs, f(k))(a.symbol)) -> r
+    case AssignDynField(lhs, fld, arrayIdx, rhs, rest) =>
+      val (f, r) = rest.fAndRest
+      ((k: Block) => AssignDynField(lhs, fld, arrayIdx, rhs, f(k))) -> r
+    case Define(defn, rest) =>
+      val (f, r) = rest.fAndRest
+      ((k: Block) => Define(defn, f(k))) -> r
+    case HandleBlock(lhs, res, par, args, cls, handlers, body, rest) =>
+      val (f, r) = rest.fAndRest
+      ((k: Block) => HandleBlock(lhs, res, par, args, cls, handlers, body, f(k))) -> r
+    case t: BlockTail => (identity: Block => Block) -> t
+  
+  lazy val flattenTopLevel = fAndRest._1(fAndRest._2)
+  
+  object FlattenAllNested extends BlockTransformer(new SymbolSubst()):
+    override def applyBlock(b: Block): Block = super.applyBlock(b.flattenTopLevel)
+  
   lazy val flatten: Block =
-    
-    def flattenConcat(a: Block, b: => Block): Block = a match
-      case Match(scrut, arms, dflt, rest) => Match(scrut, arms, dflt, flattenConcat(rest, b))
-      case Label(label, body, rest) => Label(label, body, flattenConcat(rest, b))
-      case Begin(sub, rest) => flattenConcat(sub, flattenConcat(rest, b))
-      case TryBlock(sub, finallyDo, rest) => TryBlock(sub, finallyDo, flattenConcat(rest, b))
-      case Assign(lhs, rhs, rest) => Assign(lhs, rhs, flattenConcat(rest, b))
-      case a@AssignField(lhs, nme, rhs, rest) => AssignField(lhs, nme, rhs, flattenConcat(rest, b))(a.symbol)
-      case AssignDynField(lhs, fld, arrayIdx, rhs, rest) => AssignDynField(lhs, fld, arrayIdx, rhs, flattenConcat(rest, b))
-      case Define(defn, rest) => Define(defn, flattenConcat(rest, b))
-      case HandleBlock(lhs, res, par, args, cls, handlers, body, rest) => HandleBlock(lhs, res, par, args, cls, handlers, body, flattenConcat(rest, b))
-      case End(_) => b
-      case _: BlockTail => a
-    
-    flattenConcat(this, End(""))
+    FlattenAllNested.applyBlock(this)
   
 end Block
 
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
@@ -134,61 +134,84 @@ sealed abstract class Block extends Product with AutoLocated:
         case _ => super.applyBlock(b)
     
     (transformer.applyBlock(this), defns)
-      
-  private lazy val flattenContAndTailAndRes: (Block => Block) -> BlockTail -> Block = this match
-    case Match(scrut, arms, dflt, rest) =>
-      val f -> t -> r = rest.flattenContAndTailAndRes
-      ((k: Block) => Match(scrut, arms, dflt, f(k))) ->
-      t ->
-      (if r is rest then this else Match(scrut, arms, dflt, r))
-    case Label(label, body, rest) =>
-      val f -> t -> r = rest.flattenContAndTailAndRes
-      ((k: Block) => Label(label, body, f(k)))->
-      t ->
-      (if r is rest then this else Label(label, body, r))
-    case Begin(sub, rest) =>
-      val f -> t -> r = sub.flattenContAndTailAndRes
-      t match
-        case End(_) =>
-          val f1 -> t1 -> r1 = rest.flattenContAndTailAndRes
-          ((k: Block) => f(f1(k))) ->
-          t1 ->
-          f(r1)
-        case _ => f -> t -> r
-    case TryBlock(sub, finallyDo, rest) =>
-      val f -> t -> r = rest.flattenContAndTailAndRes
-      ((k: Block) => TryBlock(sub, finallyDo, f(k))) ->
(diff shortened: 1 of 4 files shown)


hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -135,44 +135,22 @@ sealed abstract class Block extends Product with AutoLocated:
...
+      case Match(scrut, arms, dflt, rest) => Match(scrut, arms, dflt, flattenConcat(rest, b))
+      case Label(label, body, rest) => Label(label, body, flattenConcat(rest, b))
+      case Begin(sub, rest) => flattenConcat(sub, flattenConcat(rest, b))
+      case TryBlock(sub, finallyDo, rest) => TryBlock(sub, finallyDo, flattenConcat(rest, b))
+      case Assign(lhs, rhs, rest) => Assign(lhs, rhs, flattenConcat(rest, b))
+      case a@AssignField(lhs, nme, rhs, rest) => AssignField(lhs, nme, rhs, flattenConcat(rest, b))(a.symbol)
+      case AssignDynField(lhs, fld, arrayIdx, rhs, rest) => AssignDynField(lhs, fld, arrayIdx, rhs, flattenConcat(rest, b))
+      case Define(defn, rest) => Define(defn, flattenConcat(rest, b))
+      case HandleBlock(lhs, res, par, args, cls, handlers, body, rest) => HandleBlock(lhs, res, par, args, cls, handlers, body, flattenConcat(rest, b))
+      case End(_) => b
```
LPTK: Here you're discarding the current `End`, which may have contained useful info, even when what you're replacing it with is `End("")` with no info. Better make `b` an `End => Block` function starting as the identity and pass `a` here.

LPTK (changes requested):
Indeed that's better, but still not right in the way it's used (I think it was already wrongly used before this PR).

Although you're using a `lazy val`, as I suggested, that doesn't really help since none of the sub-computations use that cached value. So by calling `flatten` every time you want to pattern match a block in your pass, you re-incur an entire flattening pass.

Instead, `flatten` should probably be called _at the beginning_ of the pass for each block structure (including those nested inside definitions) and _not_ for each sub-block.

ychenfo:
Thanks, I have pushed an update which tries to further improve the implementation of `flatten`. Now there are two `lazy val`s, the first one (`flattenContAndTailAndRes`) computes for each block three things:
  1. `f`: a lambda of type `Block => Block`, which takes another block as the tail, and builds the flattened version of the current block
  2. `t`: a `BlockTail`, which is the actual tail of the current block. Now I think the original information in `End`s is preserved (for `Begin`, though, the `End` in the first sub-block is still discarded)
  3. `r`: the flattened version of the current block, and `r == f(t)`.

`flattenContAndTailAndRes` returns 3 values satisfying `r == f(t)`, while it could have only returned `f` and `t`, because I think in this way, when it is possible to directly build the flattened block, we no longer need to invoke all the accumulated lambdas to build the result.

And the `lazy val` `flatten` just calls the first one to get its third result.

`flattenContAndTailAndRes` now calls itself recursively in its computation, such that later when `flatten` is called on a block that is the `rest` of an already-flattened block, the cached result can be used. However I think I am still not fully understanding why "flatten should probably be called at the beginning of the pass for each block structure (including those nested inside definitions) and not for each sub-block", because I think to uncover the `Begin`s and inspect the real head of a block, we only need to flatten the "toplevel" blocks (which forms a linked list), and I couldn't think of an implementation such that `flatten` is not called for sub-blocks...😨

LPTK:
OMG this new version looks so complicated... Not at all sure it's worth it or I can trust whether it works correctly.

What I was suggesting was super simple: just use your previous efficient flattening function (without any `lazy val`), but it should be called once on the top-level block of each block structure (ie not in the sub-blocks). This could be done in a separate transformer pass, which would also need to traverse inside definitions like function definitions and flatten their top-level blocks as well.

Does that make more sense to you?

ychenfo:
Thanks for the explanation but I think I am still confusing myself here... Do you mean the something like the following?

```scala
  // the flatten function that preserves the information in `End` by turning `b` into a `End => Block`
  def flattenConcat(b: End => Block): Block = this match
    case Match(scrut, arms, dflt, rest) => Match(scrut, arms, dflt, rest.flattenConcat(b))
    case Label(label, body, rest) => Label(label, body, rest.flattenConcat(b))
    case Begin(sub, rest) => sub.flattenConcat(_ => rest.flattenConcat(b)) // discard the `End` in the first block `sub`
    case TryBlock(sub, finallyDo, rest) => TryBlock(sub, finallyDo, rest.flattenConcat(b))
    case Assign(lhs, rhs, rest) => Assign(lhs, rhs, rest.flattenConcat(b))
    case a@AssignField(lhs, nme, rhs, rest) => AssignField(lhs, nme, rhs, rest.flattenConcat(b))(a.symbol)
    case AssignDynField(lhs, fld, arrayIdx, rhs, rest) => AssignDynField(lhs, fld, arrayIdx, rhs, rest.flattenConcat(b))
    case Define(defn, rest) => Define(defn, rest.flattenConcat(b))
    case HandleBlock(lhs, res, par, args, cls, handlers, body, rest) => HandleBlock(lhs, res, par, args, cls, handlers, body, rest.flattenConcat(b))
    case e: End => b(e)
    case _: BlockTail => this
  
  lazy val flattenTopLevel = this.flattenConcat(identity)
  
  // the block transformer that goes through nested blocks and flatten them
  object DeepFlattenTransformer extends BlockTransformer(new SymbolSubst()):
    override def applyBlock(b: Block): Block = super.applyBlock(b.flattenTopLevel)
  
  lazy val flatten = DeepFlattenTransformer.applyBlock(this)
```

Where I believe the `DeepFlattenTransformer` traverses inside definitions like function definitions and flattens them, and calls the `flattenTopLevel` once on the top-level of the incoming block (`b.flattenTopLevel`). But here I think I am making the mistake again of writing an unhelpful `lazy val` because `flattenTopLevel` is not recursively called..?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -762,7 +762,7 @@ object TrivialStatementsAndMatch:
 object MergeMatchArmTransformer extends BlockTransformer(new SymbolSubst()):
   override def applyBlock(b: Block): Block = super.applyBlock(b) match
     case m@Match(scrut, arms, Some(dflt), rest) =>
-      dflt.flatten match
+      dflt.flattened match
```
LPTK: Did you forget to update this part or are you still completely missing the point?

Your `MergeMatchArmTransformer` currently walks through every sub-block in the program and calls `flatten` on it, giving `MergeMatchArmTransformer` quadratic complexity. You see that, right? And you know it's not acceptable, right?

That's why I've repeatedly told you that if we go with this approach, flattening should be _**a separate pass**_ of the compiler that's done on the top level and just _**once**_, every time it happens that:
* the next transformation needs the structure to be flat; and
* the previous transformation might have left it not flat.

ychenfo: Indeed it seems that I was still missing the point and confused... Sorry for all the back and forth😰

I tried to sort out my thoughts again, and here are some questions that I have:

- May I confirm my understanding that now the problem is about the place where `flattened` is called (outside its own recursive definition), instead of the definition of `flattened` itself? It is wrong that I currently call `flattened` on this `dflt` at [line 765 of `Lowering.scala`](https://github.com/CrescentonC/mlscript/blob/9d797e4eda7af7d2c1bfa925472a4a8571867e32/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala#L765).

- > flattening should be a separate pass of the compiler that's done on the top level and just once
  
  Do you mean that before `MergeMatchArmTransformer.applyBlock` is called at [line 592 of `Lowering.scala`](https://github.com/CrescentonC/mlscript/blob/9d797e4eda7af7d2c1bfa925472a4a8571867e32/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala#L592-L595), we should flattened the block that's passed to `MergeMatchArmTransformer.applyBlock` first, like below?
  ```scala
  val res =
    if lowerHandlers then HandlerLowering().translateTopLevel(stackSafe)
    else stackSafe
  val resFlattened = res.flattened
  MergeMatchArmTransformer.applyBlock(resFlattened)
  ```

  But I am not sure if I understand what you said after that:
  > every time it happens that:
  > - the next transformation needs the structure to be flat; and
  > - the previous transformation might have left it not flat.

  It is correct that I understand it like this: `val resFlattened = res.flattened` happens because
  the next transformation (`MergeMatchArmTransformer.applyBlock`) needs the structure to be flat, and the previous transformation might have left it not flat (`res` may not be flat)?

Thanks a lot for your patience to explain things again and again to me...

LPTK: Yes. No worries!

ychenfo: Thanks for the confirmation! I have made the change. The change also reveals that the pretty printer for blocks (`:slot`) forgets to print out the `rest` of `Match` blocks, and I have also fixed this.

AnsonYeung:
I wonder why flatten isn't done in the following much simpler way:
```scala
object FlattenTransformer extends BlockTransformer(new SymbolSubst()):
    case Begin(sub, rst2) => sub match
      case Match(scrut, arms, dflt, rest) =>
        applyBlock(Match(scrut, arms, dflt, Begin(rest, rst2)))
      case Return(res, true) => applyBlock(sub)
      case Return(res, false) => Begin(applyBlock(sub), applyBlock(rst2))
      case Throw(exc) => applyBlock(sub)
      case Label(label, body, rest) => applyBlock(Label(label, body, Begin(rest, rst2)))
      case Break(label) => sub
      case Continue(label) => sub
      case Begin(sub2, rest) => applyBlock(Begin(sub2, Begin(rest, rst2)))
      case TryBlock(sub, finallyDo, rest) => applyBlock(TryBlock(sub, finallyDo, Begin(rest, rst2)))
      case Assign(lhs, rhs, rest) => applyBlock(Assign(lhs, rhs, Begin(rest, rst2)))
      case sub @ AssignField(lhs, nme, rhs, rest) => applyBlock(AssignField(lhs, nme, rhs, Begin(rest, rst2))(sub.symbol))
      case AssignDynField(lhs, fld, arrayIdx, rhs, rest) => applyBlock(AssignDynField(lhs, fld, arrayIdx, rhs, Begin(rest, rst2)))
      case Define(defn, rest) => applyBlock(Define(defn, Begin(rest, rst2)))
      case HandleBlock(lhs, res, par, args, cls, handlers, body, rest) => applyBlock(HandleBlock(lhs, res, par, args, cls, handlers, body, Begin(rest, rst2)))
      case HandleBlockReturn(res) => applyBlock(sub)
      case End(msg) => Begin(sub, applyBlock(rst2))
    case _ => super.applyBlock(b)
```
Note that for the current IR, this also handles lambda in awkward positions (as Result can be Path, Path can be Value, and Value can be lambda (or array containing lambda, and so on), so actually Value can contain arbitrary nested blocks within them bc of lambdas). I don't think these are handled by the original flatten code at all.

LPTK:
Thanks for your suggestion @AnsonYeung! The code you propose is fine (except for one small bug with the End case reported by Anto) and more compact. On the other hand, it allocated _a lot_ of intermediate data structures. Maybe when this code is ported to MLscript we can use Anto's deforestation to remove them! 😁 But in the meantime, I think we'll go with the more efficient approach.

LPTK:
@AnsonYeung PS: For lambdas, it doesn't matter. As you know, we're going to remove lambdas from values. Just need someone to actually do it.

LPTK (approved):
Finally LGTM 🙂

PR #277, opened 2025-02-12
Add the generalized Pratt parsing example

chengluyu:
* Add parsing and related tests to `apps/parsing/`.
* Add [railroad-diagrams](https://github.com/tabatkins/railroad-diagrams/blob/gh-pages/railroad.js) and the needed CSS file to `apps/parsing/vendors/`.
* Add parsing’s web demo to `apps/parsing-web-demo`.
* Remove generated .mjs files, except for `Runtime`, `RuntimeJS`, and `Predef`.
* Add .gitignore file to prevent .mjs files, except for these three, from being tracked by git.
* New stds: `BetterMap`, `Char`, `TreeTracer`.
* Improve existing stds.
* Use `ImplicitResolver` in `MLsCompiler` and fix issues found by it.
* If a class is declared with `declare`, the elaborator no longer select it `.class` for pattern matching.
* When the name of a MLscript function is not a valid JavaScript identifier, it won't be used as the generated JavaScript function name.
* Solve the issue that the `render` function was accessed through `globalThis` instead of `runtime`.
* Remove the meaningless `test` function in the `Option` module.


hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala
```diff
@@ -977,7 +977,7 @@ abstract class Parser(
                 yeetSpaces match
                   case (NEWLINE, l0) :: _ =>
                     consume
-                    ???
+                    throw new NotImplementedError(s"infix on newline at ${l0.showStart}:${l0.showEnd}")
```
LPTK:
```suggestion
                    lastWords(s"infix on newline at ${l0.showStart}:${l0.showEnd}")
```

hkmc2/shared/src/test/mlscript-compile/apps/Parser.mls
```diff
@@ -0,0 +1,804 @@
...
+fun opPrec(opStr) = if
+  opStr is FloatOperator then [charPrec(opStr.at(0)), charPrec(opStr.at(0))]
+  opStr hasLetter() then [Keywords.maxPrec, Keywords.maxPrec]
+  let leftPrec = charPrec of opStr.at(0)
+  let rightOp = opStr.at(-1)
+  let rightPrec = charPrec of rightOp
+  rightOp is RightAssociative then [leftPrec, rightPrec - 1]
+  else [leftPrec, rightPrec]
+
+opPrec(";")
```
LPTK: TODO mv to diff-test

hkmc2/shared/src/test/mlscript-compile/apps/Parser.mls
```diff
@@ -0,0 +1,804 @@
...
+
+opPrec("|>")
```
LPTK: TODO mv to diff-test

hkmc2/shared/src/test/mlscript-compile/apps/Parser.mls
```diff
@@ -0,0 +1,804 @@
...
+    Infix(op: Keyword, lhs: Tree, rhs: Tree) extends Tree
+    // For `let` bindings and `if`-`then`-`else`. The last part is optional.
+    Ternary(keyword: Keyword, lhs: Tree, rhs: Tree, body) extends Tree
+  
+  fun empty = Empty()
+  fun error(message: Str) = empty Error(message)
+  fun summary(tree) =
+    fun par(text: Str, cond: Bool) = if cond then "(" + text + ")" else text
+    fun prec(tree: Tree, side: Bool) = if tree is
+      Empty(_) then INT_MAX
```
LPTK: No need for the underscores

hkmc2/shared/src/test/mlscript/backlog/UCS.mls
```diff
@@ -110,3 +110,48 @@ if 1 + 1
...
+    "a" and a(idx + 1) is [idx', res] then ???
+    "b" and b(idx + 1) is [idx', res'] then ???
+    "c" and c(idx + 1) is [idx', res''] then ???
+    "d" and d(idx + 1) is [idx', res'''] then ???
+    "e" and e(idx + 1) is [idx', res''''] then ???
+    "f" and f(idx + 1) is [idx', res'''''] then ???
+    "g" and g(idx + 1) is [idx', res'''''''] then ???
+    else [idx, ""]
+  
+  launch(0)
```
LPTK:
```suggestion
  launch(0)

// ——— ——— ———

```

LPTK (commented):
Looks like the test outputs are not up to date.

hkmc2/shared/src/test/mlscript/apps/ParserTest.mls
```diff
@@ -405,10 +415,26 @@ end
...
+//│ >       Ident of "a"
+//│ >       Ident of "b"
+//│ >   rhs = Ident of "a"
+//│ >   body = Some of Ident of "b"
+
 :todo
 printParsed of """
 let f a b =
-  if a = b then
+  if a == b then
```
LPTK: Let's add a TODO to comment on this syntax design smell in the paper.

hkmc2/shared/src/test/mlscript/apps/ParserTest.mls
```diff
@@ -472,3 +539,149 @@ let translate p dx dy =
...
+//│ >               App:
+//│ >                 callee = Ident of "sum"
+//│ >                 arguments = Stack of 
+//│ >                   Ident of "t"
+//│ >     body = None
+
+printParsed of "begin end"
+//│ > Sequence of Nil
+
+:todo
```
LPTK: Why `:todo`?

hkmc2/shared/src/test/mlscript-compile/apps/Lexer.mls
```diff
@@ -0,0 +1,271 @@
...
+  fun char(idx: Int) = if idx < str.length
+    then (Some of str.charAt of idx) else None
+  
+  // Consume a sequence of characters that satisfy a predicate.
+  // The function returns the accumulated string and the following index.
+  fun take(pred: Str -> Bool, idx: Int, acc: Str): [Int, Str] =
+    while (char of idx) is Some(ch) and pred(ch)
+      then
+        set idx = idx + 1
+        set acc = acc + ch
```
LPTK:
```suggestion
        set
          idx += 1
          acc += ch
```

hkmc2/shared/src/test/mlscript/apps/ParserTest.mls
chengluyu: The parsed tree of `let a, b = (1, 2) in a + b` is incorrect.

LPTK: Ok, better fix it then.

hkmc2/shared/src/test/mlscript-compile/TreeTracer.mls
```diff
@@ -0,0 +1,45 @@
...
+      let postfix = if lines.length > 1 then " ↵" else ""
+      i is 0 then console.log(outie + innie + line + postfix)
+      i + 1 == lines.length then console.log(outie + innieAlt + line)
+      else console.log(outie + innieAlt + line + postfix)
+  fun enter(...pieces) =
+    output of
+      if steps > 0 then "│ ".repeat(steps - 1) + "├─" else ""
+      if steps > 0 then "┮ " else "┍ "
+      "│ "
+      fold (+) of ...pieces
```
LPTK: Why the space before the first argument list?
```suggestion
      fold(+) of ...pieces
```

hkmc2/shared/src/test/mlscript/decls/Prelude.mls
```diff
@@ -36,6 +36,8 @@ declare val Infinity
 
 // declare fun typeof: (Any) -> Str
 
+fun parseInt(str: Str, radix: Int): Int
+fun parseFloat(str: Str): Num
```
LPTK: What are these doing here? 🤔

LPTK: (Never mind; just realized they're global builtin JS functions.)

hkmc2/shared/src/test/mlscript-compile/TreeTracer.mls
chengluyu: Remove the space before the first argument list of `fold`

hkmc2/shared/src/test/mlscript-compile/Iter.mls
```diff
@@ -0,0 +1,95 @@
...
+
+module Result with
+  class Next(val value) with
+    val done = false
+  object Done with
+    val done = true
+
+module Iter with ...
+
+fun derive(iterable, makeNext) = Iterable of () =>
```
LPTK: The name is quite confusing. Why not
```suggestion
fun adaptIterator(iterable, makeNext) = Iterable of () =>
```

LPTK: Also why does it require all callers to wrap things in an obfuscating `() =>`?

hkmc2/shared/src/test/mlscript-compile/apps/ParseRule.mls
```diff
@@ -27,7 +29,7 @@ fun ensurePath(pathString) =
   pathString
 
 // `process.cwd()` is different in `hkmc2AllTests / test` and
-// `~hkmc2DiffTests / Test / run`. This is a workaround to make the path.
+// `~hkmc2DiffTests / Test / run`. This is a workaround to make the
```
LPTK: the?

hkmc2/shared/src/test/mlscript-compile/Iter.mls
chengluyu: I guess you're referring to the second parameter. This function takes the old iterator and produces a function that is going to be used as the `next` method on the new iterator. That's why it's called `makeNext`.

In summary, the function `derive` does the following things:
- takes an `Iterable`,
- creates an iterator by calling its `Symbol.iterator`,
- passed the iterator to `makeNext`, and
- uses the produced function as the `next` method on the new iterator.

hkmc2/shared/src/test/mlscript-compile/apps/ParseRule.mls
chengluyu: This is a typo. I fixed it.

hkmc2/shared/src/test/mlscript-compile/Iter.mls
chengluyu: (I didn't use the Commit suggestion feature to make the changes because the references below also needs to be updated.)

hkmc2/shared/src/test/mlscript-compile/Iter.mls
```diff
@@ -23,8 +23,21 @@ module Result with
...
+fun getIterator(something) =
+  let test = something.(Symbol.iterator)
+  if test is undefined then
+    throw TypeError of "Not an iterable: " + if
+      something is undefined then "undefined"
+      something is null then "null"
+      let prototype = Reflect.getPrototypeOf of something
+      prototype is null then "object"
+      else prototype.constructor.name
+  else something.(Symbol.iterator)()
```
LPTK: Why not
```suggestion
  else test.call(something)
```
Dynamic field accesses are expensive.

hkmc2/shared/src/test/mlscript-compile/Option.mls
```diff
@@ -22,3 +22,7 @@ class Both(fst, snd)
 fun getOrElse(opt, default) = if opt is
   Some(value) then value
   None then default
+
+fun get(opt) = if opt is
+  Some(value) then value
+  None then throw Error("None.get")
```
LPTK:
```suggestion
module unsafe with
  fun get(opt) = if opt is
    Some(value) then value
    None then throw Error("None.get")
```

hkmc2/shared/src/test/mlscript-compile/apps/Parser.mls
```diff
@@ -304,18 +307,17 @@ fun matchFunctionChoice() =
...
   matchWithChoice()
   matchFunctionChoice()
   ifThenElse()
-  Choice.keyword of
-    Keywords._underscore
-    "wildcard pattern"
-    Choice.end of Tree.Underscore()
+
+val moduleRules = rule of
+  "start of the statement"
```
LPTK:
```suggestion
  "start of statement"
```

hkmc2/shared/src/test/mlscript-compile/apps/parsing/ParseRule.mls
```diff
@@ -124,19 +124,16 @@ fun rule(name, ...choices) = ParseRule(name, choices Iter.toStack())
 
 module Choice with
   class Keyword(keyword: Keyword, rest: ParseRule)
-  class Expr(process: Tree -> Tree, rest: ParseRule)
-  class Type(process: Tree -> Tree, rest: ParseRule)
-  class Terminal(process: (Token, Tree) -> Tree, token: Token, rest: ParseRule)
+  class Expr(isType: Bool, process: Tree -> Tree, rest: ParseRule)
```
LPTK: Any reason for this design? Having separate Expr and Type seems cleaner, especially when we generalize the approach to languages with more categories (and possibly some added dynamically in the program itself).

hkmc2/shared/src/test/mlscript/apps/ParseRuleVisualizerTest.mls
```diff
@@ -2,11 +2,43 @@
...
-import "../../mlscript-compile/Stack.mls"
+import "../../mlscript-compile/Iter.mls"
+import "../../mlscript-compile/XML.mls"
 
-open ParseRuleVisualizer { save, rr, render }
-open Parser { prefixRules }
-open Stack
+open ParseRuleVisualizer { render, CSS_CONTENT, filePath }
+open Parser { exprRules, moduleRules }
+open XML { html, tag, elem, style }
```
LPTK: Why not just use `open XML` in the body of `toHTML`?

chengluyu: Because not only this function uses `XML`, but the following code also uses `elem` and `style`.

LPTK: Ok. But are there any name conflicts? Why all the explicit name imports?

LPTK: (It's true that they enhance readability, which is why in the future we'll insert them automatically. But doing so manually seems a bit painful.)

hkmc2/shared/src/test/mlscript-compile/apps/parsing/Rules.mls
```diff
@@ -342,29 +343,71 @@ val typeInfixRule = rule of
...
+          (ident, _) => Some(ident)
+          "the type constructor's name"
+  // Infix rules
   Choice.Ref of
     "type"
     (lhs, compose) => compose(lhs)
     None
     None
     typeInfixRule
-
```
LPTK: noooooo

hkmc2/shared/src/test/mlscript/HkScratch.mls
LPTK, line 1: Did you forget to clean HkScratch?

hkmc2/shared/src/test/mlscript/backlog/UCS.mls
```diff
@@ -158,3 +158,25 @@ fun parse(text: Str): [Int, Str] =
...
+            other then Some(go(other))
+        else
+          console.warn("expect the choice to be a bracketed term but found " + choice)
+          None
+      else
+        console.warn("expect a the category to be an identifier but found " + category)
+        None
+    else
+      console.warn("expect the definition to be a tuple but found " + tree)
+      None
```
LPTK: What's wrong with this one? There's no error report in the test case.

LPTK:
```suggestion
      None

// ——— ——— ———

```

hkmc2/shared/src/test/mlscript/codegen/ImportMLsJS.mls
```diff
@@ -18,14 +18,6 @@ Option.isDefined(Option.None)
 Option.None Option.isDefined()
 //│ = false
 
-
```
LPTK, line 21: Pls don't remove the empty line.

chengluyu: Fixed.

hkmc2/shared/src/test/mlscript-compile/apps/parsing/Rules.mls
```diff
@@ -260,8 +246,20 @@ val typeRule = rule of
       keyword(Keywords._rightRound) of // either an identifier or nothing
         end of None
         ident of process: someFirst, name: "the type constructor's name"
-  // Infix rules
-  typeExpr of process: pipeInto, choices: typeInfixRule
+  // "infix" rule starting with <type>
+  typeExpr of
+    process: pipeInto
```
LPTK, line 250: `pipeInto` should probably be a default value of the parameter. You can implement this logic easily since it's a named parameter.

hkmc2/shared/src/test/mlscript/backlog/UCS.mls
```diff
@@ -159,34 +159,6 @@ fun parse(text: Str): [Int, Str] =
...
-        None
-    else
-      console.warn("expect the definition to be a tuple but found " + tree)
-      None
-//│ ╔══[ERROR] mismatched arity: expect 3, found 1
-//│ ║  l.173: 	            Tuple(elements) then Some(go(elements))
-//│ ╙──       	            ^^^^^^^^^^^^^^^
-
-// ——— ——— ———
-
```
LPTK: Note that I already changed this test in the other branch. Please don't mess up the conflict resolution.

chengluyu: Sure. I just made sure that this file is not in the change list.

hkmc2/shared/src/main/scala/hkmc2/MLsCompiler.scala
```diff
@@ -80,13 +82,15 @@ class MLsCompiler(preludeFile: os.Path, mkOutput: ((Str => Unit) => Unit) => Uni
...
+      val resolver = ImplicitResolver(rtl)
+      resolver.resolveBlk(blk0)(using ImplicitResolver.ICtx.empty)
       val blk = new semantics.Term.Blk(
         semantics.Import(State.runtimeSymbol, runtimeFile.toString) :: blk0.stats,
         blk0.res
       )
       val low = ltl.givenIn:
         new codegen.Lowering()
-          with codegen.LoweringSelSanityChecks
+          // with codegen.LoweringSelSanityChecks
```
chengluyu: Is this still necessary?

I briefly tested the runtime of the test. With `codegen.LoweringSelSanityChecks`:
```
[success] Total time: 22 s, completed Apr 19, 2025, 3:18:10 PM
```
Without `codegen.LoweringSelSanityChecks`:
```
[success] Total time: 22 s, completed Apr 19, 2025, 3:20:53 PM
```

hkmc2/shared/src/test/mlscript-compile/apps/parsing/Rules.mls
chengluyu: The default value or `process` is `tuple`. The second argument isn't always a function.

LPTK: Oh, I see.

hkmc2/shared/src/main/scala/hkmc2/MLsCompiler.scala
LPTK: What do you mean by "still necessary"?

Sanity checks are currently disabled in compiler tests. Of course the running time will be the same: you're running the same thing twice...

LPTK: My plan is to enable light sanity checks by default. Haven't gotten around to it, yet.

chengluyu: I didn't make myself clear. What I want to ask is whether to keep this as a comment.

> Sanity checks are currently disabled in compiler tests. Of course the running time will be the same: you're running the same thing twice...

Thanks, I just learned about this.

LPTK: Why make this change in the first place? We shouldn't be removing the sanity checks mixin.

chengluyu: I should mention the context. Previously, some issues in the sanity check led to a stack overflow, and then you introduced this fix. Now it should be reverted, right?

LPTK: Oh, I remember, now! This was so long ago... Yes, if the fix is no longer needed, it should be reverted.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -270,7 +270,7 @@ extends Importer:
     case S(cls: ClassSymbol) =>
       trm
     case S(mem: BlockMemberSymbol) =>
-      if !mem.hasLiftedClass then trm
+      if !mem.hasLiftedClass || mem.defn.exists(_.isDeclare.isDefined) then trm
```
LPTK, line 274: Why is this change needed? As you know, the elaborator can't reliably access `defn`s since they may not yet be available.

hkmc2/shared/src/test/mlscript-compile/BetterMap.mls
```diff
@@ -0,0 +1,41 @@
+import "./Option.mls"
+
+open Option { Some, None }
+
+module BetterMap with ...
+
+class Map with
```
LPTK: In MLscript, we prefer the use of module functions over methods. Any reason this is defined as a `class` with methods?

hkmc2/shared/src/test/mlscript-compile/Stack.mls
```diff
@@ -12,7 +12,7 @@ module Stack[T] with ...
 // TODO support type/module overloading
 type T[A]
 
-data class (::) Cons[A](head: A, tail: Stack[A])
+data class (::) Cons[A](head: A, tail: Cons[A] | Nil)
```
LPTK: I guess this change is to avoid the type parameter number check? How about
```suggestion
data class (::) Cons[A](head: A, tail: T[A])
```

hkmc2/shared/src/test/mlscript-compile/Stack.mls
```diff
@@ -78,5 +78,3 @@ fun filter(xs, f) = if xs is
     f(head) then head :: filter(tail, f)
     else filter(tail, f)
   Nil then Nil
-
-
```
LPTK: 👎

hkmc2/shared/src/test/mlscript-compile/apps/parsing-web-demo/index.html
```diff
@@ -0,0 +1,264 @@
...
+      cursor: help;
+    }
+
+    svg.railroad-diagram g.diagram-text:hover path.diagram-text {
+      fill: #eee;
+    }
+  </style>
+</body>
+
+</html>
```
LPTK:
```suggestion
</html>

```

hkmc2/shared/src/test/mlscript-compile/apps/parsing-web-demo/main.mls
```diff
@@ -0,0 +1,223 @@
...
+import "../parsing/Rules.mls"
+import "../parsing/vendors/railroad/railroad.mjs"
+
+open XML { elem }
+open Predef
+open Option { Some, None }
+
+declare val document
+declare val customElements
+declare class HTMLElement
```
LPTK: Maybe add these to `Prelude.mls`?

hkmc2/shared/src/test/mlscript-compile/apps/parsing-web-demo/main.mls
```diff
@@ -0,0 +1,223 @@
...
+hello "Rob" hello "Bob" goodbye
+  
+#diagram ""
+"""
+
+showExample1()
+
+editor.addEventListener of "keydown", event =>
+  if event.key is "Tab" do
+    event.preventDefault()
```
LPTK, line 54: Curious: what is this for?

hkmc2/shared/src/test/mlscript-compile/apps/parsing-web-demo/main.mls
```diff
@@ -0,0 +1,223 @@
...
+    let start = editor.selectionStart
+    let end = editor.selectionEnd
+    set editor.value = editor.value.substring(0, start) + "  " + editor.value.substring(end)
+    set editor.selectionEnd = start + 2
+    set editor.selectionStart = editor.selectionEnd
+
+selector.addEventListener of "change", event => if selector.value is
+  "example-1" do showExample1()
+  "example-2" do showExample2()
```
LPTK: This inexhaustive match looks a bit uncomfortable. Couldn't we set all examples, along with their names and titles, in one place, and then use this info in the relevant places accordingly? The HTML `select` list could start with no `option` and would be updated by the JS based on this single source of truth.

hkmc2/shared/src/test/mlscript-compile/apps/parsing-web-demo/main.mls
```diff
@@ -0,0 +1,223 @@
...
+      let indent = line.match(indentRegex).[1].length
+      let text = line.substring(indent)
+      while indent <= stack.[stack.length - 1].indent do
+        stack.pop()
+      let newNode = (text: text, children: [])
+      stack.[stack.length - 1].node.children.push of newNode
+      stack.push of node: newNode, indent: indent
+  root.children
+
+let errorDisplayStyle = """
```
LPTK, line 104: Any reason this is not defined in the HTML?

hkmc2/shared/src/test/mlscript-compile/apps/parsing/Extension.mls
```diff
@@ -0,0 +1,114 @@
...
+            Tree.Bracketed(Token.Square, Tree.Tuple(elements)) then
+              Some of tuple of
+                categoryName
+                go(elements) Choice.map(op)
+            Tree.Bracketed(Token.Square, other) then
+              Some of tuple of
+                categoryName
+                go(other :: Nil) Choice.map(op)
+        else
+          print of "Expect the choiceTree to be a bracketed term but found " + choiceTree Tree.summary()
```
LPTK: Minor suggestion:
```suggestion
          print of "Expect the choiceTree to be a bracketed term but found", choiceTree Tree.summary()
```
Same below.

hkmc2/shared/src/test/mlscript-compile/apps/parsing/Precedence.mls
LPTK, line 1: Feels like this file should be called `Keywords` and `Precedence` should be a module inside `Keywords`, not the other way around...

hkmc2/shared/src/test/mlscript-compile/apps/parsing/Rules.mls
```diff
@@ -0,0 +1,371 @@
...
+
+define("pattern-list") of term of
+  process: (head, tail) => head :: tail
+  name: "pattern"
+  choices: tuple of reference("pattern-list") of
+    process: idFirst, name: "pattern list tail"
+
+define("multiple-matching") of reference("pattern-list") of
+  process: Tree.infix of Keywords._thinArrow
+  name: "the list of patterns"
```
LPTK: "the" list of patterns? Why not
```suggestion
  name: "list of patterns"
```

hkmc2/shared/src/test/mlscript-compile/apps/parsing/vendors/railroad/railroad.mjs
LPTK, line 1: I think the containing folder should be `vendored`.

hkmc2/shared/src/test/mlscript/apps/parsing/output/rule.html
LPTK, line 1: This should probably be git-ignored.

LPTK (approved):
Beautiful! I have a few comments, but otherwise LGTM.

hkmc2/shared/src/test/mlscript-compile/apps/parsing-web-demo/index.html
chengluyu: Fixed in [1e7429a](https://github.com/hkust-taco/mlscript/pull/277/commits/1e7429adc0405bd6ade92c6ea4dcf612d70fc1a4)

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
chengluyu: If a class is declared with declare, the elaborator no longer select its `.class` for pattern matching.

> the elaborator can't reliably access defns since they may not yet be available

~~Good point. Let me think about it.~~ I think most declared classes are used after their declarations. We can address it along with the issue of class parameters after Harry's PR is completed.

hkmc2/shared/src/test/mlscript-compile/Stack.mls
chengluyu: It works.

hkmc2/shared/src/test/mlscript-compile/apps/parsing-web-demo/main.mls
chengluyu: We can start by doing it this way. But I would like to eventually have a `WebPrelude.mls` specifically for web, as there will be many things that need to be declared.

hkmc2/shared/src/test/mlscript-compile/apps/parsing-web-demo/main.mls
chengluyu: Make pressing <kbd>⇥ Tab</kbd> add two spaces at the caret's position. `event.preventDefault()` cancels the default behavior of the browser.

hkmc2/shared/src/test/mlscript-compile/apps/parsing-web-demo/main.mls
chengluyu: Done in 828589c.

hkmc2/shared/src/test/mlscript-compile/apps/parsing-web-demo/main.mls
chengluyu: The content in custom elements is placed inside the [ShadowDOM](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM), which is scoped. Therefore, styles placed directly in HTML won't be effective on elements inside custom elements. I can follow the method of cloning templates in [this tutorial](https://javascript.info/shadow-dom-style), but it's a bit complicated. Writing CSS directly here is more convenient.

hkmc2/shared/src/test/mlscript-compile/BetterMap.mls
chengluyu: I wasn't sure at the time. I just wanted to wrap the built-in `Map`.

Do you want me to change them into standalone functions? Maybe I can rename it `MapOps` and provide a bunch of functions that operate on the built-in `Map`.

hkmc2/shared/src/test/mlscript-compile/apps/parsing/vendors/railroad/railroad.mjs
chengluyu: I thought vendor can be only noun according to the dictionaries (e.g., [Cambridge](https://dictionary.cambridge.org/dictionary/english/vendor)). In actual corpora (e.g., [Linguee](https://www.linguee.com/english-french/search?query=vendored)), the use of `vendored` also seems to be relatively rare.

After all, many web frameworks are using `vendor` or `vendors` as name of the folder for third-party libraries: https://stackoverflow.com/questions/16865980/what-does-vendor-mean-in-web-file-structure

hkmc2/shared/src/test/mlscript-compile/apps/parsing/Precedence.mls
chengluyu: Done in ee2b46c.

hkmc2/shared/src/test/mlscript/apps/parsing/output/rule.html
chengluyu: Done in 06d6740.

hkmc2/shared/src/test/mlscript-compile/BetterMap.mls
LPTK: > I wasn't sure at the time. I just wanted to wrap the built-in `Map`.

Sure.

> Do you want me to change them into standalone functions? Maybe I can rename it `MapOps` and provide a bunch of functions that operate on the built-in `Map`.

I suggest to keep the name `Map` and rename the wrapper to `BetterMap`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: It's probably fine to just add a FIXME here for now. This part of the Elaborator logic will be thoroughly refactored when we add overloading.

hkmc2/shared/src/test/mlscript-compile/apps/parsing-web-demo/main.mls
LPTK: Yes, it would be better, esp. since other dialects/backends shouldn't have these in their prelude.

hkmc2/shared/src/test/mlscript-compile/apps/parsing-web-demo/main.mls
LPTK: Oh, cool. Why not document that in the code? (Is there a link that describes this? StackOverflow or other)

hkmc2/shared/src/test/mlscript-compile/apps/parsing-web-demo/main.mls
LPTK: Yes, for sure it's better than anything complicated. Please document what you wrote above in a code comment, though.

hkmc2/shared/src/test/mlscript-compile/apps/parsing-web-demo/main.mls
```diff
@@ -78,9 +43,10 @@ editor.addEventListener of "keydown", event =>
     set editor.selectionEnd = start + 2
     set editor.selectionStart = editor.selectionEnd
 
-selector.addEventListener of "change", event => if selector.value is
-  "example-1" do showExample1()
-  "example-2" do showExample2()
+selector.addEventListener of "change", event =>
+  if examples.get(selector.value) is
+    Some(example) do set editor.value = example.source
+    None do print("Example \"" + selector.value + "\" not found")
```
LPTK: Shouldn't we throw an exception so that it appears as an error in the web console?

hkmc2/shared/src/test/mlscript-compile/BetterMap.mls
chengluyu: I did. I also renamed `BetterMap` to `WrappedMap` as the implementation isn't much better.

LPTK: As I said, I think you should name this module `Map`. Code like `WrappedMap.insert(...)` looks very ugly; I want to write `Map.insert(...)`. And again, please convert the methods into module functions.

chengluyu: Oh, sorry, I misunderstood you. I'm almost done, but I found that I can't seem to change the name of this module to `Map` for two reasons:
1. Inside this module, we need to use `globalThis` to access the built-in `Map`. If the `Map` on `globalThis` is modified, it will cause problems. You've also mentioned avoiding the use of `globalThis` and better to use `Runtime`.
2. When importing it in DiffTests, it set the `Map` on `globalThis` to this module. This way, we can no longer access the built-in `Map`.

LPTK: I just noticed that the diff-tests still mess with `globalThis` when importing stuff, generating code like `this.Map = await import("...")`. This is completely wrong and unnecessary. Imports produce a symbol that can be referred to directly through the current `Scope` (see how imports are compiled for module files). Fixing this will fix your problems.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -620,7 +617,7 @@ trait JSBuilderArgNumSanityChecks(using Config, Elaborator.State)
       val paramRest = params.restParam.map(p => Scope.scope.allocateName(p.sym))
       val paramsStr = Scope.scope.allocateName(functionParamVarargSymbol)
       val functionName = JSBuilder.makeStringLiteral(name.fold("")(n => s"${JSBuilder.escapeStringCharacters(n)}"))
-      val checkArgsNum = doc"\nglobalThis.Predef.checkArgs($functionName, ${params.paramCountLB}, ${params.paramCountUB.toString}, $paramsStr.length);"
+      val checkArgsNum = doc"\nruntime.checkArgs($functionName, ${params.paramCountLB}, ${params.paramCountUB.toString}, $paramsStr.length);"
```
LPTK: Why not use the runtime symbol here?

hkmc2/shared/src/test/mlscript-compile/Predef.mls
```diff
@@ -36,8 +36,6 @@ val render = Rendering.render
 fun print(...xs) =
   console.log(...map(renderAsStr)(...xs))
 
-fun printRaw(x) =
-  console.log(render(x))
```
LPTK, line 40: You forgot to remove the empty line after this definition.

hkmc2/shared/src/test/mlscript-compile/Predef.mls
```diff
@@ -69,45 +67,10 @@ fun mkStr(...xs) =
...
-  fun log(msg) =
-    if enabled then
-      console.log("| ".repeat(indentLvl) + msg.replaceAll("\n", "\n" + "  ".repeat(indentLvl)))
-    else ()
-
-
-data class Test with
-  print("Test")
-  val y = 1
+
```
LPTK, line 78: Why so many empty lines?

hkmc2/shared/src/test/mlscript/codegen/GlobalThis.mls
```diff
@@ -44,7 +44,7 @@ foo()
 //│   }
 //│ };
 //│ foo()
-//│ ═══[RUNTIME ERROR] TypeError: Cannot read properties of undefined (reading 'checkArgs')
+//│ ═══[RUNTIME ERROR] TypeError: globalThis.Error is not a constructor
```
LPTK, line 47: Looks like a problem here

chengluyu: No, this is because previously `checkArgs` would check `globalThis.Error`, but `checkArgs` was accessed through `globalThis.Predef`. At the beginning of this test file, `globalThis` was shadowed, which made the previous error. However, now `checkArgs` is accessed directly through the local variable `Predef`, so the error disappeared.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
chengluyu: Fixed in 6b9eb01.

hkmc2/shared/src/test/mlscript-compile/Predef.mls
chengluyu: Fixed in cf8c295.

hkmc2/shared/src/test/mlscript-compile/Predef.mls
chengluyu: Fixed in cf8c295.

chengluyu:
Why does the following error only appear in CI?

```
╔══[ERROR] Module 'Math' does not contain member 'floor'
║  l.48: 	      mid = Math.floor of (begin + end) / 2
╙──      	                ^^^^^^
/!!!\ Error in parsing/Token.mls /!!!\
╔══[ERROR] Module 'Math' does not contain member 'floor'
║  l.54: 	      set mid = Math.floor of (begin + end) / 2
╙──      	                    ^^^^^^
```

OK. It's because we didn't list the members of `Math`.

But instead of this:

```
declare module Math with
   declare
    val
      E
      LN10
      LN2
      LOG10E
      LOG2E
      PI
      SQRT1_2
      SQRT2
```

can we do:

```
declare module Math with
   declare
    val
      E; LN10; LN2; LOG10E; LOG2E; PI; SQRT1_2; SQRT2
```

I tried but it didn't work.

hkmc2/shared/src/test/mlscript/codegen/GlobalThis.mls
LPTK: You need to update the comment:
```diff
-// * This one uses `globalThis.Predef` and `globalThis.Error`
-// * Notice the failed exception throw
+// * This one uses `globalThis.Error`
+// * Notice the failed exception throw
```

LPTK:
> Why does the following error only appear in CI?

Did it really only appear in CI? 🤔 

> can we do:
> 
> ```
> declare module Math with
>    declare
>     val
>       E; LN10; LN2; LOG10E; LOG2E; PI; SQRT1_2; SQRT2
> ```
> 
> I tried but it didn't work.

You already know that `;` is the effect sequencing operator and indented blocks are equivalent to the lines separated by `,`, don't you?

hkmc2/shared/src/test/mlscript/codegen/GlobalThis.mls
chengluyu: Fixed in f2dec9e.

chengluyu:
> Did it really only appear in CI? 🤔

image

I don't know why, but there really wasn't any error on my side. I find it very strange.

> You already know that ; is the effect sequencing operator and

I forgot this. :-P 

> indented blocks are equivalent to the lines separated by ,, don't you?

I found this later by looking into `hkmc2/shared/src/test/mlscript/parser/Block.mls`

hkmc2/shared/src/test/mlscript-compile/apps/parsing-web-demo/main.mls
```diff
@@ -0,0 +1,189 @@
...
+fun makeFigures(entries) = entries
+  Iter.mapping of case [name, svg] then
+    elem("figure") of elem("figcaption")(name), svg
+  Iter.joined("")
+
+fun displayRules() =
+  open ParseRuleVisualizer
+  open Rules
+  reset()
+  set query("#syntax-diagrams>main").innerHTML = Iter.joined(_, "") <| tuple of
```
LPTK: Why not just
```suggestion
  set query("#syntax-diagrams>main").innerHTML = Str.concat of
```
?

hkmc2/shared/src/test/mlscript-compile/apps/parsing/Keyword.mls
LPTK, line 1: Does this deserve its own file? Why not put it in `Keywords.mls`?

hkmc2/shared/src/test/mlscript/apps/parsing/Type.mls
LPTK, line 1: Why not call it `TypeTest.mls`?

hkmc2/shared/src/test/mlscript-compile/apps/parsing-web-demo/Examples.mls
LPTK, line 1: Please add a diff-test file that checks these work.

LPTK (approved):
Great work, thanks!

LPTK:
I made a merge commit for this one because there's _a lot_ of mostly clean history that I don't want to squash.

chengluyu:
> I made a merge commit for this one because there's a lot of mostly clean history that I don't want to squash.

Oh, thanks. I feel that the effort I put into writing commit messages has been recognized.

LPTK:
Unfortunately, there are a couple unsightly WIP commits at the very beginning https://github.com/hkust-taco/mlscript/pull/277/commits/0ab1110b3edf0d784c110d40a9c47efe62bf466f  😅

PR #279, opened 2025-02-12
Add infrastructure for quasiquote backend and runtime

NeilKleistGao:
- Add runtime definitions of terms
- Lower quoted terms
- Add `print` and `codegen` for supported quoted terms (e.g., `Lam`, `App`, simple forms of `IfLike`)
- Support cross-stage and import path resolving
- Fix Watcher's current work directory to make it consistent with the `hkmc2AllTests`
- <del>Fix unexpected indents for multi-line lambda applications</del>
- Add ADT syntax and corresponding desugaring
- Add a missing test case for BbML
- Port `id"..."` for identifier escaping (#172 )


hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -141,6 +141,8 @@ object Elaborator:
     val suid = new Uid.Symbol.State
     given State = this
     val globalThisSymbol = TopLevelSymbol("globalThis")
+    // a builtin symbol import can be used for getting current file pathes via import.meta
```
```suggestion
    // a builtin symbol import can be used for getting current file paths via import.meta
```

hkmc2/shared/src/main/scala/hkmc2/syntax/Lexer.scala
```diff
@@ -397,7 +397,15 @@ class Lexer(origin: Origin, dbg: Bool)(using raise: Raise):
...
               // * which results in token stream `|module| |P| |with|→|(|→|2|)|←|↵|1|`.
               // * So this code commutes the indent/deindent with the open/close parentheses.
-              go(CLOSE_BRACKET(Indent) -> l1.left :: (CLOSE_BRACKET(k1), l1) :: OPEN_BRACKET(Indent) -> l1.right :: rest, false, stack, acc)
+              rest match
+                case (OPEN_BRACKET(k2), l0) :: rest if k1 == Round && k2 == Round =>
+                  // * Special case for immediate lambda applications; e.g.,
+                  // * (x =>
+                  // *   x + 1)(0)
+                  // * do not insert indents around `(0)`
+                  go(CLOSE_BRACKET(Indent) -> l1.left :: (CLOSE_BRACKET(k1), l1) :: (OPEN_BRACKET(k1), l1) :: rest, false, stack, acc)
```
LPTK: This is a hack. Please remove it and either fix the real problem at its source or open an issue for it.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -141,7 +153,11 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
             return Assign(l, Call(Value.Ref(bs), Nil)(true, true), k(Value.Ref(l)))
         case S(_) => ()
         case N => () // TODO panic here; can only lower refs to elab'd symbols
-      case _ => ()
+      case _ =>
+        val subst = summon[Subst]
```
LPTK: This line can be removed.

hkmc2/shared/src/main/scala/hkmc2/utils/Scope.scala
```diff
@@ -118,8 +118,9 @@ object Scope:
   
   def scope(using scp: Scope): Scope = scp
   
-  def empty(using State): Scope =
-    Scope(N, S(S(State.globalThisSymbol)), MutMap.empty)
+  def empty(using State): Scope = 
+    val state = summon[State]
```
LPTK:
```suggestion
```

hkmc2/shared/src/test/mlscript-compile/Term.mls
```diff
@@ -0,0 +1,192 @@
...
+builtinSymbols.add("*")
+builtinSymbols.add("/")
+builtinSymbols.add("==")
+builtinSymbols.add("<")
+builtinSymbols.add(">")
+builtinSymbols.add(">=")
+builtinSymbols.add("<=")
+
+
+let names = new globalThis.Map()
```
LPTK: I don't thin this is needed
```suggestion
let names = new Map()
```

hkmc2/shared/src/test/mlscript-compile/Term.mls
```diff
@@ -0,0 +1,192 @@
...
+fun freshName(name) =
+  if names.has(name) == false do
```
LPTK:
```suggestion
  if not(names.has(name)) do
```

hkmc2/shared/src/test/mlscript-compile/Term.mls
```diff
@@ -0,0 +1,192 @@
...
+    names.set(name, 0)
+  let i = names.get(name)
+  names.set(name, i + 1)
+  Str.concat(name, "_", i.toString())
+    
+
+class Context(symbols: globalThis.Set, dependencies: globalThis.Set, printOnly: Bool) with
```
LPTK:
```suggestion
class Context(symbols: Set, dependencies: Set, printOnly: Bool) with
```

hkmc2/shared/src/test/mlscript-compile/quotes/CSPBar.mls
```diff
@@ -0,0 +1,4 @@
+import "../Example.mls"
+module CSPBar with ...
+fun res =
+  ((Example).inc)(0)
```
LPTK: The generated code should always finish with an empty line.

```suggestion
  ((Example).inc)(0)

```

hkmc2/shared/src/test/mlscript/codegen/Quasiquotes.mls
```diff
@@ -0,0 +1,214 @@
...
+
+
+term.print(x `=> `if x `== `0.0 then `1.0 else x)
+//│ > (x_9) =>
+//│ >   if 
+//│ >     let scrut_3 = (==)(x_9, 0)
+//│ >     scrut_3 is true then 1
+//│ >     else x_9
+
+import "../../mlscript-compile/QuoteExample2.mls"
```
LPTK: All these importing tests should probably be moved to a different file.

LPTK (commented):
Nice, looks promising! But please consider these comments first.

hkmc2/shared/src/main/scala/hkmc2/syntax/Lexer.scala
NeilKleistGao: This problem looks a bit tricky. I've opened an issue (#281) for it and left a `fixme` test case.

hkmc2/shared/src/test/mlscript-compile/Term.mls
LPTK: Please explain why you didn't apply the suggestion before resolving the conversation. I don't like when you do this.

https://github.com/hkust-taco/mlscript/pull/279/files#diff-070cb7224c53a3d478f0a055952d7646b2a74b9991eba9f085cf7d4f0c109eebR100

NeilKleistGao: Here, `not` is defined in `Predef.mls`. Since `Predef.mls` imported `Term.mls`, `Term.mls` importing `Predef.mls` will lead to cycle dependency (I think it was introduced in a recent commit). That's why I only used `is false` instead of `not`.

LPTK: There's no problem with such cyclic dependencies in JS actually. Please add `not` to `Predef` and use it.

And anyway, you should use the new `runtimeSymbol` approach to access the `Term` module. We want to move away from the ugly `globalThis.Predef` hack and instead use code like `Select(Value.Ref(State.runtimeSymbol), Tree.Ident("term"))`. To do this you'll have to move the `term` forwarding value from `Predef.mls` to `Runtime.mls`.

NeilKleistGao: Fixed.

hkmc2/shared/src/test/mlscript-compile/quotes/CSPBar.mls
LPTK: You should be careful when pretty-printing application terms: in JS `x.f(a)` is _not_ the same as `(x.f)(a)`.

hkmc2/shared/src/test/mlscript-compile/Term.mls
```diff
@@ -0,0 +1,183 @@
...
+class Blk(stats: Array[Statement], res: Term) extends Term
+class New(cls: Term, args: Array[Term]) extends Term
+class Region(name: Symbol, body: Term) extends Term
+class RegRef(reg: Term, value: Term) extends Term
+class Assgn(lhs: Term, rhs: Term) extends Term
+class Deref(ref: Term) extends Term
+class SetRef(ref: Term, value: Term) extends Term
+class Ret(result: Term) extends Term
+class Throw(result: Term) extends Term
+class Try(body: Term, finallyDo: Term) extends Term
```
LPTK: So... why not use the ADT syntax, here?

```fs
class Term with
  constructor
    Lit...
    ...
```

If not already done, it should be easy to desugar into the above before elaboration (in `desugStmts`).

hkmc2/shared/src/test/mlscript-compile/Term.mls
```diff
@@ -120,6 +126,11 @@ fun showPattern(p, ctx) =
   // TODO: other patterns
 
 
+fun paren(t, ctx) =
+  if (t is Ref) || (t is CSRef) || (t is Lit) || (t is Sel) then show(t, ctx)
+  else Str.concat("(", show(t, ctx), ")")
```
LPTK: Pro tip
```suggestion
  else Str.concat of "(", show(t, ctx), ")"
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -30,23 +31,34 @@ object Thrw extends TailOp:
   def apply(r: Result): Block = Throw(r)
 
 
-// * No longer in meaningful use and could be removed if we don't find a use for it:
-class Subst(initMap: Map[Local, Value]):
+// * Used by quasiquote lowering
+// * e.g., x `=> {body} will allocate a next-stage binding and store it in `x`
+// * In `{body}`, one can refer to the staged code fragment via `x`
+// * Subst maintains the mapping from symbols like `x` to the actual code data
```
LPTK: I don't see why this is needed. Why not just assign and reference the original `x` symbol? Why do we need a mapping?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -30,23 +31,34 @@ object Thrw extends TailOp:
...
+class Subst(initMap: Map[(Local, Bool), Value]):
   val map = initMap
-  /*
-  def +(kv: (Local, Value)): Subst =
+
+  def +(kv: ((Local, Bool), Value)): Subst =
```
LPTK: What is this new magical undocumented Boolean?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -30,23 +31,34 @@ object Thrw extends TailOp:
...
+
   def apply(v: Value): Value = v match
-    case Value.Ref(l) => map.getOrElse(l, v)
+    case Value.Ref(l) => map.getOrElse(l -> false, v)
     case _ => v
 object Subst:
   val empty = Subst(Map.empty)
   def subst(using sub: Subst): Subst = sub
+
+  val builtinOps = Set(
```
LPTK: Why is there a spearate set of builtin symbols? We already have such a set in the `Elaborator` companion object.

hkmc2/shared/src/test/mlscript-compile/Term.mls
```diff
@@ -0,0 +1,194 @@
...
+class Split with
+  constructor
+    Cons(head: Branch, tail: Split)
+    Let(sym: Symbol, term: Term, tail: Split)
+    Else(default: Term)
+    End
+
+
+object KeywordIf
+object KeywordWhile
```
LPTK: Why not
```suggestion
module Keyword with
  object
    If
    While
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -536,6 +553,170 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
+
+  def setupTerm(name: Str, args: Ls[Path])(k: Result => Block)(using Subst): Block =
+    k(Instantiate(Value.Ref(State.runtimeSymbol).selSN("term").selSN(name), args))
+
+  def setupFreshName(name: Str)(k: Result => Block)(using Subst): Block =
+    k(Call(
+      Value.Ref(State.runtimeSymbol).selSN("term").selSN("freshName"),
+      Value.Lit(Tree.StrLit(name)).asArg :: Nil)(true, true))
+
+  def setupObj(name: Str): Path =
```
LPTK: Obj? The naming of this function is weird.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -536,6 +553,170 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
+    val state = summon[State]
+    Value.Ref(state.importSymbol).selSN("meta").selSN("url")
+
+  def quote(t: st)(k: Result => Block)(using Subst): Block = t match
+    case Lit(lit) =>
+      setupTerm("Lit", Value.Lit(lit) :: Nil)(k)
+    case Ref(sym) if Subst.builtinOps.contains(sym.nme) => // builtin symbols
+      val l = new TempSymbol(N)
+      setupTerm("Builtin", Value.Lit(Tree.StrLit(sym.nme)) :: Nil)(k)
+    case Ref(sym) => // intermediate temp variables generated by ucs
```
LPTK: Why would this be specific to the UCS?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -536,6 +553,170 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
+          val l3 = new TempSymbol(N)
+          Assign(l3, r3, setupTerm("Branch", (l1 :: l2 :: l3 :: Nil).map(s => Value.Ref(s))): r4 =>
+            val l4 = new TempSymbol(N)
+            Assign(l4, r4, quoteSplit(tail): r5 =>
+              val l5 = new TempSymbol(N)
+              Assign(l5, r5, setupTerm("Cons", (l4 :: l5 :: Nil).map(s => Value.Ref(s)))(k))
+            )
+          )
+        )
+      )
```
LPTK: Insteaf of these big nested blocks, why not just define all temporaries at the top
```scala
      val l1, l2, l3, l4, l5 = new TempSymbol(N)
```
and then use `blockBuilder.assign(...).....`?

hkmc2/shared/src/main/scala/hkmc2/utils/Scope.scala
```diff
@@ -118,8 +118,8 @@ object Scope:
   
   def scope(using scp: Scope): Scope = scp
   
-  def empty(using State): Scope =
-    Scope(N, S(S(State.globalThisSymbol)), MutMap.empty)
+  def empty(using state: State): Scope = 
+    Scope(N, S(S(State.globalThisSymbol)), MutMap[Local, Str](state.importSymbol -> "import"))
```
LPTK: It would be better not to mess with scopes like this.

The proper way to do this is to add this symbol to Prelude. This requires adding "raw identifiers", as in:

```fs
module id"import" with
  ...
```

I remember you already implemented that feature in the old compiler. Could you port it over?

hkmc2/shared/src/main/scala/hkmc2/semantics/BlockImpl.scala
```diff
@@ -17,6 +17,10 @@ trait BlockImpl(using Elaborator.State):
         stmt.desugared match
         case PossiblyAnnotated(anns, h @ Hndl(body = N)) =>
           PossiblyAnnotated(anns, h.copy(body = S(Block(stmts)))) :: Nil
+        case TypeDef(syntax.Cls, Ident(name), rhs, S(Block(Constructor(Block(ctors)) :: rest))) =>
```
LPTK: You need to wrap this in `PossiblyAnnotated`.

hkmc2/shared/src/test/mlscript-compile/CSP.mjs
```diff
@@ -0,0 +1,37 @@
...
+    tmp7 = new runtime.term.Tup([
+      tmp4,
+      tmp5
+    ]);
+    return new runtime.term.App(tmp6, tmp7)
+  } 
+  static bar() {
+    let tmp, tmp1, tmp2, tmp3, tmp4;
+    tmp = new runtime.term.Symbol("Example");
+    tmp1 = new runtime.term.CSRef(tmp, import.meta.url, "Example.mls");
```
LPTK: What happens if that file is not in the same directory as the current one?

hkmc2/shared/src/test/mlscript-compile/Term.mls
```diff
@@ -0,0 +1,194 @@
...
+    Region(name: Symbol, body: Term)
+    RegRef(reg: Term, value: Term)
+    Assgn(lhs: Term, rhs: Term)
+    Deref(ref: Term)
+    SetRef(ref: Term, value: Term)
+    Ret(result: Term)
+    Throw(result: Term)
+    Try(body: Term, finallyDo: Term)
+
+let names = new Map()
```
LPTK: You _of course_ don't want fresh names be a **_global variable_**...

There should be a freshness context that's passed around where needed. That could possibly be the same as the `Context` you're defining below – not sure; what do you think?

You could use contextual parameters for this (https://github.com/hkust-taco/mlscript/pull/256). Note that you will still need to pass these parameters manually when constructing the lowered trees. But from the user's perspective, it could be as simpel as:

```fs
use Quotes = new Quotes

`x => `x
```

hkmc2/shared/src/test/mlscript/basics/FunDefs.mls
```diff
@@ -41,3 +41,9 @@ fun ~ concat(a, b) = a + b
 fun not(b) = !b
 
 
+:fixme
```
LPTK: This should be fixed. Please merge the main branch.

LPTK (changes requested):
Nice work! But we obviously need to avoid using a single global variable for all fresh names.

hkmc2DiffTests/src/test/scala/hkmc2/JSBackendDiffMaker.scala
```diff
@@ -48,6 +49,10 @@ abstract class JSBackendDiffMaker extends MLsDiffMaker:
     case ReplHost.Result(msg) =>
       if msg.startsWith("Uncaught") then output(s"Failed to load runtime: $msg")
     case r => output(s"Failed to load runtime: $r")
+    h.execute(s"const $termNme = (await import(\"${termFile}\")).default;") match
```
LPTK: This risks going slowing down every single diff-test, even thoguh very few use the quoting infrastructure. Can we gate this behind a flag?

hkmc2/shared/src/test/mlscript/basics/FunDefs.mls
```diff
@@ -39,5 +39,3 @@ fun ~ concat(a, b) = a + b
 
 :todo // error
 fun not(b) = !b
-
-
```
LPTK: 👎

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
NeilKleistGao: You are right. We can avoid introducing it. All use of `Subst` is removed.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
NeilKleistGao: This was because keywords are declared objects in `Term.mls`. I've renamed the function to `setupQuotedKeyword`.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
NeilKleistGao: For quoted variables introduced by `=>` or `let`, they are held by current-stage variables. For cross-stage variables, they are presented by quoted selections. This branch handles temporary variables generated by ucs (e.g., `scrut`) in `Quote` bodies.

hkmc2/shared/src/main/scala/hkmc2/utils/Scope.scala
NeilKleistGao: Good suggestion. However, I found that if I use `module id"import"`, the generated code would be `globalThis.import` or `this.import`, which yields `undefined`. So I still need to maintain this symbol. I moved the insertion to `MLsCompiler.scala` since it is not always useful for a general-purpose `Scope`, and `import` things are only used in `mjs` files. I've also ported the identifier escaping.

hkmc2/shared/src/test/mlscript-compile/CSP.mjs
NeilKleistGao: Here, `"Example.mls"` is treated as a relative path. So a file not in the same directory as the current one is fine. I added a test case where the imported module is `quotes/CSPNest.mls`. The generated code is in `quotes/CSPBaz.mls`, and it can compute the correct relative path to `quotes/CSPNest.mls` (i.e., `import "./CSPNest.mls"`).

hkmc2/shared/src/test/mlscript-compile/Term.mls
NeilKleistGao: I've moved the fresh names to the `Context` class.

hkmc2DiffTests/src/test/scala/hkmc2/JSBackendDiffMaker.scala
NeilKleistGao: I've added a `:qq` flag for it.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
LPTK: I still see a lot of `using Subst`s. Did you forget to remove them?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -549,7 +550,7 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
         subTerm_nonTail(finallyDo)(_ => End()),
         k(Value.Ref(l))
       )
-      
+    case Quoted(body) => quote(body)(k)
```
LPTK, line 554:
```suggestion
    
    case Quoted(body) => quote(body)(k)
    
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
LPTK: I have no idea:

* What you mean by "they are held by current-stage variables. For cross-stage variables, they are presented by quoted selections". I'm guessing you're wildly misusing the terms "cross-stage variables" and "presented".

* Why you think that only the UCS should be the one generating quoted variables? There is no reason for you to make such an assumption here.

hkmc2/shared/src/test/mlscript-compile/Term.mjs
LPTK, line 1: Let's not commit this file.

hkmc2DiffTests/src/test/scala/hkmc2/JSBackendDiffMaker.scala
```diff
@@ -48,6 +49,11 @@ abstract class JSBackendDiffMaker extends MLsDiffMaker:
     case ReplHost.Result(msg) =>
       if msg.startsWith("Uncaught") then output(s"Failed to load runtime: $msg")
     case r => output(s"Failed to load runtime: $r")
+    if importQQ.isSet then
+      h.execute(s"const $termNme = (await import(\"${termFile}\")).default;") match
+      case ReplHost.Result(msg) =>
+        if msg.startsWith("Uncaught") then output(s"Failed to load runtime: $msg")
+      case r => output(s"Failed to load runtime: $r")
```
LPTK: You copy-pasted the error-handling code and forgot to update references to "runtime". Please instead factor this code (including the `h.execute` above) into a common function with a parameter.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -192,7 +192,10 @@ object Elaborator:
     val suid = new Uid.Symbol.State
     given State = this
     val globalThisSymbol = TopLevelSymbol("globalThis")
+    // a builtin symbol import can be used for getting current file paths via import.meta
```
```suggestion
    // In JavaScript, `import` can be used for getting current file path, as `import.meta`
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -702,6 +705,9 @@ extends Importer:
     case Keywrd(kw) =>
       raise(ErrorReport(msg"Unexpected keyword '${kw.name}' in this position." -> tree.toLoc :: Nil))
       Term.Error
+    case Constructor(delc) =>
+      raise(ErrorReport(msg"Unsupported constructor form." -> tree.toLoc :: Nil))
```
```suggestion
      raise(ErrorReport(msg"Unsupported constructor in this position." -> tree.toLoc :: Nil))
```

hkmc2/shared/src/test/mlscript-compile/.gitignore
```diff
@@ -1,5 +1,6 @@
 *.mjs
 !/Predef.mjs
 !/Runtime.mjs
+!/Term.mjs
```
LPTK: Any reason to include this huge file?

LPTK (approved):
Otherwise LGTM. Please address the remaining comments.

hkmc2/shared/src/test/mlscript-compile/.gitignore
NeilKleistGao: No reason. I've removed it. :)

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
NeilKleistGao: Many `using Subst`s are originally in the `Lowering.scala` (e.g. `def term(t: st, inStmtPos: Bool = false)(k: Result => Block)(using Subst)`), and I didn't touch them. Shall I remove them all?

hkmc2DiffTests/src/test/scala/hkmc2/JSBackendDiffMaker.scala
NeilKleistGao: I refactored the code with a helper function `importRuntimeModule`.

NeilKleistGao:
> I have no idea:
> What you mean by "they are held by current-stage variables. For cross-stage variables, they are presented by quoted selections". I'm guessing you're wildly misusing the terms "cross-stage variables" and "presented".
> Why you think that only the UCS should be the one generating quoted variables? There is no reason for you to make such an assumption here.

```fs
fun test() = 123
fun foo() =
  `test`() `+ `1
```
For the first question, I wanted to say that a reference like ``` `test ``` in the above example would be considered a quoted selection like `SynthSel(Ref(sym: ModuleSymbol), Ident("test"))`. So it is not involved in the `case Ref(sym)` branch.
For the second question, I agree that we may have other quoted variables. I've removed the wrong comment.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
LPTK: Oh I see, they were already useless parameters. We'll deal with them later.

LPTK:
> For the first question, I wanted to say that a reference like `` `test `` in the above example would be considered a quoted selection like `SynthSel(Ref(sym: ModuleSymbol), Ident("test"))`.

Yeah, so it's a reference to a globally-accessible member, not an instance of "cross-stage" persistence.

Issue #263, opened 2025-01-22
Effect handlers (tracking issue)


LPTK:
Here are some problems left in the original PR (https://github.com/hkust-taco/mlscript/pull/257) that still need to be addressed.

### Code documentation
- [ ] https://github.com/hkust-taco/mlscript/pull/257#pullrequestreview-2557775711

### Interactions with initialization
- [x] `handle` statements should not be followed by object initialization code: https://github.com/hkust-taco/mlscript/pull/257#pullrequestreview-2560193523
- [ ] [low-priority] Once we have optional bindings, we could add special treatment to let bindings that may never run due to lack of resumption https://github.com/hkust-taco/mlscript/pull/257#pullrequestreview-2558750381

### Bug fixes
- [x] Fix bug that was introduced or uncovered after the big refactoring (https://github.com/hkust-taco/mlscript/commit/f764eb9d3664f59b8f4ea915f5f48b2a2ebdc6eb)
    <details>
    <summary>Details</summary>

    https://github.com/hkust-taco/mlscript/blob/f764eb9d3664f59b8f4ea915f5f48b2a2ebdc6eb/hkmc2/shared/src/test/mlscript/handlers/EffectsInClasses.mls#L9-L55

    </details>
- [x] Instrumentation seems to break with the following test: https://github.com/CAG2Mark/mlscript/blob/handler-fixes/hkmc2/shared/src/test/mlscript/handlers/ZCombinatorHandler.mls (edit: it's due to stuff after tail-calls not being resumed as a result of the handler tail-call optimization, it's being fixed right now)
- [x] This fails to print "b"
    ```fs
    class Lol(h) with
      print(h.perform("k"))
    
    let oops = 
      handle h = Effect with
        fun perform(arg)(k) =
          print(arg)
          k("b")
      Lol(h)
    //│ > k
    //│ oops = Lol { h: Effect$h$ {} }
    ```
- [x] Fix bug related to nested handlers and non-tail resumptive handlers (https://github.com/hkust-taco/mlscript/pull/259/commits/ea340666c7f56a74ade878314ebda52e90c0eb92)
- [x] ~~Verify assertion added in (https://github.com/hkust-taco/mlscript/pull/259/commits/ea340666c7f56a74ade878314ebda52e90c0eb92) always hold or fix those cases.~~ Rewrote from scratch.
- [ ] Fix bug related to super call being effectful or disallow effectful constructor.
[...]

PR #280, opened 2025-02-12
Miscellaneous Fixes

CAG2Mark:
## Changes:
- `Block.subBlocks` did not properly handle `Label`; this is now fixed. Now
    ```fs
    while cond do
      class Foo
      new Foo
    ```
    won't error. Also added the above test to `While.mls`.
- Updated `HandlerLowering.scala` to handle `AssignDynField`.
- By @AnsonYeung: Added broken test regarding effect handlers and modules, and fixed the incorrect handling of `AssignDynField` in `BlockTransformer.scala` and top level modules.

### DIFF
--- a/hkmc2/shared/src/test/mlscript/codegen/While.mls
+++ b/hkmc2/shared/src/test/mlscript/codegen/While.mls
@@ -275,3 +275,4 @@ while
   else 1
 //│ /!!!\ Uncaught error: scala.MatchError: InfixApp(IfLike(keyword 'while',None,Block(List(App(Ident(print),Tup(List(StrLit(Hello World)))), BoolLit(false)))),keyword 'then',App(IntLit(0),Tup(List(IntLit(0))))) (of class hkmc2.syntax.Tree$InfixApp)
 
+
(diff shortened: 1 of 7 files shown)


hkmc2/shared/src/test/mlscript/codegen/While.mls
```diff
@@ -271,4 +271,6 @@ while
   else 1
 //│ /!!!\ Uncaught error: scala.MatchError: InfixApp(IfLike(keyword 'while',None,Block(List(App(Ident(print),Tup(List(StrLit(Hello World)))), BoolLit(false)))),keyword 'then',App(IntLit(0),Tup(List(IntLit(0))))) (of class hkmc2.syntax.Tree$InfixApp)
 
-
```
LPTK: No need to remove this line.

hkmc2/shared/src/test/mlscript/codegen/While.mls
```diff
@@ -271,4 +271,6 @@ while
...
+while false do
+  class A
+  new A
```
LPTK: Doesn't seem like this belongs in the "FIXME" section.

LPTK (approved):
Nice catch! But pls address the comments first.

CAG2Mark:
Addressed comments

AnsonYeung:
I've updated https://github.com/hkust-taco/mlscript/issues/263 to add the modules and object to track the bug regarding (non-top-level) modules, and also objects.

PR #282, opened 2025-02-13
Handler runtime improvements

AnsonYeung:
Changes:
- Moved handler runtime function to `Runtime.mls`
- Massive rewrite of handler runtime logic, simplifies it a lot, removes the need for fixup code and matches closer to what is described in the report. ~~I have some further improvement in mind which should allow some program that cannot be run currently to work fine.~~ (Update: done)
  - Namely, resume function will be able to be used entirely outside of the handler, experimentation still need to be done to verify what I have in mind is correct.
- Improvement of debuggability of handler runtime, via generation of class name with source code line info so that the continuation can be traced back to the belonging code and also added a pretty printer.
- Removal of double underscore in name
- Enable tail call optimization in stack safety, this removes the need of continuation class in the handler and reduces code size. (Note: tail call optimization != optimizing tail resumptive handlers)

Problem:
I added Runtime module in `decl/Prelude.mls` just to get the class symbols. I think this is not correct as it cause `Runtime` to pass elab but I don't know how to get the class symbols otherwise. (Runtime.mls has dependence on these classes, and Predef.mls imports Runtime.mls, so I don't think these classes can go to Predef?)
* Fixed by shadowing Runtime during elab

TODO:
- [x] Right now the continuation class rely on Origin, which turns out to emit absolute path in `MLsCompiler.scala`. It should not use full path.
- [x] There is some test diff that I missed when I committed the change. Need to address these issues.
- [x] Implement further improvment mentioned above
- [x] Move `enterHandleBlock` to `Predef` (as it's intended to be used by programmers directly) and document its semantics there
- [x] Add a tests mirroring the examples at https://ocaml.org/manual/5.3/effects.html, most notably the fork/yield/xchg one

Test diffs:
- `UserThreads*.mls`: these tests are malformed, and the new implementation throws Error on them
- `LeakingEffects.mls`: the new test output matches with intended semantics
- `ZCombinator.mls`: the stack depth changes because of `Runtime.stackDepth` is a selection which gets sanity checks. The sanity check moves the selection before the stack depth is increased for the function call `console.log`.

Follow-up:
- Decouple non-local return from HandlerLowering
- HandleBlock can be separated as instantiation of the handler and binding of the handler, this may work better with IR and syntax changes.
  - e.g. have handler as a normal let declaration syntax allowing handler method, and have a handler binding syntax which behaves the same as the current `enterHandleBlock`
  - Current JS includes 2 key constructs:
    - The handler method in HandlerLowering is lowered to `(...handlerArgs) => return mkEffect(h, (k) => handlerMtdBody);`
    - The handler body is essentially `enterHandleBlock(h, () => handlerBody)`
  - So in the syntax, we only need to ensure that `h` and `k` is clear in the handler method, i.e. we can even have standalone handlers like
    ```
    let h = new Object()
    handler[h, k] fun raise(x) = k(x)
    handle h in raise(3)
    ```
    Note: `h` is a value, but `k` is binded as param, so it is kind of awkward. Also for type checking this means you retroactively add behaviour to `h` which is not great and probably won't work.
    while awkward, Object are not essential for codegen to work, `h` can be anything that can be equality checked, i.e. everything in JS is technically fine (but this is kind of awkward and might not work well with type checking).
    ```
    handler[123, k] fun raise(x) = k(x)
    handle 123 in raise(3)
    ```
    So we can also have a relatively minor adjustment to the current syntax and get (currently I prefer this syntax)
    ```
    let h = handler with
      handler fun raise(x)(k) = k(x)
    handle h in raise(3); 4
    handle h in raise(3); "abc"
    ```
    This syntax allow normal methods and other class members, and behave like an object except occurence of handler fun inside it is legal.
    It might be worthwhile to add yet another keyword for `handler fun` for clarity and conciseness.


hkmc2/shared/src/test/mlscript/handlers/EffectInHandler.mls
```diff
@@ -17,11 +18,12 @@ handle h = PrintEffect with
...
 h.aux(3)
 //│ ╔══[ERROR] Name not found: h
-//│ ║  l.15: 	    h.p(x)
+//│ ║  l.16: 	    h.p(x)
 //│ ╙──      	    ^
 //│ ╔══[ERROR] Name not found: h
-//│ ║  l.17: 	    h.p(x)
+//│ ║  l.18: 	    h.p(x)
 //│ ╙──      	    ^
+//│ ═══[RUNTIME ERROR] Error: MLscript call unexpectedly returned `undefined`, the forbidden value.
```
AnsonYeung: This diff is due to safeCall is not emitted for `cur.handlerFun(..., ...)` for some reason. Seems it is emitted for one param but not 2 params

AnsonYeung: Note: this diff is gone due to reverting to calling handlerFun with one param during refactor.

hkmc2/shared/src/test/mlscript/handlers/UserThreadsUnsafe.mls
LPTK, line 1: Why remove this file? I think it's a useful regression test file.

AnsonYeung: It just no longer works, I had combined both files into one and rewrote it to work in the new system using some unexposed primitive function at the moment.

LPTK: Wait what unexposed primitive functions are you referring to?

> It just no longer works

That doesn't matter. It's useful to keep tests that yield errors, as we also want to do regression testing against errors, not just successes.

AnsonYeung: `enterHandleBlock` used in UserThreads.mls

The semantic of that function is simply entering a handle block with handler h binded, except as opposed to handle block, it does not create the handler h

hkmc2/shared/src/test/mlscript/handlers/UserThreads.mls
```diff
@@ -0,0 +1,88 @@
...
+  fun yld()(k) =
+    this.tasks.push(k)
+    globalThis.eval("runtime").enterHandleBlock(this, () => this.tasks.shift()(this))
+  fun start(main)(k) =
+    this.tasks.push(main)
+    while this.tasks.length != 0 do
+      globalThis.eval("runtime").enterHandleBlock(this, () => this.tasks.shift()(this))
+    k()
+in
+  h.start(main)
```
AnsonYeung: This is needed, because handler method themselves cannot throw effect of the handler itself. So here it just rebinds the handler and invoke it. Previously, the handler is still binded during the handler method with some weird semantics as the order of which handler is resumed is simply very messy (refer to `RecursiveHandler.mls` L56-79). This should be the correct behaviour now.

hkmc2/shared/src/test/mlscript/handlers/UserThreads.mls
```diff
@@ -0,0 +1,88 @@
...
+
+
+// LIFO
+handle h = ThreadEffect with
+  fun fork(thread)(k) =
+    this.tasks.unshift(thread)
+    k(())
+  fun yld()(k) =
+    this.tasks.push(k)
+    globalThis.eval("runtime").enterHandleBlock(this, () => this.tasks.shift()(this))
```
LPTK: WOW what! Why not just import `Runtime.mls` and then use it? Or add this function to `Predef`.

hkmc2/shared/src/test/mlscript/handlers/UserThreads.mls
LPTK: Wait but that looks like major shift in the official semantics of effect handlers. I'm afraid it might make our support for the feature quite atypical/weird and not aligned with the standard definitions from the literature.

This seems like a significant design decision that should be properly discussed. Could you start by specifying what's the intended _semantics_ of `enterHandleBlock `? It doesn't seem to appear anywhere in your project report.

LPTK: Oh I just saw your other comment:

> The semantic of that function is simply entering a handle block with handler h binded, except as opposed to handle block, it does not create the handler h

1. Are you saying this is a way of obtaining deep handler semantics in an otherwise shallow handler implementation?

2. Could you please give examples of what happens if you forget to do it?

3. Why not implicitly wrap all handler blocks inside this call? Would that make sense?

hkmc2/shared/src/test/mlscript/handlers/UserThreadsUnsafe.mls
LPTK: Please put the broken tests back. As I said, they're important for the test suite.

> a handle block with handler h binded

I think you mean _bound_.

hkmc2/shared/src/test/mlscript/handlers/UserThreads.mls
AnsonYeung: 1. the handler is still deep, you can still raise effect multiple times in the handler body but not inside the actual handler. The actual handler method cannot raise effects.
2. This will be throwing error "Unhandled effect" as if the effect leaks.
3. Yes, this is possible. A handle block is simply `let h = new Handler(); runtime.enterHandleBlock(h, () => handlerBody);`

AnsonYeung: So what happens here is this: because this.tasks contain some function that will still raise effect due to `.fork()` just adding those function directly, those added by `.yld` won't raise effects

hkmc2/shared/src/test/mlscript/handlers/UserThreads.mls
```diff
@@ -121,3 +123,62 @@ start(main)
...
+  handle h = ThreadEffect with
+    fun fork(new_thread)(k) =
+      enqueue(k)
+      start(new_thread)
+    fun yld()(k) =
+      enqueue(k)
+      dequeue()
+    fun xchg(x)(k) =
+      if exchanger is
+        Some(Tuple2(y, k1)) then
```
LPTK: How about just
```suggestion
        Some([y, k1]) then
```

LPTK:
- [x] Some illustrative test cases to add to the tests
```fs
:re
handle h = Object with
  fun write(x)(k) =
    if x < 10 do write(10)
    [x, k(())]
h.write(1)
h.write(2)
//│ ═══[RUNTIME ERROR] Error: Unhandled effects

handle h = Object with
  fun write(x)(k) =
    if x < 10 do runtime.enterHandleBlock(this, () => write(10))
    [x, k(())]
h.write(1)
h.write(2)
//│ = [1, [2, ()]]
```

hkmc2/shared/src/main/scala/hkmc2/codegen/StackSafeTransform.scala
```diff
@@ -43,11 +43,10 @@ class StackSafeTransform(depthLimit: Int)(using State):
         .assign(tmp, res)
         .assign(tmp, Call(resetDepthPath, tmp.asPath.asArg :: curDepth.asPath.asArg :: Nil)(true, false))
         .rest(f(tmp.asPath))
-
-  def extractResTopLevel(res: Result, isTailCall: Bool, f: Result => Block, sym: Option[Symbol], curDepth: => Symbol) =
+
```
LPTK: Weird whitespace change.

hkmc2/shared/src/test/mlscript/handlers/UserThreads.mls
```diff
@@ -151,33 +149,32 @@ abstract class ThreadEffect with
   fun xchg(x: Int): Int
 
 val tasks = []
-fun enqueue(k) =
-  tasks.push(k)
+fun enqueue(k, x) =
+  tasks.push(k.bind(null, x))
```
LPTK: Why is this `bind` needed, here?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -59,7 +59,9 @@ object Elaborator:
     def nest(outer: Opt[InnerSymbol]): Ctx = Ctx(outer, Some(this), Map.empty, mode)
     
     def get(name: Str): Opt[Ctx.Elem] =
-      env.get(name).orElse(parent.flatMap(_.get(name)))
+      parent match
+      case S(parent = N) if name == "Runtime" => N
```
LPTK: This is a pretty ugly special 😛

Please remove.

hkmc2/shared/src/test/mlscript/bbml/bbPrelude.mls
```diff
@@ -30,6 +30,7 @@ declare class Str(length: Int, concat: Str -> Str)
 declare class Error(msg: Str)
 
 declare module Predef
+declare module Runtime
```
LPTK: Please remove.

hkmc2/shared/src/test/mlscript/decls/Prelude.mls
```diff
@@ -42,10 +42,12 @@ declare module Predef with
...
-  class __EffectSig(next, tail, handleBlockList, resumed, handler, handlerFun)
-  class __Return(value)
+
+declare module Runtime with
+  class EffectSig(contTrace, handler, handlerFun)
+  class Return(value)
   // This declaration is not used currently except for experimentation
-  abstract class __StackDelay with
+  abstract class StackDelay with
     fun perform()
```
LPTK: Let's not do that. `Prelude.mls` is for listing built-in things that either come with JavaScript or that represent MLscript base concepts without a JS counterpart (like `Int`).

We should also remove `Predef` from here, for what it's worth. It has nothing to do here.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -49,29 +56,28 @@ import HandlerLowering.*
...
-  private val contClsPath: Path = predefPath.selN(Tree.Ident("__Cont")).selN(Tree.Ident("class"))
-  private val retClsPath: Path = predefPath.selN(Tree.Ident("__Return")).selN(Tree.Ident("class"))
-  private val retClsSym: ClassSymbol = predefSym.tree.definedSymbols.get("__Return").get.asCls.get
-  private val appendInContPath: Path = predefPath.selN(Tree.Ident("__appendInCont"))
-  private val mkEffectPath: Path = predefPath.selN(Tree.Ident("__mkEffect"))
-  private val handleBlockImplPath: Path = predefPath.selN(Tree.Ident("__handleBlockImpl"))
+  private val runtimePath: Path = State.runtimeSymbol.asPath
+  private val runtimeSym: ModuleSymbol = ctx.builtins.Runtime
+  private val effectSigPath: Path = runtimePath.selN(Tree.Ident("EffectSig")).selN(Tree.Ident("class"))
+  private val effectSigSym: ClassSymbol = runtimeSym.tree.definedSymbols.get("EffectSig").get.asCls.get
```
LPTK: For now, just create a symbol from scratch representing `EffectSig` (you can add this symbol to `Elaborator.State`).

I think it doesn't matter that this won't be exactly the same symbol as the one we process when compiling `Runtime.mls`.

LPTK (changes requested):
Very good improvements! But let's please remove the Prelude hack.

hkmc2/shared/src/test/mlscript/handlers/UserThreads.mls
AnsonYeung: this is essentially `() => k(x)`, I guess it is confusing so I'm gonna change it to a explicit lambda.

LPTK:
> Follow-up [...]

Thanks for writing your thoughts and suggestions!

Some more thoughts on this: eventually, we'll probably want to make the effect handler and [type class/contextual parameters](https://github.com/hkust-taco/mlscript/pull/256) features work in synergy. This will allow programmers to pass handlers around implicitly instead of having to manually thread them everywhere.

Adapting an example from https://effekt-lang.org/tour/effects, we could have:

```fs
// Interface definition
abstract class Yield[A] with
  fun yield(x: A): Unit

// Interface use
fun fib(using Yield[Int]) =
  fun inner(a, b) =
    use[Yield].yield(a)
    inner(b, a + b)
  inner(0, 1)

// Handler creation (as a type class)
fun genFibs(limit: Int): Array[Int] =
  let
    count = 0
    fibs = []
  handle Yield[Int] with // Define a type class instance AND handle it
    effect yield(x) =
      if count < limit do
        count += 1
        fibs.push(x)
        resume(()) // <- we resume the computation where yield was invoked
  in fib()
  fibs
```

Of course, type classes would not be necessary to use effect handlers. We could also just use normal classes and plain `handle` statements:

```fs
// Handler creation (as a normal class)
class YieldUntil(n: Int) extends Yield[Int] with
  val results = []
  effect done() =
    results
  effect yield(x) =
    if x === 42 do done()
    results.push(x)
    resume(())

print of
  let h = YieldUntil(42)
  handle h
  // or just: `handle h = YieldUntil(42)`
  y.yield(1)
  y.yield(2)
  y.done()
  ??? // never reached
```

The type system can keep track of the fact that effect methods are not used outside of corresponding handlers. Note that for effect hygiene, it's critical that the effect be attached to a specific object instance and not just a label.

What do you think?

hkmc2/shared/src/main/scala/hkmc2/codegen/StackSafeTransform.scala
```diff
@@ -43,7 +43,7 @@ class StackSafeTransform(depthLimit: Int)(using State):
         .assign(tmp, res)
         .assign(tmp, Call(resetDepthPath, tmp.asPath.asArg :: curDepth.asPath.asArg :: Nil)(true, false))
         .rest(f(tmp.asPath))
-        
+
```
```suggestion
  
```

AnsonYeung:
I think type classes should work quite well with handlers. Does `handle h` always bind the contextual parameter of the type of `h` or only `handle Yield[Int]` binds it? the type of `h` might not be known yet.

I think it would be good to implement the explicit syntax first as the next step from here.

P.S. On the other hand,
```fs
// Handler creation (as a normal class)
class YieldUntil(n: Int) extends Yield[Int] with
  val results = []
  effect done() =
    results
  effect yield(x) =
    if x === 42 do done() // this makes effect call, which doesn't work now
    results.push(x)
    resume(())
```

For this type of recursive binding, I think the semantics for this can be:
```
  effect yield(x) =
    handle this
    if x === 42 do done() // this makes effect call
    results.push(x)
    resume(())
```

But I'm unaware of other implementation of algebraic effect doing this, raising effect in a handler body means raising an effect in the outer context.

LPTK:
> Note that for effect hygiene, it's critical that the effect be attached to a specific object instance and not just a label.

Wait, in fact I remember precisely why I proposed the original _coupled_ design in the first place. It was not accidental.

The problem is that we want to prevent programmers from being tempted to write code like this:

```fs
// In some place:
object Foo(p) with
  effect raise(x) = ... resume ...

// In another place:
fun test(f) =
  handle Foo(...)
  xs map(x => if x < 0 then Foo.raise(x) else f(x))
```

The problem is that if a user calls `foo` in a context like this:

```fs
handle Foo(arg)
test(x => ... Foo.raise ...)
```

Then there is an unhygienic interaction that may be surprising: the person who wrote `map` did not expect to unintentionally handle effects coming from `f`. This is a known problem in the literature.

So I think it's better to ensure that a fresh, unique object instance is generated whenever an effect is handled in some scope.

> I think type classes should work quite well with handlers. Does `handle h` always bind the contextual parameter of the type of `h` or only `handle Yield[Int]` binds it? the type of `h` might not be known yet.

So actually I don't think the general `handle h` form should be supported at all. One would use `handle Yield[Int]` and `handle Yield[Int] as h` (if a name is needed).

> But I'm unaware of other implementation of algebraic effect doing this, raising effect in a handler body means raising an effect in the outer context.

Yeah, it's probably better to follow the standard practice, and we could allow `handle this` as a special form specifically for this use case.

AnsonYeung:
> The problem is that we want to prevent programmers from being tempted to write code like this:
> 
> ```fsharp
> // In some place:
> object Foo(p) with
>   effect raise(x) = ... resume ...
> 
> // In another place:
> fun test(f) =
>   handle Foo(...)
>   xs map(x => if x < 0 then Foo.raise(x) else f(x))
> ```
Right, but I think the problem here lies in how the contextual parameter is handled. Suppose in `test(x => ... Foo.raise ...)` is expanded as `let h = ...; test(x => ... h.raise ...)` it would bind it directly. But if the lambda is passed around with a contextual parameter unexpanded then it can be bound accidentally within test.

AnsonYeung:
The suggested scheme doesn't involve much change for handler lowering; during lowering everything involving contextual parameter should be resolved already. Again for the example above:
> ```fs
> fun test(f) =
>   handle Foo(...)
>   xs map(x => if x < 0 then Foo.raise(x) else f(x))
> ```
The type of f can be required to be free of implicit contextual parameter, then `f(x)` cannot refer to the new contextual parameter.

> So actually I don't think the general handle h form should be supported at all. One would use handle Yield[Int] and handle Yield[Int] as h (if a name is needed).

~~Instead of `handle h` we might have `handle Yield[Int] = h`. The reason for the IR change was to allow binding of this type:~~

The reason for introducing this type of binding was due to usage of recursive handlers in the old UserThread test. We can go without this syntax while still allowing a special syntax for the semantics of the recursive binding (`handle this` at top of handler).

But the test at `RecursiveHandlers.mls:40`:
```fs
handle h = Object with
  fun write(x)(k) =
    if x < 10 do
      print(enterHandleBlock(this, () => write(10)))
    [x, k(())]
```
showed that it is possible to create many surprising patterns, including rebinding the same handler at different levels, which may result in surprising behaviour if a function rebinds a given handler. So I agree that it should probably not be introduced. On the other hand, recursive binding seems fine for now and the usage pattern seems rather restricted and cannot cause such rebinding problem, except the `resume` function (the resulting (current) behaviour is that `handle this` will not affect the execution of `resume` at all, and effect raised within the handler is handled inside the function).

LPTK:
> Right, but I think the problem here lies in how the contextual parameter is handled. Suppose in `test(x => ... Foo.raise ...)` is expanded as `let h = ...; test(x => ... h.raise ...)` it would bind it directly. But if the lambda is passed around with a contextual parameter unexpanded then it can be bound accidentally within test.

Sorry, I realize I messed up the example. I meant to make `Foo` an actual, unparameterized object.
And I was assuming there were no contextual parameters, here.

```fs
// In some place:
object Foo with
  effect raise(x) = ... resume ...

// In another place:
fun test(f) =
  handle Foo
  xs map(x => if x < 0 then Foo.raise(x) else f(x))

handle Foo
test(x => ... Foo.raise ...)
```

I can see where your remark is coming from: by using type classes like `Yield[Int]`, as in

```fs
fun test(f) =
  handle Yield[Int] with ...
  xs map(x => if x < 0 then use[Yield].yield(x) else f(x))
```

it would seem like we would lose the hygienic property I was talking about. But that's not the case.

Contextual parameters are resolved and expanded statically, at compile time. They're just as if the programmer had manually passed the parameters manually. They're not dynamic and require an explicit type annotation.

With `test` as written above, it is impossible for the mapped function `f` to capture the local `Yield` instance.

To make `f` capture the local contextual value, one would have to write something like:

```fsharp
fun test(f: (Int)(using y: Yield[Int]) ->{y} Int) =
  handle Yield[Int] with ...
  xs map(x => if x < 0 then use[Yield].yield(x) else f(x))
```

and then it would be completely clear that this capture is the intended semantics.

(Notice the `y` in `->{y}` which is needed to keep track of exactly the effect instance being thrown, without which things can't type check.)

> ~Instead of `handle h` we might have `handle Yield[Int] = h`. The reason for the IR change was to allow binding of this type:~
> 
> The reason for introducing this type of binding was due to usage of recursive handlers in the old UserThread test. 

Not sure what you are referring to. Which kind of binding specifically, and how was it used in the old UserThread test?

> except the `resume` function (the resulting (current) behaviour is that `handle this` will not affect the execution of `resume` at all, and effect raised within the handler is handled inside the function).

I guess that's as expected though because we don't bind handlers dynamically.

AnsonYeung:
Oh the type of binding I refer to is exactly the one you wrote (equivalent to arbitrary usage of `enterHandleBlock`). So yes, we should not allow binding handlers directly (and remove usage of `enterHandleBlock` once recursive handler is settled).

LPTK (approved):
LGTM, thanks!

PR #283, opened 2025-02-19
Improve module method checks

FlandiaYingman:
This PR improves the module method checks introduced in #233, and makes preparation in order to move all the checks to the resolution stage. It also fixes:

#### the problem that "module" parameters were not treated as modules
```fs
:e
fun f8(module m: M) = m
```

#### the problem that self referencing modules were not treated as modules
```fs
module Foo with
  val self = this
```

#### the problem that term definitions returning a module were not treated as modules
```fs
module A with
  fun f() = 2

module B with
  val a: module A = A

print(B.a) // B.a should be a module, and should be rejected passing to a regular parameter
```


hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1107,10 +1108,21 @@ extends Importer:
...
     case _ =>
       t.asParam(inUsing).map: (isSpd, p, t) =>
-        isSpd -> Param(FldFlags.empty, fieldOrVarSym(ParamBind, p), t.map(term(_)))
+        val sym = fieldOrVarSym(ParamBind, p)
+        val sign = t.map(term(_))
+        val param = Param(flags, sym, sign)
+        sym match
+          case sym: TermSymbol =>
+            // TODO: How can a TermSymbol accept a Declaration
+            // sym.defn = S(TermDefinition(ctx.outer, Fun, sym, Nil, Nil, sign, N, FlowSymbol(s"‹result of ${sym}›"), TermDefFlags.empty, Nil))
```
FlandiaYingman: I was a little confused about `TermSymbol`: it seems like the symbol becomes `TermSymbol` if the scope is in the class (i.e., a class's parameter), but other parts of code seem to directly create a `BlockMemberSymbol` to represent a term. Also, I found some code seems to related to it, but it was commented with some reason 🤔

https://github.com/FlandiaYingman/mlscript/blob/bb1da72d39e09bca135c590ed19a73c36a99efee/hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala#L43-L48

So I was not very sure how to deal with this here. I was thinking to change the symbol to `BlockMemberSymbol` instead but I don't know if that's appropriate.

LPTK: Right. This part of the module system is not fully/adequately worked out yet.

A couple of things influencing the current WIP design were:
* class parameters act both as normal parameters and as fields, complicating the handling of parameters
* let bindings in classes are basically private fields

But actually I realized class parameters need a very different handling than that. Consider this:

https://github.com/hkust-taco/mlscript/blob/46ae8a78c957f07f221c80a34a793bf22fdb707e/hkmc2/shared/src/test/mlscript/basics/Inheritance.mls#L50-L68

Here, the current scheme always treats the class parameter `z` symbol as a field so it always code-gens `z` as `this.z`; but this is not valid in the pre-constructor (i.e., before `super` is called). So it seems to me we should just treat all _parameters_ the same, including class parameters, but classes would additionally get some derived symbol form that represents a field induced by a class parameter.

LPTK: I think the obvious thing to do is simply to desugar each class parameter `z` into a normal parameter `z` (`VarSymbol`) along with a `val`, in the class body, that binds a field named `z` to that parameter, using a `TermDefinition`.

LPTK: Then it would have a proper `BlockMemberSymbol`.

LPTK: ie ~desugar~ elaborate
```fs
class Foo(val z) with
  blah
```
into
```fs
class Foo(z‹123›) with
  val member:z‹124› = z‹123›
  blah
```

FlandiaYingman:
```fs
class Foo(val z) with
  blah
```

should now be elaborated into

```fs
class Foo(z‹123›) with
  val member:z‹124› = z‹123›
  blah
```

I didn't differentiate `class Foo(val z)` and `class Foo(z)` (they all elaborated to an additional val binding `member:z`) because all current tests seem to use the latter. But I think this behavior can be easily modified if we want.

hkmc2/shared/src/test/mlscript/basics/Inheritance.mls
LPTK, line 1: Don't forget to remove this FIXME comment!
```scala
// FIXME: in super calls, class params should actually be viewed as a `VarSymbol`, not a `TermSymbol` wit owner
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -117,6 +117,7 @@ class InstSymbol(val origin: Symbol)(using State) extends LocalSymbol:
 
 class VarSymbol(val id: Ident)(using State) extends BlockLocalSymbol(id.name) with NamedSymbol with LocalSymbol:
   val name: Str = id.name
+  override def toLoc: Option[Loc] = id.toLoc
```
```suggestion
  override def toLoc: Opt[Loc] = id.toLoc
```

hkmc2/shared/src/test/mlscript/basics/GenericClasses.mls
```diff
@@ -25,19 +25,27 @@ class Foo[A] with
 
 
 
+// Note: the elaborator elaborates class parameters into additional field definitions;
+// this is why the following code yields two same errors.
```
LPTK: As discussed in the meeting, let's not duplicate the annotation here.

FlandiaYingman:
I changed *a lot of* tests to replace all `class` definitions that have parameters with `data class` definitions. 

I also checked them so to not add `data` modifier to some `class` definitions (especially those in `mlscript/basics`) that obviously do not require the `data` semantic.

hkmc2/shared/src/test/mlscript/handlers/ManualStackSafety.mls
```diff
@@ -117,5 +117,3 @@ test(10, 5)
 //│ > finished at offset: 5
 //│ > finished at offset: 0
 //│ = 3628800
-
-
```
LPTK: 😢

hkmc2/shared/src/test/mlscript/nofib/boyer2.mls
```diff
@@ -321,12 +322,41 @@ let rules =
...
+//│ FAILURE: Unexpected exception
```
LPTK: Whoops. When did that start happening?

hkmc2/shared/src/test/mlscript/nofib/mate.mls
```diff
@@ -557,22 +557,16 @@ fun testMate_nofib(dummy) =
...
-//│ >  if K-Q5; 2. B/KR4-KB2, K-QB5; 3. Q/QN1-KB1++
-//│ >  if K-Q7; 2. K-KB3, P/QN6xN/QR7; 3. B/KR4-K1++
-//│ >  if K-K7; 2. N/QR2-QN4,
-//│ >   if K-Q7; 3. Q/QN1-K1++
-//│ >   if K-K6; 3. Q/QN1-Q3++
-//│ >  if P/QN6xN/QR7; 2. Q/QN1-QB2, ...; 3. B/KR4-KB2++
-//│ > 
+//│ FAILURE: Unexpected runtime error
+//│ FAILURE LOCATION: mkQuery (JSBackendDiffMaker.scala:146)
+//│ ═══[RUNTIME ERROR] Error: ENOENT: no such file or directory, open 'hkmc2/shared/src/test/mlscript/nofib/input/heathcote3.prob'
```
LPTK: What happened? 🤔

FlandiaYingman: Oops, why doesn't the hkmc2AllTests report them. Let me check

hkmc2/shared/src/test/mlscript/nofib/boyer2.mls
FlandiaYingman: No idea but seems this it's not related to this PR. I checked on the hkmc2 branch and it also occured

hkmc2/shared/src/test/mlscript/handlers/ManualStackSafety.mls
FlandiaYingman: 🙈

hkmc2/shared/src/test/mlscript/HkScratch.mls
```diff
@@ -6,7 +6,6 @@
 
 :global
 :d
-:todo
```
LPTK: Hey!

hkmc2/shared/src/test/mlscript/lifter/ClassInFun.mls
```diff
@@ -1,6 +1,9 @@
 :lift
 :js
+:sjs
```
LPTK: This adds _a lot_ of output. Let's omit it.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -233,18 +233,51 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
...
     val fresh = FreshInt()
 
-    val varsMap: Map[Local, TermSymbol] = cap.map: s =>
+    val varsMap = cap.map: s =>
         val id = fresh.make
-        s -> TermSymbol(syntax.ParamBind, S(clsSym), Tree.Ident(s.nme + id + "$"))
+        val nme = s.nme + id + "$"
+        val varSym = VarSymbol(Tree.Ident(nme))
+        val fldSym = BlockMemberSymbol(nme, Nil)
+        val fldDef = TermDefinition(
```
LPTK, line 243: So what's the intuitive reason we need all that additional stuff now in the lifter?

LPTK (approved):
Nice change, thanks!

Maybe @CAG2Mark you could take a quick look at the lifter changes?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -152,7 +152,7 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
     val ignoredDefns: Set[BlockMemberSymbol] = Set.empty,
     val inScopeDefns: Map[BlockMemberSymbol, Set[BlockMemberSymbol]] = Map.empty,
     val modLocals: Map[BlockMemberSymbol, Local] = Map.empty,
-    val localCaptureSyms: Map[Local, LocalSymbol & NamedSymbol] = Map.empty,
+    val localCaptureSyms: Map[Local, (VarSymbol, TermDefinition)] = Map.empty,
```
CAG2Mark: What is the `TermDefinition` here used for? It's not used anywhere in the updated code.

If there is a reason, please update the documentation to briefly describe the `TermDefinition` here.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: It seems there's some major changes to the way `ClsLikeDefn` works. Namely, it seems that class parameters now have an associated public field which requires a term definition. When the constructor is called, the arguments are bound to a `VarSymbol` and then `preCtor` assigns the values to the term definitions. It seems `BlockMemberSymbol` can also have a reference to its definition now, so that also needs to be kept track of.

CAG2Mark (commented):
LGTM with regards to the lifter. I've left one minor comment and a reply to another comment.
(p.s. the lifter logic is quite complicated, so good job on figuring it out :P)

BTW @LPTK it seems that `VarSymbol`s are often created inside later IR passes because it's unavoidable, but you mentioned a while ago that `VarSymbol`s all come from user defined variables. I think that in the future, it would be good to add a field to `VarSymbol` such as `isUserDefined` to make this distinction more explicit.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: Ah yes, that makes sense.

Indeed, we wanted to go back to the original intent (implemented in the previous compiler) that `class Foo(x)` does _not_ introduce a public field but `class Foo(val x)` does. We also added the alternative `data class Foo(x)` which implies that field. Also, now, ctor parameters are really just (VarSymbol) parameters, cleanly distinguished from field definitions, which solves a problem we had with super calls.

CAG2Mark:
@AnsonYeung This PR may be of interest to you since it might change the way you need to call `super` in the handler lowering. I've described the changes here https://github.com/hkust-taco/mlscript/pull/283#discussion_r1986072746

AnsonYeung:
so preCtor is also for generating the field assignment, is there any reason for handling aux param differently? I don’t think it causes issue with handler lowering. In fact I think if preCtor and ctor is combined into a single block then it would be trivial to fix one of the current broken case with super call being effectful.

CAG2Mark:
> so preCtor is also for generating the field assignment, is there any reason for handling aux param differently? I don’t think it causes issue with handler lowering. In fact I think if preCtor and ctor is combined into a single block then it would be trivial to fix one of the current broken case with super call being effectful.

Probably because it's currently assumed to be handled by the front-end (but we don't have aux params in the front-end yet) but you're right. Though doing this will require even more changes in the lifter.

LPTK:
> is there any reason for handling aux param differently?

They should definitely be handled the same. The only difference is that thesae params won't get `val` semantics implicitly when using `data class`, unlike the main parameters.

> In fact I think if preCtor and ctor is combined into a single block then it would be trivial to fix one of the current broken case with super call being effectful.

Weren't you the one who introduced the `preCtor` field in the first place? 😛 

In https://github.com/hkust-taco/mlscript/commit/e37d8f8d4018d0c5fbc7d49d47ac0255dee604d7#diff-72bbb753659dd44a5e47f3cd0b3954e4ee95fda127ca491a2c267461779c60fbR126

What was the purpose of using a separate block, and does that still make sense now?

AnsonYeung:
I'm not sure about it now. Does lifter need it now (as I think it may need to insert some extra fields)? The original intent was just a hack to insert super calls before JSBuilder emits the field initializations.

In any case, I am completely fine with it being removed now.

LPTK:
@AnsonYeung Ok, we'll probably want to remove it in a later PR.

CAG2Mark:
On another note (sorry for putting so many extra comments in this PR), I think the `toString` function for classes should also be generated in an earlier phase. This could, for example, allow `toString` to benefit from stack safe recursion (which was actually a problem we ran into in our benchmarking; it would stack overflow).

LPTK:
@CAG2Mark Thanks for the suggestion. I intend for `toString` to be changed to a simpler call to some rendering function defined externally, so this should no longer be a problem. Note that we already have this problematic test case: https://github.com/hkust-taco/mlscript/blob/4e1d3cf8c04edaddac3e484ef10daf51f4c10cd0/hkmc2/shared/src/test/mlscript/basics/CyclicValues.mls#L6-L13

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -280,7 +282,7 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
       End()
     )
 
-    (defn, varsMap, varsList)
+    (defn, varsMap.mapValues(_.mapSecond(_.sym)).toMap, varsList)
```
LPTK: > method mapValues in trait MapOps is deprecated since 2.13.0: Use .view.mapValues(f). A future version will include a strict version of this method (for now, .view.mapValues(f).toMap).

LPTK (approved):
LGTM

PR #286, opened 2025-02-26
Logic subtyping and overloading

auht:
### 2025-02-26 TODO:

- [x] Add spaces like a civilized person
- [x] Check intersections for WFedness
- [x] Remove assumption that only the first argument needs disjointness
  - [x] Deal with nested `dss`
- [x] Support generalization (update `collectTVs`)
- [x] Add records ~~(for now use temporary syntax `fun foo: ["x": Int, "y": Int]`)~~

### 2025-03-06 TODO:
- [ ] don't duplicate when LB is TV on same or higher level, in which case you can make both vars bound each other (same level) or the other one bound the LB
- [ ] properly update type simplifier (and remove kludge)
- [x] remove use of nondeterministic collections

Later:
- [ ] During simplification, take UBs into account to reduce things

### 2025-03-13 TODO:
Later:
- [ ] use different propagation rules for TVs of same level
- [ ] add an option to enable/disable `:logicSub`


hkmc2/shared/src/test/mlscript/bbml/bbList.mls
```diff
@@ -43,65 +43,13 @@ region r in
 
 // * Should be an error. This definition would not be referentially transparent.
 // * The error message needs improvement, though.
-:e
```
LPTK: Why do you remove the `:e` on the test that says this should be an error? 🤔

hkmc2/shared/src/test/mlscript/bbml/bbBasics.mls
```diff
@@ -152,7 +152,7 @@ fun foofoo(x) =
 new Printer(foofoo)
 //│ Type: Printer['T]
 //│ Where:
-//│   'T <: Int
+//│   'T#'T ∨  ∧ 'T<:'x ∧ Str<:Str ∧ ⊥<:⊥
```
LPTK: Isn't this missing `'x <: Int`?

hkmc2/shared/src/test/mlscript/logicsub/MarlowWadler97.mls
```diff
@@ -53,10 +53,8 @@ test(fls, fls)
 
 
 :fixme
-:e
```
LPTK: Instead of removing this `:e`, which should be there, you can make this a `:todo` (instead of `:fixme`)

hkmc2/shared/src/test/mlscript/bbml/bbGPCE.mls
```diff
@@ -76,11 +82,21 @@ fun bind(rhs, k) = `let x = rhs `in k(x)
 bind
 //│ Type: ['cde, 'ctx, 'cde1, 'eff, 'cde2, 'ctx1] -> (CodeBase[out 'cde, out 'ctx, ?], CodeBase[in 'cde1 out 'cde1 ∨ 'cde, ?, ⊥] ->{'eff} CodeBase[out 'cde2, out 'ctx1, ?]) ->{'eff} CodeBase[out 'cde2, out 'ctx ∨ 'ctx1, ?]
 
+:e
```
LPTK, line 105: When something is not supposed to error but it does, you should use `:todo`, instead of breaking the meaning of the test.

hkmc2/shared/src/test/mlscript/bbml/bbLetRegEncoding.mls
```diff
@@ -70,8 +70,21 @@ let f = letreg(r => arg => r.ref arg)
 f
 //│ Type: 'arg ->{⊤} Ref['arg, ?]
 
+:e
```
LPTK, line 73: Same remark as above.

hkmc2/shared/src/test/mlscript/bbml/bbRec.mls
```diff
@@ -37,13 +37,10 @@ fun f(x) = f(x)
 f
 //│ Type: ['x, 'eff, 'app] -> 'x ->{'eff} 'app
 //│ Where:
-//│   'x#'x ∨ (  'x#'x ∨  ∧ 'x<:'x ∧ 'app<:'app ∧ 'eff<:'eff)
+//│   'x#'x ∨ (  'x#'x ∨  ∧ 'app<:'app ∧ 'eff<:'eff) ∧ 'x<:'x
 
 :todo
```
LPTK: Can remove the `:todo`?

hkmc2/shared/src/test/mlscript/bbml/bbList.mls
auht: I have added it back.

hkmc2/shared/src/test/mlscript/bbml/bbBasics.mls
auht: Fixed.

hkmc2/shared/src/test/mlscript/bbml/bbGPCE.mls
auht: The `:e` is there originally and I should not remove it in the first place. So I have added it back.

hkmc2/shared/src/test/mlscript/bbml/bbLetRegEncoding.mls
auht: Same here. There is a `:e` originally.

hkmc2/shared/src/test/mlscript/bbml/bbRec.mls
auht: It has been removed.

hkmc2/shared/src/test/mlscript/logicsub/DisjSub.mls
```diff
@@ -163,21 +162,18 @@ idIIBB(new Pair(1, true))
 
 :todo
```
LPTK: Is this TODO still relevant?

hkmc2/shared/src/test/mlscript/logicsub/Records.mls
```diff
@@ -0,0 +1,84 @@
...
+//│ Type: ⊤
+
+:todo
+[rcd.foo, rcd.bar]
+//│ ╔══[ERROR] Term shape not yet supported by BbML: Tup(List(Fld(‹›,Sel(Ref(rcd),Ident(foo)),None), Fld(‹›,Sel(Ref(rcd),Ident(bar)),None)))
+//│ ║  l.25: 	[rcd.foo, rcd.bar]
+//│ ╙──      	^^^^^^^^^^^^^^^^^^
+//│ Type: ⊥
+
+:todo // IntLit as label?
```
LPTK:
```suggestion
// TODO: IntLit as label?
```

hkmc2/shared/src/test/mlscript/logicsub/Records.mls
```diff
@@ -0,0 +1,84 @@
...
+
+fun id(x) = x
+//│ Type: ⊤
+
+id of (x: 1)
+//│ Type: {x: Int}
+
+:todo
+id of { x: 1 }
+//│ Type: Str
```
LPTK, line 44: wat

hkmc2/shared/src/test/mlscript/logicsub/DisjSub.mls
auht: I will remove the todo later. The temporary syntax has been implemented.

hkmc2/shared/src/test/mlscript/logicsub/Records.mls
auht: The `{ x: 1 }` is elaborated to a Tup with an ascription `["x": 1]`.
It is typed as `id("x")` since the ascription is ignored.

LPTK: I wonder why this happens. Are you sure it's an ascription? It should elaborate to a RecordField

auht: It is a `Fld` with the field `asc` set as `Some(1)`.
Similarly, `id of { x: 1, y: 2 }` reports incorrect number of arguments. It is an error in `basic/Records.mls` originally.

LPTK: Ok I'll look into it.

LPTK: Actually it's working "as intended". Currently, curly braces are equivalent to an indented block. Here one would need normal parentheses instead
```fs
id of (x: 1)
```

hkmc2/shared/src/main/scala/hkmc2/bbml/types.scala
```diff
@@ -358,17 +358,26 @@ object Type:
           val u = disjointImpl(a, p.toBasic)(prev)
           val w = disjointImpl(a, q.toBasic)(prev)
           u.flatMap(u => w.map(u ++ _))
-        case (a: InfVar, b: InfVar) if a.uid =/= b.uid => N
+        case (ComposedType(p, q, false), _) =>
+          (disjointImpl(p.toBasic, b)(prev), disjointImpl(q.toBasic, b)(prev)) match
+            case (N, w) => w
+            case (u, N) => u
+            case (S(u), S(w)) => S(u.flatMap(u => w.map(u ++ _)))
+        case (_, ComposedType(p, q, false)) =>
```
LPTK: Intersections on the right can be solved easily, no? Was this case meant to be
```suggestion
        case (_, ComposedType(p, q, true)) =>
```
?

hkmc2/shared/src/test/mlscript/bbml/bbBasics.mls
```diff
@@ -149,7 +149,12 @@ fun foofoo(x) =
   if x is Int then
     let t = x + 1 in "foo"
 foofoo
-//│ Type: (¬Int ∨ Int) -> Str
+//│ Type: ['x, 'scrut, 'eff, 'cons] -> 'x ->{'eff} 'cons
+//│ Where:
+//│   'x <: ¬Int ∨ Int
+//│   'x#Int ∨  ∧ ⊥<:'eff ∧ 'scrut<:Int ∧ Int<:Int ∧ Str<:'cons
+//│   Int ∧ 'x <: 'scrut
+//│   'scrut#Int ∨  ∧ ⊥<:'eff ∧ 'scrut<:Int ∧ Int<:Int ∧ Str<:'cons
```
LPTK: For patmat without default cases, can't we just implicitly add a case that matches the negations of all previous patterns and constrains the scrutinee to be Bot?

LPTK: (This inferred type scheme, like the original one from BbML, misinterprets the patmat as being exhaustive, thus allowing `'x = ~Int` as a – spurious – solution.)

hkmc2/shared/src/test/mlscript/bbml/bbGPCE.mls
```diff
@@ -17,15 +17,15 @@ fun id(x) = x
 
 
 run(x `=> id(x) `* x)
-//│ Type: Int -> ⊥
+//│ Type: ⊤ -> ⊥
```
LPTK: 🤔

hkmc2/shared/src/test/mlscript/logicsub/MarlowWadler97.mls
```diff
@@ -36,19 +36,21 @@ test
...
+//│   'y#Fls ∨  ∧ ⊥<:'eff2 ∧ Fls<:'res
 
 test(tru, tru)
 //│ = Tru()
 //│ Type: Tru
 
 test(tru, fls)
 //│ = Fls()
-//│ Type: Fls
+//│ Type: Fls ∨ Tru
```
LPTK: Any idea why this happens?

auht: Fixed.

hkmc2/shared/src/test/mlscript/logicsub/popl22.mls
```diff
@@ -0,0 +1,298 @@
...
+//│ Type: ⊤
+
+fun y: Any -> Tru | Fls
+//│ Type: ⊤
+
+prod(y(x => x), y(x => x)) as Prod[Tru, Tru] | Prod[Fls, Fls]
+//│ Type: Prod[Tru, Tru] | Prod[Fls, Fls]
+
+:e
+{ fst: y(x => x), snd: y(x => x) } as [ fst: Tru, snd: Tru ] | [ fst: Fls, snd: Fls ]
```
LPTK, line 102: Any reason why you use a record on the LHS and tuples on the RHS? Why are the RHS' treated like records? They should be one-element arrays. Remember that `[ fst: Tru, snd: Tru ]` is syntax sugar for `[ (fst: Tru, snd: Tru) ]`.

hkmc2/shared/src/test/mlscript/logicsub/popl22.mls
```diff
@@ -0,0 +1,298 @@
...
+let fls = new Fls()
+//│ Type: ⊤
+
+fun prod(p, q) = new Prod(p, q)
+//│ Type: ⊤
+
+fun y: Any -> Tru | Fls
+//│ Type: ⊤
+
+prod(y(x => x), y(x => x)) as Prod[Tru, Tru] | Prod[Fls, Fls]
```
LPTK: How come this type checks? Shouldn't it be `Prod[Tru | Fls, Tru | Fls]`?

auht: Because classes are merged in union. It is actually `Prod[Tru | Fls, Tru | Fls]` here.
The example using record does not type check.
The original paper uses tuple and type checks: $(y(\lambda x.x), y(\lambda x.x)): (True \times True) \lor (False \times False)$

LPTK: Wait why does it type check in the original paper, then? Do they merge the union?

auht: If I understand correctly, it is because both $y$ are applied on the same argument $\lambda x.x$, and thus have the same return type.

LPTK: Wow... that seems a bit crazy. Do they keep track of equalities between terms at the type level? Or what other sorcery do they use? It sounds very expensive.

auht: I think they use the _maximal sharing canonical forms_ to do the type checking, so the expression is something like
```
bind a = y in
bind b = x => x in
bind c = a(b) in
bind d = (c, c) in
d
```
Where `c` can be either `True` or `False`.

LPTK: Ah, right; I remember seeing this. Presumably, we could do the same thing, but that would probably not be worth it. Please document this as comments on the test case.

hkmc2/shared/src/main/scala/hkmc2/bbml/types.scala
auht: Disjointness is symmetric. This case should handle intersection (though it is incorrect).

The `disjoint` function has been rewritten and uses DNF currently.

PR #287, opened 2025-03-04
Add missing `BlockTraverser` case and fix possible mishandling of `Value.Rcd`

CAG2Mark:
Records were recently added, but it seems they are not handled in `BlockTraverser`. Also, it seems the `idx` field of `RcdArg` is not mapped in `BlockTransformer`. This has been fixed.

Note that I'm not sure if `idx` is not supposed to be mapped or if it was a mistake.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/BlockTraverser.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/BlockTraverser.scala
@@ -25,23 +25,23 @@ class BlockTraverser(subst: SymbolSubst):
     case HandleBlockReturn(res) => applyResult(res)
     case Match(scrut, arms, dflt, rst) =>
       val scrut2 = applyPath(scrut)
-      arms.map: arm =>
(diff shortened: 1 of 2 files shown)


hkmc2/shared/src/main/scala/hkmc2/codegen/BlockTraverser.scala
```diff
@@ -68,6 +68,8 @@ class BlockTraverser(subst: SymbolSubst):
     case Value.Lit(lit) => ()
     case v @ Value.Lam(params, body) => applyLam(v)
     case Value.Arr(elems) => elems.map(applyArg)
+    case Value.Rcd(fields) => fields.map:
```
LPTK: Wait why is the `BlockTraverser` using `map` everywhere? That's crazy 😅 

It should use `foreach` to avoid building useless intermediate data structures!

LPTK (approved):
Oh you're right, thanks for noticing!

PR #288, opened 2025-03-06
Non local return

AnsonYeung:
Removal of old implementation of return in handle block. This PR handles such return together with other non local return automatically. `HandleBlockReturn` is removed from the IR as a result
* there is no need for such IR node as it exist purely to distinguish non local return from exiting the handle block
* instead, Return will mean exiting the handle block now.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
@@ -1037,7 +1037,7 @@ extends Importer:
               sym.defn = S(tdf)
               
               // indicates if the function really returns a module
-              // TODO: check non-local returns
+              // TODO: check non-local returns (see [test:T4])
               val em = b.exists(ModuleChecker.evalsToModule)
               // indicates if the function marks its result as "module"
               val mm = st match
--- a/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
@@ -78,10 +78,10 @@ object Elaborator:
       case OuterCtx.Function(sym) => S(sym)
       case _ => parent.flatMap(_.getNonLocalRetHandler)
     // Returns N if no return handler is required (a direct `return` is possible).
-    // Returns S(N) if the return handler is required but not found.
+    // Returns S(N) if there is no function in scope.
     def getRetHandler: Opt[Opt[TempSymbol]] = outer match
       case OuterCtx.Function(sym) => N
-      case OuterCtx.LambdaOrHandlerBlock | OuterCtx.InnerScope(_) => S(getNonLocalRetHandler)
+      case _: (OuterCtx.LambdaOrHandlerBlock.type | OuterCtx.InnerScope) => S(getNonLocalRetHandler)
       case _ =>
         // If parent is empty, we are at the top-level context.
         parent.fold(S(N))(_.getRetHandler)
@@ -629,7 +629,7 @@ extends Importer:
     case Modified(Keyword.`return`, kwLoc, body) =>
       ctx.getRetHandler match
       case S(S(sym)) =>
-        tl.log(s"Non local return: $sym")
+        tl.log(s"Non-local return: $sym")
         val rs = FlowSymbol("‹app-res›")
         val retMtdTree = new Tree.Ident("ret")
         val argTree = new Tree.Tup(body :: Nil)
@@ -643,7 +643,6 @@ extends Importer:
           ErrorReport(msg"Return statement outside of a function." -> tree.toLoc :: Nil)
         Term.Error
       case N =>
-        tl.log(s"Normal return: $ctx")
         Term.Ret(term(body))
     case Modified(Keyword.`throw`, kwLoc, body) =>
       Term.Throw(term(body))
@@ -1013,13 +1012,13 @@ extends Importer:
                 then rhs.map(term(_)(using newCtx))
                 else S(Term.Missing)
               val nb: Opt[Term] = if nonLocalRetHandler.directRefs.isEmpty then b else b.map: inner =>
-                val clsSym = ClassSymbol(Tree.DummyTypeDef(Cls), Tree.Ident("‹non local return effect›"))
+                val clsSym = ClassSymbol(Tree.DummyTypeDef(Cls), Tree.Ident("‹non-local return effect›"))
                 val valueSym = VarSymbol(Ident("value"))
                 val resumeSym = VarSymbol(Ident("resume"))
                 val mtdSym = BlockMemberSymbol("ret", Nil, true)
                 val td = TermDefinition(
                   N, Fun, mtdSym, PlainParamList(Param(FldFlags.empty, valueSym, N) :: Nil) :: Nil,
-                  N, N, S(valueSym.ref(Ident("value"))), FlowSymbol(s"‹result of non local return›"), TermDefFlags.empty, Nil)
+                  N, N, S(valueSym.ref(Ident("value"))), FlowSymbol(s"‹result of non-local return›"), TermDefFlags.empty, Nil)
                 val htd = HandlerTermDefinition(resumeSym, td)
                 Term.Handle(nonLocalRetHandler, state.nonLocalRetHandlerTrm, Nil, clsSym, htd :: Nil, inner)
               val r = FlowSymbol(s"‹result of ${sym}›")
@@ -1028,7 +1027,7 @@ extends Importer:
               sym.defn = S(tdf)
               
               // indicates if the function really returns a module
-              // TODO: check non local returns
+              // TODO: check non-local returns
               val em = b.exists(ModuleChecker.evalsToModule)
               // indicates if the function marks its result as "module"
               val mm = st match
--- a/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
@@ -38,11 +38,21 @@ object Elaborator:
     op -> op).toMap
 
   val reservedNames = binaryOps.toSet ++ aliasOps.keySet + "NaN" + "Infinity"
+
+  enum OuterCtx:
+    case Function(returnHandlerSymbol: TempSymbol)
+    case InnerScope(innerSymbol: InnerSymbol)
+    case LocalScope
+    case LambdaOrHandlerBlock
+
+    def inner: Opt[InnerSymbol] = this match
+      case InnerScope(inner) => S(inner)
+      case _ => N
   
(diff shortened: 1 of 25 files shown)


hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -103,6 +103,15 @@ class TempSymbol(val trm: Opt[Term], dbgNme: Str = "tmp")(using State) extends B
   override def toString: Str = s"$$${super.toString}"
   override def subst(using s: SymbolSubst): TempSymbol = s.mapTempSym(this)
 
+class WeakTempSymbol(trm: Opt[Term], dbgNme: Str = "weak")(using State) extends TempSymbol(trm, dbgNme):
+  var used: Bool = false
```
LPTK: Just make use of `directRefs`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -39,8 +39,10 @@ object Elaborator:
 
   val reservedNames = binaryOps.toSet ++ aliasOps.keySet + "NaN" + "Infinity"
   
+  // isFunLike specifies the current context begins a function-like control flow where `return` would exit from the flow.
+  // nonLocalRetHandler specifies the current context is a binding site of `return`.
   case class Ctx(outer: Opt[InnerSymbol], parent: Opt[Ctx], env: Map[Str, Ctx.Elem], 
-    mode: Mode):
+    isFunLike: Bool, nonLocalRetHandler: Opt[WeakTempSymbol], mode: Mode):
```
LPTK: Please try to only allow representing legal states. This should probably be combined with `outer`.

```scala
enum OuterCtx:
  case Function(returnHandlerSymbol)
  case InnerScope(innerSymbol)
  case LocalScope
  case LambdaOrHandlerBlock
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -53,27 +63,28 @@ object Elaborator:
...
     def get(name: Str): Opt[Ctx.Elem] =
       env.get(name).orElse(parent.flatMap(_.get(name)))
-    def getOuter: Opt[InnerSymbol] = outer.orElse(parent.flatMap(_.getOuter))
-    def getNonLocalRetHandler: Opt[WeakTempSymbol] =
-      nonLocalRetHandler.orElse(parent.flatMap(_.getNonLocalRetHandler))
+    def getOuter: Opt[InnerSymbol] = outer.inner.orElse(parent.flatMap(_.getOuter))
+    def getNonLocalRetHandler: Opt[TempSymbol] = outer match
+      case OuterCtx.Function(sym) => S(sym)
+      case _ => parent.flatMap(_.getNonLocalRetHandler)
     // Returns S(N) if the return handler is required but not found.
```
```suggestion
    // Returns N if no return handler is required (a direct `return` is possible).
    // Returns S(N) if the return handler is required but not found.
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -626,10 +637,13 @@ extends Importer:
       ctx.getRetHandler match
       case S(S(sym)) =>
         val rs = FlowSymbol("‹app-res›")
+        val retMtdTree = new Tree.Ident("ret")
+        val argTree = new Tree.Tup(body :: Nil)
+        val dummyIdent = new Tree.Ident("return").withLocOf(tree)
```
```suggestion
        val dummyIdent = new Tree.Ident("return").withLocOf(kwLoc)
```

hkmc2/shared/src/test/mlscript/handlers/ReturnInHandler.mls
```diff
@@ -4,6 +4,7 @@
 abstract class Effect with
   fun f(): ()
 
+// return within handle blocks are implemented as usual non local return
```
LPTK:
```suggestion
// `return` within handle blocks is implemented as a non-local return
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -965,12 +1012,23 @@ extends Importer:
               val b = if ctx.mode != Mode.Light
                 then rhs.map(term(_)(using newCtx))
                 else S(Term.Missing)
+              val nb: Opt[Term] = if nonLocalRetHandler.directRefs.isEmpty then b else b.map: inner =>
+                val clsSym = ClassSymbol(Tree.DummyTypeDef(Cls), Tree.Ident("‹non local return effect›"))
```
```suggestion
                val clsSym = ClassSymbol(Tree.DummyTypeDef(Cls), Tree.Ident("‹non-local return effect›"))
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -965,12 +1012,23 @@ extends Importer:
...
+                val valueSym = VarSymbol(Ident("value"))
+                val resumeSym = VarSymbol(Ident("resume"))
+                val mtdSym = BlockMemberSymbol("ret", Nil, true)
+                val td = TermDefinition(
+                  N, Fun, mtdSym, PlainParamList(Param(FldFlags.empty, valueSym, N) :: Nil) :: Nil,
+                  N, N, S(valueSym.ref(Ident("value"))), FlowSymbol(s"‹result of non local return›"), TermDefFlags.empty, Nil)
```
```suggestion
                  N, N, S(valueSym.ref(Ident("value"))), FlowSymbol(s"‹result of non-local return›"), TermDefFlags.empty, Nil)
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -965,12 +1012,23 @@ extends Importer:
...
+                val htd = HandlerTermDefinition(resumeSym, td)
+                Term.Handle(nonLocalRetHandler, state.nonLocalRetHandlerTrm, Nil, clsSym, htd :: Nil, inner)
               val r = FlowSymbol(s"‹result of ${sym}›")
-              val tdf = TermDefinition(owner, k, sym, pss, tps, s, b, r, 
+              val tdf = TermDefinition(owner, k, sym, pss, tps, s, nb, r, 
                 TermDefFlags.empty.copy(isModMember = isModMember), annotations)
               sym.defn = S(tdf)
               
               // indicates if the function really returns a module
+              // TODO: check non local returns
```
LPTK: Can you add a test case that this missing check causes an issue with?

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -103,7 +103,6 @@ class TempSymbol(val trm: Opt[Term], dbgNme: Str = "tmp")(using State) extends B
   override def toString: Str = s"$$${super.toString}"
   override def subst(using s: SymbolSubst): TempSymbol = s.mapTempSym(this)
 
-
```
LPTK: 😢

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
AnsonYeung: just added it in ModuleMethods test

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1076,7 +1076,7 @@ extends Importer:
         val sym = members.getOrElse(nme.name, lastWords(s"Symbol not found: ${nme.name}"))
         var newCtx = S(td.symbol).collectFirst:
             case s: InnerSymbol => s
-          .fold(ctx.nestLocal)(ctx.nestInner(_))
+          .fold(ctx.nest(OuterCtx.LambdaOrHandlerBlock))(ctx.nestInner(_))
```
LPTK: It looks like the only time we don't have an inner symbol here would be when we are looking at a type alias (`Als`). Why would you set the context to `LambdaOrHandlerBlock` here? I think we should add a new case to the enum (call it `NonReturnContext`) where using `return` is simply forbidden. Surely we don't want to allow things like `type T = return "lol"`

hkmc2/shared/src/test/mlscript/handlers/NonLocalReturns.mls
```diff
@@ -82,3 +82,41 @@ fun f() =
     print("Bad")
 f()
 //│ = 100
+
+// broken due to handler does not handle object properly
```
LPTK:
```suggestion
// broken due to handlers currently not handling `object` properly
```

LPTK: BTW do you know why it works with a class but not with an object?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: In fact we should add `type T = return "lol"` as a negative test case in `NonLocalReturns.mls`!

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -78,10 +78,10 @@ object Elaborator:
       case OuterCtx.Function(sym) => S(sym)
       case _ => parent.flatMap(_.getNonLocalRetHandler)
     // Returns N if no return handler is required (a direct `return` is possible).
-    // Returns S(N) if the return handler is required but not found.
+    // Returns S(N) if there is no function in scope.
```
LPTK: A function in scope would mean that there's a sbiling `fun` statement in the current block or in an outer block.

```suggestion
    // Returns S(N) if we are not in a function scope.
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -103,6 +103,7 @@ class TempSymbol(val trm: Opt[Term], dbgNme: Str = "tmp")(using State) extends B
   override def toString: Str = s"$$${super.toString}"
   override def subst(using s: SymbolSubst): TempSymbol = s.mapTempSym(this)
 
+
```
LPTK, line 106:
```suggestion
```

hkmc2/shared/src/test/mlscript/handlers/NonLocalReturns.mls
AnsonYeung: object instantiation emitted in JSBuilder is not checked for effect.

LPTK: Why not? Would it be hard to fix?

LPTK: Right, I just remembered there's already a corresponding item in https://github.com/hkust-taco/mlscript/issues/263

LPTK (approved):
Nice simplifications!

The code is also much cleaner now, thanks to `OuterCtx` and `ReturnHandler`.

PR #289, opened 2025-03-07
Preliminary deforestation implementation for hkmc2

ychenfo:


hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -668,284 +698,291 @@ class Deforest(using TL, Raise, Elaborator.State):
...
-        case CtorFinalDest.Sel(s) => None
-      }.toMap
+    private val toBeReplacedForAllBranches = mutable.Map.empty[ResultId, Map[ResultId, Symbol]].withDefaultValue(Map.empty)
+    d.filteredCtorDests.values.foreach:
+      case CtorFinalDest.Match(scrut, expr, selInArms, selMaps) =>
+        toBeReplacedForAllBranches += scrut -> (toBeReplacedForAllBranches(scrut) ++ selMaps._2)
     
+    def apply(scrutExprId: ResultId, m: Match) = store.getOrElseUpdate(
+      scrutExprId,
+      {
```
LPTK:
```suggestion
      locally:
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -668,284 +698,291 @@ class Deforest(using TL, Raise, Elaborator.State):
...
+        val matchExpr@Match(Value.Ref(l), arms, dflt, rest) = m
+        val parentMatchRest = d.matchScrutToParentMatchScrut(m.scrut.uid).flatMap: p =>
+          if !d.filteredDtors.contains(p) then Some(d.matchScrutToMatchBlock(p).rest)
+          else None
+        val selReplacementNotForThisSel = replaceSelInfo -- toBeReplacedForAllBranches(scrutExprId).keys
+        
+        val traverser = DeforestationFreeVarTraverser(using nonFreeVars + l, selReplacementNotForThisSel, toBeReplacedForAllBranches(scrutExprId).values)
+        traverser.applyPath(m.scrut)
+        (arms.map(_._2) ++ dflt).foreach: a =>
+          // dflt may just be `throw error``, and `rest` may use vars assigned in non default arms.
```
LPTK: ~~I don't understand what this comment means.~~ BTW it's duplicated from earlier in the file...

LPTK: It looks like the two copies of the comment have diverged. Is that really intentional?

hkmc2/shared/src/test/mlscript/deforest/simple.mls
```diff
@@ -5,13 +5,13 @@
...
 object B
 object C
-class AA(aa)
-class BB(bb)
-class AAA(x, y)
-class BBB(x, y)
-class CCC(c)
+class AA(val aa)
+class BB(val bb)
+class AAA(val x, val y)
```
LPTK: Why not just make these
```suggestion
data class AAA(x, y)
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: Indeed previously I didn't fully figure out the proper logic in these two places. I have updated the comments [here](https://github.com/CrescentonC/mlscript/blob/2519a808b00bed7e25a508c2e14e8088df0041b4/hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala#L116-L117) and [here](https://github.com/CrescentonC/mlscript/blob/2519a808b00bed7e25a508c2e14e8088df0041b4/hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala#L214-L216). Now these two places are mostly the same except that the one for un-transformed `Match` blocks also takes care of the `rest`s of its parent `Match` blocks.

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: Thanks! I have updated my uses of curly braces to `locally:`.

hkmc2/shared/src/test/mlscript/deforest/simple.mls
ychenfo: Thanks! I've made the update to use `data class` in all the tests now.

hkmc2/shared/src/test/mlscript/deforest/append.mls
LPTK, line 1: Maybe it's time we finally make these fuse 😬 

After all we had a plan to make them work, right?

ychenfo: Oh right, I remember the reification of recursive strategies into actual computations. I will look into this.

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -735,44 +735,45 @@ class Deforest(using TL, Raise, Elaborator.State):
...
-                    assert(getClsFields(c).map(_.id).contains(fs.field))
-                    fieldNameToSymToBeReplaced.updateWith(fs.field):
-                      case Some(v) => Some(v)
-                      case None => Some(if varSymInsteadOfTempSym
-                        then VarSymbol(Tree.Ident(s"_deforest_${c.name}_${fs.field.name}"))
-                        else TempSymbol(N, s"_deforest_${c.name}_${fs.field.name}"))
-                    val sym = fieldNameToSymToBeReplaced(fs.field)
+          handledMatches.getOrElseUpdate(
+            scrutRef.uid -> currentCtorCls,
+            locally:
```
LPTK: I don't think this serves any purpose. Just remove it?

LPTK: (You shouldn't need braces either.)

ychenfo: Ah indeed, I have removed this `locally:`. Thanks!

hkmc2DiffTests/src/test/scala/hkmc2/JSBackendDiffMaker.scala
LPTK, line 1: It looks like you introduced _**a lot**_ of duplication in this file, which already contained some duplicated logic (there were two instances of `should always ends with "undefined" (TODO: check)` – now there are four!!).

Every duplicated implementation is more maintenance burden on future developers. Please be more careful about that and try to share and ruse the logic. Even better if you can leave the place in a _better_ state than you found it, for example by removing the original duplication.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -459,6 +468,11 @@ sealed abstract class Result extends AutoLocated:
     case DynSelect(qual, fld, arrayIdx) => qual.freeVarsLLIR ++ fld.freeVarsLLIR
     case Value.Rcd(args) => args.flatMap(arg => arg.idx.fold(Set.empty)(_.freeVarsLLIR) ++ arg.value.freeVarsLLIR).toSet
   
+  lazy val uid =
```
LPTK: Please document that this is for Lumberhack.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -406,6 +406,15 @@ enum Case:
 
 sealed trait TrivialResult extends Result
 
+type ResultId = Uid[Result]
+object ResultUidHandler extends Uid.Handler[Result]
+object ResultUid extends ResultUidHandler.State:
```
LPTK: The fact that this piece of state is global is a bit concerning... If the actual `uid` value does not influence the result and is only used to distinguish things, you might as well use the object's identity and avoid adding a field to `Result`, no?

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -0,0 +1,1104 @@
+package hkmc2
+package codegen
+
+import semantics.*
+import semantics.Elaborator.State
+import syntax.{Literal, Tree}
+import utils.{TL, tl, SymbolSubst}
```
LPTK:
```suggestion
import utils.*
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -0,0 +1,1104 @@
...
+        rest.willBeNonEndTailBlock
+    case _: End => ()
+    case _: BlockTail => flag = true
+    case _ => super.applyBlock(b)
+  def analyze(b: Block): Bool =
+    applyBlock(b)
+    flag
+
+class ReplaceLocalSymTransformer(freeVarsAndTheirNewSyms: Map[Symbol, Symbol]) extends
+  BlockTransformer(new SymbolSubst()):
```
LPTK: Why would you break the line _after_ `extends` and indent the `BlockTransformer...` exactly the same as the statements in the class body?

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -0,0 +1,1104 @@
...
+trait StratVarTrait(stratState: StratVarState):
+  this: ProdVar | ConsVar =>
+  
+  lazy val asProdStrat = stratState.asProdStrat
+  lazy val asConsStrat = stratState.asConsStrat
+  lazy val uid = stratState.uid
+
+final case class NotDeforestableException(msg: String) extends Exception(msg)
+
+// Compute free vars for a block, without considering deforestation, used on transformed blocks
```
LPTK: What do you mean by "transformed blocks" in this context?

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -0,0 +1,1104 @@
...
+
+case class Ctor(ctor: ClsOrModSymbol, args: Map[TermSymbol, ProdStrat], expr: CtorExpr) extends ProdStrat
+case class ProdFun(l: Ls[ConsStrat], r: ProdStrat) extends ProdStrat
+case class ProdVar(s: StratVarState) extends ProdStrat with StratVarTrait(s)
+case object NoProd extends ProdStrat
+
+
+
+case class Dtor(scrut: ResultId)(val expr: Match, val outterMatch: Option[ResultId])(using d: Deforest) extends ConsStrat:
+  assert(scrut === expr.scrut.uid)
```
LPTK: What's the point of this redundant info? It might be better to use a normal class along with an `unapply` method?.

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -0,0 +1,1104 @@
...
+  d.matchScrutToMatchBlock.updateWith(scrut):
+    case None => Some(expr)
+    case Some(exist) => ??? // should only update once
```
LPTK: `???` is for missing implementation. You probably mean to use `die` or `lastWords`?

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -0,0 +1,1104 @@
...
+  
+  def hasExplicitRet: Boolean =
+    HasExplicitRetTraverser.analyze(b)
+  
+  def willBeNonEndTailBlock(using d: Deforest): Bool =
+    WillBeNonEndTailBlockTraverser().analyze(b)
+    
+class Deforest(using TL, Raise, Elaborator.State):
+  
+  object StratVarUidHandler extends Uid.Handler[StratVar]()
```
LPTK: Why not move this out and use `StratVarUidHandler.State` everywhere instead of the extremely awkward `Uid.Handler[StratVar]#State`? Also the `()` is unnecesasary.

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -0,0 +1,1104 @@
...
+            funSymsWithDefn += fun.sym
+            super.applyFunDefn(fun)
+            
+        FreshVarForAllVars.applyBlock(p)
+      // `NoProd` to block fusion for those functions that are imported from elsewhere
+      usedFunSym.diff(funSymsWithDefn).foreach: funSymsWithoutDefn =>
+        constrain(NoProd, store(funSymsWithoutDefn).asConsStrat)
+      
+    
+    // TODO: ctor as a function?
```
LPTK: What does this mean? Is it still to do?

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -0,0 +1,1104 @@
...
+    case Begin(sub, rest) =>
+      processBlock(sub)
+      processBlock(rest)
+    case Define(defn, rest) =>
+      defn match
+        case FunDefn(_, sym, params, body) =>
+          val funSymStratVar = symToStrat(sym)
+          val param = params.head match
+            case ParamList(flags, params, restParam) => params
+          val funStrat = constrFun(param, body) // TODO: handle mutiple param list
```
LPTK: If you intend to do it in a later PR, please at least crash here instead of silently doing the wrong thing!! And then add a TODO in a tracking issue for Lumberhack (similar to, eg, https://github.com/hkust-taco/mlscript/issues/263).

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -0,0 +1,1104 @@
...
+      processBlock(rest)
+    case Begin(sub, rest) =>
+      processBlock(sub)
+      processBlock(rest)
+    case Define(defn, rest) =>
+      defn match
+        case FunDefn(_, sym, params, body) =>
+          val funSymStratVar = symToStrat(sym)
+          val param = params.head match
+            case ParamList(flags, params, restParam) => params
```
LPTK: You can't just ignore some of the parameters... `restParam` is a real thing.

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -0,0 +1,1104 @@
...
+    constraints.foreach(c => handle(c)(using mutable.Set.empty))
+  
+  
+  // ======== after resolving constraints ======
+    
+  lazy val resolveClashes =
+    type CtorToDtor = Map[CtorExpr, CtorDest]
+    type DtorToCtor = Map[DtorExpr, DtorSource]
+    
+    def removeCtor(ctorDests: CtorToDtor, dtorSources: DtorToCtor, rm: Set[CtorExpr]): CtorToDtor -> DtorToCtor =
```
LPTK: This code looks extremely inefficient, rebuilding tons of intermediate data structures. As the author of Lumberhack, you should have a clue that this isn't great... 😛 

Couldn't you just keep track of ctors and dtors by using a couple of mutable fields in the corresponding ctors and dtors?

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -0,0 +1,1104 @@
...
+  
+  def freshVar(nme: String = "")(using vuid: Uid.Handler[StratVar]#State) =
+    val newId = vuid.nextUid
+    val s = StratVarState(newId, nme)
+    val p = s.asProdStrat
+    val c = s.asConsStrat
+    p -> c
+
+
+type CtorExpr = ResultId
```
LPTK: Why not use `ResultId` everywhere? `CtorExpr` gives the wrong impression that this is an actual expression node/AST.

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -0,0 +1,1104 @@
...
+      defn match
+        case FunDefn(_, sym, params, body) =>
+          val funSymStratVar = symToStrat(sym)
+          val param = params.head match
+            case ParamList(flags, params, restParam) => params
+          val funStrat = constrFun(param, body) // TODO: handle mutiple param list
+          constrain(funStrat, funSymStratVar.asConsStrat)
+          funSymStratVar
+        case v: ValDefn => throw NotDeforestableException("No support for `ValDefn` yet")
+        // only handle code inside module for now to show the
```
LPTK: Show the what?

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -0,0 +1,1104 @@
...
+        // todo case of if scrut being not the same as what the user writes
```
LPTK: What?

LPTK (changes requested):
Just a preliminary quick review with already a few problems found.

hkmc2DiffTests/src/test/scala/hkmc2/JSBackendDiffMaker.scala
ychenfo: Thanks for the suggestion! In the new commits I tried to deduplicate the logic in this file. As for the check `should always ends with "undefined" (TODO: check)`, there is now only one place that this needs to be done, but it seems that currently it's not valid that all programs end with "undefined" (some others end with "Unit { ... }")

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
ychenfo: I have added a comment indicating that this is for Lumberhack.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
ychenfo: Thanks for the suggestion! Indeed the actual `uid` value does not influence the result and is only used to distinguish things. I have pushed the update to remove the global state, and now this `uid` is just `System.identityHashCode`.

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: Yeah I did this unintentionally... I have updated the formatting here. Thanks for catching this!

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: Here I mean blocks after the deforestation transformation, for which when we compute the free variables, we don't need to consider the change of variables brought by deforestation.

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: Thanks for the suggestion! Indeed it's redundant here and I have updated it to use a normal class along with an `unapply` method.

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: Thanks for the suggestion. I have updated this and some other inappropriate uses of `???`.

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: Indeed this awkward type projection here and the `()` is unnecessary, I have updated them according to your suggestion.

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: I meant to refer to cases like

```
fun app(ctor) = ctor(1)
if app(AA) is AA(x) then x
```

Where the constructor `AA` is passed around like a function. Indeed this is still a TODO because we cannot fuse the program above. I have updated the comment here and also improved the handling of cases like this.

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: Thanks for the suggestion. I have updated here so that now it crashes when `restParam` is not `None`, and will open a tracking issue for this.

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: Sorry that this is an obsolete comment. The handling for `ClsLikeDefn` used to be here to show a test case in `todo.mls` which uses module. But I realize that to show the point of that test we don't necessarily need to use a module, and can just use [this test](https://github.com/CrescentonC/mlscript/blob/daf8749d16852151becac4f4f6e271d2d26fc025/hkmc2/shared/src/test/mlscript/deforest/todos.mls#L19-L25). So I removed that test and the insufficient handling for `ClsLikeDefn` here, making this comment obsolete.

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: Thanks for the suggestion! I am looking into optimizing this part.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
LPTK: Ok, cool. As long as the result is still deterministic and does not depend on the order of iteration of the hashed data structures, it should be fine. (It's a subtle "if", tho – please check.)

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -468,10 +467,14 @@ sealed abstract class Result extends AutoLocated:
     case DynSelect(qual, fld, arrayIdx) => qual.freeVarsLLIR ++ fld.freeVarsLLIR
     case Value.Rcd(args) => args.flatMap(arg => arg.idx.fold(Set.empty)(_.freeVarsLLIR) ++ arg.value.freeVarsLLIR).toSet
   
-  lazy val uid =
-    val id = ResultUid.nextUid
-    ResultUid.uidToResult.addOne(id -> this)
-    id
+  // for deforestation
+  import Result.*
+  lazy val uidValue: ResultId = ResultId(System.identityHashCode(this))
```
LPTK: Ok, no. Again, you don't want to rely on an ordering that's nondeterministic (will change whenever the VM decides to allocate things differently, which can happen on version changes or when unrelated things change, or just depending on the system). I assume this ordering could change the output of diff-tests, right?

If you really do need an ordered UID, just have it. It's not good to have global state, and it's better to follow [this pattern](https://github.com/hkust-taco/mlscript/blob/b81146d3936d54ffd0390dff27ffa5fadb73cbeb/hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala#L56), when possible. But the global state is probably not that bad here as long as it's only for deterministic ordering.

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -25,50 +26,52 @@ class StratVarState(val uid: StratVarId, val name: Str = ""):
   
   override def toString(): String = s"${if name.isEmpty() then "var" else name}@${uid}"
 
+object StratVarUidHandler extends Uid.Handler[StratVar]
 object StratVarState:
-  
-  def freshVar(nme: String = "")(using vuid: Uid.Handler[StratVar]#State) =
+  def freshVar(nme: String = "")(using vuid: StratVarUidHandler.State) =
```
LPTK: Why do you remove all empty lines around these definitions, but use empty lines lower, down?

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -415,19 +428,19 @@ class Deforest(using TL, Raise, Elaborator.State):
         case FunDefn(_, sym, params, body) =>
           val funSymStratVar = symToStrat(sym)
           val param = params.head match
-            case ParamList(flags, params, restParam) => params
-          val funStrat = constrFun(param, body) // TODO: handle mutiple param list
+            case ParamList(flags, params, N) => params // TODO: handle mutiple param list
```
LPTK: You're still silently discarding extra parameter lists!!

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
LPTK: Ok, please clarify in the comment (my question was a prompt for you to do so).

LPTK: FYI "transformed block" can be ambiguous – is it the block that is being transformed or the result of the transformation?

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
LPTK: > also improved the handling of cases like this

I'm curious what: what aspect of this handling did you improve?

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: I see, thanks for the explanation! I have updated the comment here to say "blocks after the deforestation transformation"

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: I realized that passing a constructor may cause deforestation to produce wrong programs, and now this is fixed.

For a program like below:

```
fun app(f) = f(1)
fun c(x) = if x is
  AA(i) then i
c(app(AA)) + c(AA(2))
```

Here the consumer is used twice, once on `app(AA)` (cannot be fused) and once on `AA(2)` (can be fused). Previously the pattern matching in `c` and the constructor call `AA(2)` are fused, but `app(AA)` is not fused. So the program after transformation is wrong. Now this is fixed by blocking the fusion (by assigning `NoProd` when seeing a constructor without being applied).

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: I have updated this part so that the functions `removeCtor` and `removeDtor` directly remove entries in the maps instead of allocating new intermediate maps.

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: Oh this is also part of the [obsolete comment](https://github.com/hkust-taco/mlscript/pull/289#discussion_r2059455049) that I have mentioned below and have deleted now. Sorry I missed this earlier.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
ychenfo: Thanks for the explanation! Indeed after testing and checking I found nondeterministic behaviors caused by how the VM allocates things. And I also found that a global state to handle UID causes nondeterminism, I think it's because tests are executed in parrallel.

I think that maybe making `Result` dependent on `Deforest` for a non-global state requires a lot of changes just to add the ordered UID, and I figured that [using `LinkedHashMap` here to preserve the order of insertion](https://github.com/CrescentonC/mlscript/blob/f4f7f95771b22e7bf222f73388a3392e50fe786d/hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala#L536) can fix the nondeterminism here. So currently I fix this by using `LinkedHashMap`.

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: Sorry I am not sure if I fully get your question. I think I just removed this empty line (line 29) without thinking about the reason...😨 and by "but use empty lines lower, down" do you mean empty lines at line 35-36? For those I also just randomly kept them there...

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: Sorry I missed this earlier, now it only works when `params` has exactly one element, and that element's `restParam` is `None`. Otherwise it crashes instead of silently doing the wrong thing.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
ychenfo: Indeed I found non deterministic behaviors caused by this, and I have explained them [here](https://github.com/hkust-taco/mlscript/pull/289#discussion_r2063017838).

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: https://github.com/hkust-taco/mlscript/pull/289#discussion_r2063017996

https://github.com/hkust-taco/mlscript/issues/304#:~:text=restParam%20and%20multiple%20parameter%20lists

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
LPTK: I'm not sure exactly what I meant by "lower, down" either, but your use of whitespace looks weird and inconsistent, mainly around this new `StratVarUidHandler` definition, which comes right in between the `StratVarState` class and its companion and is somehow _not_ separated by whitespace below it. By the way, why not renamr `StratVarUidHandler` to `StratVar` and move it next to the homonym type definition?

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -0,0 +1,1102 @@
...
+  
+  // ======== after resolving constraints ======
+    
+  lazy val resolveClashes =
+    val ctorToDtor = ctorDests.ctorDests
+    val dtorToCtor = dtorSources.dtorSources
+    
+    def removeCtor(rm: Set[ResultId]): Unit =
+      if rm.isEmpty then ()
+      else
```
LPTK:
```suggestion
      if rm.nonEmpty then
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
```diff
@@ -0,0 +1,1102 @@
...
+        tl.log("rm ctor: " + rm.map(c => c.getClsSymOfUid.nme).mkString(" | "))
+        val toDeleteDtors = rm.flatMap(r => ctorToDtor.remove(r)).flatMap:
+          case CtorDest(mat, sels, _) => mat.keySet.map(s => DtorExpr.Match(s)) ++ sels.map(s => DtorExpr.Sel(s.expr))
+        removeDtor(toDeleteDtors)
```
LPTK: Still... why do you need flat-mapping over concatenated collections just to effectfully remove the corresponding elements? Why isn't this just using nested `foreach`calls? Do you really like useless intermediate collections that much?

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
LPTK: Well we dodged a bullt here.

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: Thanks I see. I did the renaming, and I also realize that all the handler objects seem to be defined in [`Uid.scala`](https://github.com/CrescentonC/mlscript/blob/817d40819f2088623b005abdb01d7f8b32449361/hkmc2/shared/src/main/scala/hkmc2/Uid.scala#L16), so I move the definition there.

hkmc2/shared/src/main/scala/hkmc2/codegen/Deforestation.scala
ychenfo: Thanks! I think now I understand. I have pushed the update to remove intermediate collections and directly remove each element.

hkmc2/shared/src/test/mlscript/deforest/new-todo.mls
```diff
@@ -0,0 +1,96 @@
+:js
+
+
+data class AA(x)
+data class BB(x)
+object A
+object B
+
+// TODO: should fuse `A` and `B` with the pattern match in `f`
```
LPTK: Any idea why this doesn't happen?

hkmc2/shared/src/test/mlscript/deforest/new-todo.mls
```diff
@@ -0,0 +1,96 @@
...
+//│ ---------- deforest summary ----------
+//│ Call(Ref(member:AA),List(Arg(false,Ref(member:A))))@1@inst_0_tsni --> Ref(x)@inst_0_tsni@AA
+//│ Call(Ref(member:BB),List(Arg(false,Ref(member:B))))@2@inst_0_tsni --> Ref(x)@inst_0_tsni@BB
+//│ -------------- executing -------------
+//│ = 1
+//│ <<<<<<<<<<<<<<<<<<<<<<<<<<< Deforestation <<<<<<<<<<<<<<<<<<<<<<<<<<<
+
+
+// TODO: fusion opportunity is lost
+// when instantiation ids of dtors are also tracked
```
LPTK: I don't understand the comment. Care to explain?

hkmc2/shared/src/test/mlscript/deforest/append.mls
LPTK: I guess it's too much for this PR, right? Let's add it to an MLscript-Lumberhack tracking issue.

hkmc2/shared/src/test/mlscript/deforest/new-todo.mls
ychenfo: When I was looking into the explanation for this I realized that this may not be the best example to illustrate the issue here. I think a better one looks like this:
```
fun f1(x) = if x is A then 1 else 2
fun f2(x) = if x is A then 2 else 3
fun f3(x) = if x is A then 3 else 4
let p = if true then AA(A) else BB(B)
if p is
  AA(x) then f1(x)
  BB(x) then f2(x) + f3(x)
```

The fusion doesn't happen now (but can happen some time before) is because now the symbol for the first pattern variable (`x` in the source program) in both branches (`is AA(x) then ...; is BB(x) then ...`) is shared, which means that the program after lowering looks like:
```
if p is
  AA then
    param0 = p.x
    x1 = param0
    f1(x1)
  BB then
    param0 = p.x
    x2 = param0
    f2(x2) + f3(x2)
```

The shared `param0` in both branches leads to both `A` (in `AA(A)`) and `B` (in `BB(B)`) flowing into all `f1`, `f2` and `f3`, causing clash. Before, `param0` is not shared (the first branch uses `param0` and the second branch uses `param1`), so that `A` in `AA(A)` could be fused, and only `B` in `BB(B)` couldn't get fused.

I said this program in the difftest file is not a good example because the fusion can also happen if we do not distinguish two call sites for the consumer `f`.

I am trying to address this issue by generating different type variables for the shared `param0` in different branches.

hkmc2/shared/src/test/mlscript/deforest/new-todo.mls
ychenfo: Sure, here I mean that if we do not distinguish different call sites for consumers, we may be able to fuse again. Now this cannot fuse because there are two call sites for the `inner` function (which turned out to be only a consumer), so two different `inner` functions are considered to be called at these two call sites, causing strategy clash. This may also be related to the merging of consumers (in the next program in `new-todo.mls`).

hkmc2/shared/src/test/mlscript/deforest/append.mls
ychenfo: Yes I think so... I thought I had an idea for the reification, but it didn't work... I will add this to the tracking issue. Thanks for the reminder!

ychenfo:
Superseded by #335

PR #290, opened 2025-03-11
Defunkify lifter and related middle-IR code

LPTK:
There was a bunch of funky stuff happening both in the lifter and in related middle-IR code.

The only reason I made this a PR is for @CAG2Mark to have a chance to review it and learn a bit in the process.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
@@ -303,23 +303,27 @@ sealed abstract class Defn:
     case ClsLikeDefn(preCtor = preCtor, ctor = ctor, methods = mtds) =>
       preCtor :: ctor :: mtds.flatMap(_.subBlocks)
   
+  // * Note that `privateFields` abd `publicFields` can't possibly be free since they are never
+  // * referred to directly (they are only accessed through selections).
+  // * At some point we'll want to make `Local` more specific than `Symbol` to express this
+  // * in the type system.
   lazy val freeVars: Set[Local] = this match
     case FunDefn(own, sym, params, body) => body.freeVars -- params.flatMap(_.paramSyms) - sym
     case ValDefn(owner, k, sym, rhs) => rhs.freeVars
     case ClsLikeDefn(own, isym, sym, k, paramsOpt, auxParams, parentSym, 
-      methods, privateFields, publicFields, preCtor, ctor) =>
+        methods, privateFields, publicFields, preCtor, ctor) =>
       preCtor.freeVars
         ++ ctor.freeVars ++ methods.flatMap(_.freeVars)
-        -- privateFields -- publicFields.map(_.sym) -- auxParams.flatMap(_.paramSyms)
+        -- auxParams.flatMap(_.paramSyms)
   
   lazy val freeVarsLLIR: Set[Local] = this match
     case FunDefn(own, sym, params, body) => body.freeVarsLLIR -- params.flatMap(_.paramSyms) - sym
     case ValDefn(owner, k, sym, rhs) => rhs.freeVarsLLIR
     case ClsLikeDefn(own, isym, sym, k, paramsOpt, auxParams, parentSym, 
-      methods, privateFields, publicFields, preCtor, ctor) =>
+        methods, privateFields, publicFields, preCtor, ctor) =>
       preCtor.freeVarsLLIR
         ++ ctor.freeVarsLLIR ++ methods.flatMap(_.freeVarsLLIR)
-        -- privateFields -- publicFields.map(_.sym) -- auxParams.flatMap(_.paramSyms)
+        -- auxParams.flatMap(_.paramSyms)
   
 final case class FunDefn(
     owner: Opt[InnerSymbol],
@@ -347,12 +351,10 @@ final case class ClsLikeDefn(
     parentPath: Opt[Path],
     methods: Ls[FunDefn],
     privateFields: Ls[TermSymbol],
-    publicFields: Ls[TermDefinition],
+    publicFields: Ls[BlockMemberSymbol],
     preCtor: Block,
     ctor: Block,
 ) extends Defn:
-  publicFields.foreach: f =>
-    require(f.owner.contains(isym))
   val innerSym = S(isym)
 
 final case class Handler(
@@ -405,6 +407,7 @@ sealed abstract class Result:
     case Value.Lit(lit) => Set.empty
     case Value.Lam(params, body) => body.freeVars -- params.paramSyms
     case Value.Arr(elems) => elems.flatMap(_.value.freeVars).toSet
+    case Value.Rcd(elems) => elems.flatMap(_.value.freeVars).toSet
     case DynSelect(qual, fld, arrayIdx) => qual.freeVars ++ fld.freeVars
 
   lazy val freeVarsLLIR: Set[Local] = this match
@@ -416,6 +419,7 @@ sealed abstract class Result:
     case Value.Lit(lit) => Set.empty
     case Value.Lam(params, body) => body.freeVarsLLIR -- params.paramSyms
     case Value.Arr(elems) => elems.flatMap(_.value.freeVarsLLIR).toSet
+    case Value.Rcd(elems) => elems.flatMap(_.value.freeVarsLLIR).toSet
     case DynSelect(qual, fld, arrayIdx) => qual.freeVarsLLIR ++ fld.freeVarsLLIR
   
 // type Local = LocalSymbol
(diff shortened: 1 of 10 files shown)


hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -347,12 +351,10 @@ final case class ClsLikeDefn(
     parentPath: Opt[Path],
     methods: Ls[FunDefn],
     privateFields: Ls[TermSymbol],
-    publicFields: Ls[TermDefinition],
+    publicFields: Ls[BlockMemberSymbol],
```
LPTK, line 354: The only reason we were storing `TermDefinition`s here was to remember which things are supposed to be public fields. But clearly it's wrong to store elaborated terms in a middle-IR data structure. All we needed here was just the symbol.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockTransformer.scala
```diff
@@ -147,14 +147,17 @@ class BlockTransformer(subst: SymbolSubst):
     if (own2 is fun.owner) && (sym2 is fun.sym) && (params2 is fun.params) && (body2 is fun.body)
       then fun else FunDefn(own2, sym2, params2, body2)
   
+  def applyValDefn(defn: ValDefn): ValDefn =
```
LPTK, line 150: (Only reason I extracted this is because it used to be needed in my previous refactoring; I kept it for consistency with `FunDefn`.)

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockTraverser.scala
```diff
@@ -10,9 +10,12 @@ import os.write.over
 // These all work like BlockTransformer and its derivatives, but do not rewrite the block. See BlockTransformer.scala.
 // Please use this instead of BlockTransformer for static analysis.
 
-class BlockTraverser(subst: SymbolSubst):
```
LPTK, line 13: Why would you need a symbol substitution in a block _traverser_? 🤔

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -153,7 +153,7 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
     val ignoredDefns: Set[BlockMemberSymbol] = Set.empty,
     val inScopeDefns: Map[BlockMemberSymbol, Set[BlockMemberSymbol]] = Map.empty,
     val modLocals: Map[BlockMemberSymbol, Local] = Map.empty,
-    val localCaptureSyms: Map[Local, (VarSymbol, BlockMemberSymbol)] = Map.empty,
```
LPTK, line 156: It seems the `BlockMemberSymbol` info was redundant.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -234,55 +236,38 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
...
-          Nil
-        )
-        fldSym.defn = S(fldDef)
-        s -> (
-          varSym,
-          fldDef,
-        )
-      .toMap
-
-    val varsList = cap.toList
```
LPTK, line 260: You can't `toList` a set to get some arbitrary order, because that order is not well-defined. It could change depending on the object's memory address next time the compiler runs after minor modifications. But we need stable and reproducible compilation outputs.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -234,55 +236,38 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
     val FreeVars(_, cap) = ctx.usedLocals(f.sym)
 
     val fresh = FreshInt()
-
-    val varsMap = cap.map: s =>
```
LPTK, line 238: The creation and subsequent repeated querying of this intermediate map was entirely gratuitous and resulted in needless overhead.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -234,55 +236,38 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
...
+      varSym.decl = S(p) // * Currently this is only accessed to create the class' toString method
+      
+      val vd = ValDefn(
+        S(clsSym),
+        syntax.ImmutVal,
+        fldSym,
+        Value.Ref(varSym)
+      )
+      
+      (sym -> varSym, p, vd)
```
LPTK, line 257: I now compute all the relevant things in one place that's much easier to inspect and understand.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -234,55 +236,38 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
...
     val defn = ClsLikeDefn(
       None, clsSym, BlockMemberSymbol(nme, Nil),
       syntax.Cls,
-      S(PlainParamList(varsList.map: s => 
-        val sym = varsMap(s)._1
-        val p = Param(FldFlags.empty.copy(value = true), sym, None)
-        sym.decl = S(p)
-        p
-      )),
+      S(PlainParamList(sortedVars.iterator.map(_._2).toList)),
```
LPTK, line 262: I use `iterator` here and below to avoid allocating intermediate collections.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -234,55 +236,38 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
     val FreeVars(_, cap) = ctx.usedLocals(f.sym)
 
     val fresh = FreshInt()
-
-    val varsMap = cap.map: s =>
-        val id = fresh.make
-        val nme = s.nme + id + "$"
-        val varSym = VarSymbol(Tree.Ident(nme))
-        val fldSym = BlockMemberSymbol(nme, Nil)
-        val fldDef = TermDefinition(
```
LPTK, line 243: @FlandiaYingman turns out we didn't need that crazy dummy term definition at all. And we didn't even need a `ValDefn`, either.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -223,7 +223,9 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
     *  and `BlockLocalSymbol` (for the class fields) in the class, and
```
FlandiaYingman: and the funny comment here about `BlockLocalSymbol`

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
FlandiaYingman: I guess maybe we want to the funny comment about the `BlockMemberSymbol` right above for `localCaptureSyms`, which was added by me 😬

LPTK: Ah you're right, the documentation is now out of sync.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockTransformer.scala
CAG2Mark: For consistency, we should also probably extract `applyClsLikeDefn`. Note that this may require some changes to overridden `applyDefn` methods.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockTraverser.scala
```diff
@@ -120,12 +127,12 @@ class BlockTraverser(subst: SymbolSubst):
     applySubBlock(lam.body)
   
   def applyTermDefinition(td: TermDefinition): Unit =
```
CAG2Mark: This can now be removed.

CAG2Mark (approved):
Looks good. There's just some small things you may have overlooked in addition to @FlandiaYingman's comments.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockTransformer.scala
LPTK: I know but ain't nobody got time for dat.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockTraverser.scala
LPTK: Good catch, I missed that one!

PR #291, opened 2025-03-13
Use handlers to implement debugging functions

CAG2Mark:
- The runtime now prints out a handler trace (similar to a stack trace) when there are unhandled effects (by @AnsonYeung)
- Added a function to continuation classes that can get all in-scope local variables
- [x] Port fixes from benchmarking branch
- [x] getLocals cause excessive codegen which is not needed. Generating local information could be done with additional flags so they don't pollute as much test outputs.
- [x] Implement the interface to invoke getLocal and obtain local variables information.
- [x] decode pc into source information.


hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -294,10 +342,12 @@ class HandlerLowering(paths: HandlerPaths)(using TL, Raise, Elaborator.State, El
    * 3. float out definitions
    */
   
-  private def translateBlock(b: Block, h: HandlerCtx): Block =
-    given HandlerCtx = h
+  private def translateBlock(b: Block, extraLocals: Set[Local], h: HandlerCtx, debugNme: Str): Block =
```
CAG2Mark: Could `debugNme` be moved to `DebugInfo`?

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -342,18 +392,19 @@ class HandlerLowering(paths: HandlerPaths)(using TL, Raise, Elaborator.State, El
           ResultPlaceholder(res, freshId(), c2, k(Value.Ref(res)))
         case r => super.applyResult2(r)(k)
       override def applyLam(lam: Value.Lam): Value.Lam =
-        Value.Lam(lam.params, translateBlock(lam.body, functionHandlerCtx(s"Cont$$lambda$$")))
+        Value.Lam(lam.params, translateBlock(lam.body, lam.params.paramSyms.toSet, functionHandlerCtx(s"Cont$$lambda$$"), "‹lambda›"))
```
CAG2Mark: Note that lambdas are desugared before they get to this pass. Maybe we could move the `‹lambda›` name inside the symbols by modifying the lambda desugaring pass.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -26,56 +26,75 @@ object HandlerLowering:
...
+        debugInfo: DebugInfo,
+        linkAndHandle: LinkState => Block,
+    ):
+    def nestDebugScope(locals: Set[Local], localsFn: Path) = copy(debugInfo = DebugInfo(
+      debugInfo.inScopeLocals ++ locals, S(localsFn)))
+  
+  // inScopeLocals: Local variables that are in scope.
+  // prevLocalsFn: The function that gets the outer function's locals.
+  private case class DebugInfo(
+      inScopeLocals: Set[Local],
```
CAG2Mark: These should probably only have one more level of indentation (the point of having more indentation is to have them on a different indentation level than the class members and methods).

CAG2Mark (commented):
Some small comments on things I noticed.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -393,6 +393,9 @@ class HandlerLowering(paths: HandlerPaths)(using TL, Raise, Elaborator.State, El
           ResultPlaceholder(res, freshId(), c2, k(Value.Ref(res)))
         case r => super.applyResult2(r)(k)
       override def applyLam(lam: Value.Lam): Value.Lam =
+        // This should normally be unreachable, but we can just emit a warning and proceed with the transformation
+        raise(WarningReport(msg"BUG: Unexpected lambda during handler lowering" -> lam.toLoc :: Nil,
```
LPTK: The correct thing to report here is an
```suggestion
        raise(InternalError(msg"Unexpected lambda during handler lowering" -> lam.toLoc :: Nil,
```

hkmc2/shared/src/test/mlscript/handlers/StackSafety.mls
```diff
@@ -130,21 +131,21 @@ sum(10000)
 //│   return prev
 //│ };
 //│ sum1 = function sum(n) {
-//│   let scrut, tmp, tmp1, curDepth, stackDelayRes, getLocals2, Cont$func$sum$StackSafety$_mls_L118_4_57$1;
+//│   let scrut, tmp, tmp1, curDepth, stackDelayRes, getLocals2, Cont$func$sum$StackSafety$_mls_L119_4_57$1;
 //│   getLocals2 = function getLocals() {
 //│     let n1, prev, thisInfo, tmp2;
 //│     prev = getLocals1();
-//│     n1 = new runtime.LocalVarInfo.class("n", "697", n);
+//│     n1 = new runtime.LocalVarInfo.class("n", "699", n);
```
LPTK: Q: Why are these numbers changing all the time?

AnsonYeung: It is the uid of the symbol of the local var

LPTK: Why do we need  this?

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
AnsonYeung: the name generated is kind of weird... not sure about how to change it atm as it also affect a lot of other test output

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -24,7 +24,28 @@ sealed abstract class Block extends Product with AutoLocated:
   
   def ~(that: Block): Block = Begin(this, that)
   
-  protected def children: Ls[Located] = ??? // Maybe extending AutoLocated is unnecessary
+  protected def children: Ls[Located] = this match
```
LPTK, line 27: Wouldn't this better be implemented with a `BlockTraverser`? It would be more concise and efficient.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -387,7 +410,19 @@ enum Case:
 
 sealed trait TrivialResult extends Result
 
-sealed abstract class Result:
+sealed abstract class Result extends AutoLocated:
+
+  protected def children: List[Located] = this match
```
LPTK, line 414: I guess this would be handled by the same `BlockTraverser` mentioned above.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -41,7 +62,9 @@ sealed abstract class Block extends Product with AutoLocated:
     case Define(defn, rst) =>
       val rest = rst.definedVars
       if defn.isOwned then rest else rest + defn.sym
-    case HandleBlock(lhs, res, par, args, cls, hdr, bod, rst) => bod.definedVars ++ rst.definedVars + lhs
+    // we do not consider lhs and anything inside the body, as those are inside the handle block, which
+    // be moved into a new function definition
```
```suggestion
    // Note that the handler's LHS and body are not part of the current block, so we do not consider them here.
```

hkmc2/shared/src/test/mlscript/handlers/StackSafety.mls
LPTK: IIRC in a previous meeting we concluded this ID thing is not useful yet... Could you please remind me whether we really intend to use it later, and for what?

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -294,13 +344,16 @@ class HandlerLowering(paths: HandlerPaths)(using TL, Raise, Elaborator.State, El
...
+    val getLocalsFn = createGetLocalsFn(b, extraLocals)(using h)
+    given HandlerCtx = h.nestDebugScope(b.userDefinedVars ++ extraLocals, getLocalsFn.sym.asPath)
+    
     val stage1 = firstPass(b)
-    val stage2 = secondPass(stage1)
+    val stage2 = secondPass(stage1, getLocalsFn)
     if h.isTopLevel then stage2 else thirdPass(stage2)
   
   private def firstPass(b: Block)(using HandlerCtx): Block =
+    val getLocalSym = ctx.builtins.debug.getLocals
```
```suggestion
    val getLocalsSym = ctx.builtins.debug.getLocals
```

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -420,21 +490,23 @@ class HandlerLowering(paths: HandlerPaths)(using TL, Raise, Elaborator.State, El
...
-  private def genContClass(b: Block)(using HandlerCtx): Opt[ClsLikeDefn] =
+  private def genContClass(b: Block)(using h: HandlerCtx): Opt[ClsLikeDefn] =
     val clsSym = ClassSymbol(
       Tree.DummyTypeDef(syntax.Cls),
       Tree.Ident(handlerCtx.contName)
     )
     
     val pcVar = VarSymbol(pcIdent)
+
+    // Also used to check if the block is trivial
```
LPTK: Could you please elaborate on this comment? (Not here; in the comment.)

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -449,7 +521,7 @@ class HandlerLowering(paths: HandlerPaths)(using TL, Raise, Elaborator.State, El
     val actualBlock = handlerCtx.ctorThis match
       case N => prepareBlock(b)
       case S(thisPath) => Begin(prepareBlock(b), Return(thisPath, false))
-    if trivial then return N
+    if pcToLoc.isEmpty then return N
```
LPTK, line 527: Why does the absence of Loc mean triviality? 🤔

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -504,7 +504,7 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
 
     val modLocal = d match
       case c: ClsLikeDefn if modOrObj(c) && !ctx.ignored(c.sym) => parentCls match
-        case None => S(VarSymbol(Tree.Ident(c.sym.nme + "$")))
+        case None => S(FlowSymbol(c.sym.nme + "$"))
```
LPTK: Why this change? `FlowSymbol` is mostly meant for type checking/analysis.

hkmc2/shared/src/test/mlscript-compile/Runtime.mls
```diff
@@ -1,5 +1,4 @@
-import "./RuntimeJS.mjs" 
-
```
LPTK, line 2: Why this change?

hkmc2/shared/src/test/mlscript-compile/RuntimeJS.mjs
```diff
@@ -1,9 +1,11 @@
+import Predef from "./Predef.mjs";
 
 const RuntimeJS = {
   try_catch(computation, onError) {
     try { return computation() }
     catch (error) { return onError(error) }
-  }
+  },
+  Predef
```
LPTK: This is a bit funny. Why re-export Predef here?

hkmc2/shared/src/test/mlscript/handlers/BadHandlers.mls
```diff
@@ -21,11 +21,12 @@ x
 //│ ║  l.19: 	x
 //│ ╙──      	^
 
+// Getter
```
LPTK, line 25:
```suggestion

// Getter
```

LPTK (approved):
Nice changes!

hkmc2/shared/src/test/mlscript/handlers/StackSafety.mls
CAG2Mark: I added it in case the debug client finds it useful, i.e. so you can jump to a certain variable's definition site by clicking it in the sidebar. But I think we mentioned that this is fine because you need to add a certain flag for all the debug stuff to show (it should be hidden in this file).

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
CAG2Mark: In this case I'm not sure, as it's not always obvious exactly which fields to include in the list. For example in `ValDefn`, `owner` should not be included, but these things are easy to miss when using a BlockTraverser.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
AnsonYeung: Updated the comment

hkmc2/shared/src/test/mlscript-compile/RuntimeJS.mjs
AnsonYeung: This is the workaround for the cyclic import problem discussed in the meeting

hkmc2/shared/src/test/mlscript/handlers/StackSafety.mls
AnsonYeung: Well you can include the location instead for that purpose anyway? And during editing the uid will change a lot depending on the surrounding while the location might still be useful if there is some smart way of updating the location based on the text inserted.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
AnsonYeung: AutoLocated contains some caching mechanism so that it is computed at most once for each block.

LPTK: Ok, I guess it wouldn't be that beneficial to use `BlockTraverser`, then.

hkmc2/shared/src/test/mlscript-compile/RuntimeJS.mjs
LPTK: Right, and IIRC, in the meeting I explained that the proper way of solving this is to move the rendering logic to a `Rendering` module that is imported both by `Predef` and `Runtime` (where `Predef` re-exports some of its contents), not to create some cursed cyclic import.

AnsonYeung: Moved it to a new file. I also moved some other function depended by render function (map, fold, pass, etc)

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
AnsonYeung: cc: @CAG2Mark

PR #293, opened 2025-03-18
Clean up Middle IR and variable assignment/reassignment

LPTK:
I'm trying to refactor the currently messy handling of variables in MIR.

- [X] Distinguish initial variable assignments from mutable reassignments
- [ ] Add a way of specifying where variables are bound. Is this really necessary, though, with the distinction above? Now we can differentiate when a loop allocates a variable within its body versus when it changes an already defined one.
- [ ] Later: clean up the messy ad-hoc logic around local `val`s (that have no owner), private fields defined with `let` and `val`-less class parameters, etc.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
@@ -67,10 +67,10 @@ class HandlerLowering(paths: HandlerPaths)(using TL, Raise, Elaborator.State, El
   private def funcLikeHandlerCtx(ctorThis: Option[Path], isHandlerMtd: Bool, nme: Str) =
     HandlerCtx(false, false, nme, ctorThis, state =>
       blockBuilder
-        .assignFieldN(state.res.contTrace.last, nextIdent, Instantiate(
+        .reassignFieldN(state.res.contTrace.last, nextIdent, Instantiate(
           state.cls.selN(Tree.Ident("class")),
           Value.Lit(Tree.IntLit(state.uid)) :: Nil))
-        .assignFieldN(state.res.contTrace, lastIdent, state.res.contTrace.last.next)
+        .reassignFieldN(state.res.contTrace, lastIdent, state.res.contTrace.last.next)
         .ret(state.res))
   private def functionHandlerCtx(nme: Str) = funcLikeHandlerCtx(N, false, nme)
   private def topLevelCtx(nme: Str) = HandlerCtx(true, false, nme, N, _ => rtThrowMsg("Unhandled effects"))
@@ -270,12 +270,15 @@ class HandlerLowering(paths: HandlerPaths)(using TL, Raise, Elaborator.State, El
       case Assign(lhs, rhs, rest) =>
         val PartRet(head, parts) = go(rest)
         PartRet(Assign(lhs, rhs, head), parts)
-      case blk @ AssignField(lhs, nme, rhs, rest) =>
+      case Reassign(lhs, rhs, rest) =>
         val PartRet(head, parts) = go(rest)
-        PartRet(AssignField(lhs, nme, rhs, head)(blk.symbol), parts)
-      case AssignDynField(lhs, fld, arrayIdx, rhs, rest) =>
+        PartRet(Reassign(lhs, rhs, head), parts)
+      case blk @ ReassignField(lhs, nme, rhs, rest) =>
         val PartRet(head, parts) = go(rest)
-        PartRet(AssignDynField(lhs, fld, arrayIdx, rhs, head), parts)
+        PartRet(ReassignField(lhs, nme, rhs, head)(blk.symbol), parts)
+      case ReassignDynField(lhs, fld, arrayIdx, rhs, rest) =>
+        val PartRet(head, parts) = go(rest)
+        PartRet(ReassignDynField(lhs, fld, arrayIdx, rhs, head), parts)
       case Return(_, _) => PartRet(blk, Nil)
       // ignored cases
       case TryBlock(sub, finallyDo, rest) => ??? // ignore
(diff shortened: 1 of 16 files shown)


LPTK:
@CAG2Mark The lifter is not really updated. There are tests failing as it does the wrong thing in some cases (see the diff of `FunInFun.mls`).

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -455,8 +457,13 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
     blockPreamble(p.main) -> (imps.mkDocument(doc" # ") :/: returningTerm(p.main, endSemi = false).stripBreaks)
   
   def blockPreamble(t: Block)(using Raise, Scope): Document =
+    
     // TODO document: mutable var assnts require the lookup
     val vars = t.definedVars.toSeq.filter(scope.lookup(_).isEmpty).sortBy(_.uid).iterator.map(l =>
+    
+    // TODO: now should be possible to just write: (currently causes problems to handlers)
+    // val vars = t.definedVars.toSeq.sortBy(_.uid).iterator.map(l =>
```
LPTK, line 465: @CAG2Mark @AnsonYeung There are probably many instances where the effect handler transformation reassigns non-mutable symbols instead of simply assigning them. In principle it should be possible to make this change, but it breaks code that uses handlers.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -486,7 +489,7 @@ class HandlerLowering(paths: HandlerPaths)(using TL, Raise, Elaborator.State, El
     
     val resumedVal = VarSymbol(Tree.Ident("value$"))
 
-    def createAssignment(sym: Local) = Assign(sym, resumedVal.asPath, End())
+    def createAssignment(sym: Local) = Reassign(sym, resumedVal.asPath, End()) // TODO: or is this `Assign`?
```
LPTK, line 492: Cc @CAG2Mark @AnsonYeung

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -601,24 +601,27 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
...
           val initial = blk.assign(local, createCall(bms, ctx))
           ctx.defns(bms) match
-            case c: ClsLikeDefn => initial.assignFieldN(local.asPath, Tree.Ident("class"), bms.asPath)
+            case c: ClsLikeDefn => initial.reassignFieldN(local.asPath, Tree.Ident("class"), bms.asPath)
             case _ => initial
       
       val remaining = rewritten match
-        case Assign(lhs: InnerSymbol, rhs, rest) => ctx.getIsymPath(lhs) match
+        
+        // TODO (LP:) not sure about the semantics of these "Reassign" cases
```
LPTK, line 609: @CAG2Mark 👀

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
AnsonYeung: Indeed the continuation class implicitly captures the variables so all assignments within continuation should be reassignments really (including the this.pc assignments), even though they might not be assigned or initialized in the first place yet.

A brief remark of why capturing is needed: any function could define other nested function that capture the same variable, and it is important for the continuation class to capture the same one.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
AnsonYeung: This is always a reassignment, as the old symbol would hold the previous EffectSig

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
LPTK: > Indeed the continuation class implicitly captures the variables so all assignments within continuation should be reassignments really (including the this.pc assignments), even though they might not be assigned or initialized in the first place yet.

I'm not following. Doesn't the continuation class just duplicate the function's implementation? If the function was assigning the variable for the first time (maybe the variable is even immutable), then the continuation class should also be assigning that variable for the first time in the corresponding part of its duplicated code. No?

AnsonYeung: Oh right, so we need proper binding site probably. With the changes in the comment above it would make the initial assignment to create a local variable at the function declaration which is incorrect as the resume function can return a effectsig and resume later which uses the same variable again.

LPTK: Ah right, I see. That's probably why it doesn't work currently.

We should list the variables that are bound within a given MIR block body.

LPTK:
In today's meeting, we figured that we currently don't actually need this change. We just need to change the way `Label` treats assignments, so it is basically treated similarly to a local function with its own scope.

LPTK:
In a more recent meeting, we figured that the semantics of `Label` may not need to change. (Note: other places, like the UCS, use `Label` (for join points) without needing it to have its own scope – I guess it's always the case when `loop = false`, though.)

We decided that the more general way of going at it is to simply:
* desugar `while` to use a recursive function
* rely on the lifter to move that function moved to the top level
* compile mutually-tail-recursive functions to label-loops

Alternatively, we could at least palce the bodies of `while` loops in local functions.

The downside of keeping the current `Label` semantics and just working around it is that it makes inlining generally unsafe... so the whole approach seems doomed. We really should have a way of specifying sub-scopes when needed.

Issue #272, opened 2025-02-07
Resolver & Type Classes Tracking Issue


FlandiaYingman:
This issue tracks the problems that are _neither urgent nor pressing_ related to the resolver & type classes feature.

### Summer UROP

- [x] Unsoundness of partial application.
- [x] Support explicitly specifying contextual argument.
- [x] Change `use` keyword into a predef function.
- [x] Reused mutable terms, mostly of UCS normalization. (see https://github.com/hkust-taco/mlscript/pull/297#discussion_r2060561339). 

---

- [ ] Separate `Value.Ref` into two cases to make fewer illegal states representable.
- [ ] https://github.com/hkust-taco/mlscript/pull/343#discussion_r2522287364

### Later

- [ ] Handle cyclic resolution
  ```fs
  M.List.map
  
  module M with
    val List: M2.List = M2.List
  
  module M2 with
    val List: M.List = M.List
  ```

- [ ] Move selections' symbol resolution to the resolution phase and refactor the `resolvedSymbol` part.
  - [x] Before that, move UCS normalization to/after the resolution phase.

<details><summary>
<h3> Completed Tasks </h3>
</summary>
- [x] Resolution possibly leading to more symbol resolution.
  ```fs
  fun foo[A](using foo: Foo[A]): A
  
  use Foo[List]
  foo.map
  ```

- [x] Local functions as module methods (as their definitions always exist).

- [x] It resolves implicit application through `Sel` and `SynthSel` (because they may lead to an implicit module method call), but `Ref`.
  ```fs
  Sel(M, f) // is resolved to App(Sel(M, f), CtxArgs)
  SynthSel(M, f) // is resolved to App(SynthSel(M, f), CtxArgs)
  Ref(M.f) // is not, and is expected to be, resolved to App(Ref(M.f), CtxArgs)
  ```

- [x] Module information of definitions that returns a module
  ```fs
  module A with
    fun f() = 2
  
  module B with
    val a: module A = A
  
  print(B.a) // B.a should be a module, and should be rejected passing to a regular parameter
  ```

- [x] #274

- [x] #294 

- [x] Add symbols to all possibly moduleful terms. (https://github.com/hkust-taco/mlscript/issues/294#issuecomment-2745300607)

- [x] Support `use[...]` syntax as in (see https://github.com/hkust-taco/mlscript/pull/282#issuecomment-2676133410)
  ```fs
  fun fib(using Yield[Int]) =
    fun inner(a, b) =
[...]

PR #295, opened 2025-03-22
Move module method checks

### LINKED ISSUE #294: Move all module method checking logic to the resolution stage.
We have the following rules:

- [x] Module parameters must have an explicit type.
- [x] Module parameters must have a concrete type.
- [x] Function returning module values must have an explicit return type.
- [x] Function returning module values must have a concrete return type.
- [x] The return type of functions returning module values must be prefixed with `module` keyword.
- [x] Only module arguments (values) may be passed to module parameters.
- [x] Only module methods may return module values.

Now we want to move most of them to the resolution stage, which used to be in the elaboration stage. This is because at elaboration the elaborator doesn't always know if a term is a _module_ or not.

Some simple checks are still preserved in the elaboration stage: 

- [x] Terms with `module` modifiers must have an explicit and concrete type.

And we will move the following checks to the resolution stage:

- [x] Module terms can only occur at the following positions.
  - [x] As a _module parameter_.
  - [x] As the RHS of a term definition that is modified by `module`.
  - [x] After a `return` in a function that returns a module.

A module term is some term that evaluates to a module. They include

- A `Ref` to a `ModuleSymbol`.
- A `Ref` to a `TermDefinition` that returns some module term.
  - Note: a `Ref` to a function returning module without `App` is also considered "moduleful".
- A `App` to a `TermDefinition` that returns some module term.
- A `Ref` to a module parameter.

Note that there doesn't exist a definition for let-bindings, so it couldn't be bind to a module value for now.

FlandiaYingman:
Close #294.


hkmc2/shared/src/test/mlscript/ucs/examples/EitherOrBoth.mls
```diff
@@ -91,3 +91,9 @@ fun eobToString[A, B](eob: EitherOrBoth[A, B]): Str =
     Left(left) then "Left(" ~ Str.from(left) ~ ")"
     Right(right) then "Right(" ~ Str.from(right) ~ ")"
     Both(left, right) then "Both(" ~ Str.from(left) ~ ", " ~ Str.from(right) ~ ")"
+//│ FAILURE: Unexpected type error
+//│ FAILURE LOCATION: resolve (ImplicitResolver.scala:122)
+//│ ╔══[ERROR] Unexpected moduleful reference.
+//│ ║  l.89: 	fun eobToString[A, B](eob: EitherOrBoth[A, B]): Str =
+//│ ║        	                                                ^^^
+//│ ╙── Function must be marked as 'module' in order to have a module signature.
```
FlandiaYingman: Here's a problem with overloaded definitions. There is a `Str` class declaration in the prelude and a `Str` module in `mlscript-compile/Str.mls`. What I expect is that there is a `BlockMemberSymbol` of the two `Str`, but currently there are two `BlockMemberSymbol` and the imported one from `mlscript-compile/Str.mls` shadows the one from prelude. The consequence here (and in some other test cases) is that `Str` is recognized as a module instead of a class. If it were one `BlockMemberSymbol` containing one `ClassSymbol` and one `ModuleSymbol`, then we should be able to tell which kind of symbol has priority here.

hkmc2/shared/src/test/mlscript/nofib/constraints.mls
```diff
@@ -78,8 +78,23 @@ fun inconsistencies(csp, as_) = if csp is CSP(vars, vals, rel) then
...
+//│ ╔══[ERROR] Expected 2 arguments, got 1
+//│ ║  l.80: 	fun test(csp) = filter(consistent(csp))
+//│ ╙──      	                       ^^^^^^^^^^
 
 fun solver(csp) = test(csp, generate(csp))
+//│ FAILURE: Unexpected type error
+//│ FAILURE LOCATION: resolveApp (ImplicitResolver.scala:168)
+//│ ╔══[ERROR] Expected 1 arguments, got 2
+//│ ║  l.92: 	fun solver(csp) = test(csp, generate(csp))
+//│ ╙──      	                  ^^^^
```
FlandiaYingman: Since now all global functions have the ability of module methods, it checks the arity now. In `nofib` there are some places that the arity does not match, which I am not sure if that's intended 🤡

hkmc2/shared/src/test/mlscript/codegen/Spreads.mls
```diff
@@ -11,19 +11,39 @@ let b = [0, ...a]
 fun foo(w, x, y, z) = [w, x, y, z]
 
 foo(...b)
```
FlandiaYingman, line 13: I'll have to fix this later. Just adding a comment for record.

hkmc2/shared/src/test/mlscript/codegen/Spreads.mls
```diff
@@ -11,19 +11,39 @@ let b = [0, ...a]
...
+//│ FAILURE: Unexpected type error
+//│ FAILURE LOCATION: resolveApp (ImplicitResolver.scala:168)
+//│ ╔══[ERROR] Expected 4 arguments, got 1
+//│ ║  l.13: 	foo(...b)
+//│ ╙──      	^^^
 //│ = [0, 1, 2, 3]
```
FlandiaYingman: I'll have to fix this later. Just adding a comment for record.

hkmc2/shared/src/test/mlscript/basics/ModuleMethods.mls
FlandiaYingman, line 1: I rewrote this test file so that it tests more cases (while keeping the previous ones).

hkmc2/shared/src/test/mlscript/basics/MemberProjections.mls
```diff
@@ -19,45 +19,142 @@ foo.n(1)
...
+//│ 	at: mlscript.utils.package$.lastWords(package.scala:230)
+//│ 	at: hkmc2.semantics.ImplicitResolver.resolveApp$$anonfun$3$$anonfun$2(ImplicitResolver.scala:175)
+//│ 	at: scala.runtime.function.JProcedure1.apply(JProcedure1.java:15)
+//│ 	at: scala.runtime.function.JProcedure1.apply(JProcedure1.java:10)
+//│ 	at: scala.collection.immutable.List.foreach(List.scala:334)
+//│ 	at: hkmc2.semantics.ImplicitResolver.resolveApp$$anonfun$3(ImplicitResolver.scala:165)
+//│ 	at: hkmc2.semantics.ImplicitResolver.resolveApp$$anonfun$adapted$1(ImplicitResolver.scala:179)
+//│ 	at: hkmc2.utils.TraceLogger.trace(TraceLogger.scala:17)
+//│ 	at: hkmc2.semantics.ImplicitResolver.resolveApp(ImplicitResolver.scala:179)
+//│ 	at: hkmc2.semantics.ImplicitResolver.resolve$$anonfun$3(ImplicitResolver.scala:132)
```
FlandiaYingman: Here's an error that I didn't have time to look at. Will take a look later.

FlandiaYingman (commented):
This is what I currently have and there are some problems in the test cases. 

Stricter module method rules have been enforced, to name a few:

```fs
// module-typed parameter without module modifier
fun f6(m: M) = m
// module-modified parameter without a module type
fun f6(module m: Int) = m
```

Two examples above used to pass the checks and they no more pass them now.

hkmc2/shared/src/test/mlscript/ucs/examples/EitherOrBoth.mls
LPTK: That's not a problem with overloaded definitions, though. Just conflicting imports/builtins. Overloads must be defined in the very same scope, otherwise it's madness.

hkmc2/shared/src/test/mlscript/nofib/constraints.mls
LPTK: Whoops cc @CrescentonC

hkmc2/shared/src/test/mlscript/ucs/examples/EitherOrBoth.mls
FlandiaYingman: I see. Then I'll have to rename the `Str` module in `mlscript-compile/Str.mls`. I guess it should be an error to _define_ a module named `Str` here.

FlandiaYingman: But we still have to deal with overloaded definitions. A symbol used to be recognized as a module as long as one of its overloaded definition is a symbol. 

I think it should be more clever in the sense that it should be able to know if the user wants a class type or module type:

```fs
class Str
module Str

// a module
fun f: module Str

// a class
fun f: Str
```

How do you think about this behavior? We will still perform checks to prevent user from really returning a modueful or non-moduleful term that doesn't match the signature.

LPTK: Yes, that's the intended behavior. `module Str` disambiguates the overloaded definition by selecting the module part. Same as other constructs, when not referred to in their expected context. For example, in
```
class Foo
fun Foo
```
then `f(Foo)` refers to the `fun`, but `f(class Foo)` refers to the class (using it as a first-class value).

LPTK: I've been wanting to add `Disambiguate(term, kind)(innerSymbol)` nodes to `Term` which would make it perfectly clear what we're looking at. Just haven't had time yet.

LPTK: > I see. Then I'll have to rename the `Str` module in `mlscript-compile/Str.mls`.

Note that eevntually `Str` should just refer to a module with helper functions along with an overloaded type declaration, probably both defined in Predef. `Str` actually has nothing to do in `Prelude`. We can fix these details later, when overloaded is supported.

> I guess it should be an error to _define_ a module named `Str` here.

I don't see why. We don't want to (and reasonably can't) prevent people from defining modules with the names they choose.

FlandiaYingman: > I don't see why. We don't want to (and reasonably can't) prevent people from defining modules with the names they choose.

You said it is "conflicting". So I guess this should be an error? (or shadowing the previous definition should be the intended behavior?)

Given that you said eventually `Str` should be a module with helper functions, I guess we don't rename `Str`, and currently let's just define it as an `object` so that it doesn't fail the module checks.

LPTK: Yes that's a fine plan for now.

> You said it is "conflicting". So I guess this should be an error? (or shadowing the previous definition should be the intended behavior?)

Yes, it's a name conflict. No, it shouldn't be an error to name something `Str`, obviously. And yes, if you import something that shadows something else, you shouldn't make a surprised pikatchu face when the old symbol is shadowed.

hkmc2/shared/src/test/mlscript/nofib/constraints.mls
ychenfo: Thanks for reporting this! Indeed this is an arity mismatch here and I missed it. This problem was not revealed before in the runtime because `test` and `solver` (and also `consistent`, `inconsistencies` and `generate`, which are called either directly or indirectly by `test`) are actually dead code (I confirmed that they are indeed dead code in the [original nofib haskell code](https://github.com/ghc/nofib/blob/master/spectral/constraints/Main.hs)...). So I think we can simply delete to fix the problem.

FlandiaYingman:
This should be ready for review. 

I didn't add symbols for all possibly moduleful terms as you suggested here (https://github.com/hkust-taco/mlscript/issues/294#issuecomment-2745202096). I tracked it in #272, and I'll try and address it in a following PR. (Otherwise there might be another conflict hell 🤡)

@CrescentonC would you look at dd3bb009d00e56b67ae71da9b7bb88a4380ca6cf? I removed the dead code you mentioned and some additional ones.

hkmc2/shared/src/test/mlscript/nofib/ansi.mls
```diff
@@ -44,7 +44,7 @@ fun returnn(s, consume) = consume(reverse(s))
 
 :...
 //│ ————————————————————————————————————————————————————————————————————————————————
-fun deletee(n, s, l, consume) = if n > 0 then writeString(nofibStringToList("BS_BS"), loop(n - 1, tail(s), l, consume)) else ringBell(loop(0, nofibStringToList(""), l, consume))
+fun deletee(n, s, l, consume) = if n > 0 then writeString(nofibStringToList("BS_BS"), loop(n - 1, tail(s), l, consume)) else ringBell(loop(0, nofibStringToList(""), l, consume), Nil)
```
ychenfo:
```suggestion
fun deletee(n, s, l, consume, d) = if n > 0 then writeString(nofibStringToList("BS_BS"), loop(n - 1, tail(s), l, consume), d) else ringBell(loop(0, nofibStringToList(""), l, consume), d)

fun loop(n, s, l, consume) = x => readChar of
  returnn(s, consume)
  (c, d) => if
            c == "B" then deletee(n, s, l, consume, d) 
            c == "D" then deletee(n, s, l, consume, d)
            c == "`" then returnn(s, consume)(d)
```

Thanks for the changes! I suggest this update to `ansi.mls` to further fix it. The number of arguments passed to `writeString` is also wrong, and `d` should also be used in the first 3 branches in `loop`. Indeed the problem is not noticed before also because the first 3 branches are never executed.

(It seems that github does not allow me refer to line 52 - 54, which also need to be updated. So I include my updates to line 52 - 54 here)

hkmc2/shared/src/test/mlscript/nofib/fish.mls
```diff
@@ -129,21 +129,6 @@ fun pseudocorner(arg, q6, q7) = quartet(corner2, side2, (a, b, c) => rot(side2,
...
-    (a1, a2, a3) => above of
-      1
-      1
-      (b5, b6, b7) => beside(1, 2, p4, (b5, b6, b7) => beside(1, 1, p5, p6, b5, b6, b7), b5, b6, b7)
-      (b5, b6, b7) => beside(1, 2, p7, (b5, b6, b7) => beside(1, 1, p8, p9, b5, b6, b7), b5, b6, b7)
-      a1
-      a2
-      a3
-    arg1
-    arg2
```
LPTK:
```suggestion
```

hkmc2/shared/src/test/mlscript/nofib/constraints.mls
```diff
@@ -44,42 +44,6 @@ fun maxLevel(ls) = if ls is
...
-      lscomp2(reverse(as_))
-  
-  lscomp1(as_)
-
-
-fun consistent(csp, x) = null_(inconsistencies(csp, x))
-
-fun test(csp) = filter(consistent(csp))
-
-fun solver(csp) = test(csp, generate(csp))
```
LPTK:
```suggestion
```

hkmc2/shared/src/test/mlscript/std/PredefTest.mls
```diff
@@ -172,6 +172,9 @@ fold(+) of
 :todo
 :re
```
LPTK, line 173:
```suggestion
:todo
:re
:e
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -173,7 +260,22 @@ class ImplicitResolver(tl: TraceLogger)
...
+    if p.flags.mod then
+      if p.sign.isEmpty then
+        raise(ErrorReport(msg"Module parameter must have explicit type." -> p.sym.toLoc :: Nil))
+    p.sign.foreach:
+      resolve(_, 
+        expectModule = S(p.flags.mod), 
+        errMsg = if p.flags.mod 
+          then S(msg"Module parameter must have a module type.")
+          else S(msg"Non-module parameter must have a non-module type."),
+        soft = true,
```
LPTK: ???

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -211,21 +317,46 @@ class ImplicitResolver(tl: TraceLogger)
...
+            
+            pss.foreach(_.params.foreach(resolveParam(_)))
+            tps.getOrElse(Nil).flatMap(_.subTerms).foreach(resolve(_))
+            sign.foreach(resolve(
+              _,
+              expectModule = S(isModTyped),
+              errMsg = if isModTyped
+                then S(msg"Function marked as 'module' must have a module signature.")
+                else S(msg"Function must be marked as 'module' in order to have a module signature."),
+              soft = true,
```
LPTK: ???

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -104,20 +105,39 @@ class ImplicitResolver(tl: TraceLogger)
...
-    case Apps(Term.TyApp(base, targs), argss) if argss.nonEmpty =>
-      resolveApp(base, S(targs), argss)
-      t.subTerms.foreach(resolve(_))
-    case Apps(base, argss) if argss.nonEmpty =>
-      resolveApp(base, N, argss)
-      t.subTerms.foreach(resolve(_))
-    case _ => 
-      t.subTerms.foreach(resolve(_))
+  def resolve(t: Term, 
+    expectModule: Opt[Bool] = S(false),
```
LPTK:
```suggestion
    expectModule: Opt[Bool],
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -104,20 +105,39 @@ class ImplicitResolver(tl: TraceLogger)
...
-      t.subTerms.foreach(resolve(_))
-    case Apps(base, argss) if argss.nonEmpty =>
-      resolveApp(base, N, argss)
-      t.subTerms.foreach(resolve(_))
-    case _ => 
-      t.subTerms.foreach(resolve(_))
+  def resolve(t: Term, 
+    expectModule: Opt[Bool] = S(false), 
+    errMsg: => Opt[Message] = N,
+    soft: Bool = false,
```
LPTK: No default argument!

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -278,12 +409,17 @@ object ModuleChecker:
...
     .exists(_.isInstanceOf[TyParam])
   
-  /** Checks if a term evaluates to a module value. */
-  def evalsToModule(t: Term): Bool =
+  /** 
+   * Checks if a term evaluates to a module value. 
+   * 
+   * @param soft If true, the term is not a module if there is any overloaded non-module definition.
+   */
+  def evalsToModule(t: Term, soft: Bool = false): Bool =
```
LPTK: No default argument!

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
LPTK: Please document

LPTK (commented):
Nice changes! But please no default arguments in core functions...

ychenfo (changes requested):

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1007,7 +975,7 @@ extends Importer:
           case R(id) =>
             val sym = members.getOrElse(id.name, die)
             val owner = ctx.outer.inner
-            val isModMember = owner.exists(_.isInstanceOf[ModuleSymbol])
+            val isModMember = owner.forall(_.isInstanceOf[ModuleSymbol]) // forall because a top-level def is also a module method
```
LPTK: Actually, this also includes nested functions, not only top-level ones.

```fs
fun main() =
  fun f(module m: M): module M = m
  f(M).x
```

Indeed, this should also work. But `f` is definitely not a "module method".

So maybe it'd make more sense to distinguish normal functions (within modules or without) from "methods" aka "object methods", which are the ones that are meant to be limited.

LPTK: (Perhaps we should even use a different keyword for object methods, at this point, such as `method`...)

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -104,20 +105,65 @@ class ImplicitResolver(tl: TraceLogger)
...
-    case Apps(Term.TyApp(base, targs), argss) if argss.nonEmpty =>
-      resolveApp(base, S(targs), argss)
-      t.subTerms.foreach(resolve(_))
-    case Apps(base, argss) if argss.nonEmpty =>
-      resolveApp(base, N, argss)
-      t.subTerms.foreach(resolve(_))
-    case _ => 
-      t.subTerms.foreach(resolve(_))
+  enum Expect:
+    case Module(reason: Opt[Message] = N)
```
LPTK: Please, no default args in core logic.

hkmc2/shared/src/test/mlscript-compile/Option.mls
```diff
@@ -6,7 +6,7 @@ open Predef
 // TODO
 // type Option[A] = Option.Some(A) | Option.None
 
-module Option with ...
+object Option with ...
```
LPTK: Why?

hkmc2/shared/src/test/mlscript-compile/Stack.mls
```diff
@@ -6,7 +6,7 @@ open Predef
 // TODO
 // type Stack[A] = Stack.Cons[A] | Stack.Nil
 
-module Stack[T] with ...
+object Stack[T] with ...
```
LPTK: We most definitely want this to be a module...

hkmc2/shared/src/test/mlscript/backlog/Lifter.mls
```diff
@@ -108,6 +108,9 @@ fun foo() =
 :todo
 :expect "NN"
 foo(0, 2)
```
LPTK: Looks like this was meant to call the other `foo` defined above. We should rename it to avoid the name clash.

hkmc2/shared/src/test/mlscript/basics/BadModuleUses.mls
```diff
@@ -4,13 +4,20 @@
 module M with
   fun mtd() = 42
 
+:e
 module Example with
   fun foo(): module null = M
+//│ ╔══[ERROR] Expected a module, found non-moduleful null literal.
+//│ ╙── Function marked as 'module' must have a module signature.
```
LPTK: This function is not "marked as 'module'" 🤔

hkmc2/shared/src/test/mlscript/basics/ModuleMethods.mls
```diff
@@ -1,115 +1,166 @@
-:js
```
LPTK, line 1: Why no JS?

hkmc2/shared/src/test/mlscript/codegen/Hygiene.mls
```diff
@@ -146,10 +146,10 @@ cls.bar
 
 fun foo() =
   if false then
-    module A
```
LPTK, line 150: I think the point of this test was to use a nested module. Can we just change the definition to `fun foo(): module M` for some `M`?

hkmc2/shared/src/test/mlscript/codegen/SelfReferences.mls
```diff
@@ -23,9 +23,8 @@ Foo.self
 :e
 module Foo with
   val self = this
-//│ ╔══[ERROR] Functions returning module values must have explicit return types.
-//│ ║  l.25: 	  val self = this
-//│ ╙──      	      ^^^^
+//│ ╔══[ERROR] Unexpected moduleful reference.
+//│ ╙── Function must be marked as 'module' in order to return a module.
```
LPTK: There is no function here 🤔

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
LPTK, line 1: As discussed in the last meeting, could you please document `ICtx` and explain why `tEnv` is needed?

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
FlandiaYingman: Isn't that this default argument fine? The only thing it does is to provide an optional reason, and it does not affect the behavior of the compiler, only the error message produced when the expectation of modulefulness is not met.

hkmc2/shared/src/test/mlscript-compile/Option.mls
FlandiaYingman: Some function used to have a signature that returns an `Option`, but `Option` used to be module, so it requires users to write `fun f: module Option` which I guess is not what we want... I don't know why no function returns an option now. Maybe we removed them in some following commits.

I just realized that this is wrong and it should be a part of definition overloading, i.e. there should be a type called `Option` that is different from the module `Option`. As it doesn't cause error now, I'll change it back to an module.

hkmc2/shared/src/test/mlscript-compile/Stack.mls
FlandiaYingman: This is similar to https://github.com/hkust-taco/mlscript/pull/295#discussion_r2029700651 as we are now using `Stack` as both the module `Stack` and the type `Stack`. At `GeneratorStack.mls:8` there is a reference to `Stack` which is used as a non-module type, and it crashes because of our module method checks. 

I will change this back to module. I think the way to do it is to just remove the type signature `Stack[Int]` at `GeneratorStack.mls:8` until we implement the overloading.

hkmc2/shared/src/test/mlscript/backlog/Lifter.mls
FlandiaYingman: I see. But don't we have (or plan to do) function overloading?

hkmc2/shared/src/test/mlscript/basics/ModuleMethods.mls
FlandiaYingman: This is because module method related stuffs are done completely before the lowering stage. Things that related to runtime (e.g. type classes) are in another test file.

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
LPTK: Error messages are part of the behavior of the compiler!

First, is there any reason to not always include a reason?

Second, the default here buys you exactly one character at the call sites (`N`). Why risk forgetting to pass a reason at a given call site just to save on ONE character?

hkmc2/shared/src/test/mlscript/basics/BadModuleUses.mls
FlandiaYingman: How do you say that, "function signature marked as 'module'..."?

hkmc2/shared/src/test/mlscript/codegen/SelfReferences.mls
FlandiaYingman: 😬

hkmc2/shared/src/test/mlscript-compile/Stack.mls
LPTK: Instead of removing the type signature, please change `Stack[Int]` to `Stack.T[Int]`, where `T` is defined as
```fs
type T[A] // TODO support type/module overloading
```
in the `Stack` module.

hkmc2/shared/src/test/mlscript/backlog/Lifter.mls
LPTK: We'll probably have this sort of arity-based overloading later, but it won't work when the several variants are defined in different diff-test blocks. Anyway, testing overloading is clearly not the intent of this test case.

hkmc2/shared/src/test/mlscript/basics/ModuleMethods.mls
LPTK: Ok, fair enough.

It's just that my default strategy is "always turn in JS to add more coverage to our JS codegen tests", even if the tested feature is about type/module checking.

hkmc2/shared/src/test/mlscript/basics/BadModuleUses.mls
LPTK: No... it's not the signature (which includes parameter types) that's marked. You want to say
```suggestion
//│ ╙── Function marked as returning a 'module' must have a module return type.
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
FlandiaYingman: Okay maybe making it explicit is better...

> First, is there any reason to not always include a reason?

Because of the white-list strategy, we are not able to provide reasons for all situations (but of course there is a default message saying something like "unexpected moduleful term ...").

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -104,22 +124,75 @@ class ImplicitResolver(tl: TraceLogger)
...
+      case blk: Term.Blk => 
+        resolveBlk(blk)
+      case Apps(Term.TyApp(base, targs), argss) =>
+        resolveApp(base, S(targs), argss)
+      case Apps(base, argss) if argss.nonEmpty =>
+        resolveApp(base, N, argss)
+      case t: (Term.Sel | Term.SynthSel) =>
+        t.subTerms.foreach(resolve(_, expect = Any))
+      case t: Term.IfLike =>
+        t.subTerms.foreach(resolve(_, expect = Any))
```
LPTK: The subterms of an `IfLike` are very heterogeneous – you get all sorts of things, including scrutinees and the bodies of branches. Why would it be correct to use `Any` here for all of them? That doesn't seem to make sense.

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -104,22 +124,75 @@ class ImplicitResolver(tl: TraceLogger)
...
+    
+    t match
+      case blk: Term.Blk => 
+        resolveBlk(blk)
+      case Apps(Term.TyApp(base, targs), argss) =>
+        resolveApp(base, S(targs), argss)
+      case Apps(base, argss) if argss.nonEmpty =>
+        resolveApp(base, N, argss)
+      case t: (Term.Sel | Term.SynthSel) =>
+        t.subTerms.foreach(resolve(_, expect = Any))
```
LPTK: Similarly, it's better to just access the terms explicitly. What if someone later add a subterm to Sel that shouldn't be handled with `Any`?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -780,16 +751,26 @@ extends Importer:
         t
       case (_, Nil) => t
     
+    /**
+     * An extractor that extracts the (tree) definition of a module method.
+     */
+    object FunctionTreeDef:
+      def unapply(t: Term): Opt[Tree.TermDef] = t.symbol match
```
LPTK, line 758: Why are we still trying to access symbol definitions in the elaborator? Won't this break if `t` is a selection, since its symbol might not be resolved yet? (Esp. when module forwarders and the like are involved.)

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -194,11 +348,15 @@ class ImplicitResolver(tl: TraceLogger)
...
+          // Case: Fun/Val definition. 
           // Add the contextual parameters to the context and use the context to resolve the body.
           // Resolve other subterms with the original context.
-          case t @ TermDefinition(_, Fun, _, pss, _, _, _, _, _, _) =>
-            log(s"Resolving function definition $t")
+          case t @ TermDefinition(_, Fun | ImmutVal, _, pss, tps, sign, body, _, TermDefFlags(isMethod, isModTyped), annotations) =>
+            log(s"Resolving ${t.k.desc} definition $t")
+            
+            if isMethod && isModTyped then
+              raise(ErrorReport(msg"${t.k.desc.capitalize} returning modules should not be a method." -> t.toLoc :: Nil))
```
LPTK: It seems this case is never reached in the test. It also seems that the error message would be quite weird.

hkmc2/shared/src/test/mlscript/basics/BadModuleUses.mls
```diff
@@ -4,13 +4,20 @@
 module M with
   fun mtd() = 42
 
+:e
 module Example with
   fun foo(): module null = M
+//│ ╔══[ERROR] Expected a module, found non-moduleful null literal.
+//│ ╙── Function marked as returning a 'module' must have a module return type.
```
LPTK, line 11: You forgot to include locations along the messages. In fact, locations are even more important than specific messages and should _not_ be optional, here. Imagine if this were a 10,000-line file.

LPTK (changes requested):
Thanks for your changes! The logic is still quite wrong, though (see my comments and the new `ForwardTypeClassUses.mls` test file).

hkmc2/shared/src/test/mlscript/basics/BadModuleUses.mls
FlandiaYingman: Of course I did include the locations and they are there by default (and cannot be `None`!) 😕 the problem here is that keyword literals such as `true`, `false`, `null` and `this` do not come with a location when they were parsed

LPTK: That's just a temporary issue in the parser (which we are about to rewrite anyway). The point is that you should have that location field. Also, more often than not, it will produce something a bit useful, which is much better than nothing at all.

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -312,50 +502,18 @@ object ModuleChecker:
...
           // At this point all member symbols should have definition,
           // except for the class(-like) that are currently being elaborated.
-          // TODO: We will fix this by deferring the checks to the resolution stage.
           false
       case sym => 
         lastWords(s"Unsupported symbol kind ${sym}")
     t match
       case Term.Blk(_, res) => evalsToModule(res)
-      case Term.App(lhs, rhs) => lhs.symbol match
+      case Term.IfLike(`if`, split) => split.results.exists(evalsToModule(_))
```
FlandiaYingman, line 532: Here's the place where the modulefulness of `IfLike` is evaluated

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
FlandiaYingman: Ah this checking logic is a kind of funny. Let me explain by this example:

```fs
fun f = if true then M else M
```

Here, we know that the body of the function should not be moduleful, so when the resolver goes to the body of the function, it checks the `IfLike` term and immediately evaluate its modulefulness by recursively evaluating the modulefulness of its bodies. See https://github.com/hkust-taco/mlscript/pull/295/commits/bbae9dc429ae41bf1a3fce961069b65d5b6ff547#r2032655221 for how it checks the modulefulness of a `IfLike`.

So here we are not evaluating the modulefulness of the `IfLike` term, we are just traversing the sub-terms.

The reason why I implement it in this way is that there might be `return` statements in arbitrary depth, so I can't just rely solely on the `expect` parameter. I have to "look behind" for arbitrary depth when I see a term such as `Blk` or `IfLike`.

FlandiaYingman: (Although I haven't implemented the checking logic of `return` statements 😛)

hkmc2/shared/src/test/mlscript/basics/BadModuleUses.mls
FlandiaYingman: The location field is there automatically at around `ImplicitResolver.scala:163`

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: I will fix this in a follow-up PR

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
FlandiaYingman: and directly answering your question:

- scrutinee should be `Any` because it doesn't pass the term to somewhere else that erases the modulefulness; moreover, it is required for `IfLike` expressions such as `M is ...`
- bodies of branches should by `Any` because we already checked them at the beginning of the function (which is what I just explained).

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
```diff
@@ -104,22 +124,85 @@ class ImplicitResolver(tl: TraceLogger)
...
+      raise(ErrorReport(msg"Expected a module, found non-moduleful ${t.describe}." -> t.toLoc 
+        :: expect.message))
+    if expect.nonModule && evalsToModule then
+      raise(ErrorReport(msg"Unexpected moduleful ${t.describe}." -> t.toLoc 
+        :: expect.message))
+    
+    t match
+      // The modulefulness of t is already checked, 
+      // even for the block-like terms such as Blk and IfLike. 
+      // In other words, at this point, even t is a nested Blk and IfLike,
```
LPTK:
```suggestion
      // In other words, at this point, even if t is a nested Blk and IfLike,
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ImplicitResolver.scala
LPTK: Ok, thx for the explanation and documentation.

But why would one pattern match on a module? It doesn't make sense and your lack of check there is unsound:
```fs
module M with { fun whoops(using Int) = 1 }
if M is m then m.whoops
//│ = [function whoops]
```

hkmc2/shared/src/test/mlscript/basics/BadModuleUses.mls
LPTK: Ah sorry, I had missed that part.

LPTK (changes requested):
Looks like there was a miscommunication or I misremembered and the type class changes will be made in a separate PR. Good!

But your module check for pattern matching is currently broken.

hkmc2/shared/src/test/mlscript/codegen/SelfReferences.mls
```diff
@@ -28,9 +28,8 @@ Foo.self
 :e
 module Foo with
   val self = this
-//│ ╔══[ERROR] Functions returning module values must have explicit return types.
-//│ ║  l.30: 	  val self = this
-//│ ╙──      	      ^^^^
```
LPTK, line 33: How come we lost this location?

FlandiaYingman: It used to check the term definitions, so the location is from the method. Now it checks the terms themselves. so the location is from the keyword `this`, and `this` doesn't come with a location 🤕

LPTK (approved):
LGTM

PR #296, opened 2025-03-24
Add additional test cases

AnsonYeung:
These are problems I found in the benchmarking branch. All the test contain the relevant detail for why it is failing.

Currently for the benchmark branch, I either removed the usage of relevant features from the benchmarking files or included either a hack or a fix to these problems.

cc: @CAG2Mark


hkmc2/shared/src/test/mlscript/lifter/EffectHandlers.mls
```diff
@@ -0,0 +1,20 @@
...
+    f()
+    val a = 1
+//│ /!!!\ Uncaught error: hkmc2.InternalError: `this` not in scope: class:Test
+
+// caused by: ValDefn inside continuation class was still using the old `this`
+:fixme
+:effectHandlers
+class Test with
+  f()
+  val aaaaaaa = 1
```
LPTK: Appropriate name.

hkmc2/shared/src/test/mlscript/codegen/Classes.mls
```diff
@@ -86,3 +86,21 @@ data class Foo(static)
...
+:fixme
+data class Foo(v')
+//│ JS (unsanitized):
+//│ let Foo13;
+//│ Foo13 = function Foo(v$_1) {
+//│   return new Foo.class(v$_1);
+//│ };
+//│ Foo13.class = class Foo12 {
+//│   constructor(v$_) {
+//│     this.v' = v$_;
```
LPTK: Here we should just generate
```suggestion
//│     this["v'"] = v$;
```
It should be an easy fix.

hkmc2/shared/src/test/mlscript/lifter/EffectHandlers.mls
```diff
@@ -0,0 +1,20 @@
...
+
+fun f() = 3
+
+:fixme
+:effectHandlers
+module A with
+  data class Test with
+    f()
+    val a = 1
+//│ /!!!\ Uncaught error: hkmc2.InternalError: `this` not in scope: class:Test
```
LPTK: What happens here? Using `:sjs` would help.

hkmc2/shared/src/test/mlscript/lifter/FunInFun.mls
```diff
@@ -604,3 +604,48 @@ fun f() =
...
+//│     return tmp19(v$_$0)
+//│   }
+//│ };
+//│ A1 = function A(v$_$01) {
+//│   return new A.class()(v$_$01);
+//│ };
+//│ A1.class = class A {
+//│   constructor() {
+//│     return (v$_$0) => {
+//│       this.v'$0 = v$_$0;
```
LPTK: Same fix as above.

LPTK (commented):
Thanks, that's useful!

> Currently for the benchmark branch, I either removed the usage of relevant features from the benchmarking files or included either a hack or a fix to these problems.

It would be nice to port the non-hacky fixes to this PR before we merge it.

Also, do you intend for all these problems to be fixed before the PR is merged, or is this just to add progression test cases?

AnsonYeung:
given some of them is relatively easy to fix I guess I will fix them before merge. But some of them like private vars might be problematic.

CAG2Mark:
Thanks, I've been a bit sick the past week and have been quite busy with other things (have two midterms this Thursday and had a presentation on Monday). I'll have a look after Thursday

hkmc2/shared/src/test/mlscript/lifter/EffectHandlers.mls
AnsonYeung: :sjs will not show anything as the error originate in JSBuilder, but as mentioned it is due to the continuation class copying the ValDefn over, the continuation class is lifted out, but the ValDefn still refer to the class

hkmc2/shared/src/test/mlscript/lifter/ModulesObjects.mls
```diff
@@ -286,3 +286,89 @@ fun foo(x) =
...
+//│   }
+//│   static get g() {
+//│     return 1;
+//│   } 
+//│   static get f() {
+//│     return lambda;
+//│   }
+//│   static toString() { return "A"; }
+//│ };
+//│ ═══[RUNTIME ERROR] TypeError: Cannot read properties of undefined (reading 'g')
```
AnsonYeung: I wonder if adding a fix for this would be great: i.e.
```js
class A9 {
  static {
    A10 = A9;
    A9.f()
    ...
```

AnsonYeung: I've implemented this on the latest commit. Let me know if this solution is fine or not.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -613,6 +613,12 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
         case Assign(t: TermSymbol, rhs, rest) if t.owner.isDefined =>
           ctx.getIsymPath(t.owner.get) match
             case Some(value) if !belongsToCtor(t.owner.get) =>
+              if (t.k is syntax.LetBind) && !t.owner.forall(_.isInstanceOf[semantics.TopLevelSymbol]) then
+                // TODO: can we do better?
+                raise(ErrorReport(
+                  msg"Usage of private fields cannot be lifted." -> N :: Nil,
```
CAG2Mark:
```suggestion
                  msg"Usages of private fields cannot yet be lifted." -> N :: Nil,
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -650,7 +662,14 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
         case _ => super.applyPath(p)
       case Value.Ref(t: TermSymbol) if t.owner.isDefined =>
         ctx.getIsymPath(t.owner.get) match
-          case Some(value) if !belongsToCtor(t.owner.get) => Select(value.asPath, t.id)(N)
+          case Some(value) if !belongsToCtor(t.owner.get) =>
+            if (t.k is syntax.LetBind) && !t.owner.forall(_.isInstanceOf[semantics.TopLevelSymbol]) then
+              // TODO: can we do better?
+              raise(ErrorReport(
+                msg"Usage of private fields cannot be lifted." -> N :: Nil,
```
CAG2Mark:
```suggestion
                msg"Usages of private fields cannot yet be lifted." -> N :: Nil,
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -613,6 +613,12 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
         case Assign(t: TermSymbol, rhs, rest) if t.owner.isDefined =>
           ctx.getIsymPath(t.owner.get) match
             case Some(value) if !belongsToCtor(t.owner.get) =>
+              if (t.k is syntax.LetBind) && !t.owner.forall(_.isInstanceOf[semantics.TopLevelSymbol]) then
+                // TODO: can we do better?
```
CAG2Mark: It's possible that we can ignore nested functions/classes which reference private variables, but this will make the lifter even more complicated than it already is. For now I think this is acceptable.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -623,6 +629,12 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
             case None => super.applyBlock(rewritten)
             case Some(value) => Assign(value, applyResult(rhs), applyBlock(rest))
         
+        case Define(d: ValDefn, rest: Block) if d.owner.isDefined =>
+          ctx.getIsymPath(d.owner.get) match
+            case Some(value) if !belongsToCtor(d.owner.get) =>
+              AssignField(value.asPath, Tree.Ident(d.sym.nme), applyResult(d.rhs), applyBlock(rest))(N)
```
CAG2Mark: Could the FieldSymbol parameter here be assigned?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -613,6 +613,12 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
         case Assign(t: TermSymbol, rhs, rest) if t.owner.isDefined =>
           ctx.getIsymPath(t.owner.get) match
             case Some(value) if !belongsToCtor(t.owner.get) =>
+              if (t.k is syntax.LetBind) && !t.owner.forall(_.isInstanceOf[semantics.TopLevelSymbol]) then
+                // TODO: can we do better?
+                raise(ErrorReport(
```
CAG2Mark, line 618: I think we mentioned before that `ErrorReport` is reserved for errors in the user's code. This should be `InternalError`.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -650,7 +662,14 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
         case _ => super.applyPath(p)
       case Value.Ref(t: TermSymbol) if t.owner.isDefined =>
         ctx.getIsymPath(t.owner.get) match
-          case Some(value) if !belongsToCtor(t.owner.get) => Select(value.asPath, t.id)(N)
+          case Some(value) if !belongsToCtor(t.owner.get) =>
+            if (t.k is syntax.LetBind) && !t.owner.forall(_.isInstanceOf[semantics.TopLevelSymbol]) then
+              // TODO: can we do better?
+              raise(ErrorReport(
```
CAG2Mark, line 668: Same comment as before about `ErrorReport`.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -218,19 +224,28 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
...
               val pss = ctorAuxParams.map(_.map(_._2))
               val newCtorCode = doc"$ctorCode # return this;"
               val ctorBraced = doc"${ braced(newCtorCode) }"
               val funBod = pss.foldRight(ctorBraced):
                 case (psDoc, doc) => doc"(${psDoc.mkDocument(", ")}) => $doc"
 
-              doc"${ braced(doc" # return $funBod") }"
+              doc" # return $funBod";
+            
+            val ctorBod = if isModule then
```
CAG2Mark, line 241: Just a thought: could a similar change be applied for objects?

hkmc2/shared/src/test/mlscript/lifter/ClassInFun.mls
```diff
@@ -593,6 +594,7 @@ class A(a) with
     fun g() = x
     g()
 A(2).f()
+//│ ═══[COMPILATION ERROR] Usage of private fields cannot be lifted.
```
CAG2Mark: This test should probably be moved to the backlog.

CAG2Mark (approved):

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK:
```suggestion
                  msg"Uses of private fields cannot yet be lifted." -> N :: Nil,
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
AnsonYeung: Should InternalError not crash the compiler? Right now it seems it stop the codegen entirely if InternalError is raised.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
AnsonYeung: This is just for showing something when a private variable is encountered instead of just failing silently. Not meant to be fixing private variables in this PR.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
AnsonYeung: No, because the constructor of object could be used by somewhere else...
That probably means the fix for module might be incorrect too? if module constructor could be used multiple times

AnsonYeung: and another thing: I just tested it and it also caused problem with inheritance, as you can only access `this` after the super call.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: Whoops. `InternalError` _should_ crash the compiler; it is raised when the compiler finds it has a bug, for things that should never happen (failed assumptions). That's why it is purposely not handled in the diff tests.

This should be an `ErrorReport`. Though one could make a case that a separate kind of errors could be made to represent current compiler limitations (rather than user errors).

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
LPTK: But the
```scala
              case N =>
                braced(doc" # ${getVar(sym)} = ${getVar(isym)};$ctorCode")
```
case still made sense, right?

To me that just shows that we should always insert that (even when there is an owner) and make sure self-references refer to that variable while we're in the ctor before `super`, no?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK:
```suggestion
                // TODO: improve the error message
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
AnsonYeung: There is no problem of crashing the compiler here anyway. The generated code would be wrong for private variables.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
AnsonYeung: Oh, I am probably confusing myself here. The change (hack) was that we assign the original name to the name generated by JSCodegen.

So what I currently know is that this fix cannot be applied to object, as its constructor might be called to create new instances that is not the original instance and also this change insert it at the beginning, meaning that super is not called yet.

However, module "constructor" is just the static block currently, which is probably fine to have the assignment there? Even if it has an owner the code in the static block will only run once instead of the case for the object.

AnsonYeung: I'm not sure whether module will have its constructor exposed and called in a similar way to how object can. If it is the case, then applying this hack will cause problem later.

LPTK: I'm 90% sure that at least normal modules won't have their ctors exposed in this way. These modules don't really support inheritance, since all their members are static and they always use named references like `TheModule` instead of `this`.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: @AnsonYeung it looks like you missed my suggestion when marking this conversation as "resolved".

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -650,7 +662,14 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
         case _ => super.applyPath(p)
       case Value.Ref(t: TermSymbol) if t.owner.isDefined =>
         ctx.getIsymPath(t.owner.get) match
-          case Some(value) if !belongsToCtor(t.owner.get) => Select(value.asPath, t.id)(N)
+          case Some(value) if !belongsToCtor(t.owner.get) =>
+            if (t.k is syntax.LetBind) && !t.owner.forall(_.isInstanceOf[semantics.TopLevelSymbol]) then
+              // TODO: can we do better?
```
LPTK: Same suggestion as above.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -218,19 +224,28 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
...
               val funBod = pss.foldRight(ctorBraced):
                 case (psDoc, doc) => doc"(${psDoc.mkDocument(", ")}) => $doc"
 
-              doc"${ braced(doc" # return $funBod") }"
+              doc" # return $funBod"
+            
+            val ctorBod = if isModule then
+              ownr match
+              case S(owner) =>
+                braced(doc" # ${mkThis(owner)}.${sym.nme} = ${getVar(isym)};$ctorCode")
```
LPTK: This is a bit messy. We get code like:
```fs
//│     this.Baz = class Baz1 {
//│       static {
//│         this.Baz = Baz1;
```
where the outer assignment is completely redundant. We should only do the outer assignment if the inner one is not already performed.

LPTK: And actually, we have exactly the same redundancy in the other case:
```fs
//│ let Bar3;
//│ Bar3 = class Bar2 {
//│   static {
//│     Bar3 = Bar2;
```

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
LPTK: Ok @AnsonYeung, what's preventing us from just adding a similar early-initializer for `object` but just after the `super` call and using `this` instead of the module/object's name?

hkmc2/shared/src/test/mlscript/lifter/ModulesObjects.mls
LPTK: Pls see https://github.com/hkust-taco/mlscript/pull/296/files#r2029701237

hkmc2/shared/src/test/mlscript/lifter/ModulesObjects.mls
```diff
@@ -286,3 +288,87 @@ fun foo(x) =
...
+
+// private fields in modules
+:fixme
+module A with
+  let x = 1
+  fun mtd =
+    fun nested() = x
+    nested()
+A.mtd
+//│ /!!!\ Uncaught error: hkmc2.InternalError: Uses of private fields cannot yet be lifted.
```
LPTK: This is not an internal error, as discussed.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
AnsonYeung: The constructor might be called again, but we do not want to change the binding when a new object is constructed

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
AnsonYeung: It caused problem if the class is at the top level. Seems that you're not allowed to have the same class name if it is just
```
class Baz { ... }
class Baz { ... }
```
whereas
```
Baz1 = class Baz { ... }
```
does not actually bind the name Baz

LPTK: Couldn't you just make it
```js
(class Baz { ... })
(class Baz { ... })
```

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
LPTK: Ah, right. We could check and only assign when whether the binding is `undefined`, but that's a bit hacky. Perhaps better to leave it as is.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -300,13 +300,13 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
                 assert((kind is syntax.Pat) || paramsOpt.isEmpty)
                 // doc"${mkThis(owner)}.${sym.nme} = new ${clsJS}"
                 if isModule
-                then doc"${mkThis(owner)}.${sym.nme} = ${clsJS};"
+                then doc"(undefined, ${clsJS});"
```
LPTK: What purpose does this `undefined` serve?

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
AnsonYeung: right, it works

LPTK (approved):

PR #297, opened 2025-04-10
Move All Resolution to the Resolution Stage

FlandiaYingman:
This PR is based on #295 so that there are some duplicated commits. I will fix them once #295 is merged.

It moves the whole implicit arguments resolution to the resolution stage, supporting forward type classes usage.

The features in this PR are complete but there may have some styling or designing issues that require refactorization.


FlandiaYingman:
@Copilot pro suggestions

hkmc2/shared/src/main/scala/hkmc2/semantics/Split.scala
```diff
@@ -38,8 +38,7 @@ enum Split extends AutoLocated with ProductWithTail:
     case Split.End => Nil
   
   def subTerms: Ls[Term] = this match
-    case Split.Cons(Branch(scrutinee, pattern, continuation), tail) => 
-      scrutinee :: pattern.subTerms ++ continuation.subTerms ++ tail.subTerms
+    case Split.Cons(head, tail) => head.scrutinee :: head.pattern.subTerms ++ tail.subTerms
```
LPTK: whoops

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -169,174 +163,245 @@ class ImplicitResolver(tl: TraceLogger)
...
-            split(tail)
-          case Split.Else(default) =>
-            resolve(default, expect = Any)
-          case Split.End =>
-        split(s)
-      case Term.Sel(p, _) =>
-        resolve(p, expect = Any)
-      case Term.SynthSel(p, _) =>
-        resolve(p, expect = Any)
+      case t: Term.IfLike =>
```
LPTK: whoops

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -198,7 +198,16 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
     def warnStmt = if inStmtPos then
       raise:
         WarningReport(msg"Pure expression in statement position" -> t.toLoc :: Nil, S(t))
+    
+    val withImplicitApp = t match
```
LPTK: whoops

hkmc2/shared/src/test/mlscript/codegen/TraceLogIndent.mls
```diff
@@ -3,16 +3,12 @@
...
 
 :traceJS
-:e
 fun g(z) =
   if
     z < -1 then g(z + 1)
     z == -1 then 42
     z == 0 then plus(0)
     z > 0 then g(z - 1)
-//│ ╔══[ERROR] Expected 2 arguments, got 1
```
LPTK, line 13: whoops

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -504,10 +490,10 @@ class Resolver(tl: TraceLogger)
           // Case: Fun/Val definition. 
           // Add the contextual parameters to the context and use the context to resolve the body.
           // Resolve other subterms with the original context.
-          case t @ TermDefinition(_, Fun | ImmutVal, _, pss, tps, sign, body, _, TermDefFlags(isMethod, isModTyped), annotations) =>
+          case t @ TermDefinition(_, Fun | ImmutVal, _, pss, tps, sign, body, _, TermDefFlags(isMethod), modulefulness, annotations) =>
             log(s"Resolving ${t.k.desc} definition $t")
-            
-            if isMethod && isModTyped then
+            val isModuleful = modulefulness.isModuleful
```
LPTK: It seems this binding is only used once. Inline it?

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -318,17 +318,27 @@ final case class DefineVar(sym: LocalSymbol, rhs: Term) extends Statement
...
     val flags = Buffer.empty[String]
     if isMethod then flags += "method"
-    if isModTyped then flags += "modTyped"
     flags.mkString(" ")
   override def toString: String = "‹" + showDbg + "›"
 
-object TermDefFlags { val empty: TermDefFlags = TermDefFlags(false, false) }
+object TermDefFlags { val empty: TermDefFlags = TermDefFlags(false) }
+
+final case class Modulefulness(msym: Opt[MemberSymbol[ModuleDef]])(val modified: Bool):
```
LPTK: The name `modified` is ambiguous. Pick a better one and document its meaning?

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -328,18 +328,38 @@ final case class TermDefFlags(isMethod: Bool):
...
 
-final case class Modulefulness(msym: Opt[MemberSymbol[ModuleDef]])(val modified: Bool):
-  
-  
+/**
+ * A case class representing the modulefulness of a declaration.
+ *
+ * @param msym if None, the declaration is non-moduful; if Some of
+ * symbol, the declaration is moduleful and the symbol is the symbol of
+ * the module
```
```suggestion
 * the module's type
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -328,18 +328,38 @@ final case class TermDefFlags(isMethod: Bool):
...
   
+  /**
+   * Return Some of the module definition if it is moduleful; None
+   * otherwise.
+   */
   def mdef: Opt[ModuleDef] = msym match
     case S(msym) => msym.defn match
       case S(defn: ModuleDef) => S(defn)
-      case _ => lastWords(s"no module definition for moduleful symbol $msym")
+      case _ => lastWords(s"no kmodule definition for moduleful symbol $msym")
```
LPTK: `kmodule`? 🤔

hkmc2/shared/src/test/mlscript-compile/Rendering.mjs
```diff
@@ -105,9 +105,9 @@ let Rendering1;
       tmp = Rendering.fold(lambda);
       tmp1 = Rendering.interleave(", ");
       tmp2 = Rendering.map(Rendering.render);
-      tmp3 = tmp2(...arg);
-      tmp4 = tmp1(...tmp3);
-      return tmp("[", ...tmp4, "]")
+      tmp3 = runtime.safeCall(tmp2(...arg));
```
LPTK: Oh noes why did these come back?

FlandiaYingman: I did some funny thing to the symbols. Let me fix it later

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -67,11 +105,13 @@ enum Term extends Statement:
...
     derivedClsSym: ClassSymbol, defs: Ls[HandlerTermDefinition], body: Term)
   
-  lazy val symbol: Opt[Symbol] = this match
+  def symbol: Opt[Symbol] = this match
     case Ref(sym) => S(sym)
     case sel: SynthSel => sel.sym
     case sel: Sel => sel.sym
     case sel: SelProj => sel.sym
+    case App(lhs, _) => lhs.symbol
+    case TyApp(lhs, _) => lhs.symbol
```
LPTK: These two lines are wrong, as discussed in the meeting.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -28,17 +28,55 @@ enum Annot extends AutoLocated:
...
   case Missing // Placeholder terms that were not elaborated due to the "lightweight" elaboration mode `Mode.Light`
   case Lit(lit: Literal)
   case Builtin(id: Tree.Ident, nme: Str)
-  case Ref(sym: Symbol)(val tree: Tree.Ident, val refNum: Int)
-  case App(lhs: Term, rhs: Term)(val tree: Tree.App, val resSym: FlowSymbol)
-  case TyApp(lhs: Term, targs: Ls[Term])
-  case Sel(prefix: Term, nme: Tree.Ident)(val sym: Opt[FieldSymbol])
-  case SynthSel(prefix: Term, nme: Tree.Ident)(val sym: Opt[FieldSymbol])
+  case Ref(sym: Symbol)(val tree: Tree.Ident, val refNum: Int) extends Term with Resolvable
+  case App(lhs: Term, rhs: Term)(val tree: Tree.App, var resultSym: Opt[FieldSymbol], val resSym: FlowSymbol) extends Term with Resolvable
```
LPTK: Why not call this `sym` for consistency?

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -28,17 +28,55 @@ enum Annot extends AutoLocated:
     case Trm(trm) => trm :: Nil
     case _: Modifier | Untyped => Nil
 
+sealed trait Resolvable:
+  t: Term =>
+  
+  var iargsLs: Ls[Term.Tup] = Nil
```
LPTK: This should be
```suggestion
  var iargsLs: Opt[Ls[Term.Tup]] = N
```

LPTK (changes requested):

hkmc2/shared/src/test/mlscript/codegen/FunnyOpen.mls
FlandiaYingman, line 1: Do you really want to change these tests?

LPTK: Of course. There's a risk the test is not testing the correct thing if we pick symbols that also exist in `Predef`.

hkmc2/shared/src/test/mlscript/codegen/FunnyOpen.mls
```diff
@@ -3,70 +3,70 @@
 
 module Mod with
   fun print(x) = console.log(x)
```
LPTK: So actually this should also be
```suggestion
  fun printt(x) = console.log(x)
```

hkmc2/shared/src/main/scala/hkmc2/bbml/bbML.scala
```diff
@@ -442,15 +442,15 @@ class BBTyper(using elState: Elaborator.State, tl: TL):
             effBuff += eff
             ctx += sym -> rhsTy
             goStats(stats)
-          case TermDefinition(_, Fun, sym, ps :: Nil, _, sig, S(body), _, _, _) :: stats =>
+          case TermDefinition(_, Fun, sym, ps :: Nil, _, sig, S(body), _, _, _, _) :: stats =>
```
LPTK: Instead of adding more meaningless underscores, better leave this place in a better state than you found it by using this recent Scala 3 feature:
```suggestion
          case td @ TermDefinition(k = Fun, params = ps :: Nil, body = S(body)) :: stats =>
```
(and use `td.sym` instead of `sym`)

hkmc2/shared/src/test/mlscript/bbml/bbBasics.mls
```diff
@@ -1,5 +1,5 @@
 :bbml
-//│ Type: ⊤
+//│ Type: ⊥
```
FlandiaYingman: I just realized that for some reason the bbml test cases were not resolved by the resolver. I fixed it by modifying `hkmc2DiffTests/src/test/scala/hkmc2/MLsDiffMaker.scala`. 

IDK why this funny T symbol (and in other bbml files) became upside down 🤦 I guess this is related to the automatically inserted import statement. Just take a note and I will take a look at it later.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -314,8 +323,24 @@ class Resolver(tl: TraceLogger)
     defn match
     
     // Case: instance definition. Add the instance to the context.
-    case defn @ TermDefinition(k = Ins, sym = sym, sign = sign) =>
+    case defn @ TermDefinition(_, Ins, sym, pss, tps, sign, body, _, TermDefFlags(isMethod), modulefulness, annotations) =>
       log(s"Resolving instance definition ${defn.showDbg}")
+      
+      pss.foreach(_.params.foreach(resolveParam(_)))
```
LPTK: Why does everyone always forget about the `restParam`? 😫

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -72,8 +76,8 @@ enum Term extends Statement:
   case Missing // Placeholder terms that were not elaborated due to the "lightweight" elaboration mode `Mode.Light`
   case Lit(lit: Literal)
   case Builtin(id: Tree.Ident, nme: Str)
-  case Ref(sym: Symbol)(val tree: Tree.Ident, val refNum: Int) extends Term with Resolvable
-  case App(lhs: Term, rhs: Term)(val tree: Tree.App, var resultSym: Opt[FieldSymbol], val resSym: FlowSymbol) extends Term with Resolvable
+  case Ref(var sym: Symbol)(val tree: Tree.Ident, val refNum: Int) extends Term with Resolvable
```
LPTK: Do we ever need to _change_ the symbol of a `Ref`? 🤔

hkmc2/shared/src/test/mlscript/backlog/NamedArgs.mls
```diff
@@ -35,15 +35,19 @@ id(a: 0)
 //│ Elaborated tree:
 //│ Blk:
 //│   stats = Nil
-//│   res = App{Some(member:id), None}:
```
LPTK: 👀

hkmc2/shared/src/test/mlscript/bbml/bbBasics.mls
LPTK: Lol ⊤ is top and ⊥ is bottom. The fact top became bottom is somewhat concerning (an understatement).

LPTK: In BbML, ⊤ is used for "unit type", basically. Which is wrong, but somehow @NeilKleistGao didn't feel like adding a real unit type?

hkmc2/shared/src/test/mlscript/bbml/bbGetters.mls
```diff
@@ -1,8 +1,20 @@
 :js
 :noSanityCheck // * For some reason, these cause problems
-
```
LPTK: Bad idea.

(The way BbML is set up is very hacky and this needs to be done this way for now.)

FlandiaYingman: I didn't tend to do that, just didn't review the diff for bbml because of the T and upside down T problem; plan to investigate it today.

hkmc2/shared/src/test/mlscript/backlog/NamedArgs.mls
FlandiaYingman: What is the problem? The `member:id` symbol used to refer to the LHS of an app, which was not the symbol we wanted. The result symbol of this app is none because the result is not a module or something else that can be resolved.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
FlandiaYingman: Yes, because `Ref` might be an implicit application which could be elaborated to some `App`, and the actual symbol should (probably) also change accordingly:

```fs
fun f(using Int): module M
f // Ref(f), should be Ref(M)
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: 😣

hkmc2/shared/src/test/mlscript/bbml/bbBasics.mls
FlandiaYingman: It's funny that the problem was I forgot that I changed `bbPrelude` 🤡

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
FlandiaYingman: Hmm... no then it would be lowered to something else... Probably we have to add a new field for Ref?

hkmc2/shared/src/main/scala/hkmc2/MLsCompiler.scala
FlandiaYingman, line 1: I adapted the changes from Luyu's PR so that the resolver works as well for MLsCompiler.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -67,11 +109,14 @@ enum Term extends Statement:
   case Handle(lhs: LocalSymbol, rhs: Term, args: List[Term],
     derivedClsSym: ClassSymbol, defs: Ls[HandlerTermDefinition], body: Term)
   
-  lazy val symbol: Opt[Symbol] = this match
+  def symbol: Opt[Symbol] = this match
+    case ref: Ref if ref.resSym.nonEmpty => ref.resSym
```
FlandiaYingman: I suppose this `symbol` method is returning the symbol of the thing that the term evaluates to. So I changed Ref's symbol to the new field `resSym`, which will be resolved by the resolver. If the resolver hasn't yet resolved the symbol or the resolver cannot resolve the new symbol, it falls back to the `sym` field of `Ref`

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -67,11 +108,15 @@ enum Term extends Statement:
...
     case Ref(sym) => S(sym)
-    case sel: SynthSel => sel.sym
     case sel: Sel => sel.sym
+    case sel: SynthSel => sel.sym
     case sel: SelProj => sel.sym
+    case app: App => app.sym
+    case tyApp: TyApp if tyApp.sym.nonEmpty => tyApp.sym
+    case TyApp(prefix, _) => prefix.symbol
```
FlandiaYingman: TyApp is similar to Ref. It always has the symbol from its LHS, except in the case that there is an implicit application on the TyApp. E.g., 

```fs
fun foo[A](using ...): module M
foo[Int]
```

The symbol of the Ref of foo is `foo` but the symbol of the TyApp of Ref of foo is `M`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -153,7 +199,7 @@ sealed trait Statement extends AutoLocated with ProductWithExtraInfo:
     case Rcd(stats) => stats.flatMap(_.subTerms)
     case Quoted(term) => term :: Nil
     case Unquoted(term) => term :: Nil
-    case New(_, args, rft) => args ::: rft.toList.flatMap(_._2.blk.subTerms)
+    case New(cls, args, rft) => cls :: args ::: rft.toList.flatMap(_._2.blk.subTerms)
```
FlandiaYingman, line 224: Is it correct to include `cls` to the subStatements?

hkmc2/shared/src/test/mlscript/ctx/CascadingTypeClasses.mls
```diff
@@ -6,87 +6,133 @@ class Bar[B]
 class Baz
 
 
-fun foo[A](using Foo[A], A)
+fun foo[A](using Foo[A], A) = 42
```
FlandiaYingman, line 9: It is funny that if we don't provide the body, the lowering part will throw a "not in scope" error below, which has confused me for a long time (I thought it was a resolver problem)

FlandiaYingman (commented):
LGTM

hkmc2/shared/src/test/mlscript-compile/Runtime.mls
```diff
@@ -29,7 +29,7 @@ class EffectHandle(_reified) with
   fun resumeWith(value) =
     Runtime.try(() => resume(reified.contTrace)(value))
   fun raise() =
-    topLevelEffect(reified)
+    topLevelEffect(reified, null)
```
FlandiaYingman: I added a `null` argument here because `topLevelEffect` takes 2 arguments.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -0,0 +1,766 @@
...
+              head.pattern.subTerms.foreach(traverse(_, expect = NonModule(N)))
+              split(head.continuation)
+              split(tail)
+            case Split.Let(sym, term, tail) =>
+              traverse(term, expect = NonModule(N))
+              split(tail)
+            case Split.Else(default) =>
+              traverse(default, expect = Any)
+            case Split.End =>
+          split(t.normalized)
```
FlandiaYingman, line 292: I changed `t.desugared` to `t.normalized` because it appears that it's the normalized term being lowered later

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
LPTK: Yeah, don't change the symbol that a reference refers to, that's crazy!

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
LPTK: Nooo don't make the method return the wrong result if it's called too early, just make it cleanly crash. It's like you're begging for bugs to bite us when we inevitably misuse something down the line :P

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
LPTK: Yes, sure.

hkmc2/shared/src/test/mlscript/ctx/CascadingTypeClasses.mls
LPTK: Yes, it's hilarious.

hkmc2/shared/src/test/mlscript-compile/Runtime.mls
LPTK: Smart. `null` is the best value.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
LPTK: Indeed, that's right!

hkmc2/shared/src/main/scala/hkmc2/bbml/bbML.scala
```diff
@@ -442,16 +442,16 @@ class BBTyper(using elState: Elaborator.State, tl: TL):
...
-          case TermDefinition(_, Fun, sym, Nil, _, sig, S(body), _, _, _) :: stats =>
-            typeFunDef(sym, body, sig, ctx)  // * may be a case expressions
+          case (td @ TermDefinition(k = Fun, params = Nil, sign = sig, body = S(body))) :: stats =>
+            typeFunDef(td.sym, body, sig, ctx)  // * may be a case expressions
             goStats(stats)
-          case TermDefinition(_, Fun, sym1, _, _, S(sig), None, _, _, _) :: (td @ TermDefinition(_, Fun, sym2, _, _, _, S(body), _, _, _)) :: stats
-            if sym1 === sym2 => goStats(td :: stats) // * avoid type check signatures twice
-          case TermDefinition(_, Fun, sym, _, _, S(sig), None, _, _, _) :: stats =>
-            ctx += sym -> typeType(sig)
+          case (td1 @ TermDefinition(k = Fun, sign = S(sig), body = None)) :: (td2 @ TermDefinition(k = Fun, body = S(body))) :: stats
```
LPTK, line 451: Nice improvements!

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -198,7 +198,12 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
     def warnStmt = if inStmtPos then
       raise:
         WarningReport(msg"Pure expression in statement position" -> t.toLoc :: Nil, S(t))
+
```
LPTK, line 202: If you add a space here, also add one above `def warnStmt`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -28,17 +28,58 @@ enum Annot extends AutoLocated:
...
+  def termDefn: Opt[TermDefinition] = t.defn match
+    case S(td: TermDefinition) => S(td)
+    case _ => N
+  
+  def typeDefn: Opt[ClassLikeDef] = t.defn match
+    case S(td: ClassLikeDef) => S(td)
+    case _ => N
+  
+  def withIArgs(iargsLs: Ls[Term.Tup]): Term = 
+    this.iargsLs = S(iargsLs)
```
LPTK, line 73:
```suggestion
    require(this.iargsLs.isEmpty)
    this.iargsLs = S(iargsLs)
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -502,18 +498,16 @@ extends Importer:
...
     case tree @ App(lhs, rhs) =>
       val sym = FlowSymbol("‹app-res›")
       val lt = term(lhs, inAppPrefix = true)
       val rt = term(rhs)
-      maybeApp:
-        Term.App(lt, rt)(tree, sym)
+      Term.App(lt, rt)(tree, N, sym)
     case SynthSel(pre, nme) =>
       val preTrm = term(pre)
       val sym = resolveField(nme, preTrm.symbol, nme)
```
LPTK, line 510: Can we just delete this code and `resolveField`? Is there a reason to resolve any field here as opposed to just doing it in the resolver?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -251,7 +251,7 @@ extends Importer:
         case N =>
           raise(ErrorReport(msg"${cls.k.desc.capitalize} '${cls.symbol.nme
             }' does not contain member '${nme.name}'" -> srcTree.toLoc :: Nil))
-          N
+          S(ErrorSymbol(nme.name, srcTree))
```
LPTK, line 254: Why is it useful to return an `ErrorSymbol` here?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1018,17 +946,20 @@ extends Importer:
...
                 val mtdSym = BlockMemberSymbol("ret", Nil, true)
                 val td = TermDefinition(
-                  N, Fun, mtdSym, PlainParamList(Param(FldFlags.empty, valueSym, N) :: Nil) :: Nil,
-                  N, N, S(valueSym.ref(Ident("value"))), FlowSymbol(s"‹result of non-local return›"), TermDefFlags.empty, Nil)
+                  N, Fun, mtdSym, PlainParamList(Param(FldFlags.empty, valueSym, N, Modulefulness.none) :: Nil) :: Nil,
+                  N, N, S(valueSym.ref(Ident("value"))), FlowSymbol(s"‹result of non-local return›"), TermDefFlags.empty, Modulefulness.none, Nil)
                 val htd = HandlerTermDefinition(resumeSym, td)
                 Term.Handle(nonLocalRetHandler, state.nonLocalRetHandlerTrm, Nil, clsSym, htd :: Nil, inner)
               val r = FlowSymbol(s"‹result of ${sym}›")
-              // indicates if the function marks its result as "module"
```
LPTK, line 1027: Maybe keep the comment? It was a bit helpful. And you could add an example, like `fun f(): module M`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1150,8 +1082,8 @@ extends Importer:
               case S(tree) =>
                 val (patternParams, extractionParams) = ps match // Filter out pattern parameters.
                   case S(ParamList(_, params, _)) => params.partition:
-                    case param @ Param(FldFlags(false, false, false, false, true, false), _, _) => true
-                    case param @ Param(FldFlags(_, _, _, _, false, _), _, _) => false
+                    case param @ Param(FldFlags(false, false, false, true, false), _, _, _) => true
+                    case param @ Param(FldFlags(_, _, _, false, _), _, _, _) => false
```
LPTK: How about we use the named pattern field syntax here too?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1251,21 +1183,22 @@ extends Importer:
     else VarSymbol(id)
   
   def param(t: Tree, inUsing: Bool, inDataClass: Bool): Ctxl[Opt[Opt[Bool] -> Param]] =
-    def go(t: Tree, inUsing: Bool, flags: FldFlags): Ctxl[Opt[Opt[Bool] -> Param]] = t match
+    // mm: `module` modified
```
```suggestion
    // mm: `module`-modified
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -0,0 +1,766 @@
...
+   *    environment.
+   *
+   * 2. If it sees an instance definition, it adds the instance to the
+   *    instance environment.
+   *
+   * 3. If it sees an application with contextual arguments to be
+   *    resolved and type arguments specified, it resolves the
+   *    contextual arguments from the instance environment, and adds the
+   *    type arguments to the type environment.
+   *
```
LPTK, line 43:
```suggestion
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -0,0 +1,766 @@
...
+      case _ => false
+    
+    def nonModule = this match
+      case Expect.NonModule(_) => true
+      case _ => false
+    
+  end Expect
+  import Expect.*
+  
+  type Resolvable = Term & IResolvable
```
LPTK: Why not just
```suggestion
  type ResblTerm = Term & Resolvable
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -0,0 +1,766 @@
...
+          case sym: Sym => sym
+          case App(sym, _) => sym
+    
+    final case class Instance(sym: Symbol)
+    
+    val empty = ICtx(N, Map.empty, Map.empty)
+    
+  def ictx(using ICtx) = summon[ICtx]
+
+type IResolvable = Resolvable
```
LPTK: 😢

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -0,0 +1,766 @@
...
+  * Here are some notes that I used for reasoning about the design.
+  *
+  * ### Static Members vs. Class Members
+  *
+  * An important feature in MLscript is that the program should be
+  * compilable without type checking, and the user experience of doing
+  * so should be good and not error prone. (See:
+  * https://github.com/hkust-taco/mlscript-design-docs/blob/main/wiki/module-methods.md)
+  * A consequence of this design is that, the definition of a method may
+  * not always present at compilation-time, because the type information
```
```suggestion
  * not always be present at compilation-time, because the type information
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -0,0 +1,766 @@
...
+  *
+  * Here are some notes that I used for reasoning about the design.
+  *
+  * ### Static Members vs. Class Members
+  *
+  * An important feature in MLscript is that the program should be
+  * compilable without type checking, and the user experience of doing
+  * so should be good and not error prone. (See:
+  * https://github.com/hkust-taco/mlscript-design-docs/blob/main/wiki/module-methods.md)
+  * A consequence of this design is that, the definition of a method may
```
```suggestion
  * A consequence of this design is that the definition of a method may
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -0,0 +1,766 @@
...
+  * not always present at compilation-time, because the type information
+  * of the object is absent.
+  *
+  * In contrast, for a (global or local) function, or a module method,
```
LPTK: There are actually no global functions. Also, let's call "module methods" just _functions_.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -0,0 +1,766 @@
...
+  * the definition is always present at compilation-time. This is
```
```suggestion
  * the definition is always known at the call sites at compilation time. This is
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -0,0 +1,766 @@
...
+  * important because most call-site features require the definition to
+  * be present. For example, by-name, lazy and implicit parameters all
```
```suggestion
  * be known. For example, by-name, lazy and implicit parameters all
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -0,0 +1,766 @@
...
+  * require the definition so that the compiler can perform the correct
+  * elaboration and resolution.
+  *
+  * ### Module Methods & Modulefulness Check
+  *
+  * In addition to (global or local) functions, module methods also
```
LPTK: Same remarks as above.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -0,0 +1,766 @@
...
+  * other names. In order to maintain the correctness of this system,
+  * the resolver performs the extra checks to ensure that the definition
+  * of a module is never lost during re-binding. 
+  *
+  * 1. Only declarations that have a signature and marked as `module`
+  *    may bind to a module value.
+  * 2. Only declarations of static members may bind to a module value.
+  *
+  * If a term evaluates to a module value (e.g., an application to a
+  * function returning a module, a reference to a variable of module),
```
```suggestion
  * function returning a module, a reference to a module variable),
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: (If this causes too much breakage, it's better to do it in a follow-up PR.)

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -0,0 +1,766 @@
...
+    case Any
+    
+    def message: Ls[Message -> Opt[Loc]] = this match
+      case Expect.Module(msg) => msg.toList.map(_ -> N)
+      case Expect.NonModule(msg) => msg.toList.map(_ -> N)
+      case Expect.Any => Nil
+    
+    def module = this match
+      case Expect.Module(_) => true
+      case _ => false
```
```suggestion
    def module = isInstanceOf[Module]
```

hkmc2/shared/src/test/mlscript/basics/CyclicModuleForwarders.mls
```diff
@@ -1,14 +1,12 @@
 :js
 
-
-fun f = M.List.map
-module M with
-  fun List: module M2.List = M2.List
```
LPTK, line 6: I'd like to also keep this nonsensical example as a regression test.

hkmc2/shared/src/test/mlscript/basics/CyclicModuleForwarders.mls
```diff
@@ -1,14 +1,12 @@
...
-  fun List: module M.List = M.List
+  fun List: module M1 = M1.List
 
-:re
+:todo
+:e
 f
 //│ ═══[RUNTIME ERROR] RangeError: Maximum call stack size exceeded
-
-
```
LPTK, line 14: Y u hate space

hkmc2/shared/src/test/mlscript/basics/Records.mls
```diff
@@ -113,53 +113,12 @@ id of ({ x: 1, y: 2 })
...
-//│ >       rect x: 71, y: 82, width: 80, height: 90
-//│ >       ellipse x: 81, y: 92, rx: 40, ry: 60
-
-let display = t("group", x: 0, y: 0) of
-  t("circle", x: 10, y: 10, r: 42)
-//│ display = [function]
-
-display(x: 88, y: 99, i: 0)
-//│ > group x: 88, y: 99 with
-//│ >   circle x: 98, y: 109, r: 42
```
LPTK: Whats' the reason for removing all these tests? 🤔

hkmc2/shared/src/test/mlscript/ctx/ForwardTypeClassUses.mls
```diff
@@ -11,47 +11,51 @@ module M with
 bar
 //│ = "ok"
 
-
```
LPTK, line 14: 😭

LPTK (commented):
Nice work! I did not yet have time to review the whole diff, but here are preliminary comments.

hkmc2/shared/src/test/mlscript/basics/CyclicModuleForwarders.mls
FlandiaYingman: Maybe I should write a script with MLscript to automatically check the white spaces removed before committing🤦

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: Currently, there are two symbol resolution stages: one is in the elaborator and one is in the resolver. If the field does not exist in the definition at the elaboration stage, it doesn't exist in the resolution stage as well; this raises duplicate errors during compilation, which I guess is not preferred. 

Of course, if we implement your suggestion at https://github.com/hkust-taco/mlscript/pull/297#discussion_r2055681451, then we wouldn't need `ErrorSymbol`, which I think it is the correct way to do it. Let me work on it and see if that's feasible.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: My thought was that `Resolvable` and `Resolvable & Term` were equivalent because `Resolvable` is sealed and has a self-type of `Term`. It's just that Scala couldn't deduce it. So I created this funny alias, overriding the original name. 😣

hkmc2/shared/src/test/mlscript/basics/CyclicModuleForwarders.mls
LPTK: Or maybe just don't remove whitespace for no reason 😬

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
LPTK: Sure. But maybe just rename the trait to `ResolvableImpl` and next to it add `type Resolvable = Term & ResolvableImpl`.

Local type aliases that _redefine_ already-used names are not a very good idea.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
LPTK: I really don't like that the official symbol _changes_ over the phases of the compiler. Fix it like for references?

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
FlandiaYingman: I just realized that there is probably a hole in the implementation... In the codebase, there are some places (mostly UCS normalization) that use the same term in different places as if they were immutable. So we couldn't simply apply this change. We may need a proper duplication mechanism to resolve this problem. 

A more problematic thing is that the implicit arguments of a term might be resolved differently if the same term was used in different places... I changed the condition to `this.iargsLs.isEmpty || this.iargsLs.get == iargsLs` and fortunately it doesn't report any problem.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: As discussed in the group meeting, let's do it in a follow-up PR

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: I think this also needs to be resolved in a follow-up PR

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
FlandiaYingman: I tried, and it seemed that we couldn't just make it crash (or even make it return `None`) at the time because so many places are using this `symbol` method before elaboration. I think the proper way is to clearly separate the preliminary symbols and the symbols after resolution because even in the resolution stage we need the symbols to determine what to do next...

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -637,7 +632,12 @@ class Resolver(tl: TraceLogger)
             case t: Term.Ref => sym.map(sym => t.resSym = S(sym))
           log(s"Resolved symbol for ${t}: ${lhsDefn.sym}")
         case _ =>
-    case _ => 
+    case _ =>
+    
+    t match
+    case t: Term.Ref if t.resSym.isEmpty =>
```
LPTK, line 641: How about you document what's going on here?

From what I can surmise, it's something like
```suggestion
    // If a reference was not resolved to take implicit arguments or return some module type,
    // then its result symbol is the same as the symbol it refers to.
    case t: Term.Ref if t.resSym.isEmpty =>
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
LPTK: Good catch. But I think this will be fine once the UCS normalization is moved to/after the resolution phase: we will probably not even need to duplicate the terms, which will be already resolved.

Cc @chengluyu 

> mostly UCS normalization

Are there any other places where this happens?

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
LPTK: I agree, we should clearly separate the different kinds of symbols available at different times. But really, after we move all resolution to after elaboration, the only symbols we should really be looking at before resolution are the original symbols stored in Ref's main parameter list, no?

So, in an case, we shouldn't use a `symbol` method like the one you're defining here whose semantics _changes_ over the course of compilation.

The original `symbol` method was just a shorthand to check if a given term had a symbol (mostly, if it was a Ref) without having to perform pattern matching. Let's restore it to that use case.

LPTK (changes requested):

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
FlandiaYingman: Yes, a few places in the elaborator. I remember it is somewhere in elaborating `SynthSel`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -115,15 +115,29 @@ enum Term extends Statement:
...
+  def resolvedSymbol: Opt[Symbol] = this match
+    case ref: Ref => ref.resSym
+    case sel: Sel => sel.sym
+    case sel: SynthSel => sel.sym
+    case sel: SelProj => sel.sym
     case app: App => app.sym
-    case tyApp: TyApp if tyApp.sym.nonEmpty => tyApp.sym
-    case TyApp(prefix, _) => prefix.symbol
+    case tyApp: TyApp => tyApp.sym
     case _ => N
```
FlandiaYingman: I restored the `symbol` for the original use case, but kept the case that for `TyApp` inherits the symbol from it LHS because AFAIK applying types to a term never changes its symbol.

I added a new method `resolvedSymbol` that returns the resolved symbol of a term. The places that require symbols after or during resolution should call this method instead. 

For `Sel` they are available from both `symbol` and `resolvedSymbol` because of the reason we discussed in the metting.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -635,8 +635,16 @@ class Resolver(tl: TraceLogger)
     case _ =>
     
     t match
+    // If a reference was not resolved to take implicit arguments or
+    // return some module type, then its result symbol is the same as
+    // the symbol it refers to.
     case t: Term.Ref if t.resSym.isEmpty =>
       t.resSym = S(t.sym)
+    // If a type application was not resolved to take implicit
+    // arguments, then its result symbol is the same as the symbol its
```
```suggestion
    // arguments, then its result symbol is the same as the symbol of its
```

LPTK (approved):
Thanks for another great contribution!

PR #302, opened 2025-04-25
Reimplement operator apps, operator splits, and "placeholder shorthand" lambdas

LPTK:
@chengluyu This is the WIP of my changes. Can you please look into updating the UCS and pattern compilation implementations?


hkmc2/shared/src/test/mlscript/backlog/UCS.mls
```diff
@@ -88,10 +88,12 @@ else
...
 //│ ╟── Because the previous split is full.
-//│ ║  l.85: 	if audits.has(1) === true do
-//│ ║        	   ^^^^^^^^^^^^^^^^^^^^^^^^^
 //│ ║  l.86: 	  print("ok")
-//│ ╙──      	^^^^^^^^^^^^^
+//│ ║        	  ^^^^^^^^^^^
+//│ ║  l.87: 	else
+//│ ║        	^^^^
+//│ ║  l.88: 	  1
+//│ ╙──      	^^^
```
LPTK: @chengluyu BTW what causes this?

LPTK: @chengluyu is this resolved?

LPTK:
@chengluyu I think I'm done with my changes. Could you please review them and also review your own changes?

hkmc2/shared/src/test/mlscript/backlog/UCS.mls
LPTK: Ah, I think this was resolved in Florent's PR.

hkmc2/shared/src/main/scala/hkmc2/syntax/Lexer.scala
```diff
@@ -595,10 +595,12 @@ object Lexer:
     "virtual"
   )
   
+  private val SEP = "┊"
+  
   def printToken(tl: TokLoc): Str = tl match
     case (SPACE, _) => " "
     case (COMMA, _) => ","
-    case (SEMI, _) => ";"
+    // case (SEMI, _) => ";"
```
chengluyu:
```suggestion
```
Why not remove the case if `SEMI` has already been removed?

hkmc2/shared/src/main/scala/hkmc2/syntax/Keyword.scala
```diff
@@ -138,9 +138,14 @@ object Keyword:
...
   
-  type Infix = `and`.type | `or`.type | `then`.type | `else`.type | `is`.type | `:`.type | `->`.type |
-    `=>`.type | `extends`.type | `restricts`.type | `as`.type | `do`.type | `where`.type
-
+  type Infix =
+    `is`.type | `:`.type | `->`.type | `=>`.type | `extends`.type | `restricts`.type | `as`.type | `do`.type | `where`.type |
+    `and`.type | `or`.type | `then`.type | `else`.type
+  
+  type InfixSplittable =
+    `is`.type | `:`.type | `->`.type | `=>`.type | `extends`.type | `restricts`.type | `as`.type | `do`.type | `where`.type |
```
chengluyu, line 146: Does “splittable” mean people can use it in UCS splits? Why are `extends`, `restricts` splittable?

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -2,6 +2,7 @@
 
 module Rendering with ...
 
+
```
chengluyu, line 5: Only this one blank line was added to this file?

hkmc2/shared/src/test/mlscript/ucs/syntax/Do.mls
chengluyu, line 1: This file only has changes to blank lines. Was this intended?

hkmc2/shared/src/test/mlscript/codegen/SetStmt.mls
```diff
@@ -0,0 +1,21 @@
...
+g
+//│ = [0, 1, 2]
+
+set g.0 += 1
+
+set g.0 = g.0 + 1
+
+g
+//│ = [2, 1, 2]
+
```
chengluyu, line 20: It would be interesting if we can add:

```suggestion

set g.0 |>= Math.sign

set g.2 |>= Math.pow(_, 5)

g
//│ = [1, 1, 32]
```

hkmc2/shared/src/test/mlscript/basics/PureTermStatements.mls
```diff
@@ -12,15 +12,15 @@ x => x, 0
...
 //│ = 0
 
 :w
 case x then x
 case x then x
 //│ ╔══[WARNING] Pure expression in statement position
 //│ ║  l.19: 	case x then x
-//│ ╙──      	            ^
+//│ ╙──      	     ^^^^^^^^
 //│ = [function]
```
chengluyu, line 24: Shouldn't `case` be included in location to make it more accurate?

hkmc2/shared/src/test/mlscript/codegen/Juxtapositions.mls
```diff
@@ -121,37 +121,55 @@ inc(2) * 2
 // :pt
 // :elt
 // :sjs
+:todo // ?
```
chengluyu: Better explain `// ?`?

The same remark applies to the following tests with `// ?`.

hkmc2/shared/src/main/scala/hkmc2/codegen/Printer.scala
```diff
@@ -19,7 +19,9 @@ object Printer:
       // ts.trmTree
     case ts: semantics.InnerSymbol => ts.nme
     case ts: semantics.BuiltinSymbol => ts.nme
-    case _ => summon[Scope].lookup_!(l)
+    case _ => summon[Scope].lookup(l) match
+      case S(str) => str
+      case N => s"‹not in scope: ${l}›"
```
chengluyu, line 24: Does this mean that if the symbol is not found in the scope, we just generate `s"‹not in scope: ${l}›"` as the identifier directly in the code?

hkmc2/shared/src/test/mlscript/backlog/NonReturningStatements.mls
```diff
@@ -0,0 +1,38 @@
+:js
+
+
+// TODO `fun` does not return anything; the returned value should be the result of `bar + 1`
+fun foo =
+  bar + 1
+  fun bar = 1
+
+foo
```
chengluyu, line 9: Do you plan to hoist local functions into the beginning?

hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala
```diff
@@ -493,14 +535,16 @@ abstract class Parser(
   
   
   // TODO: rm `allowIndentedBlock`? Seems it can always be `true`
```
chengluyu: Is it removed or just renamed to `allowNewlines`?

hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala
chengluyu, line 1: I see 14 `allowNewlines = true`, 15 `allowNewlines = false`, and 67 `allowNewlines = allowNewlines` in this file. It might be possible to refactor this by introducing an implicit parameter for the "newline mode", which indicates how the parsing functions should handle newlines.

hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala
```diff
@@ -79,15 +83,20 @@ object Parser:
   }
   val prefixOps: Set[Str] = Set("!", "+", "-", "~", "@")
   
+  object Indent_Curly:
+    def unapply(t: Curly.type | Indent.type): true = true
+  
+  object NEWLINE_COMMA:
+    def unapply(t: NEWLINE.type | COMMA.type): true = true
```
chengluyu: Why not just `type NEWLINE_COMMA = NEWLINE.type | COMMA.type`? IIRC it can function as patterns in Scala 3.

hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala
```diff
@@ -949,12 +1057,12 @@ abstract class Parser(
...
         val res = App(acc, Tup(as).withLoc(S(loc)))
-        exprCont(res, prec, allowNewlines)
-      case (KEYWORD(Keyword.`of`), _) :: _ =>
+        exprCont(res, prec, allowNewlines = allowNewlines)
+      case (KEYWORD(kw @ Keyword.`of`), _) :: _ if prec <= AppPrec =>
         consume
         val as = blockMaybeIndented
         val res = App(acc, Tup(as))
-        exprCont(res, prec, allowNewlines)
+        exprCont(res, prec, allowNewlines = allowNewlines)
```
chengluyu: `kw` is not used at all?

hkmc2/shared/src/test/mlscript/syntax/AbusiveSuspensions.mls
chengluyu, line 1: What does "abusive suspension" mean?

chengluyu (changes requested):
Nice PR! I’ve left a few minor suggestions and questions.

I’m glad to see that the syntax for operator and newline has been determined. Also, the parser can now handle operator splits in a better way.

This PR has also eliminated many parse errors, making the parser more robust. I think like I can now update the parser used by the VSCode extension.

(Let's merge this PR so that I can resolve the conflicts in my PR. 😁)

hkmc2/shared/src/main/scala/hkmc2/codegen/Printer.scala
LPTK: This is just the pretty-printer!

hkmc2/shared/src/main/scala/hkmc2/syntax/Keyword.scala
LPTK: Not just UCS splits, but any operator split/block. I just copy from `Infix` to `InfixSplittable` everything that's not _not_ splittable, for regularity. The semantics is left unspecificed. We may find a use for it later.

hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala
LPTK: Nah, of course I would have used an implicit if I'd thought it was appropriate. I made it epxlicit because the choice really does need to be made carefully at every call site. With an implicit, we would simply forget to use the correct value when the correct value isn't the one already in scope.

hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala
LPTK: Yes, good suggestion!

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
LPTK: Yup. I always separate the `module ‹...› with ...` part with at least two empty lines, so it stands out.

hkmc2/shared/src/test/mlscript/backlog/NonReturningStatements.mls
LPTK: That's already happening. But my comment here is about the semantics; how it happens behind the scenes is irrelevant.

hkmc2/shared/src/test/mlscript/basics/PureTermStatements.mls
LPTK: Of course. But we _still_ haven't implemented keyword locations in this iteration of the parser. Actually, this PR makes progress towards it by making the End rule store a thunk rather than a single shared result, which had been the blocker.

hkmc2/shared/src/test/mlscript/codegen/Juxtapositions.mls
LPTK: It's just used to question whether we really want to do this.

hkmc2/shared/src/test/mlscript/ucs/syntax/Do.mls
LPTK: Naturally.

hkmc2/shared/src/test/mlscript/syntax/AbusiveSuspensions.mls
LPTK: Means this is a bit of an abuse of this syntax 😉

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
chengluyu: I see. Thanks.

chengluyu (approved):
LGTM!

PR #303, opened 2025-04-27
Benchmarking subproject

AnsonYeung:
WIP
- [ ] Fix tests
- [ ] Port all the test
- [ ] nofib `mate.mls` bug for using comparision for the wrong type.


hkmc2/shared/src/test/mlscript/HkScratch.mls
```diff
@@ -9,4 +9,3 @@
 :todo
 
 
-
```
LPTK, line 12: 👀

LPTK:
@AnsonYeung it's better to leave a remark when closing, for the record.

Basically, the important changes in this PR were already integrated into other PRs, except for a part that was used to auto-copy nofib test files in a way that doesn't seem very useful anymore.

PR #306, opened 2025-05-13
Record patterns, conjunction, disjunction added to ucs

Chronarak:
## Features
- [X] record patterns `{ident:p, ...}`
- [X] implements pattern conjunction (&) in the ucs
- [X] implements pattern disjunction (|) in the ucs

## Implementation
Both operators are implemented as syntactic sugar for sequentialization or new splits :
```
if e is
  p | q then sequel
  alternatives
```
is transformed into
```
if e is
  p then sequel
  q then sequel
  alternatives
```
in the same way
```
if e is
  p & q then sequel
  alternatives
```
is transformed into
```
if e is
  p then if e is
    q then sequel
    alternatives
  alternatives
```

## Possible improvements
- [ ] propagate more thouroughly information that a scrutinee is an object to avoid the runtime checks repetition


hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
```diff
@@ -22,15 +22,12 @@ enum Pattern extends AutoLocated:
   case Synonym(symbol: PatternSymbol, patternArguments: Ls[(split: DeBrujinSplit, tree: Tree)])
   
   case Tuple(size: Int, inf: Bool)
-
```
LPTK: Please don't mess with the whitespace 😬

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
```diff
@@ -50,3 +47,7 @@ enum Pattern extends AutoLocated:
     case Record(Nil) => "{}"
     case Record(entries) =>
       entries.iterator.map(_.name + ": " + _).mkString("{ ", ", ", " }")
+
+  def isRecord = this match
+    case _:Record => true
```
LPTK:
```suggestion
    case _: Record => true
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
```diff
@@ -50,3 +47,7 @@ enum Pattern extends AutoLocated:
     case Record(Nil) => "{}"
     case Record(entries) =>
       entries.iterator.map(_.name + ": " + _).mkString("{ ", ", ", " }")
+
```
LPTK:
```suggestion
  
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -188,6 +219,16 @@ class Normalization(elaborator: Elaborator)(using raise: Raise, ctx: Ctx):
...
+              thatPattern match
+              case thatPattern :Pattern.Record => // so the type system is happy
+                // we can use information if pattern is itself a record, or if it is a constructor with arguments
+                val simplifiedRecord = thatPattern assuming pattern
+                if simplifiedRecord.entries.isEmpty then
+                  tail
+                else
+                  Split.Cons(Branch(thatScrutinee, simplifiedRecord, continuation), tail)
+              case _ => tail // impossible
             else if pattern <:< thatPattern then
```
LPTK: Why not
```suggestion
            else thatPattern match
            case thatPattern: Pattern.Record => // so the type system is happy
              // we can use information if pattern is itself a record, or if it is a constructor with arguments
              val simplifiedRecord = thatPattern assuming pattern
              if simplifiedRecord.entries.isEmpty then
                tail
              else
                Split.Cons(Branch(thatScrutinee, simplifiedRecord, continuation), tail)
            case _ =>  if pattern <:< thatPattern then
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -249,7 +290,11 @@ object Normalization:
     case (Lit(Tree.StrLit(_)), ClassLike(blt.`Str`, _, _, _)) => true
     case (Lit(Tree.DecLit(_)), ClassLike(blt.`Num`, _, _, _)) => true
     case (Lit(Tree.BoolLit(_)), ClassLike(blt.`Bool`, _, _, _)) => true
-    case (_, _) => false
+    case (_:Synonym, _) => false
```
LPTK:
```suggestion
    case (_: Synonym, _) => false
```

hkmc2/shared/src/test/mlscript/ucs/patterns/ConjunctionPattern.mls
```diff
@@ -0,0 +1,97 @@
...
+suffixes(range(0, -1))
+//│ = Cons(Nil, Nil)
+
+suffixes(range(0, 0))
+//│ = Cons(Cons(0, Nil), Cons(Nil, Nil))
+
+if 5 is a & b & c then a+b+c
+//│ = 15
+
+:e
```
LPTK: The error is spurious, right? If so, this should be
```suggestion
:fixme
```

hkmc2/shared/src/test/mlscript/ucs/syntax/PlainConditionals.mls
```diff
@@ -28,6 +27,7 @@ fun foo(x) = x is Pair(a, b) and a > b
 
 fun foo(x) = if x is Pair(a, b) then a > b else false
 
+:todo // function prototype defintion is not there yet
```
LPTK: What does this mean?

hkmc2/shared/src/test/mlscript/ucs/syntax/PlainConditionals.mls
```diff
@@ -50,23 +50,14 @@ fun foo(x) = x is
...
 fun foo(x) = x is Pair(a, b) | Int
-//│ ╔══[ERROR] Name not found: |
-//│ ║  l.54: 	fun foo(x) = x is Pair(a, b) | Int
-//│ ╙──      	                             ^
-//│ ╔══[ERROR] Cannot use this identifier as an extractor
-//│ ║  l.54: 	fun foo(x) = x is Pair(a, b) | Int
-//│ ╙──      	                             ^
 
 :todo
 fun foo(x) = x is (Pair(a, b) and a > b) | Int
```
LPTK: TBH we should probably use another keyword here, such as `where`, as in
```suggestion
fun foo(x) = x is (Pair(a, b) where a > b) | Int
```

LPTK (commented):
Wow, this already looks quite promising!

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
Chronarak: because I did not think about it ^^.
I'll change it.

hkmc2/shared/src/test/mlscript/ucs/patterns/ConjunctionPattern.mls
Chronarak: Yes, it was a temporary thing, but it should had been ``:fixme``.
In any case, it has been fixed.

hkmc2/shared/src/test/mlscript/ucs/syntax/PlainConditionals.mls
Chronarak: The following function is defined using 3 functions ``f``, ``g`` and ``h`` who only have a prototype and no definition. So trying to do anything other than parsing the term results in an error.

hkmc2/shared/src/test/mlscript/ucs/syntax/PlainConditionals.mls
Chronarak: Agreed, I'll try to implement it.

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
Chronarak: I'll try to be more cautious ><

hkmc2/shared/src/test/mlscript/ucs/syntax/PlainConditionals.mls
Chronarak: It should now be implemented.

hkmc2/shared/src/test/mlscript/ucs/patterns/ConjunctionPattern.mls
```diff
@@ -52,46 +50,71 @@ suffixes(range(0, 0))
...
+//│           Block of Ls of 
+//│             Pun:
+//│               eql = false
+//│               id = Ident of "a"
+//│           Ident of "C"
+//│     kw = keyword 'then'
+//│     rhs = Ident of "a"
+//│ = 1
+
+:todo // parser does not recognize the pun??
```
LPTK: Lol it's due to @chengluyu's code randomly swallowing keywords:
https://github.com/hkust-taco/mlscript/blob/12479ae279843801357bace68e4e87d55a9a947e/hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala#L671-L675

Repro:

```fs
:dp
:pt
C & {:a}
//│ |C| |&| |{|:|a|}|
//│ @ blockOf(start of statement)    [at syntax/Parser.scala:212]
//│ │ ? blockOfImpl		inspects |C| |&| |{|:|a|}|    [at syntax/Parser.scala:277]
//│ │ @ simpleExpr(1)    [at syntax/Parser.scala:345]
//│ │ │ ? simpleExprImpl		inspects |C| |&| |{|:|a|}|    [at syntax/Parser.scala:503]
//│ │ │ ! simpleExprImpl		consumes |C|    [at syntax/Parser.scala:527]
//│ │ │ @ exprCont(1,`Ident(C)`,true)    [at syntax/Parser.scala:536]
//│ │ │ │ ? exprContImpl		inspects | |&| |{|:|a|}|    [at syntax/Parser.scala:726]
//│ │ │ │ ! exprContImpl		consumes | |    [at syntax/Parser.scala:852]
//│ │ │ │ @ exprCont(1,`Ident(C)`,true)    [at syntax/Parser.scala:855]
//│ │ │ │ │ ? exprContImpl		inspects |&| |{|:|a|}|    [at syntax/Parser.scala:726]
//│ │ │ │ │ ! exprContImpl		consumes |&|    [at syntax/Parser.scala:810]
//│ │ │ │ │ ? exprContImpl		inspects | |{|:|a|}|    [at syntax/Parser.scala:812]
//│ │ │ │ │ ! exprContImpl		consumes | |    [at syntax/Parser.scala:812]
//│ │ │ │ │ found an infix operator: &
//│ │ │ │ │ ? exprContImpl		inspects |{|:|a|}|    [at syntax/Parser.scala:817]
//│ │ │ │ │ ! exprContImpl		consumes |{|:|a|}|    [at syntax/Parser.scala:819]
//│ │ │ │ │ > @ concludeWith(())    [at syntax/Parser.scala:821]
//│ │ │ │ │ > │ @ split()    [at syntax/Parser.scala:821]
//│ │ │ │ │ > │ │ @ item(index = 1)    [at syntax/Parser.scala:666]
//│ │ │ │ │ > │ │ │ ? item		inspects |:|a|    [at syntax/Parser.scala:667]
//│ │ │ │ │ > │ │ │ ! item		consumes |:|    [at syntax/Parser.scala:671]
//│ │ │ │ │ > │ │ │ @ parseRule(0,start of statement ::= Kw(keyword 'let') | Kw(keyword 'set') | Kw(keyword 'handle') | Kw(keyword 'new') | Kw(keyword 'in') | Kw(keyword 'if') | Kw(keyword 'while') | Kw(keyword 'else') | Kw(keyword 'case') | Kw(keyword 'region') | Kw(keyword 'outer') | Kw(keyword 'constructor') | Kw(keyword 'fun') | Kw(keyword 'val') | Kw(keyword 'use') | Kw(keyword 'type') | Kw(keyword 'pattern') | Kw(keyword 'class') | Kw(keyword 'trait') | Kw(keyword 'module') | Kw(keyword 'object') | Kw(keyword 'open') | Kw(keyword 'abstract') | Kw(keyword 'mut') | Kw(keyword 'do') | Kw(keyword 'virtual') | Kw(keyword 'override') | Kw(keyword 'declare') | Kw(keyword 'data') | Kw(keyword 'public') | Kw(keyword 'private') | Kw(keyword 'out') | Kw(keyword 'return') | Kw(keyword 'throw') | Kw(keyword 'import') | Kw(keyword 'using') | Kw(keyword 'true') | Kw(keyword 'false') | Kw(keyword 'undefined') | Kw(keyword 'null') | Kw(keyword 'this') | Kw(keyword '_') | Expr(expression ::= End(())) | Blk(block ::= End(())))    [at syntax/Parser.scala:675]
//│ │ │ │ │ > │ │ │ │ ? parseRuleImpl		inspects |a|    [at syntax/Parser.scala:367]
//│ │ │ │ │ > │ │ │ │ @ simpleExpr(0)    [at syntax/Parser.scala:345]
//│ │ │ │ │ > │ │ │ │ │ ? simpleExprImpl		inspects |a|    [at syntax/Parser.scala:503]
//│ │ │ │ │ > │ │ │ │ │ ! simpleExprImpl		consumes |a|    [at syntax/Parser.scala:527]
//│ │ │ │ │ > │ │ │ │ │ @ exprCont(0,`Ident(a)`,true)    [at syntax/Parser.scala:536]
//│ │ │ │ │ > │ │ │ │ │ │ ? exprContImpl		inspects ||    [at syntax/Parser.scala:726]
//│ │ │ │ │ > │ │ │ │ │ │ @ exprJux(0,`Ident(a)`,false)    [at syntax/Parser.scala:1028]
//│ │ │ │ │ > │ │ │ │ │ │ │ ? exprJuxImpl		inspects ||    [at syntax/Parser.scala:1035]
//│ │ │ │ │ > │ │ │ │ │ │ │ stops at the end of input
//│ │ │ │ │ > │ │ │ │ │ │ = Ident(a)
//│ │ │ │ │ > │ │ │ │ │ = Ident(a)
//│ │ │ │ │ > │ │ │ │ = Ident(a)
//│ │ │ │ │ > │ │ │ │ @ parseRule(0,expression ::= End(()))    [at syntax/Parser.scala:347]
//│ │ │ │ │ > │ │ │ │ │ ? parseRuleImpl		inspects ||    [at syntax/Parser.scala:367]
//│ │ │ │ │ > │ │ │ │ = Some(())
//│ │ │ │ │ > │ │ │ = Some(Ident(a))
//│ │ │ │ │ > │ │ │ @ exprCont(0,`Ident(a)`,true)    [at syntax/Parser.scala:675]
//│ │ │ │ │ > │ │ │ │ ? exprContImpl		inspects ||    [at syntax/Parser.scala:726]
//│ │ │ │ │ > │ │ │ │ @ exprJux(0,`Ident(a)`,false)    [at syntax/Parser.scala:1028]
//│ │ │ │ │ > │ │ │ │ │ ? exprJuxImpl		inspects ||    [at syntax/Parser.scala:1035]
//│ │ │ │ │ > │ │ │ │ │ stops at the end of input
//│ │ │ │ │ > │ │ │ │ = Ident(a)
//│ │ │ │ │ > │ │ │ = Ident(a)
//│ │ │ │ │ > │ │ = Ident(a)
//│ │ │ │ │ > │ │ ? splitItem		inspects ||    [at syntax/Parser.scala:681]
//│ │ │ │ │ > │ │ ! end of split
//│ │ │ │ │ > │ = List(Ident(a))
//│ │ │ │ │ > │ ? concludeWithImpl		inspects ||    [at syntax/Parser.scala:229]
//│ │ │ │ │ > │ Concluded with List(Ident(a))
//│ │ │ │ │ > = List(Ident(a))
//│ │ │ │ = App(Ident(&),Tup(List(Ident(C), Block(List(Ident(a))))))
//│ │ │ = App(Ident(&),Tup(List(Ident(C), Block(List(Ident(a))))))
//│ │ = App(Ident(&),Tup(List(Ident(C), Block(List(Ident(a))))))
//│ │ @ parseRule(1,expression ::= End(()))    [at syntax/Parser.scala:347]
//│ │ │ ? parseRuleImpl		inspects ||    [at syntax/Parser.scala:367]
//│ │ = Some(())
//│ │ ? blockOfImpl		inspects ||    [at syntax/Parser.scala:331]
//│ │ ? blockContOf		inspects ||    [at syntax/Parser.scala:272]
//│ = List(App(Ident(&),Tup(List(Ident(C), Block(List(Ident(a)))))))
//│ ? parseAll		inspects ||    [at syntax/Parser.scala:216]
//│ Parsed tree:
//│ App:
//│   lhs = Ident of "&"
//│   rhs = Tup of Ls of 
//│     Ident of "C"
//│     Block of Ls of 
//│       Ident of "a"
```

hkmc2/shared/src/test/mlscript/ucs/syntax/PlainConditionals.mls
LPTK: Elaboration and resolution should work, though, not just parsing.

In the old compiler we used to have a check in the diff-tests to avoid code-gen for snippets using missing definitions, but it's no longer done 🤷‍♂️

hkmc2/shared/src/test/mlscript/ucs/syntax/PlainConditionals.mls
LPTK: Nice, that was fast!

hkmc2/shared/src/test/mlscript/ucs/syntax/PlainConditionals.mls
Chronarak: It does indeed work unless you try to execute/compile it in any way, that is what I meant, sorry for the imprecision.
It just felt wrong to leave a test that really wasn't working just beccause we were not testing it enough.
But I guess just testing the syntax is a reasonable thing.

I can remove the todo, it would cause no errors at the moment.
I'll also add a test file to ensure the behaviour of only having prototyped functions.

LPTK: > It just felt wrong to leave a test that really wasn't working just beccause we were not testing it enough.

I can't actually understand what you mean here. Note that I don't have any context; I just saw this comment and couldn't understand what it's trying to say. So it's the comment that needs improvement.

Also note that `:todo` is to temporarily disable test failures. If we just have something to do later and there are no test failures, we just use a `// TODO:` comment. Here, `:todo` should be avoided as it could mask a regression.

BTW, there is no such thing as "function prototypes" in reasonable languages; functions simply have function _types_ 😉

Chronarak: > So it's the comment that needs improvement.

> Here, :todo should be avoided as it could mask a regression.

Got it 👍 

> BTW, there is no such thing as "function prototypes" in reasonable languages; functions simply have function types

What I mean is really the type signature. Wikipedia tells me that the word prototype is used in C/C++ for function types, it's probably were i picked it up.

hkmc2/shared/src/main/scala/hkmc2/utils/Scope.scala
```diff
@@ -96,7 +96,7 @@ class Scope
             case t: syntax.Tree.TypeDef => t.head.toLoc
           .flatten.orElse(l.toLoc)
         case other => other.toLoc
-      raise(InternalError(msg"Not in scope: ${l.toString} (${l.getClass.toString})" -> loc :: Nil,
+      raise(ErrorReport(msg"No definition found in scope for: ${l.toString} (${l.getClass.toString})" -> loc :: Nil,
```
LPTK: Tip: error reports should be user-readable and shouldn't include implementation details like symbol identifiers and runtime class names. So I'd use:
```suggestion
      raise(ErrorReport(msg"No definition found in scope for '${l.nme}'" -> loc :: Nil,
        extraInfo = Some(l -> l.getClass),
        source = Diagnostic.Source.Compilation))
```
since `extraInfo` is the parameter for additional debugging info, which will gives us:
```fs
fun f: Int

f
//│ FAILURE LOCATION: lookup_! (Scope.scala:99)
//│ FAILURE INFO: Tuple2:
//│   _1 = member:f
//│   _2 = class hkmc2.semantics.BlockMemberSymbol
//│ ╔══[COMPILATION ERROR] No definition found in scope for 'f'
//│ ║  l.12: 	fun f: Int
//│ ╙──      	    ^^^^^^
//│ FAILURE LOCATION: mkQuery (JSBackendDiffMaker.scala:149)
//│ ═══[RUNTIME ERROR] ReferenceError: f is not defined
```

hkmc2/shared/src/test/mlscript/ucs/syntax/PlainConditionals.mls
LPTK: Yeah, that's what I mean, it's used in unreasonable languages that don't have actual function types. Don't worry I guessed your meaning 😉

hkmc2/shared/src/test/mlscript/ucs/patterns/ConjunctionPattern.mls
LPTK: (The fix is to just move the `consume` inside the `case S(subRule)` branch.)

Chronarak: I doesn't look like it is enough.
It does not look like it wants to try the right rule to get a pun, I think?
I'm looking into it, but understanding the parser is taking a bit of time.

Chronarak: I think I managed to fix it.
The issue was that the parser saw a keyword, namely `:`, when trying prefixRulesAllowIndentedBlock, and it assumed it meant a special expression, like a let binding or an if statement.
Whereas what we really wanted was to not consume anything and parse as an expression.
I added a small exception, which solved it.

It had the interesting side effect of allowing the following
`fun foo(x) = :x`
It is an odd syntax, but coherent with the fact that record are just blocks, and it is clear that something special is going on. I personally find it acceptable.

LPTK: Thanks for looking into it. The current parser architecture is an intermediate result of some wild experimentation from Luyu and me. The handling of all cases around indentation etc. is really poor (too complicated). Good on you for figring it out!

hkmc2/shared/src/test/mlscript/basics/Puns.mls
```diff
@@ -8,11 +8,8 @@ let a = 1
 //│ = {a: 1}
 
 
-:fixme // should we support this?
+
```
LPTK: Extra new line here 😬

LPTK:
BTW, regarding your "updated tests" commits – ideally they wouldn't be needed: the diff-test maker should keep track of the running offset in the line numbers from its previous updates, and just provide the correct number to the parser. It's just that no one had time to fix this long-standing annoyance yet.

hkmc2/shared/src/test/mlscript/ucs/patterns/RecordPattern.mls
```diff
@@ -143,3 +143,18 @@ foo(1 :: 2 :: 4 :: Nil)
...
 //│ = "at least three of them"
+
+fun should_reject_null(x) = if x is { a: 0 } then 42 else -42
+
+should_reject_null(null)
+//│ = -42
+
+:todo
+:sjs
+fun is_record(x) = if x is {} then 0 else 1
```
LPTK: Oh, this is an empty split.

In MLscript, records don't need braces. Braces can be used, but their use kind of conflicts with splits. Maybe we could just allow `{}` to denote empty records, which would be an exception to the more-or-less general rule that braces and indentation should be interchangeable – in this case it's probably fine, as there's no such thing as ab empty indentation block.

Chronarak: The current plan was to consider it as an empty record, as it visually looks like it, and this is more natural in the code.

That said, @chengluyu raised an interesting issue : what does matching against an empty record mean?

We currently have 2 options :
- equivalent to a wildcard because ``forall field in the record, having field`` is always true for an empty record.
- equivalent to checking if the scrutinee is of type object, because trying to match against any record can lead to searching for fields, and therefore thr scrutinee shall be of object type.
  In some sense, matching the empty record is the prerequisite to matching any record, so it should enforce being an object.

LPTK: Yes the latter one is the only correct interpretation. BTW Scala also treats `{}` in types as meaning Object (not Any) for similar reasons.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -393,19 +393,22 @@ enum Case:
   case Lit(lit: Literal)
   case Cls(cls: ClassLikeSymbol, path: Path)
   case Tup(len: Int, inf: Bool)
-  case Field(name: Tree.Ident) // checks field existence
+  /** checks field existence
+    * @param safe true will omit the Object type check
+  */
+  case Field(name: Tree.Ident, safe: Bool = false)
```
LPTK: Default arguments are an anti-pattern in core logic, and should only be used in API functions to make end usage easier.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -491,14 +491,17 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
-                entries.foldRight
-                  (go(tail, topLevel = false)) {
+                val objectSym = summon[Ctx].builtins.Object
+                mkMatch( // checking that we have an object
+                  Case.Cls(objectSym, Value.Ref(BuiltinSymbol(objectSym.nme, false, false, true, false))),
+                  entries.foldRight(go(tail, topLevel = false)) {
                     case ((fieldName, fieldSymbol), blk) =>
                       mkMatch(
-                        Case.Field(fieldName),
+                        Case.Field(fieldName, safe = true), // we have checked object type already
```
LPTK: Is this true even for virtual classes? Ask @chengluyu if you don't know what they are.

Chronarak: It looks like my comment is misleading.
What I mean is that we know we can try for fields, without checking if we are of type object first.
e.g. we have already chekced that we are not an Int or something like that, but an object (typeof x === 'object')
I'll make a better comment.

LPTK: What I meant is that not all values matched by `Case.Cls` are objects.

```fs
:lot
:sjs
if 1 is
  Object then "o"
  Int then "i"
//│ JS (unsanitized):
//│ let scrut;
//│ scrut = 1;
//│ if (scrut instanceof this.Object) {
//│   "o"
//│ } else if (globalThis.Number.isInteger(scrut)) {
//│   "i"
//│ } else {
//│   throw new this.Error("match error");
//│ }
//│ Lowered:
//│ Program:
//│   imports = Nil
//│   main = Assign:
//│     lhs = $scrut
//│     rhs = Lit of IntLit of 1
//│     rest = Assign: \
//│     lhs = $selRes
//│     rhs = Select{member:Object}:
//│       qual = Ref of globalThis:globalThis
//│       name = Ident of "Object"
//│     rest = Match: \
//│     scrut = Ref of $selRes
//│     arms = Ls of 
//│       Tuple2:
//│         _1 = Lit of UnitLit of false
//│         _2 = Throw of Instantiate:
//│           cls = Select:
//│             qual = Ref of globalThis:globalThis
//│             name = Ident of "Error"
//│           args = Ls of 
//│             Lit of StrLit of "Access to required field 'Object' yielded 'undefined'"
//│     dflt = N
//│     rest = Match: \
//│     scrut = Ref of $scrut
//│     arms = Ls of 
//│       Tuple2:
//│         _1 = Cls:
//│           cls = class:Object
//│           path = Ref of $selRes
//│         _2 = Assign:
//│           lhs = $block$res
//│           rhs = Lit of StrLit of "o"
//│           rest = Return: \
//│           res = Lit of UnitLit of false
//│           implct = true
//│       Tuple2:
//│         _1 = Cls:
//│           cls = class:Int
//│           path = Select:
//│             qual = Ref of $runtime
//│             name = Ident of "unreachable"
//│         _2 = Assign:
//│           lhs = $block$res
//│           rhs = Lit of StrLit of "i"
//│           rest = Return: \
//│           res = Lit of UnitLit of false
//│           implct = true
//│     dflt = S of Throw of Instantiate:
//│       cls = Select:
//│         qual = Ref of globalThis:globalThis
//│         name = Ident of "Error"
//│       args = Ls of 
//│         Lit of StrLit of "match error"
//│     rest = End of ""
//│ = "i"
```

This is because `class:Int` is "virtual", not a real class.

chengluyu: The code here turns multiple consecutive field cases into a `Object` case followed by several field cases. This is done to reduce repeated object instance checking.

https://github.com/hkust-taco/mlscript/blob/e35e4c185100c806f732807caba798d9ac46bb98/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala#L493-L504

So it doesn't seem to be related to other virtual classes.

Chronarak: Right, I saw that some checks were hardcoded (e.g. Int).
If I'm not mistaken, a virtual class member should not match an empty record (because it is not an object, that's the point), this is currently the observed behaviour.
These tests are preceded by a class test against Object (``foo instanceof Object``), so a virtual class will fail the match before reaching the fields tests.
This optimization is currently done only when we generate the matchs for a whole record, which is not virtual.
The plan is to extend the optimization at some point, atthat moment we should be wary of virtual classes.
Hopefuly this answers your worries.

LPTK: Ah, I read the context too fast and thought this was a check to optimize things like `C & (a: x)` where `C` is a class being matched – that's another optimization we could do (except, ofc, when `C` is virtual).

> a virtual class member

Minor nitpick: you mean "instance". A member is something like a field or method.

chengluyu: I brought this up during my discussion with @Chronarak. But now I realize that we don't need to check whether the scrutinee is an object. We only need to make sure it's neither `null` nor `undefined`. 

- Using `obj instanceof Object` is not comprehensive, because it fails to recognize objects created via `Object.create(null)`, and it also doesn't work for objects from other realms, browser frames, or V8 contexts.
- Using `typeof obj === "object" && obj !== null` doesn't always work either, because it doesn't accept functions objects.
- No matter we use `"key" in obj` or `Reflect.has(obj, "key")` to check field inclusion, the only restriction is that `obj` must not be `null` or `undefined`. So `obj !== undefined && obj !== null` should be what `JSBuilder` should produce.

LPTK:
```
> 'a' in 42
Uncaught TypeError: Cannot use 'in' operator to search for 'a' in 42
> Reflect.has(42, "a")
Uncaught TypeError: Reflect.has called on non-object
    at Reflect.has (<anonymous>)
```

chengluyu: Oh, right. I somehow omitted the primitives. 

We can use `obj instanceof Object || (typeof obj === "object" && obj !== null)`. This can be further simplified to `obj instanceof Object || typeof obj === "object"`.

LPTK: That looks a bit expensive. We should make sure to avoid doing it when we can (eg when the thing was already matched against a concrete class pattern)

LPTK: (But this optimization can be implemented later, in a future PR.)

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -54,9 +54,7 @@ class Normalization(elaborator: Elaborator)(using raise: Raise, ctx: Ctx):
             fieldName1 == fieldName2 && p1 =:= p2
         }
       case (Pattern.Synonym(sym1, args1), Pattern.Synonym(sym2, args2)) =>
-        // TODO : there is surely some other conditions
-        // and we may want to accept equality beween a synonym and its input pattern
-        sym1 === sym2
+        args1 == args2 && sym1.params == sym2.params && sym1.body == sym2.body
```
LPTK: Pro-tip `==` is deprecated (use `===` instead). There used to be a WartRemover warning in Scala 2, but we lost it since we moved to Scala 3.

Chronarak: Is there a reason for this? If my linter is right, `===` is actually a wrapper for `==`.
In any case,  this is solved in https://github.com/hkust-taco/mlscript/pull/306/commits/93050a32f0f764949f6ab9e9bbae2241296f205f

LPTK: Scala's typing of `==` is too permissive and leads to bug when code is changed/refactored. See also the compromise in https://docs.scala-lang.org/scala3/reference/contextual/multiversal-equality.html, but last time I tried it (a couple of years ago), it wasn't really usable yet.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -490,6 +490,18 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
                       (cse, Assign(arg, Select(sr, param.id/*FIXME incorrect Ident?*/)(S(param)), blk))
                   mkMatch(mkArgs(clsParams.iterator.zip(args).collect { case (s1, S(s2)) => (s1, s2) }.toList))
               case Pattern.Tuple(len, inf) => mkMatch(Case.Tup(len, inf) -> go(tail, topLevel = false))
+              case Pattern.Record(entries) =>
+                val objectSym = summon[Ctx].builtins.Object
```
```suggestion
                val objectSym = ctx.builtins.Object
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -490,6 +490,18 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
+                mkMatch( // checking that we have an object
+                  Case.Cls(objectSym, Value.Ref(BuiltinSymbol(objectSym.nme, false, false, true, false))),
+                  entries.foldRight(go(tail, topLevel = false)) {
+                    case ((fieldName, fieldSymbol), blk) =>
+                      mkMatch(
+                        Case.Field(fieldName, safe = true), // we know we have an object, no need to check again
+                        Assign(fieldSymbol, Select(sr, fieldName)(N), blk)
+                      )
+                  }
```
```suggestion
                  entries.foldRight(go(tail, topLevel = false)):
                    case ((fieldName, fieldSymbol), blk) =>
                      mkMatch(
                        Case.Field(fieldName, safe = true), // we know we have an object, no need to check again
                        Assign(fieldSymbol, Select(sr, fieldName)(N), blk)
                      )
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
```diff
@@ -26,11 +26,8 @@ enum Pattern extends AutoLocated:
   case Record(entries: List[(Ident -> BlockLocalSymbol)])
   
   def subTerms: Ls[Term] = this match
-    case Lit(_) => Nil
     case ClassLike(_, t, _, _) => t :: Nil
-    case Synonym(_, _) => Nil
-    case Tuple(_, _) => Nil
-    case Record(_) => Nil
+    case _:(Lit | Synonym | Tuple | Record) => Nil
```
LPTK:
```suggestion
    case _: (Lit | Synonym | Tuple | Record) => Nil 
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
```diff
@@ -50,3 +47,7 @@ enum Pattern extends AutoLocated:
     case Record(Nil) => "{}"
     case Record(entries) =>
       entries.iterator.map(_.name + ": " + _).mkString("{ ", ", ", " }")
+  
+  def isRecord = this match
```
LPTK: This is now unused.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
```diff
@@ -7,11 +7,12 @@ import mlscript.utils.*, shorthands.*
 import Message.MessageContext
 import utils.TraceLogger
 import syntax.Literal
-import Keyword.{as, and, `do`, `else`, is, let, `then`}
+import Keyword.{as, and, `do`, `else`, is, let, `then`, where}
 import collection.mutable.{HashMap, SortedSet}
 import Elaborator.{ctx, Ctxl}
 import scala.annotation.targetName
 import hkmc2.semantics.ClassDef.Parameterized
+import hkmc2.codegen.Case.Lit
```
LPTK: I don;t think this is used. (It would be out of place.)

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/HelperExtractors.scala
```diff
@@ -13,6 +13,12 @@ object HelperExtractors:
       case App(Ident("|"), Tup(lhs :: rhs :: Nil)) => S(lhs, rhs)
       case _ => N
   
+  /** A helper extractor for matching the tree of `x a y`.*/
+  object and:
+    infix def unapply(tree: Tree): Opt[(Tree, Tree)] = tree match
```
LPTK:
```suggestion
    infix def unapply(tree: App): Opt[(Tree, Tree)] = tree match
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -47,7 +48,15 @@ class Normalization(elaborator: Elaborator)(using raise: Raise, ctx: Ctx):
       case (c1: Pattern.ClassLike, c2: Pattern.ClassLike) => c1.sym === c2.sym
       case (Pattern.Lit(l1), Pattern.Lit(l2)) => l1 === l2
       case (Pattern.Tuple(n1, b1), Pattern.Tuple(n2, b2)) => n1 === n2 && b1 === b2
-      case (_, _) => false
+      case (Pattern.Record(ls1), Pattern.Record(ls2)) =>
+        ls1.zip(ls2)
+           .forall{case ((fieldName1, p1), (fieldName2, p2)) =>
+            fieldName1 == fieldName2 && p1 =:= p2
+        }
```
LPTK: * `lazyZip` is more efficient
* `=:=` is just a less used alternative to `===`
* `==` shouldn't be used.

```suggestion
        ls1.lazyZip(ls2).forall:
          case ((fieldName1, p1), (fieldName2, p2)) =>
            fieldName1 === fieldName2 && p1 === p2
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -47,7 +48,15 @@ class Normalization(elaborator: Elaborator)(using raise: Raise, ctx: Ctx):
...
+      case (Pattern.Synonym(sym1, args1), Pattern.Synonym(sym2, args2)) =>
+        args1 === args2 && sym1.params === sym2.params && sym1.body === sym2.body
+      case (_:Pattern.ClassLike, _) | (_:Pattern.Lit, _) |
+        (_:Pattern.Tuple, _) | (_:Pattern.Synonym, _) | (_:Pattern.Record, _) => false
```
LPTK: Please use spaces consistently with the rest of the code base.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -68,6 +77,26 @@ class Normalization(elaborator: Elaborator)(using raise: Raise, ctx: Ctx):
...
       case _ => ()
   
+  extension (lhs: Pattern.Record)
+    /** reduces the record pattern `lhs` assuming we have matched `rhs`.
+      * It removes field matches that may now be unnecessary
+      */
+    infix def assuming(rhs: Pattern): Pattern.Record = rhs match
+      case Pattern.Record(rhsEntries) =>
+        val filteredEntries = lhs.entries.filter { (fieldName1, _) =>
+          rhsEntries.forall { (fieldName2, _) => ! (fieldName1 === fieldName2)}
```
LPTK:
```suggestion
          rhsEntries.forall { (fieldName2, _) => !(fieldName1 === fieldName2)}
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -68,6 +77,26 @@ class Normalization(elaborator: Elaborator)(using raise: Raise, ctx: Ctx):
       case lhs: Pattern.ClassLike => lhs.refined = true
       case _ => ()
   
+  extension (lhs: Pattern.Record)
+    /** reduces the record pattern `lhs` assuming we have matched `rhs`.
+      * It removes field matches that may now be unnecessary
+      */
+    infix def assuming(rhs: Pattern): Pattern.Record = rhs match
+      case Pattern.Record(rhsEntries) =>
+        val filteredEntries = lhs.entries.filter { (fieldName1, _) =>
```
LPTK:
```suggestion
        val filteredEntries = lhs.entries.filter: (fieldName1, _) =>
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -68,6 +77,26 @@ class Normalization(elaborator: Elaborator)(using raise: Raise, ctx: Ctx):
...
+          rhsEntries.forall { (fieldName2, _) => ! (fieldName1 === fieldName2)}
+        }
+        Pattern.Record(filteredEntries)
+      case Pattern.ClassLike(sym = cls : ClassSymbol) =>
+        cls.defn match
+        case S(ClassDef.Parameterized(params = paramList)) =>
+          val filteredEntries = lhs.entries.filter { (fieldName1, _) =>
+          paramList.params.forall { (param:Param) => ! (fieldName1 === param.sym.id)}
```
LPTK: Please use indentation consistently with the rest of the code base.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -68,6 +77,26 @@ class Normalization(elaborator: Elaborator)(using raise: Raise, ctx: Ctx):
...
+    infix def assuming(rhs: Pattern): Pattern.Record = rhs match
+      case Pattern.Record(rhsEntries) =>
+        val filteredEntries = lhs.entries.filter { (fieldName1, _) =>
+          rhsEntries.forall { (fieldName2, _) => ! (fieldName1 === fieldName2)}
+        }
+        Pattern.Record(filteredEntries)
+      case Pattern.ClassLike(sym = cls : ClassSymbol) =>
+        cls.defn match
+        case S(ClassDef.Parameterized(params = paramList)) =>
+          val filteredEntries = lhs.entries.filter { (fieldName1, _) =>
```
LPTK:
```suggestion
          val filteredEntries = lhs.entries.filter: (fieldName1, _) =>
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -154,8 +183,9 @@ class Normalization(elaborator: Elaborator)(using raise: Raise, ctx: Ctx):
   
   /**
     * Specialize `split` with the assumption that `scrutinee` matches `pattern`.
-    * If `matchOrNot` is `true`, the function _keeps_ branches that agree on
-    * `scrutinee` matches `pattern`. Otherwise, the function _removes_ branches
+    * If `mode` is `+`, the function _keeps_ branches that agree on
+    * `scrutinee` matches `pattern` and simplifies the record patterns it sees if the fields were already matched.
+    * Otherwise (if `mode is `-`), the function _removes_ branches
```
LPTK:
```suggestion
    * `scrutinee` matching `pattern` and simplifies the record patterns it sees if the fields were already matched.
    * Otherwise (if `mode` is `-`), the function _removes_ branches
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -188,7 +218,16 @@ class Normalization(elaborator: Elaborator)(using raise: Raise, ctx: Ctx):
...
+            else thatPattern match
+            case thatPattern :Pattern.Record =>
+              log(s"Case 1.1.4: $thatPattern is a record")
+              // we can use information if pattern is itself a record, or if it is a constructor with arguments
+              val simplifiedRecord = thatPattern assuming pattern
+              if simplifiedRecord.entries.isEmpty then
+                tail
+              else
+                Split.Cons(Branch(thatScrutinee, simplifiedRecord, continuation), tail)
+            case _ => (if pattern <:< thatPattern then
```
LPTK: Why parenthesize this? Maybe just use normal indentation and no extra parens.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -249,7 +290,17 @@ object Normalization:
     case (Lit(Tree.StrLit(_)), ClassLike(blt.`Str`, _, _, _)) => true
     case (Lit(Tree.DecLit(_)), ClassLike(blt.`Num`, _, _, _)) => true
     case (Lit(Tree.BoolLit(_)), ClassLike(blt.`Bool`, _, _, _)) => true
-    case (_, _) => false
+    case (_: Synonym, _: Synonym) => lhs =:= rhs
+    case (Record(entries1), Record(entries2)) =>
+      entries1.forall { (fieldName1, _) => entries2.exists { (fieldName2, _) => fieldName1 === fieldName2 } }
+    case (Record(entries), ClassLike(cs: ClassSymbol, _, _, _)) =>
```
LPTK:
```suggestion
    case (Record(entries), ClassLike(sym = cs: ClassSymbol)) =>
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -249,7 +290,17 @@ object Normalization:
...
+      val clsParams = cs.defn match
+        case S(Parameterized(params = paramList)) => paramList.params
+        case (S(_)|N) => List()
```
LPTK: `Nil` is more efficient

```suggestion
        case (S(_) | N) => Nil
```

hkmc2/shared/src/main/scala/hkmc2/syntax/Keyword.scala
```diff
@@ -59,7 +59,7 @@ object Keyword:
   val `..` = Keyword("..", N, N)
   val `...` = Keyword("...", N, N)
   // val `;` = Keyword(";", ascPrec, eqPrec)
-
```
LPTK: 😢

hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala
```diff
@@ -401,7 +401,7 @@ abstract class Parser(
               tryEmpty(tok, loc)
           case N =>
             tryEmpty(tok, loc)
-      case N =>
+      case S(_) | N => // should only include the case S(Keyword.`:`) and N
```
LPTK:
```suggestion
      case S(kw) | N =>
        assert(kw is Keyword.`:`)
```

hkmc2/shared/src/test/mlscript/backlog/UCS.mls
```diff
@@ -81,18 +81,10 @@ if
 let audits = new Set()
 //│ audits = Set{}
 
-:fixme
 if audits.has(1) === true do
```
LPTK: This fixed test case can now be moved out of the backlog into a normal test file (pick a meaningful one that maybe already tests `do`).

hkmc2/shared/src/test/mlscript/ucs/patterns/where.mls
```diff
@@ -0,0 +1,60 @@
+:js
+
+data class Pair[A, B](fst: A, snd: B)
+
+
+fun orderedPair(p) = p is
+    Pair(a, b) where a <= b
```
LPTK: Here and everywhere else, please use two spaces for indentation.

LPTK (approved):
Awesome changes, thanks!

I only have minor comments (and so many about good formatting – or lack thereof). Once they're addressed, please assign me and I'll merge the PR.

hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala
Chronarak: Scala is not happy about not always binding the same number of variables, so I inlined the check in the pattern, see https://github.com/hkust-taco/mlscript/pull/306/commits/5dd6ed6ae1dbf9a3056696af2398e8a75db0cfd0.

LPTK: Oh yeah of course, here any language should be unhappy. It was my mistake.

hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala
```diff
@@ -401,8 +401,10 @@ abstract class Parser(
               tryEmpty(tok, loc)
           case N =>
             tryEmpty(tok, loc)
-      case N =>
+      case S(Keyword.`:`) | N =>
         tryParseExp(prec, tok, loc, rule)
+      case S(kw) =>
+        lastWords(s"invalid keyword $kw")
```
LPTK: How is this an invalid keyword? This is a case that should be unreachable – and if it is ever reached, it's because we messed up, not because some keyword is "invalid" 😉

LPTK: So, here, it's much simpler to just handlethe ``case S(Keyword.`:`)`` on its own before the `case S(kw)` and remove the latter's guard, which should make the Scala compiler happy without needing an unsafe assertion.

hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala
Chronarak: Oh yeah I agree it should be refused.
I just thought that maybe there was a way to make Scala understand that the match was exhaustive, but i guess not.
The only way I see is to enumerate all the possible keywords on the first branch, which is just not a good idea.

hkmc2/shared/src/main/scala/hkmc2/syntax/Parser.scala
Chronarak: My issue with this solution is that it duplicates the associated code. Currently it's very simple, but it sounds less future proof to me.

LPTK: Actually I pushed a solution that doesn't.

PR #308, opened 2025-05-17
Move UCS normalization before lowering

chengluyu:


hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
```diff
@@ -89,16 +89,11 @@ class Desugarer(val elaborator: Elaborator)
     def ++(fallback: Split): Split =
       if fallback == Split.End then
         split
-      else if split.isFull then
-        raise:
-          ErrorReport:
-            msg"The following branches are unreachable." -> fallback.toLoc ::
-            msg"Because the previous split is full." -> split.toLoc :: Nil
-        split
       else (split match
```
LPTK:
```suggestion
      else split match
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
```diff
@@ -89,16 +89,11 @@ class Desugarer(val elaborator: Elaborator)
...
         case Split.Cons(head, tail) => Split.Cons(head, tail ++ fallback)
         case Split.Let(name, term, tail) => Split.Let(name, term, tail ++ fallback)
-        case Split.Else(_) /* impossible */ | Split.End => fallback)
+        case Split.Else(_) => split
```
LPTK: Is it impossible or not? Should we use `die` instead?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
```diff
@@ -658,6 +647,70 @@ class Desugarer(val elaborator: Elaborator)
...
+      fallback: Split,
+      consequent: Split,
+  ): Split = scoped("ucs:rp"):
+    log(s"SYNONYM: ${scrutinee.showDbg} is $symbol")
+    import DeBrujinSplit.*, PatternStub.*
+    val pattern2 = ClassLike(ConstructorLike.Instantiation(symbol, arguments))
+    val mainSplit = Binder(Branch(Outermost, pattern2, Accept(42), Reject))
+    log(s"the initial split:\n${mainSplit.display}")
+    val (normalizedMainSplit, idSplitMap) = scoped("ucs:rpn"):
+      mainSplit.normalize(using elaborator.tl)
```
LPTK: Doesn't this normalization require everything to be resolved? But you're currently calling `compilePattern` from the _desugarer_!

Why is this no longer done in normalization?

hkmc2/shared/src/test/mlscript/backlog/UCS.mls
```diff
@@ -81,19 +81,6 @@ if
 let audits = new Set()
 //│ audits = Set{}
 
-:fixme
-if audits.has(1) === true do
-  print("ok")
-else
-  1
```
LPTK: You removed the test case instead of moving it to a different file?!

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
chengluyu: Resolved in d0d221c.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
chengluyu: It's possible because I removed test `split.isFull` before it.

But now I think I should keep that test. I added it back and use `die` in the case.

Resolved in d0d221c.

hkmc2/shared/src/test/mlscript/backlog/UCS.mls
chengluyu: 😢 Oh! I shouldn't have done this. Thanks for the catch.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
```diff
@@ -89,11 +89,13 @@ class Desugarer(val elaborator: Elaborator)
...
         split
-      else (split match
+      else if split.isFull then
+        split
+      else split match
         case Split.Cons(head, tail) => Split.Cons(head, tail ++ fallback)
         case Split.Let(name, term, tail) => Split.Let(name, term, tail ++ fallback)
-        case Split.Else(_) => split
-        case Split.End => fallback)
+        case Split.Else(_) => die
```
LPTK: Ok, now I see that this was supposed to be impossible because of the `split.isFull` check that you added back...

It's not very good style anyway. I'd prefer just:
```suggestion
        case Split.Else(_) => split // Shouldn't actually happen because of the `split.isFull` check above
```
This way, if we refactor the code and this case actually does trigger, it will still work and not crash.

hkmc2/shared/src/test/mlscript/ucs/syntax/Else.mls
```diff
@@ -83,4 +83,13 @@ fun f(x, y) =
   if x
     then y
 
+// ——— ——— ———
```
LPTK: Why? This test file doesn't use such separators.

hkmc2/shared/src/main/scala/hkmc2/bbml/bbML.scala
```diff
@@ -295,9 +295,10 @@ class BBTyper(using elState: Elaborator.State, tl: TL):
       val nestCtx1 = ctx.nest
       val nestCtx2 = ctx.nest
       val patTy = pattern match
-      case Pattern.ClassLike(sym, _, _, _) =>
-        val (clsTy, tv, emptyTy) = sym.asCls.flatMap(_.defn) match
-        case S(cls) =>
+      case pat @ Pattern.ClassLike(_, _, _, _) =>
```
LPTK: Bunches of underscores are an anti-pattern
```suggestion
      case pat: Pattern.ClassLike =>
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -466,29 +466,28 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
               )
             pat match
               case Pattern.Lit(lit) => mkMatch(Case.Lit(lit) -> go(tail, topLevel = false))
-              case Pattern.ClassLike(cls: ClassSymbol, _trm, _args0, _refined)
-                  // Do not elaborate `_trm` when the `cls` is virtual.
-                  if Elaborator.ctx.builtins.virtualClasses contains cls =>
+              // Do not elaborate `_trm` when the `cls` is virtual.
+              case pat @ Pattern.ClassLike(ctor, _, _, _) if pat.isVirtualClass =>
```
LPTK: Bunches of underscores in patterns are an _anti_-pattern.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -466,29 +466,28 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
-                  val args = args0.getOrElse(Nil)
-                  val clsParams = cls match
-                    case cls: ClassSymbol => cls.tree.clsParams
-                    case _: ModuleSymbol => Nil
-                  assert(args0.isEmpty || clsParams.length === args.length)
+                mkMatch(Case.Cls(pat.ctorSym.asClsOrMod.get, unreachableFn) -> go(tail, topLevel = false))
+              case Pattern.ClassLike(ctor, args0, _, _refined) =>
+                subTerm_nonTail(ctor): st =>
+                  val args = args0.map(_.map(_.scrutinee)).getOrElse(Nil)
+                  val (ctorSym, clsParams) = ctor.symbol.flatMap(_.asClsOrMod) match // TODO(ucs)
```
LPTK: Seems you forgot to include the None case, which should probably be a compilation error.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -466,29 +466,28 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
+                    case S(cls: ClassSymbol) => (cls, cls.tree.clsParams)
+                    case S(mod: ModuleSymbol) => (mod, Nil)
+                  // assert(args0.isEmpty || clsParams.length === args.length) // TODO(ucs): Is this really needed?
```
LPTK: Isn't this invariant supposed to hold? Why comment the assertion?

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
```diff
@@ -13,34 +14,85 @@ enum Pattern extends AutoLocated:
...
-    case ClassLike(_, t, _, _) => t :: Nil
+    case ClassLike(ctor, _, _, _) => ctor :: Nil
     case Tuple(_, _) => Nil
     case Record(_) => Nil
   
   def children: Ls[Located] = this match
     case Lit(literal) => literal :: Nil
-    case ClassLike(_, t, args, _) =>
-      t :: args.fold(Nil)(_.collect { case S(symbol) => symbol })
+    case ClassLike(ctor, scruts, _, _) => ctor :: scruts.map(_.map(_.scrutinee)).getOrElse(Nil)
```
LPTK: Why did you replace a `fold` with the less efficient `map`/`getOrElse`?

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
```diff
@@ -13,34 +14,85 @@ enum Pattern extends AutoLocated:
...
     case Tuple(fields, _) => Nil
     case Record(entries) => entries.flatMap { case (nme, als) => nme :: als :: Nil }
   
   def showDbg: Str = this match
     case Lit(literal) => literal.idStr
-    case ClassLike(sym, t, ps, rfd) => (if rfd then "refined " else "") +
-      sym.nme + ps.fold("")(_.iterator.map(_.fold("_")(_.toString)).mkString("(", ", ", ")"))
+    case ClassLike(ctor, args, _, rfd) =>
+      def showCtor(ctor: Term): Str = ctor match
+        case Term.Ref(sym: BlockMemberSymbol) => sym.nme // w/o `refNum`` and prefix
```
LPTK: What does this comment mean?

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
```diff
@@ -13,34 +14,85 @@ enum Pattern extends AutoLocated:
...
+  /** A class-like pattern whose symbol is resolved to a module. */
+  object Module:
+    def unapply(p: Pattern): Opt[ModuleSymbol] = p match
+      case Pattern.ClassLike(ctor, _, _, _) => ctor.symbol.flatMap(_.asModOrObj)
+      case _ => N
+
+  private[Pattern] sealed trait ClassLikeImpl:
+    p: Pattern.ClassLike =>
+    
+    def ctorSym: ClassSymbol | ModuleSymbol | PatternSymbol =
```
LPTK: You use this in many places. Some of them might be reachable in case there's an error in the user prorgam. In such cases, the compiler shouldn't crash.

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
```diff
@@ -13,34 +14,85 @@ enum Pattern extends AutoLocated:
...
+        lastWords("Pattern.ClassLike: constructor is not a class or module")
+
+    def isVirtualClass(using Ctx): Bool = ctorSym match
+      case cls: ClassSymbol => ctx.builtins.virtualClasses contains cls
+      case _: (ModuleSymbol | PatternSymbol) => false
+  
+  enum MatchMode:
+    /** The default mode. If the constructor resolves to:
+     *  - a class symbol, then check if the scrutinee is an instance;
+     *  - a module symbol, then check if the scrutinee is the module itself;
```
LPTK: Actually pattern matching on modules is not intentionally supported. Or does that refer to `object`s?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
```diff
@@ -443,77 +447,105 @@ class Desugarer(val elaborator: Elaborator)
...
-          compilePattern(ref, pat, patArgs.zip(args), fallback, sequel(ctx))
-      case S(_: PatternSymbol) =>
-        makeUnapplyBranch(ref, clsTrm, sequel(ctx))(fallback)
-      case _ =>
-        // Raise an error and discard `sequel`. Use `fallback` instead.
-        raise(ErrorReport(msg"Cannot use this ${ctor.describe} as an extractor" -> ctor.toLoc :: Nil))
-        fallback
+    def dealWithCtorCase(ctor: Ctor, mode: MatchMode)(fallback: Split): Sequel = ctx =>
+      Branch(ref, Pattern.ClassLike(term(ctor), N, mode, false)(ctor), sequel(ctx)) ~: fallback
+    // @deprecated("Remove before merging the PR.")
```
LPTK: 👀

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
```diff
@@ -443,77 +447,105 @@ class Desugarer(val elaborator: Elaborator)
...
+              error(msg"Ill-formed pattern argument" -> td.toLoc); N
+            case _ => N
+          (symbol, tree, argument)
+      .toList
+      Branch(
+        ref,
+        Pattern.ClassLike(term(ctor), S(matches), mode, false)(app), // TODO: refined?
+        subMatches(matches, sequel)(Split.End)(ctx)
+      ) ~: fallback
+    // @deprecated("Remove before merging the PR.")
```
LPTK: 👀

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
```diff
@@ -528,11 +560,12 @@ class Desugarer(val elaborator: Elaborator)
         val aliasSymbol = VarSymbol(id)
         val ctxWithAlias = ctx + (nme -> aliasSymbol)
         Split.Let(aliasSymbol, ref, sequel(ctxWithAlias) ++ fallback)
-      case ctor: Ctor => dealWithCtorCase(ctor, false)
-      case Annotated(Ident("compile"), ctor: Ctor) => dealWithCtorCase(ctor, true)
+      case ctor: Ctor => dealWithCtorCase(ctor, MatchMode.Default)
+      case Annotated(compiled @ Ident("compile"), ctor: Ctor) =>
```
LPTK: I don't think you can correctly interpret the annotation now. The user might have imported a different `compile` annotation; it might also be selected out of its parent module. Why not just keep annotations until normalization, which will intepret them based on their resolved symbol?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
```diff
@@ -550,44 +583,45 @@ class Desugarer(val elaborator: Elaborator)
...
         ) ~: fallback
       // Negative numeric literals
       case App(Ident("-"), Tup(IntLit(value) :: Nil)) => fallback => ctx =>
         Branch(ref, Pattern.Lit(IntLit(-value)), sequel(ctx)) ~: fallback
       case App(Ident("-"), Tup(DecLit(value) :: Nil)) => fallback => ctx =>
         Branch(ref, Pattern.Lit(DecLit(-value)), sequel(ctx)) ~: fallback
       // A single constructor pattern.
-      case Annotated(Ident("compile"), app @ App(ctor: Ctor, Tup(args))) =>
-        dealWithAppCtorCase(app, ctor, args, true)
+      case Annotated(compiled @ Ident("compile"), app @ App(ctor: Ctor, Tup(args))) =>
```
LPTK: Same remark as above.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -81,28 +98,249 @@ class Normalization(using raise: Raise, tl: TL, ctx: Ctx, state: State):
...
+              log("BROKEN")
+              normalizeImpl(alternative)
+            else
+              // TODO(ucs): deduplicate [1]
+              val whenTrue = normalize(specialize(consequent ++ alternative, +, scrutinee, pattern))
+              val whenFalse = normalizeImpl(specialize(alternative, -, scrutinee, pattern).clearFallback)
+              Branch(scrutinee, pattern.selectClass, whenTrue) ~: whenFalse
+          case S(pat: PatternSymbol) => mode match
+            case MatchMode.Default =>
+              // TODO(rp): `argsOpt` is not useless for now
```
LPTK: What does that TODO mean?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
chengluyu: This has been removed in 3090e81.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
chengluyu: This has been removed in 3090e81, too.

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
chengluyu: Yes. That refers to `object`s.

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
chengluyu: The comment explains why `BlockMemberSymbol` needs to be treated specially: I don't need to print the `refNum` (it's 666 most of the time) and the "member:" prefix in the UCS debug information.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
chengluyu: Fix in 1b9a477.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
chengluyu: Fix in 1b9a477.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
chengluyu: Implemented in af5b41a.

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
chengluyu: This is removed and replaced with case analysis on `pattern.constructor.symbol` with proper error messages in e49bd1d.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
chengluyu: Clarified in ab3a69c.

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
```diff
@@ -64,6 +60,16 @@ enum Pattern extends AutoLocated:
...
 object Pattern:
+  /** Represent the type of arguments in `ClassLike` patterns. This type alias
+   *  is used to reduce repetition in the code.
+   * 
+   *  - Field `pattern` is for error messages.
+   *  - Field `split` is for pattern compilation.
+   *    **TODO(ucs/rp)**: Replace with suitable representation when implement
+   *    the new pattern compilation.
+   */
+  type Argument = (scrutinee : BlockLocalSymbol, pattern : Tree, split : Opt[DeBrujinSplit])
```
LPTK: What's with the weird space before `:`?

hkmc2/shared/src/test/mlscript/ucs/patterns/Parameters.mls
```diff
@@ -104,3 +104,28 @@ point(Point(2, 4))
...
+// Validation on object patterns
+
+object A
+
+:e
+fun foo(x) = if x is A() then 0
+//│ ╔══[ERROR] `A` is an object.
+//│ ║  l.110: 	object A
+//│ ║         	       ^
+//│ ╟── Its pattern cannot have a parameter list.
```
LPTK:
```suggestion
//│ ╟── Its pattern cannot have an argument list.
```

hkmc2/shared/src/test/mlscript/ucs/patterns/Parameters.mls
```diff
@@ -104,3 +104,28 @@ point(Point(2, 4))
...
+//│ ║  l.113: 	fun foo(x) = if x is A() then 0
+//│ ╙──       	                     ^^^
+
+:e
+fun foo(x) = if x is A(y) then y
+//│ ╔══[ERROR] `A` is an object.
+//│ ║  l.110: 	object A
+//│ ║         	       ^
+//│ ╟── Its pattern cannot have parameters.
```
LPTK:
```suggestion
//│ ╟── Its pattern cannot have arguments.
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
chengluyu: Fixed in 4a9dd48.

hkmc2/shared/src/test/mlscript/ucs/patterns/Parameters.mls
chengluyu: Fixed in 8b2af81.

hkmc2/shared/src/test/mlscript/ucs/patterns/Parameters.mls
chengluyu: Fixed in 8b2af81.

hkmc2/shared/src/test/mlscript/ucs/examples/LeftistTree.mls
```diff
@@ -1,44 +1,27 @@
...
-    "(" ++ show(l) ++ " " ++ toString(v) ++ " " ++ show(r) ++ ")"
+    "(" + show(l) + " " + String(v) + " " + show(r) + ")"
   Empty then "•"
 
 fun singleton(x) = Node(x, Empty, Empty, 1)
 fun rank(t) = if t is
   Empty then 0
   Node(_, _, _, r) then r
 
-// This can be improved. This can be better.
```
LPTK, line 41: So it could not be improved? 👀

hkmc2/shared/src/test/mlscript/ucs/normalization/UnifySubScrutinees.mls
```diff
@@ -17,6 +19,7 @@ fun sum(acc, xs) =
     x :: xs then sum(acc + x, xs)
     Nil     then acc
 
+// :ucs normalized
```
LPTK, line 22: Do we need this comment? Is this a TODO?

LPTK: Same remarks for similar commented lines in other files.

hkmc2/shared/src/test/mlscript/ucs/papers/OperatorSplit.mls
```diff
@@ -1,231 +1,37 @@
...
-fun abs(x) = if x < 0 then -x else x
-
-:pt
-:elt
-fun example(args) =
-  if foo(args)
+// :pt
+// :elt
+// :ucs normalized
+:todo
```
LPTK: Remove the `:todo`. The `:fixme` below already does the job.

hkmc2/shared/src/test/mlscript/ucs/syntax/Do.mls
```diff
@@ -58,13 +58,6 @@ f(42)
 f(41)
 //│ = Some(41)
 
-let audits = new Set()
```
LPTK, line 64: I guess you're removing this because it's already in `Else.mls`, but that file does not have `:js` at the moment.

hkmc2/shared/src/test/mlscript/ucs/hygiene/HygienicBindings.mls
```diff
@@ -45,7 +39,8 @@ fun h1(a) =
     a is None then 0
 
 // FIXME
```
LPTK: Is this still a FIXME?

hkmc2/shared/src/test/mlscript/ucs/normalization/UnifySubScrutinees.mls
chengluyu: This is just to make it easier to uncomment, similar to some of the commented commands in HkScratch.mls. By the way, do you think we can keep a few desugared or normalized splits in test files?

hkmc2/shared/src/test/mlscript/ucs/examples/LeftistTree.mls
chengluyu: Done in 828a84a.

hkmc2/shared/src/test/mlscript/ucs/papers/OperatorSplit.mls
chengluyu: It turned out that this is just a coding mistake. Fixed in 95bb793.

hkmc2/shared/src/test/mlscript/ucs/syntax/Do.mls
chengluyu: Yes. I've added `:js` to that test file in 166e5d0.

hkmc2/shared/src/test/mlscript/ucs/hygiene/HygienicBindings.mls
chengluyu: No. It's no longer a FIXME. I've clarified the comments.

hkmc2/shared/src/test/mlscript/ucs/normalization/UnifySubScrutinees.mls
LPTK: Sure, why not? As long as it's not many.

hkmc2/shared/src/test/mlscript/ucs/syntax/Else.mls
```diff
@@ -98,3 +93,31 @@ if audits.has(1) === true do
...
+  then
+    5
+  else 1
+) + 3
+//│ ╔══[PARSE ERROR] Unexpected 'else' keyword here
+//│ ║  l.108: 	  else 1
+//│ ╙──       	  ^^^^
+//│ = 8
+
+:fixme
```
LPTK, line 115: I'm not so sure we want to support that... Anyway, this will be addressed in the parser rewrite (no need to change it).

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/DeBrujinSplit.scala
```diff
@@ -315,18 +317,32 @@ extension (split: DeBrujinSplit)
         // lastWords(s"Arity mismatch in expand: ${split.unbind}")
         Reject // TODO: report mismatched arity
   
+    /** To perform a reverse lookup for a term that references a symbol in the current context. */
+  def reference(target: ClassSymbol | ModuleSymbol)(using Ctx, State): Opt[Term] =
```
LPTK, line 321: It seems you moved this function from `Elaborator`. But your recent borked merge conflict resolution re-introduced it there.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -105,78 +115,288 @@ class Normalization(elaborator: Elaborator)(using raise: Raise, ctx: Ctx):
...
+      // are forged in the elaborator state, so it does not have `defn`.
+      // TODO(ucs): do not make special case for them
+      if ctorSymbol == State.matchResultClsSymbol then
+        (ctorSymbol.id: Located) -> matchResultClassParamOpt
+      else if ctorSymbol == State.matchFailureClsSymbol then
+        (ctorSymbol.id: Located) -> matchFailureClassParamOpt
+      else
+        ctorSymbol.defn match
+          case N => lastWords(s"Class ${ctorSymbol.name} does not have a definition")
+          case S(cd) => ctorSymbol.id -> cd.paramsOpt
```
chengluyu: I’d like to know if there’s a solution to this.

chengluyu: Can we get symbols for `MatchResult` and `MatchFailure` by adding them to `Prelude.mls` and then using `assumeBuiltinCls`?

LPTK: The simplest solution is to just put the definitiosn there yourself. Any problem with that approach? Otherwise, we'd have to declare these classes in the `Prelude` and special-case their lowering, but that seems a bit troublesome.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -105,78 +115,288 @@ class Normalization(elaborator: Elaborator)(using raise: Raise, ctx: Ctx):
...
+    case MatchMode.Default | _: MatchMode.StringPrefix => ()
+    case MatchMode.Annotated(annotation) => annotation.symbol.flatMap(_.asObj) match
+      case S(symbol) if symbol === ctx.builtins.compile =>
+        warn(msg"Cannot compile `${ctorSymbol.name}`," -> Loc(annotation :: ctorTerm :: Nil),
+          msg"because it is a ${ctorSymbol.tree.k.desc}." -> ctorSymbol.toLoc,
+          msg"Note: only patterns can be compiled." -> N)
+      case S(_) =>
+        warn(msg"Unknown annotation on pattern" -> annotation.toLoc,
+        msg"Note: only `@compile` is supported on patterns." -> N)
+      // Name resolution should have already reported an error.
```
LPTK:
```suggestion
      // `Resolver` should have already reported an error.
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Translator.scala
```diff
@@ -137,19 +116,31 @@ class Translator(val elaborator: Elaborator)
...
+      val postfixSymbol = new TempSymbol(N, "postfix")
+      val mode = MatchMode.StringPrefix(prefixSymbol, postfixSymbol)
+      val pattern = Pattern.ClassLike(ctorTrm, N, mode, false)(ctor)
+      Branch(scrut(), pattern, inner(Map.empty, () => postfixSymbol.ref())) ~: Split.End
+      
+      // Because we don't resolve the symbol of `ctor` here, we don't know if
+      // it is a class, a module, or a pattern symbol. We no longer need to
+      // make the unapply branch here and it will be done in normalization.
+      // - - - - - //
+      // OLD CODE FOR REFERENCE:
```
LPTK: Is it useful to keep this code commented here?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Translator.scala
```diff
@@ -219,21 +209,26 @@ class Translator(val elaborator: Elaborator)
...
-      case Split.Else(Term.Error) =>
-        makeMatcher("unapplyStringPrefix", scrutSym, failure)
-      case split =>
-        val topmost = split ~~: failure
-        log(s"Translated `unapplyStringPrefix`: ${display(topmost)}")
-        makeMatcher("unapplyStringPrefix", scrutSym, topmost)
-    unapply :: unapplyStringPrefix :: Nil
+    if patternParams.nonEmpty then
+      // Temporarily disable the translation of pattern with pattern parameters.
+      // TODO(rp): pass pattern parameters as to the `unapply` function
```
LPTK: as to?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/package.scala
```diff
@@ -1,12 +1,12 @@
 package hkmc2
 package semantics
 
-import sourcecode.{Name, Line, FileName}
+import sourcecode.{FileName, Line, Name}
 
 package object ucs:
-  def error(using Raise, Line, Name, FileName)(msgs: (Message, Option[Loc])*): Unit =
+  def error(using Line, FileName)(msgs: (Message, Option[Loc])*)(using Raise)(using Name): Unit =
```
LPTK: Why this change?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/package.scala
```diff
@@ -1,12 +1,12 @@
...
     raise(ErrorReport(msgs.toList))
   
-  def warn(using Raise, Line, Name, FileName)(msgs: (Message, Option[Loc])*): Unit =
+  def warn(using Line, FileName)(msgs: (Message, Option[Loc])*)(using Raise)(using Name): Unit =
```
LPTK: Why this change?

hkmc2/shared/src/test/mlscript/HkScratch.mls
LPTK, line 1: Please move tests here to a proper test file and leave this file unchanged.

hkmc2/shared/src/test/mlscript/decls/Prelude.mls
```diff
@@ -104,6 +104,7 @@ declare module debug with
   fun printStack
   fun getLocals
 
+declare object compile
```
LPTK: Adding this here without care is dangerous:
```
compile
//│ ═══[RUNTIME ERROR] Error: Access to required field 'compile' yielded 'undefined'
```
It would make more sense to add it inside some `annotations` "virtual module" and add a check to the sequence of checks here:

https://github.com/hkust-taco/mlscript/blob/55e4008eab9095b4b57de0b0127198ace0da2d1b/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala#L223

LPTK (approved):
Otherwise, LGTM!

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Translator.scala
chengluyu: No, I'm going to remove it.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/package.scala
chengluyu: It's not intended. I did the same change but the parameters are in different order on the operator split branch. 

~~First, `Name` is not required by `ErrorReport`, so it can be removed. Then,~~ I'll put all implicit parameters in a single parameter list.

chengluyu: I just remembered I mimicked the parameter order of `ErrorReport`.

```scala
object ErrorReport:
  def apply(using Line, FileName)
      (msgs: Ls[Message -> Opt[Loc]], extraInfo: => Opt[Any] = N, source: Source = Source.Typing)
      (using Name): ErrorReport =
    ErrorReport(msgs.head._1.show, msgs, () => extraInfo, source)
```

May I know why implicit parameters are ordered like this?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
chengluyu: Cool. There’s no problem. I just want to confirm before taking action.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/package.scala
LPTK: I don't think there's a good reason for `Name` to be at the end or `FileName` to be at the beginning. But `Line` must be _before_ the other parameters, or we'll get inaccurate line number information. `ErrorReport` is probably like this because during a refactoring I lazily added a param list I was also adding to many places.

hkmc2/shared/src/test/mlscript/syntax/annotations/Pattern.mls
```diff
@@ -1,33 +1,32 @@
...
-//│ ║  l.6: 	fun foo(x) = if x is @compile Box then "yes" else "no"
-//│ ║       	                      ^^^^^^^^^^^
-//│ ╟── because it is a class.
-//│ ║  l.3: 	class Box()
-//│ ║       	      ^^^
-//│ ╙── Note: only patterns can be compiled.
+//│ ╔══[WARNING] Unknown annotation on this class.
+//│ ║  l.8: 	fun foo(x) = if x is @compile Box then "yes" else "no"
+//│ ║       	                      ^^^^^^^
+//│ ╙── Note: `@compile` is only supported on patterns.
```
LPTK: But this _is_ a pattern. It's a class instance pattern.

hkmc2/shared/src/test/mlscript/syntax/annotations/Pattern.mls
```diff
@@ -1,33 +1,32 @@
...
 :w
 fun foo(x) = if x is @compile Box then "yes" else "no"
-//│ ╔══[WARNING] Cannot compile `Box`,
-//│ ║  l.6: 	fun foo(x) = if x is @compile Box then "yes" else "no"
-//│ ║       	                      ^^^^^^^^^^^
-//│ ╟── because it is a class.
-//│ ║  l.3: 	class Box()
-//│ ║       	      ^^^
-//│ ╙── Note: only patterns can be compiled.
+//│ ╔══[WARNING] Unknown annotation on this class.
```
LPTK: The error message seems to have regressed significantly.

hkmc2/shared/src/test/mlscript/syntax/annotations/Pattern.mls
```diff
@@ -1,33 +1,32 @@
...
 
 :w
 fun foo(x) = if x is @do_not_compile Box then "yes" else "no"
-//│ ╔══[WARNING] Unknown annotation on pattern
-//│ ║  l.18: 	fun foo(x) = if x is @do_not_compile Box then "yes" else "no"
+//│ ╔══[WARNING] Unknown annotation on this class.
+//│ ║  l.17: 	fun foo(x) = if x is @do_not_compile Box then "yes" else "no"
 //│ ║        	                      ^^^^^^^^^^^^^^
-//│ ╙── Note: only `@compile` is supported on patterns.
+//│ ╙── Note: `@compile` is only supported on patterns.
```
LPTK: This change is backwards. The previous one was correct and this one is not.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/package.scala
chengluyu: OK! I got it. I have already put `Line` at the beginning. Now the parameter lists look like:

```scala
def error(using Line, FileName, Name, Raise)(msgs: (Message, Option[Loc])*): Unit
```

hkmc2/shared/src/test/mlscript/syntax/annotations/Pattern.mls
```diff
@@ -6,27 +6,47 @@ class Box()
...
+:w
+fun foo(x) = if x is @compile Singleton then "yes" else "no"
+//│ ╔══[WARNING] `@compile` cannot annotate the object instance pattern
+//│ ║  l.17: 	fun foo(x) = if x is @compile Singleton then "yes" else "no"
+//│ ║        	                      ^^^^^^^
+//│ ╙── Note: The `@compile` annotation is intended for pattern compilation.
+
+:e
+fun foo(x) = if x is @compile 42 then "yes" else "no"
+//│ ╔══[ERROR] Unrecognized pattern (annotated)
```
LPTK, line 25: Why is the error different here than the ones above?

hkmc2/shared/src/test/mlscript/syntax/annotations/Pattern.mls
```diff
@@ -6,27 +6,47 @@ class Box()
 
 :w
 fun foo(x) = if x is @compile Box then "yes" else "no"
-//│ ╔══[WARNING] Unknown annotation on this class.
+//│ ╔══[WARNING] `@compile` cannot annotate the class instance pattern
```
LPTK:
```suggestion
//│ ╔══[WARNING] `@compile` cannot be used on class instance pattern
```

hkmc2/shared/src/test/mlscript/syntax/annotations/Pattern.mls
```diff
@@ -6,27 +6,47 @@ class Box()
...
 //│ ║  l.8: 	fun foo(x) = if x is @compile Box then "yes" else "no"
 //│ ║       	                      ^^^^^^^
-//│ ╙── Note: `@compile` is only supported on patterns.
+//│ ╙── Note: The `@compile` annotation is intended for pattern compilation.
```
LPTK:
```suggestion
//│ ╙── Note: The `@compile` annotation is for compiling pattern definitions.
```

hkmc2/shared/src/test/mlscript/syntax/annotations/Pattern.mls
```diff
@@ -6,27 +6,47 @@ class Box()
...
 
 object do_not_compile
 
 :w
 fun foo(x) = if x is @do_not_compile Box then "yes" else "no"
-//│ ╔══[WARNING] Unknown annotation on this class.
-//│ ║  l.17: 	fun foo(x) = if x is @do_not_compile Box then "yes" else "no"
-//│ ║        	                      ^^^^^^^^^^^^^^
-//│ ╙── Note: `@compile` is only supported on patterns.
+//│ ╔══[WARNING] The annotation is not supported on the class instance pattern.
```
LPTK:
```suggestion
//│ ╔══[WARNING] This annotation is not supported on class instance patterns.
```

hkmc2/shared/src/test/mlscript/syntax/annotations/Pattern.mls
```diff
@@ -6,27 +6,47 @@ class Box()
...
 
 pattern Box' = Box
 
 fun foo(x) = if x is @compile Box' then "yes" else "no"
 
 :w
 fun foo(x) = if x is @do_not_compile Box' then "yes" else "no"
-//│ ╔══[WARNING] Unknown annotation on this pattern.
-//│ ║  l.28: 	fun foo(x) = if x is @do_not_compile Box' then "yes" else "no"
+//│ ╔══[WARNING] The annotation is not supported here.
```
LPTK:
```suggestion
//│ ╔══[WARNING] This annotation is not supported here.
```

hkmc2/shared/src/test/mlscript/syntax/annotations/Pattern.mls
chengluyu: Only annotated constructors are desugared. Because only `Pattern.ClassLike` has a field for annotations.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
chengluyu: I removed the special cases in f1ca85d.

PR #310, opened 2025-05-23
`use[T]` as Summon

FlandiaYingman:
This PR supports the usage of `use[T]` that is equivalent to Scala's `summon[T]`. 

```fs
fun fib(using Yield[Int]) =
  fun inner(a, b) =
    use[Yield].yield(a)
    inner(b, a + b)
  inner(0, 1)
```
(from https://github.com/hkust-taco/mlscript/pull/282#issuecomment-2676133410)


hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -651,6 +651,27 @@ extends Importer:
...
+      // We try to pull out the [T] part from the tree.
+      def go(t: Tree): Tree = t match
+        case App(lhs, rhs) => App(go(lhs), rhs)
+        case InfixApp(lhs, kw, rhs) => InfixApp(go(lhs), kw, rhs)
+        case Sel(lhs, rhs) => Sel(go(lhs), rhs)
+        case SynthSel(lhs, rhs) => SynthSel(go(lhs), rhs)
+        case Jux(lhs, rhs) => Jux(go(lhs), rhs)
+        case tyArgs: Tup =>
+          TermDef(Ins, tyArgs, N)
+      term(go(head))
```
FlandiaYingman: Here is a funny fix of a parsing issue: `use[T].foo` actually gets parsed into `use ([T].foo)`, i.e.,

```fs
:pt
use[T].foo
//│ Parsed tree:
//│ TermDef:
//│   k = Ins
//│   head = Sel:
//│     prefix = Tup of Ls of 
//│       Ident of "T"
//│     name = Ident of "foo"
//│   rhs = N
```

This seems not possible to be fixed at the ParseRule level because `[T].foo` together forms an expression. The funny 🤡 workaround here is to recursively go down to the LHS of the trees, and "pull out" the `[T]` part from the tree.

I think this can somehow be fixed by adding a special case in the parser that only parses an expression surrounded by a pair of square brackets, but I think I wouldn't have time to study into how the parser works because of the _useless_ final exams 🫡

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -917,6 +938,13 @@ extends Importer:
           raise(ErrorReport(msg"Unrecognized definitional assignment left-hand side: ${lhs.describe}"
             -> lhs.toLoc :: Nil)) // TODO BE
           go(sts, Nil, Term.Error :: acc)
+      // A use[T] construct: analogous to Scala's summon[T].
+      case (tree @ TermDef(Ins, _, N)) :: sts =>
+        val res = annotations.foldLeft(term(tree)):
+          case (acc, ann) => Term.Annotated(ann, acc)
+        sts match
+        case Nil => (mkBlk(acc, S(res), hasResult), ctx)
+        case _ => go(sts, Nil, res :: acc)
```
FlandiaYingman: Here's a duplication of the fallback case that handles the trees (as opposed to statements)

```scala
      case (st: Tree) :: sts =>
        // TODO reject plain term statements? Currently, `(1, 2)` is allowed to elaborate (tho it should be rejected in type checking later)
        val res = annotations.foldLeft(term(st)):
          case (acc, ann) => Term.Annotated(ann, acc)
        sts match
        case Nil => (mkBlk(acc, S(res), hasResult), ctx)
        case _ => go(sts, Nil, res :: acc)
```

I think it is not beautiful but seems there is no other way to work around it.

hkmc2/shared/src/test/mlscript/ctx/Summon.mls
```diff
@@ -0,0 +1,71 @@
...
+use Int = 42
+f[Int]
+//│ ╔══[ERROR] Missing instance for contextual parameter of type `T` in this call
+//│ ║  l.62: 	f[Int]
+//│ ║        	^^^^^
+//│ ╟── Required by contextual parameter declaration: 
+//│ ║  l.60: 	fun f[T](using T) = use[T]
+//│ ║        	               ^
+//│ ╙── Expected: T; Available: (Int, Int, Int, Str, Str, Some[Int])
+//│ ═══[RUNTIME ERROR] Error: Function 'f' expected 1 argument but got 0
```
FlandiaYingman: > This doesn't work because this doesn't work. 🤡

I can recall (maybe) at some point we decided not to allow implicit parameters of non-concrete type; or maybe I mixed it up with the "module parameter must be of concrete type" restriction. Anyway I believe it is easy to fix, and I will fix it at some point in the (near) future.

FlandiaYingman (commented):
LGTM as no extra whitespaces introduced 🫡

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: oh no

LPTK:
Thanks! I'll look into the issues you raised when I have more time, probably next week.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
LPTK, line 1: FYI, this crashes when it shouldn't:

```fs
:elt
use[T].foo
//│ ╔══[ERROR] Name not found: T
//│ ║  l.12: 	use[T].foo
//│ ╙──      	    ^
//│ Elaborated tree:
//│ Blk:
//│   stats = Nil
//│   res = Sel:
//│     prefix = Summon:
//│       ty = Error
//│       iargsLs = N
//│     nme = Ident of "foo"
//│     iargsLs = N
//│ ═══[ERROR] Expected a type symbol, got <error>
//│ /!!!\ Uncaught error: java.lang.Exception: Internal Error: unresolved summon: Summon(Error)
```

FlandiaYingman: What should be the correct behavior if the name `T` is not found 🤔

FlandiaYingman: Oh nvm

FlandiaYingman: Fixed

hkmc2/shared/src/test/mlscript/ctx/Summon.mls
FlandiaYingman: Never mind... it was a bug since the very beginning and I didn't realize it until now 🤡😁. I fixed it and the behavior is now

```fs
fun summon[T](using t: T) = t

// should reject because cannot query an instance of unspecified type
summon

// ok
summon[Int]

fun summonSome[T](using t: Some[T]) = t

// ok because Some[T] is not completely unspecified
summonSome

// ok
summonSome[Int]
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: Because I think in the future there would be a rewrite for the parser? So I didn't spend much time on trying to fix the parser

LPTK: Following our discussion, let's change the syntax of used instances to `using T = t`.

FlandiaYingman (commented):
Done!

LPTK (approved):
Great, thanks!

Do we have test cases exercising sunsupported syntaxes like:
```fs
using n: Int
fun foo(using Int = 1)
```
?

FlandiaYingman:
No and now I added to the `BadTypeClasses.mls` file

PR #311, opened 2025-05-24
Contextual Params for Constructors

FlandiaYingman:


FlandiaYingman:
Can't believe the implementation was accepting `using Some[A]` but not `using ..., Some[A]` 🤡

FlandiaYingman:
Some abstraction is probably needed to abstract over type constructors and functions.

FlandiaYingman:
It's more difficult than I expected ☠️

To resolve an implicit parameter, the resolver has to match the arity of parameters and type parameters. This directly rejects a huge number of test cases 🤡 Because of our duplication trick for class parameters (duplicate the parameter into a field so that in the body it can refer to the parameter by referring to the field), I had to do more tricks to find the duplicated symbol and correctly refer to that symbol.

hkmc2/shared/src/test/mlscript/basics/MultiParamListClasses.mls
```diff
@@ -0,0 +1,89 @@
...
+Foo(1, 2)(3)
+//│ = Foo(1, 2)
+
+
+new Foo(1, 2)
+//│ = [function]
+
+:fixme
+new Foo(1, 2)(3)
+//│ ═══[RUNTIME ERROR] TypeError: tmp1 is not a constructor
```
FlandiaYingman: The current JS trick to implement multiple parameter lists for classes does not work for `new` operator.

hkmc2/shared/src/test/mlscript/rp/MatchResult.mls
```diff
@@ -19,7 +19,11 @@ if MatchFailure("oops") is MatchFailure(x) then x
 
 data class MatchResult(x, y, z)
 
+:e
 MatchResult(0)
+//│ ╔══[ERROR] Expected 3 arguments, got 1
+//│ ║  l.23: 	MatchResult(0)
+//│ ╙──      	           ^^^
 //│ = MatchResult(0, undefined, undefined)
```
FlandiaYingman: I am not sure what is this test case testing. I suppose it should be an error?

FlandiaYingman (commented):
The feature should be working. 

But there is a funny problem that references to class at type positions are also resolved. For example,

```cs
class Foo(using val Int)

:rt
val v: Foo = Foo
//│ FAILURE LOCATION: resolveArg (Resolver.scala:776)
//│ ╔══[ERROR] Cannot query instance of type Int for call: 
//│ ║  l.14: 	val v: Foo = Foo
//│ ║        	       ^^^
//│ ╟── Required by contextual parameter declaration: 
//│ ║  l.11: 	class Foo(using val Int)
//│ ║        	                    ^^^
//│ ╙── Missing instance: Expected: Int; Available: ()
//│ FAILURE LOCATION: resolveArg (Resolver.scala:776)
//│ ╔══[ERROR] Cannot query instance of type Int for call: 
//│ ║  l.14: 	val v: Foo = Foo
//│ ║        	             ^^^
//│ ╟── Required by contextual parameter declaration: 
//│ ║  l.11: 	class Foo(using val Int)
//│ ║        	                    ^^^
//│ ╙── Missing instance: Expected: Int; Available: ()
//│ Resolved tree:
//│ Blk:
//│   stats = Ls of 
//│     TermDefinition:
//│       owner = N
//│       k = ImmutVal
//│       sym = member:v
//│       params = Nil
//│       tparams = N
//│       sign = S of Ref{member:Foo}:
//│         sym = member:Foo
//│         iargsLs = S of Ls of 
//│           Tup of Ls of 
//│             Fld:
//│               flags = ()
//│               term = Error
//│               asc = N
//│       body = S of Ref{member:Foo}:
//│         sym = member:Foo
//│         iargsLs = S of Ls of 
//│           Tup of Ls of 
//│             Fld:
//│               flags = ()
//│               term = Error
//│               asc = N
//│       resSym = ‹result of member:v›
//│       flags = ()
//│       modulefulness = Modulefulness of N
//│       annotations = Nil
//│ v = Foo(undefined)
```

and it seems that there is no easy way to differentiate a type `Foo` and a constructor `Foo`. I think we need the disambiguate term that we discussed in a previous meeting to deal with this problem

hkmc2/shared/src/test/mlscript/basics/MultiParamListClasses.mls
LPTK: Actually it does, it's just that our parsing of `new` is currently broken (any fix welcome; in any case this can be done separately).

```fs
(new Foo(1, 2))(3)
//│ = Foo(1, 2)
```

hkmc2/shared/src/test/mlscript/rp/MatchResult.mls
LPTK: Yeah IDK, let's ask @chengluyu :)

LPTK: But I'm guessing it's just reprdocuing the cursed `undefined`s that pop up in such situations.

LPTK:
> But there is a funny problem that references to class at type positions are also resolved.

That's not a blocker, right?

I don't think we need to worry about that for now. Later, we'll add a mode for elaborating types, as they sometimes should be elaborated differently than terms, and then we'll be able to solve this.

hkmc2/shared/src/test/mlscript/basics/MultiParamListClasses.mls
FlandiaYingman: Wow again a parsing problem

FlandiaYingman:
Yes, I don't think it is a pressing problem that has to be fixed right now. I just mention it for tracking and future handling.

hkmc2/shared/src/test/mlscript/basics/MultiParamListClasses.mls
FlandiaYingman: Actually I think it should be an elaboration problem because the parse tree seems correct

```
new Foo(40)(2)
//│ Parsed tree:
//│ New:
//│   body = S of App:
//│     lhs = App:
//│       lhs = Ident of "Foo"
//│       rhs = Tup of Ls of 
//│         IntLit of 40
//│     rhs = Tup of Ls of 
//│       IntLit of 2
//│   rft = N
```

but the elaboration tree is a bit wrong

```
//│ Elaborated tree:
//│ Blk:
//│   stats = Nil
//│   res = New:
//│     cls = App:
//│       lhs = Ref:
//│         sym = member:Foo
//│         iargsLs = N
//│       rhs = Tup of Ls of 
//│         Fld:
//│           flags = ()
//│           term = Lit of IntLit of 40
//│           asc = N
//│       iargsLs = N
//│     args = Ls of 
//│       Lit of IntLit of 2
//│     rft = N
```

I could try to fix it in this PR

hkmc2/shared/src/test/mlscript/basics/Inheritance.mls
```diff
@@ -89,3 +89,27 @@ data class Some[out T](x: T) extends Option[T]
...
+Bazz(42).f
+//│ ═══[RUNTIME ERROR] Error: Access to required field 'f' yielded 'undefined'
+
+:todo
+new Barr(40)(2) with
+  fun f = 43
+//│ ╔══[COMPILATION ERROR] Inheriting class with multiple parameter lists is not supported
+//│ ║  l.108: 	new Barr(40)(2) with
+//│ ╙──       	    ^^^^
+//│ = [function]
```
FlandiaYingman: I found that the current implementation for curried classes does not support inheritance (I explicitly reject it for now and added some cases for tracking).

FlandiaYingman (commented):
Now the `new` operator supports multiple argument lists

hkmc2/shared/src/test/mlscript/basics/Inheritance.mls
```diff
@@ -89,3 +89,27 @@ data class Some[out T](x: T) extends Option[T]
 //│ = [true, true, false, false]
 
 
+:todo
+data class Barr(x: Int)(y: Int) with
```
LPTK, line 92: Why is there a TODO here?

hkmc2/shared/src/test/mlscript/basics/MultiParamListClasses.mls
```diff
@@ -0,0 +1,88 @@
...
+//│ ║  l.38: 	  constructor(x, y)(z)
+//│ ╙──      	                 ^
+//│ ╔══[ERROR] Name not found: z
+//│ ║  l.38: 	  constructor(x, y)(z)
+//│ ╙──      	                    ^
+
+
+class Foo(x, y)(using z: Int)
+
+// * it to be understood as
```
LPTK:
```suggestion
// class Foo(using z: Int)
// * is to be understood as
```

hkmc2/shared/src/test/mlscript/basics/MultiParamListClasses.mls
```diff
@@ -0,0 +1,88 @@
...
+fun Foo(using z: Int) = new Foo0 !> res =>
+  set res.z = z
+
+using Int = 123
+
+Foo
+//│ = Foo0
+
+// class Foo(x, y)(using z)
+// * it to be understood as
```
LPTK:
```suggestion
// * is to be understood as
```

hkmc2/shared/src/test/mlscript/basics/MultiParamListClasses.mls
```diff
@@ -0,0 +1,88 @@
...
+Foo(1, 2)
+//│ = Foo0
+
+:sjs
+let f = Foo // should compile to `(x, y) => Foo(x, y)(use[Int])`
+//│ JS (unsanitized):
+//│ let f; f = Foo9;
+//│ f = [function Foo]
+
+// FIXME: unsoundness
```
LPTK, line 83: How about this FIXME? Is it tracked somewhere? Will we address it soon?

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -526,7 +593,12 @@ class Resolver(tl: TraceLogger)
           // Other: spread arguments
           case _ => (false, 0)
           
+          log(s"${defn.sym}, ${state.matchResultClsSymbol.defn}, ${state.matchFailureClsSymbol.defn}")
           (ps.paramCountUB, argCountUB) match
+            // Don't check the internal classes MatchResult and MatchFailure
+            case _ if state.matchResultClsSymbol.defn.exists(cdef => defn.sym == cdef.bsym) => ()
+            case _ if state.matchFailureClsSymbol.defn.exists(cdef => defn.sym == cdef.bsym) => ()
```
```suggestion
            case _ if state.matchResultClsSymbol.defn.exists(cdef => defn.sym is cdef.bsym) => ()
            case _ if state.matchFailureClsSymbol.defn.exists(cdef => defn.sym is cdef.bsym) => ()
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -526,7 +593,12 @@ class Resolver(tl: TraceLogger)
           // Other: spread arguments
           case _ => (false, 0)
           
+          log(s"${defn.sym}, ${state.matchResultClsSymbol.defn}, ${state.matchFailureClsSymbol.defn}")
           (ps.paramCountUB, argCountUB) match
+            // Don't check the internal classes MatchResult and MatchFailure
```
LPTK: What's the reason for this special-case? It should be documented here. (Or, better yet, just make the generated code use the classes correctly.)

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -474,7 +538,10 @@ class Resolver(tl: TraceLogger)
           case _ => N
         val newICtx = (tparams, targs) match
           case (S(tparams), S(targs)) =>
-            if tparams.length != targs.length then
+            // Don't check the number of type arguments for class-like
+            // definitions because they are already checked in the
+            // VarianceTraverser.
```
LPTK: `VarianceTraverser` is a legacy piece of junk. It will be ported/removed soon. Can we just comment its use for now and do things properly here, instead of adding special-cases?

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -350,9 +386,38 @@ class Resolver(tl: TraceLogger)
...
-      )(using resolveCtxParams(pss)))
+      )(using withCtxParams(pss)))
       annotations.flatMap(_.subTerms).foreach(traverse(_, expect = NonModule(N)))
     
+    def traverseClassLikeDef(cld: ClassLikeDef) =
+      def withCtxParams: ICtx = cld.paramsOpt
+        .filter(_.flags.ctx)
+        .foldLeft(ictx): (ictx, ps) => 
+          ps.params.foldLeft(ictx): (ictx, p) => 
+            p.sign match
```
LPTK, line 409: Can you pls explain what this does? And add an examples.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -219,6 +253,8 @@ class Resolver(tl: TraceLogger)
   import tl.*
   import Resolver.ICtx
   import Resolver.ictx
+  import Resolver.CallableDefinition
+  import Resolver.callableDefn
```
```suggestion
  import Resolver.*
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1087,22 +1083,29 @@ extends Importer:
               
           val ctxWithFields = ctx
             .withMembers(
-              fields.fold(Nil)(_.collect:
-                case f: TermDefinition => f.sym.nme -> f.sym // class fields
-              ),
+              fields.foldRight(Nil):
```
LPTK: Use `fields.iterator.collect` instead?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1087,22 +1083,29 @@ extends Importer:
...
+                  // class fields
+                  f.sym.nme -> f.sym :: acc
+                case (_, acc) =>
+                  acc
+              ,
               ctx.outer.inner
-            ) ++ fields.fold(Nil)(_.collect:
-              case d: LetDecl => d.sym.nme -> d.sym // class params
-            )
+            ) ++ fields.foldRight(Nil):
```
LPTK: Use `fields.iterator.collect` instead?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -551,17 +557,34 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
+                Assign(ts, Call(inner, args)(true, true), acc(Value.Ref(ts)))
+          val ts = TempSymbol(N)
+          Assign(ts, Instantiate(sr, asr), z(Value.Ref(ts)))
       
-    case New(cls, as, S((isym, rft))) =>
+    case New(cls, ass, S((isym, rft))) =>
       subTerm(cls): clsp =>
         val sym = new BlockMemberSymbol(isym.name, Nil)
         val (mtds, publicFlds, privateFlds, ctor) = gatherMembers(rft)
+        if ass.length > 1 then
```
LPTK: Why is this error logic duplicated witht the one above?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -177,12 +177,18 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
         case S(ext) =>
           assert(k isnt syntax.Mod) // modules can't extend things and can't have super calls
           subTerm(ext.cls): clsp =>
+            if ext.argss.length > 1 then 
+              raise:
+                ErrorReport(
+                  msg"Inheriting class with multiple parameter lists is not supported" -> ext.toLoc :: Nil,
```
```suggestion
                  msg"Extending a class with multiple parameter lists is not supported" -> ext.toLoc :: Nil,
```

hkmc2/shared/src/test/mlscript/ctx/ClassCtxParams.mls
```diff
@@ -1,40 +1,76 @@
...
+//│ ╙── Missing instance: Expected: Str; Available: (Int)
+//│ = Bar()
 
-using Int = 1
 
-:re
-foo()
-//│ ═══[RUNTIME ERROR] TypeError: tmp is not a function
 
+fun summon(using i: Int) = i
```
LPTK, line 58: This reminds me that in Scala, there is no special term for summon, it's a normal definition. Could we make `use` not a keyword and simply define it in Predef?

LPTK (commented):
Nice work! I have a few remarks below.

hkmc2/shared/src/test/mlscript/rp/MatchResult.mls
chengluyu: This is testing whether the compiled `unapply` method can use the correct `MatchResult` when the `MatchResult` in `Runtime` is shadowed. But I might have gotten the number of arguments wrong. You can change the parameters of the `MatchResult` above to just one.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: Internally, the compiler generates constructor calls to `MatchResult` and `MatchFailture` with zero or one argument, but the definition of `MatchResult` and `MatchFailure` has only one parameter.

```
pattern Zero = "0"
//│ FAILURE LOCATION: newDefn (Resolver.scala:603)
//│ ═══[ERROR] Expected 1 arguments, got 0
//│ FAILURE LOCATION: newDefn (Resolver.scala:603)
//│ ═══[ERROR] Expected 1 arguments, got 0
```

I will try to correct the generated code later today.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: If `VarianceTraverser` would be ported or removed later then we could just remove this funny special case. I put the special case here intended to mute the duplicate errors reported.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
chengluyu: Argh, I can help you just fix the misuse.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
FlandiaYingman: Oh because the two pieces of lowering logic are just identical in the beginning. There was a comment `TODO dedup with New case` later

```
            val pctor = // TODO dedup with New case
              plainArgs(ext.argss.headOr(Nil)): args =>
                Return(Call(Value.Ref(State.builtinOpsMap("super")), args)(true, true), implct = true)
```

hkmc2/shared/src/test/mlscript/ctx/ClassCtxParams.mls
FlandiaYingman: Oh yes, before we changed the syntax it was impossible because `use` was a keyword. Now yes, I think we can change it to a function.

We can do it in a separate PR.

hkmc2/shared/src/test/mlscript/basics/Inheritance.mls
FlandiaYingman: Oh no, this shouldn't be a TODO. The cases after it should be TODO

hkmc2/shared/src/test/mlscript/basics/MultiParamListClasses.mls
FlandiaYingman: Yes I can address it in a later PR. In fact I think the abstraction I added for constructor implicit parameters support can naturally address it.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
chengluyu: Done.

hkmc2/shared/src/test/mlscript/rp/MatchResult.mls
chengluyu: I fixed this.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/DesugaringBase.scala
```diff
@@ -77,7 +77,7 @@ trait DesugaringBase(using Ctx, State):
     app(matchResultClass, tup(fld(captures)), "result of `MatchResult`")
     
   protected final def makeMatchFailure =
-    app(matchFailureClass, tup(), "result of `MatchFailure`")
+    app(matchFailureClass, tup(fld(Term.Lit(UnitLit(true)))), "result of `MatchFailure`")
```
FlandiaYingman, line 80: wow i didn't know it could be fixed by simply adding a default argument

thanks Luyu

chengluyu: This argument is reserved for error messages in case the match fails. But the logic to generate error messages has been implemented yet, so let's just put a `null` here for now.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
LPTK: You need to understand that every special case you add is more maintenance burden on us down the line. This special case you add now will cause confusion when we later find out, probably long after we remove the old VarianceChecker, that the resolver is _**unexpectedly not checking some cases**_!!

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
LPTK: Ok, in that case it's good practice to not make things worse but rather leave the place in a better state than you found it (ie dedup this code 😉)

hkmc2/shared/src/test/mlscript/ctx/ClassCtxParams.mls
LPTK: Ok. Do we have a todo list or teaching issue for things related to contextual parameters?

hkmc2/shared/src/test/mlscript/basics/MultiParamListClasses.mls
LPTK: Cool! Please don't forget to add this to the tracking issue.

hkmc2/shared/src/test/mlscript/ctx/ClassCtxParams.mls
FlandiaYingman: Yes we have one https://github.com/hkust-taco/mlscript/issues/272

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: Luyu committed a small fix so we don't need this piece of special cases.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: Wow I didn't notice that this was just a simple `collect`! But why `fields.iterator.collect`? I think it's just `fields.collect`?

FlandiaYingman: Changed to `fields.collect`

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
FlandiaYingman: Deduped

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: Removed, and maybe _expectedly_, it is already _unexpectedly_ not checking some cases, which leads to some extra errors 🤔🤡 I fixed them and pointed them out as separate comments below.

hkmc2/shared/src/test/mlscript/decls/Prelude.mls
```diff
@@ -7,7 +7,7 @@ declare type Nothing
 declare type untyped
 
 declare class Object
-declare class Array
+declare class Array[T]
```
FlandiaYingman, line 10: I added a type parameter for `Array` so that we can specify the type of array.

hkmc2/shared/src/test/mlscript-compile/Option.mls
```diff
@@ -8,10 +8,11 @@ open Predef
 
 module Option with ...
 
+// See https://github.com/hkust-taco/mlscript/pull/295#discussion_r2029692230
 // TODO support type/module overloading
 type T[A]
 
-data class Some(value)
+data class Some[T](value)
```
FlandiaYingman, line 15: I added a type parameter for `Some` so that we can specify the type of `Some`.

hkmc2/shared/src/test/mlscript-compile/Term.mls
```diff
@@ -45,7 +45,7 @@ class Pattern with
     ClassLike(val sym: Symbol, val trm: Term, val parameters: Array[Symbol] | null)
     Synonym(val symbol: Symbol, val patternArguments: Array[DebrujinSplit])
     Tuple(val size: Int, val inf: Bool)
-    Record(val entities: Array[(Str, Symbol)])
+    Record(val entities: Array[[Str, Symbol]])
```
FlandiaYingman, line 48: I think tuple types in MLscript use brackets? Parentheses surrounded tuples are parsed into `Block` which surprised me.

FlandiaYingman (commented):
Done! Some errors are duplicated because both `VarianceTraverser` and `Resolver` check them. We might want to remove `VarianceTraverser` in the future.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: `fields.collect` will allocate a new list in memory, where every element is wrapped in a separate cons cell. Using an iterator means we'll only allocate an iterator object (which uses mutation internally).

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1080,23 +1080,14 @@ extends Importer:
...
-              case (_, acc) =>
-                acc
+          
+          val valParams = fields.collect:
+            case f: TermDefinition =>
+              f.sym.nme -> f.sym
+          val params = fields.collect:
+            case (f: LetDecl) =>
+              f.sym.nme -> f.sym
+          val ctxWithFields = ctx.withMembers(valParams,ctx.outer.inner) ++ params
```
LPTK: However, binding iterators to variables is error-prone, as they have state, so better keep them hidden from the current scope:
```suggestion
          val ctxWithFields =
            val valParams = fields.collect:
              case f: TermDefinition =>
                f.sym.nme -> f.sym
            val params = fields.collect:
              case (f: LetDecl) =>
                f.sym.nme -> f.sym
            ctx.withMembers(valParams, ctx.outer.inner) ++ params
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -354,9 +354,17 @@ class Resolver(tl: TraceLogger)
...
   def resolveDefn(defn: Definition)(using ICtx): ICtx =
   trace(s"Resolving definition: $defn"):
-    def traverseTermDef(tdf: TermDefinition) = tdf match
-    case TermDefinition(_, _, _, pss, tps, sign, body, _, TermDefFlags(isMethod), modulefulness, annotations) =>
-      def withCtxParams(pss: Ls[ParamList]): ICtx = pss
+    def traverseTermDef(tdf: TermDefinition) =
+      val TermDefinition(_, _, _, 
+        pss, tps, sign, body, 
+        _, TermDefFlags(isMethod), modulefulness, annotations
+      ) = tdf
```
LPTK: Ouch. Why not just access these fields as `tdf.X`? Doing this makes further changes to `TermDefinition` (such as adding or removing a field) more painful. Underscoreitis.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -396,12 +409,18 @@ class Resolver(tl: TraceLogger)
...
-                  // elaborator in order for the lowering stage to
-                  // correctly lower the program.
+                  // elaborator. This is because (the symbol of) the
+                  // class parameters are only accessible in the
+                  // constructor of the class-like definition in the
+                  // lowering stage. So, in the elaborator we duplicate
+                  // the class parametes into a private/public fields
+                  // (depending on if the parameter is modified by `val`
+                  // or not). We have to use the field symbols instead
+                  // of the parameter symbols here.
```
LPTK: Ok, I see what this is doing now. But this is a very ugly workaround; surely there's a better way of doing this... How about we just store the relevant information in the appropriate place, so we don't have to go fishing for it later?

LPTK: I.e., why doesn't the elaborator store the field symbol of the parameter to `Param` directly?

hkmc2/shared/src/test/mlscript-compile/Term.mls
LPTK: Huh, good catch! Cc @NeilKleistGao

hkmc2/shared/src/test/mlscript/basics/Inheritance.mls
```diff
@@ -89,16 +89,15 @@ data class Some[out T](x: T) extends Option[T]
...
 
 :todo
 data class Bazz(z: Int) extends Bar(z % 10)(z / 10) with
   fun g = z
-//│ ╔══[COMPILATION ERROR] Inheriting class with multiple parameter lists is not supported
-//│ ║  l.97: 	data class Bazz(z: Int) extends Bar(z % 10)(z / 10) with
-//│ ╙──      	                                ^^^^^^^^^^^^^^^^^^^
+//│ ╔══[COMPILATION ERROR] Extending a class with multiple parameter lists is not supported
+//│ ║  l.96: 	data class Bazz(z: Int) extends Bar(z % 10)(z / 10) with
+//│ ╙──      	                                    ^^^^^^^^^^^^^^
```
LPTK: I wonder why this location changed 🤔

hkmc2/shared/src/test/mlscript/ctx/ClassCtxParams.mls
```diff
@@ -76,3 +76,17 @@ Foo.bar
...
+
+:e
+Foo
+//│ ╔══[ERROR] Cannot query instance of type Str for call: 
+//│ ║  l.84: 	Foo
+//│ ║        	^^^
+//│ ╟── Required by contextual parameter declaration: 
+//│ ║  l.79: 	class Foo(using Int)(using Str) with
+//│ ║        	                           ^^^
+//│ ╙── Missing instance: Expected: Str; Available: (Int, Int)
```
LPTK: This `(Int, Int)` would parse into a block, which surprised me 🤡

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: I intended to write so because here we are traversing all subterms. If we make further changes to `TermDefinition`, like if we added some new subterms to `TermDefinition`, we would likely want to change here too!

hkmc2/shared/src/test/mlscript/basics/Inheritance.mls
FlandiaYingman: I deduped the logic for extending classes and `new`ing anonymous classes in the lowering stage, and I found it would be more natural to report the location of param lists

hkmc2/shared/src/test/mlscript/ctx/ClassCtxParams.mls
FlandiaYingman: Wow

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
LPTK: Hmm fair enough. But then it would be good to at least name these underscore (eg use `_k: Kind` instead of `_`), just so we can be sure we're not missing anything important.

hkmc2/shared/src/test/mlscript/ctx/ClassCtxParams.mls
LPTK: Why not properly display these without making them look like a tuple and also don't show duplicates? 😬

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
```diff
@@ -290,7 +288,7 @@ object PlainTup:
 object Apps:
   def apply(t: Tree, as: Ls[Tup]): Tree = as match
     case Nil => t
-    case arg :: args => App(Apps(t, args), arg)
+    case arg :: args => Apps(App(t, arg), args)
```
FlandiaYingman, line 291: It seems that the order of argument lists is wrong and I fixed it here

hkmc2/shared/src/test/mlscript/ctx/Summon.mls
```diff
@@ -52,9 +46,9 @@ f
 :e
 fun f(using Int) = use[Num]
 //│ ╔══[ERROR] Cannot query instance for use-expression of type Num
-//│ ║  l.53: 	fun f(using Int) = use[Num]
+//│ ║  l.47: 	fun f(using Int) = use[Num]
 //│ ║        	                       ^^^
-//│ ╙── Missing instance: Expected: Num; Available: (Int, Int, Int, Str, Str)
+//│ ╙── Missing instance: Expected: Num; Available: [Int, Str]
```
FlandiaYingman: I changed the `()` to `[]` and removed the duplicates. Is it what you mean?

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -631,6 +637,8 @@ final case class TyParam(flags: FldFlags, vce: Opt[Bool], sym: VarSymbol) extend
 
 final case class Param(flags: FldFlags, sym: VarSymbol, sign: Opt[Term], modulefulness: Modulefulness) 
 extends Declaration with AutoLocated:
+  var fldSym: Opt[FieldSymbol] = N
+  
   def subTerms: Ls[Term] = sign.toList
   override protected def children: List[Located] = sym :: sign.toList
   def showDbg: Str = flags.toString + sym + sign.fold("")(": " + _.showDbg)
```
FlandiaYingman: I used a `var` to store the field symbols instead of fishing them later. I didn't make it a case class param because usually the class symbol etc. that are required to create the field symbol are not available at the time we create a `Param`. So it is more natural to make it a `var` and set it afterwards (just like what we did for `sym.defn`).

FlandiaYingman (commented):
Fixed

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -95,7 +95,7 @@ object Resolver:
     def showEnv: Str =
       iEnv.values
         .flatMap(_.map((typ, instance) => s"${typ.show}"))
-        .mkString("(", ", ", ")")
+        .mkString("[", ", ", "]")
```
LPTK: Why the insistence on brackets? It looks like an array now, but is not an array. Wouldn't it look better as:
```suggestion
        .mkString(", ")
```

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
LPTK: Wow! Good catch. This was copilot-generated slop 😬

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
LPTK: Though we should then also add a special-case for the empty list.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1087,21 +1089,26 @@ extends Importer:
                   p.modulefulness,
                   Nil
                 )
+                p.fldSym = S(fsym)
```
LPTK, line 1093:
```suggestion
                assert(p.fldSym.isEmpty)
                p.fldSym = S(fsym)
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -637,6 +637,8 @@ final case class TyParam(flags: FldFlags, vce: Opt[Bool], sym: VarSymbol) extend
 
 final case class Param(flags: FldFlags, sym: VarSymbol, sign: Opt[Term], modulefulness: Modulefulness) 
 extends Declaration with AutoLocated:
+  var fldSym: Opt[FieldSymbol] = N
+
```
```suggestion
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
LPTK: Ok, that's already much better.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
FlandiaYingman: Wow

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: Applied. Plus I made it output `<no available>` when there is no instance:

```cs
M.f
//│ ╔══[ERROR] Cannot query instance of type Int for call: 
//│ ║  l.13: 	M.f
//│ ║        	^^^
//│ ╟── Required by contextual parameter declaration: 
//│ ║  l.10: 	  fun f(using Int)
//│ ║        	              ^^^
//│ ╙── Missing instance: Expected: Int; Available: <no available>
```

LPTK: Oops but it should be  `‹none available›`, not `<no available>`!

hkmc2/shared/src/test/mlscript/backlog/ToTriage.mls
```diff
@@ -352,8 +352,11 @@ mkStr of ... // hello
 
 data class Foo(x, y)
 
-// TODO missing sanity check
+:e
 Foo(1)
+//│ ╔══[ERROR] Expected 2 arguments, got 1
+//│ ║  l.356: 	Foo(1)
+//│ ╙──       	   ^^^
```
LPTK: The intent of this test was to show the lack of runtime sanity checks. So you should revert it to something like
```suggestion
// TODO missing sanity check
:re
id(Foo)(1)
```

hkmc2/shared/src/test/mlscript/basics/Inheritance.mls
LPTK: But the message says "a class with multiple parameter lists", so it should show _a class with multiple parameter lists_, not just _multiple parameter lists_.

LPTK (approved):
Otherwise LGTM! Wow!

hkmc2/shared/src/test/mlscript/basics/Inheritance.mls
FlandiaYingman: Hmm okay

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -97,7 +97,7 @@ object Resolver:
         .flatMap(_.map((typ, instance) => s"${typ.show}"))
         .toList
         .distinct
-        .mkStringOr(", ", els = "<no available>")
+        .mkStringOr(", ", els = "<none available>")
```
LPTK: You didn't copy my unicode characters 😢

FlandiaYingman: What they are different? My bad 😢

LPTK: Can you pls just apply my change (with the unicode characters) so we can finally merge this PR?

FlandiaYingman: I did the change but I just forgot to push it 🤥 my bad

PR #309, opened 2025-05-23
Merge InvalML changes, with temporary support for ADTs

NeilKleistGao:
This is a temporary solution. Related code will be rewritten or removed after we truly support ADTs correctly.

PR #313, opened 2025-06-10
Fix parsing of `new` and refactor parsing of `with` and other things

LPTK:
`new` is a strange keyword: it has a very high precedence that sits between that of selection and that of application. Indeed, `new Foo().bar` should parse as `(new Foo()).bar`, not `new (Foo().bar)`, but `new Foo.Bar` should parse as `new (Foo.Bar)`.

I just wanted to fix the precedence of `new`, but this lead to a formidable yak shave. Notably ended up having to make `with` parse as an infix keyword (because of the `new ... with ...` syntax and the incomparable precedences) and having to temporarily refactor the `constructor` desugaring hack (which will be changed again by https://github.com/hkust-taco/mlscript/pull/309).

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
@@ -145,8 +145,8 @@ object Elaborator:
           case huh => wat(huh)
         protected def assumeObject(nme: Str): BlockMemberSymbol =
           module.tree.definedSymbols.get(nme).getOrElse:
-            throw new NoSuchElementException:
-              s"builtin module symbol source.$nme. we have"
+            throw new NoSuchElementException(
+              s"builtin module symbol source.$nme")
       object source extends VirtualModule(assumeBuiltinMod("source")):
         val line = assumeObject("line")
         val name = assumeObject("name")
@@ -202,17 +202,18 @@ object Elaborator:
     val importSymbol = new VarSymbol(syntax.Tree.Ident("import"))
     val runtimeSymbol = TempSymbol(N, "runtime")
     val termSymbol = TempSymbol(N, "Term")
-    val effectSigSymbol = ClassSymbol(TypeDef(syntax.Cls, Dummy, N, N), Ident("EffectSig"))
+    val effectSigSymbol = ClassSymbol(DummyTypeDef(syntax.Cls), Ident("EffectSig"))
     val nonLocalRetHandlerTrm =
       val id = new Ident("NonLocalReturn")
-      val sym = ClassSymbol(TypeDef(syntax.Cls, Dummy, N, N), id)
+      val sym = ClassSymbol(DummyTypeDef(syntax.Cls), id)
       Term.Sel(runtimeSymbol.ref(), id)(S(sym))
     val nonLocalRet =
       val id = new Ident("ret")
       BlockMemberSymbol(id.name, Nil, true)
     val matchResultClsSymbol =
       val id = new Ident("MatchResult")
-      val td = TypeDef(syntax.Cls, App(id, Tup(Ident("captures") :: Nil)), N, N)
+      // val td = DummyTypeDef(syntax.Cls)
+      val td = TypeDef(syntax.Cls, App(id, Tup(Ident("captures") :: Nil)), N)
       val cs = ClassSymbol(td, id)
       val flag = FldFlags.empty.copy(value = true)
       val ps = PlainParamList(Param(flag, VarSymbol(Ident("captures")), N, Modulefulness(N)(false)) :: Nil)
@@ -221,7 +222,7 @@ object Elaborator:
       cs
     val matchFailureClsSymbol =
       val id = new Ident("MatchFailure")
-      val td = TypeDef(syntax.Cls, App(id, Tup(Ident("errors") :: Nil)), N, N)
+      val td = DummyTypeDef(syntax.Cls)
       val cs = ClassSymbol(td, id)
       val flag = FldFlags.empty.copy(value = true)
       val ps = PlainParamList(Param(flag, VarSymbol(Ident("errors")), N, Modulefulness(N)(false)) :: Nil)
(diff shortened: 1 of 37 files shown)


FlandiaYingman:
> `new Foo().bar` should parse as `(new Foo()).bar`

wow, i thought in other classic languages like C++ and Java they require explicit parentheses here

LPTK:
Wow are you new to programming? 😛

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
```diff
@@ -252,14 +261,14 @@ enum Tree extends AutoLocated:
       case _ => S(N, Ident(""), S(inner))
   
   def isModuleModifier: Bool = this match
-    case Tree.TypeDef(Mod, _, N, N) => true
+    case td @ Tree.TypeDef(Mod, _, rhs) => rhs.isEmpty && td.extension.isEmpty && td.withPart.isEmpty
```
FlandiaYingman, line 264: Why not

```suggestion
    case td @ Tree.TypeDef(Mod, _, N) => td.extension.isEmpty && td.withPart.isEmpty
```

hkmc2/shared/src/test/mlscript/backlog/Lifter.mls
```diff
@@ -160,7 +160,13 @@ fun foo(x, y) =
...
 //│ ╔══[COMPILATION ERROR] No definition found in scope for 'M'
 //│ ║  l.154: 	  module M with
-//│ ╙──       	         ^
+//│ ║         	         ^^^^^^
+//│ ║  l.155: 	    fun foo() =
+//│ ║         	^^^^^^^^^^^^^^^
+//│ ║  l.156: 	      set y = 2
+//│ ║         	^^^^^^^^^^^^^^^
+//│ ║  l.157: 	      x + y
+//│ ╙──       	^^^^^^^^^^^
```
FlandiaYingman, line 169: Do you think the location is too... long? 🤔

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -202,17 +202,18 @@ object Elaborator:
...
+      val sym = ClassSymbol(DummyTypeDef(syntax.Cls), id)
       Term.Sel(runtimeSymbol.ref(), id)(S(sym))
     val nonLocalRet =
       val id = new Ident("ret")
       BlockMemberSymbol(id.name, Nil, true)
     val matchResultClsSymbol =
       val id = new Ident("MatchResult")
-      val td = TypeDef(syntax.Cls, App(id, Tup(Ident("captures") :: Nil)), N, N)
+      // val td = DummyTypeDef(syntax.Cls)
+      val td = TypeDef(syntax.Cls, App(id, Tup(Ident("captures") :: Nil)), N)
```
FlandiaYingman, line 216:
```suggestion
      val td = TypeDef(syntax.Cls, App(id, Tup(Ident("captures") :: Nil)), N)
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -585,7 +586,10 @@ extends Importer:
       )
     case tree @ Tup(fields) =>
       Term.Tup(fields.map(fld(_)))(tree)
-    case New(body, rfto) => // TODO handle Under
+    // case New(c, rfto) =>
+    //   assert(rfto.isEmpty)
+    //   Term.New(cls(subterm(c), inAppPrefix = inAppPrefix), params.map(subterm(_)), bodo).withLocOf(tree)
+    case ProperNew(body, rfto) => // TODO handle Under
```
FlandiaYingman, line 592:
```suggestion
    case ProperNew(body, rfto) => // TODO handle Under
```

hkmc2/shared/src/main/scala/hkmc2/syntax/Keyword.scala
```diff
@@ -131,19 +133,27 @@ object Keyword:
...
-  val `=>` = Keyword("=>", S(_curPrec + charPrecList.length), eqPrec)
+  val `=>` = Keyword("=>", S(maxPrec.get + charPrecList.length), eqPrec)
+  
+  // * `new` is a strange keyword:
+  // * it has a very high precedence that sits between that of selection and that of application.
+  // * Indeed, `new Foo().bar` should parse as `(new Foo()).bar`, not `new (Foo().bar)`,
+  // * but `new Foo.Bar` should parse as `new (Foo.Bar)`.
+  val newRightPrec = S(maxPrec.get + charPrecList.length - 1)
+  // * ^ maxPrec.get + charPrecList.length is the precedence of selection
+  val `new` = Keyword("new", N, newRightPrec)
```
FlandiaYingman, line 144: Wow, how does this mysterious precedence definition make it possible to parse `new` before application but after selection 🤔

FlandiaYingman (approved):
Wow amazing. I just have some questions about the new new parsing strategy

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
LPTK: wow

hkmc2/shared/src/test/mlscript/backlog/Lifter.mls
LPTK: Yeah, in other prototypes we had logic to abbreviate long locations. We should port that over at some point.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: Actually I think this is good now:
```suggestion
    case ProperNew(body, rfto) =>
```

hkmc2/shared/src/main/scala/hkmc2/syntax/Keyword.scala
LPTK: This logic could be improved for clarity, for sure.

Chec out how `charPrecList` reserves a slot for `new` that sits between the slot for `"."` and that for applications.

PR #315, opened 2025-06-24
Support explicitly specifying contextual arguments

FlandiaYingman:
I created a new term form `CtxTup`, which represents an argument tuple prefixed by `using`, e.g., 

```cs
foo(using 42)
   ^^^^^^^^^^
```

If the resolver sees a `App(base, iargs: CtxTup)`, it skips the implicit argument resolution of `base`. It checks the arity of the implicit arguments against the parameter declarations, *but it doesn't check if the types of the arguments match the declarations*.

After resolution, the compiler just compiles `CtxTup` to regular `Tup`, which essentially provides the implicit arguments to the function.

I tried to add a flag to `Tup` to indicate if the tuple is prefixed by `using` rather than using `CtxTup`, but it appears that there are a huge number of places that I have to update. So I gave up and just created `CtxTup`.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
@@ -467,18 +467,28 @@ class Resolver(tl: TraceLogger)
     *    the semantic of the term, so it has to done before the symbol
     *    resolution.
     *
+    * @param inCtxPrefix if true, the currently resolving term is the
+    * prefix of an App where the implicit arguments are explicitly
+    * specified, e.g., `f(using 42)`. The implicit arguments should be
+    * resolved on the the App `f(using 42)`, but not on the base of the
+    * App `f`.
     * @param inTyPrefix if true, the currently resolving term is the
-    * prefix of an TyApp which the implicit arguments shouldn't be
-    * resolved on it, but on the TyApp instead.
+    * prefix of an TyApp, e.g., `f[Int]`. The implicit arguments should
+    * be resolved on the the TyApp `f[Int]`, but not on the base of the
+    * TyApp `f`.
     */
-  def resolve(t: Resolvable, inTyPrefix: Bool = false)(using ICtx): (Opt[CallableDefinition], ICtx) =
+  def resolve(t: Resolvable, inCtxPrefix: Bool = false, inTyPrefix: Bool = false)(using ICtx): (Opt[CallableDefinition], ICtx) =
   trace[(Opt[CallableDefinition], ICtx)](s"Resolving resolvable term: ${t}, (inPrefix = ${inTyPrefix})", _ => s"~> ${t}"):
     // Resolve the sub-resolvable-terms of the term. 
     val (defn, newICtx1) = t match
       // Note: the arguments of the App are traversed later because the
       // definition is required.
-      case Term.App(lhs: Resolvable, _) =>
-        val result = resolve(lhs)
+      case Term.App(lhs: Resolvable, args) =>
+        val result = args match
+          case t @ Term.CtxTup(_) => 
+            resolve(lhs, inCtxPrefix = true, inTyPrefix = inTyPrefix)
+          case _ => 
+            resolve(lhs, inCtxPrefix = inCtxPrefix, inTyPrefix = inTyPrefix)
         resolveSymbol(t)
         result
       case Term.App(lhs, _) =>
@@ -547,7 +557,7 @@ class Resolver(tl: TraceLogger)
       //
       // For example: In `fun f(a)(b) = 42`, f's definition should
       // indicate that it accepts two argument lists; f(42)'s definition
-      // should indicate that it accepts one argument list; f(42, 43)'s
+      // should indicate that it accepts one argument list; f(42)(43)'s
       // definition should indicate that it accepts zero argument lists.
       //
       // Currently, only parameters are processed for these new term
@@ -559,18 +569,22 @@ class Resolver(tl: TraceLogger)
         case S(defn @ CallableDefinition(params = ps :: pss)) =>
           val (argCountUB, argCountLB) = as match
           // Tup: regular arguments
-          case tup: Term.Tup => (
-            !tup.fields.exists(_.isInstanceOf[Spd]),
-            tup.fields.map:
-              case Fld(asc = S(_)) => 0
-              case _: Fld => 1
-              case _: Spd => 0
-            .sum +
-            tup.fields.exists:
-              case Fld(asc = S(_)) => true
-              case _ => false
-            .into(if _ then 1 else 0)
-          )
+          case tup: (Term.Tup | Term.CtxTup) => 
+            val fields = tup match
+              case Term.Tup(fs) => fs
+              case Term.CtxTup(fs) => fs
+            (
+              !fields.exists(_.isInstanceOf[Spd]),
+              fields.map:
+                case Fld(asc = S(_)) => 0
+                case _: Fld => 1
+                case _: Spd => 0
+              .sum +
+              fields.exists:
+                case Fld(asc = S(_)) => true
+                case _ => false
+              .into(if _ then 1 else 0)
+            )
           // Other: spread arguments
           case _ => (false, 0)
           
@@ -633,6 +647,7 @@ class Resolver(tl: TraceLogger)
           
           val args = as match
             case Term.Tup(args) => args
+            case Term.CtxTup(args) => args
             case spd => Spd(true, spd) :: Nil
           
           // The lhs of the App is already traversed by the recursive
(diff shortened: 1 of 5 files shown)


LPTK:
> _it doesn't check if the types of the arguments match the declarations_

Why not?

FlandiaYingman:
Do we want to do that? I thought it would involve something like type inference in complex cases, which is what we don't want; but indeed we can do it in simple cases where types exactly match (align with the instance resolution rules).

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -559,18 +569,22 @@ class Resolver(tl: TraceLogger)
         case S(defn @ CallableDefinition(params = ps :: pss)) =>
           val (argCountUB, argCountLB) = as match
           // Tup: regular arguments
```
LPTK: there should be an indentation here, after the `match`. It took me a while to figure out what was going on here, because I'd missed the line above.

LPTK:
Ah I see what you mean, never mind. This will be checked by the type checker.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -559,18 +569,22 @@ class Resolver(tl: TraceLogger)
...
+            val fields = tup match
+              case Term.Tup(fs) => fs
+              case Term.CtxTup(fs) => fs
+            (
+              !fields.exists(_.isInstanceOf[Spd]),
+              fields.map:
+                case Fld(asc = S(_)) => 0
+                case _: Fld => 1
+                case _: Spd => 0
+              .sum +
```
LPTK: Instead of `.map .sum`, why not use `.count`?

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -559,18 +569,22 @@ class Resolver(tl: TraceLogger)
...
+          case tup: (Term.Tup | Term.CtxTup) => 
+            val fields = tup match
+              case Term.Tup(fs) => fs
+              case Term.CtxTup(fs) => fs
+            (
+              !fields.exists(_.isInstanceOf[Spd]),
+              fields.map:
+                case Fld(asc = S(_)) => 0
+                case _: Fld => 1
+                case _: Spd => 0
```
LPTK: Is this case really sound? Maybe fix it now? Or add it to the issue tracker (if not already there).

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -559,18 +569,22 @@ class Resolver(tl: TraceLogger)
...
+              .sum +
+              fields.exists:
+                case Fld(asc = S(_)) => true
+                case _ => false
+              .into(if _ then 1 else 0)
```
LPTK: Use `.count`?

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -633,6 +647,7 @@ class Resolver(tl: TraceLogger)
           
           val args = as match
             case Term.Tup(args) => args
+            case Term.CtxTup(args) => args
             case spd => Spd(true, spd) :: Nil
```
LPTK: Is this exhaustive? Is `spd` really a spread? Are error cases potentially missing, here?

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -467,18 +467,28 @@ class Resolver(tl: TraceLogger)
...
+    * App `f`.
     * @param inTyPrefix if true, the currently resolving term is the
-    * prefix of an TyApp which the implicit arguments shouldn't be
-    * resolved on it, but on the TyApp instead.
+    * prefix of an TyApp, e.g., `f[Int]`. The implicit arguments should
+    * be resolved on the the TyApp `f[Int]`, but not on the base of the
+    * TyApp `f`.
     */
-  def resolve(t: Resolvable, inTyPrefix: Bool = false)(using ICtx): (Opt[CallableDefinition], ICtx) =
+  def resolve(t: Resolvable, inCtxPrefix: Bool = false, inTyPrefix: Bool = false)(using ICtx): (Opt[CallableDefinition], ICtx) =
```
LPTK: I can't find anywhere `inTyPrefix` is passed as `true`. If it is, the call site is missing `inTyPrefix =`.

I used to think this pattern was one of the only places where default arguments are fine in internal code, but even this usage feels to error-prone to me, now.

It would be better to remove the defaul arguments and instead use different functions, like `resolveOther`, `resolveCtxPrefix`, and `resolveTyPrefix`, which pass the correct arguments to `resolve`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: I don't see any unsoundness here? It's just the lower bound of number of arguments

LPTK: Ah, I see. In that case, it's probably fine.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: This is different from the above case: all fields with ascriptions are aggregated to one record and passed to the function. So I think `.count` does not work here

LPTK: Right. How about
```scala
              fields.collectFirst:
                case Fld(asc = S(_)) => 1
              .getOrElse(0)
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: I remember I saw a comment in the code base saying any `Term` occurring as the RHS of an `App` is a spread. Let me find it out

FlandiaYingman: Here it is

https://github.com/hkust-taco/mlscript/blob/f2c2999f96f4a3f4299256f0d6eb82142fb68584/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala#L325-L332

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: It's here

https://github.com/hkust-taco/mlscript/blob/f2c2999f96f4a3f4299256f0d6eb82142fb68584/hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala#L498-L502

I could refactor to the pattern you mentioned

FlandiaYingman: No wait, I think it couldn't be refactored like that because there are places inheriting the `inPrefix` state to the next recursive call. Instead I just removed the default parameters

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
LPTK: Ah, then please document that here too!

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -496,7 +496,7 @@ class Resolver(tl: TraceLogger)
         (t.callableDefn, ictx)
       
       case Term.TyApp(lhs: Resolvable, targs) =>
-        resolve(lhs, inTyPrefix = true)
+        resolve(lhs, inCtxPrefix= false, inTyPrefix = true)
```
```suggestion
        resolve(lhs, inCtxPrefix = false, inTyPrefix = true)
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
LPTK: Ok, I'm more than happy about that!

hkmc2/shared/src/test/mlscript/ctx/ExplicitlySpecifying.mls
```diff
@@ -0,0 +1,87 @@
...
+
+:expect 42
+f(using 32)(using 8)(using 2)
+//│ = 42
+
+
+fun f(using i: Int)(j: Int)(using k: Int)(l: Num): Int = i + j + k + l
+
+:expect [function]
+f(using 32)
```
LPTK, line 66: How is that allowed? Shouldn't this be """moduleful""", since the returned function won't work? (It will fail to resolve its implicit parameter implicitly.)

FlandiaYingman: It doesn't bind the function to some name, so it is allowed by design.

For example for simple moduleful terms we have

```fs
module M
fun f: module M

M // allowed
//| = M
f // allowed
//| = M
```

So I think this should be allowed too!

LPTK: Ah, ok, why not. Weird, but not unsound!

LPTK (approved):

PR #316, opened 2025-06-30
The Ultimate Pattern Syntax

chengluyu:


hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -311,7 +311,9 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
         case _: sem.BuiltinSymbol => true
         case sym: sem.BlockMemberSymbol =>
           sym.trmImplTree.fold(sym.clsTree.isDefined)(_.k is syntax.Fun)
-        case _ => false
+        // Do not perform safe check on `MatchResult` and `MatchFailure`.
```
```suggestion
        // Do not perform safety check on `MatchResult` and `MatchFailure`.
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -311,7 +311,9 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
         case _: sem.BuiltinSymbol => true
         case sym: sem.BlockMemberSymbol =>
           sym.trmImplTree.fold(sym.clsTree.isDefined)(_.k is syntax.Fun)
-        case _ => false
```
LPTK, line 344: Removing the `_` case seems unsafe (and probably raises a warning) 🤔

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1112,33 +1109,54 @@ extends Importer:
         case Pat =>
           val patSym = td.symbol.asInstanceOf[PatternSymbol] // TODO improve `asInstanceOf`
           val owner = ctx.outer.inner
+          // OK. It seems that the context after `nestInner` already have all parameters.
```
LPTK: I don't understand this comment. Why does it say "OK"?

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -498,7 +498,17 @@ case class PatternDef(
...
-    body: ObjBody,
+    /** The pattern parameters, for example, `T` in
+     *  `pattern Nullable(pattern T) = null | T`. */
+    patternParams: Ls[Param],
+    /** The extraction parameters, for example, `x` in
+     *  `pattern PairLike(x, y) = [x, y] | Pair(x, y)`. */
+    extractionParams: Ls[Param],
+    /** The elaborated pattern right-hand side. */
+    pattern: Pattern,
+    // Here, `ObjBody` contains methods `unapply` and `unapplyStringPrefix`,
```
LPTK: I'm not sure this is the right place to store these. Why aren't these methods generated in Lowering? They are implementation details, whereas `Term` should contain user-relevant semantic details.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -117,39 +119,63 @@ class Normalization(using tl: TL)(using Raise, Ctx, State) extends DesugaringBas
...
           case S(N) =>
-            // The constructor is not a class-like symbol. Report the error and skip the branch.
-            error(msg"Cannot use this ${ctor.describe} as a pattern" -> ctor.toLoc)
-            normalizeImpl(alternative)
+            // The constructor is not a class-like symbol. But it might be a
+            // `VarSymbol` referencing to a pattern parameter.
+            ctor.symbol match
+              case S(symbol: VarSymbol) => symbol.decl match
+                case S(param @ Param(flags = FldFlags(pat = true))) =>
+                  // We're missing two checks here. The first one is to make
```
LPTK: Is this reported in a tracking issue? Or meant to be addressed later in this current PR?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Translator.scala
LPTK, line 1: Let's rename this file to something more specific, such as `DirectTranslator` or `NaiveTranslator`.

hkmc2/shared/src/main/scala/hkmc2/semantics/ups/Compiler.scala
```diff
@@ -0,0 +1,449 @@
...
+import syntax.{Keyword, LetBind, Tree}, Tree.{DecLit, Ident, IntLit, StrLit, UnitLit}
+import Term.{Blk, IfLike, Rcd, Ref, SynthSel}
+import Pattern.{Instantiation, Head}
+import Elaborator.{Ctx, State, ctx}, utils.TL
+import ucs.{DesugaringBase as Base, FlatPattern, safeRef}
+
+import collection.mutable.{Queue, Map as MutMap}, collection.immutable.{Set, Map}
+
+
+class Compiler(using Context)(using tl: TL)(using Ctx, State, Raise) extends Base:
```
LPTK: Why not document what this class is for?

hkmc2/shared/src/main/scala/hkmc2/semantics/ups/Context.scala
```diff
@@ -0,0 +1,22 @@
...
+      case Some(pattern) => pattern
+      case None =>
+        bug(msg"The pattern is not instantiated." -> instantiation.toLoc)
+        Never
+
+object Context:
+  extension (instantiation: Instantiation)
+    /** Get the body of the instantiated pattern definition. */
+    def body(using Context, Raise): Pat =
+      summon[Context].get(instantiation)
```
LPTK:
```suggestion
      summon[Context].get(instantiation)

```

hkmc2/shared/src/main/scala/hkmc2/syntax/Token.scala
```diff
@@ -52,6 +52,18 @@ final case class COMMENT(text: String) extends Token with Stroken
...
+
+(foo
+  bar)
+
+(foo INDENT bar) DEINDENT
+
+Bracket[Round]{foo Bracket[INDENT]{bar}}
+
+
+*/
```
LPTK: Looks like you committed this by mistake.

hkmc2/shared/src/main/scala/hkmc2/utils/utils.scala
```diff
@@ -102,3 +102,41 @@ extension [A](self: Opt[A])
...
+    case 3 => "three" case 4 => "four"      case 5 => "five"
+    case 6 => "six"   case 7 => "seven"     case 8 => "eight"
+    case 9 => "nine"  case _ => n.toString
+
+extension (str: String)
+  /** Converts a singular noun to its plural form using English pluralization
+   *  rules. The rules should be updated as needed. */
+  def pluralize: String =
+    // This is not a complete list (nor is it intended to be).
+    val irregularPlurals = Map("datum" -> "data", "index" -> "indices")
```
LPTK: Are you crazy? Why would you create this map (and the sets below) _every single time_ this function is called?

hkmc2/shared/src/main/scala/hkmc2/utils/utils.scala
```diff
@@ -102,3 +102,41 @@ extension [A](self: Opt[A])
...
+    val ves = Set("proof") // Add more as needed.
+    val o = Set("zero") // Add more as needed.
+    irregularPlurals.get(str).getOrElse:
+      // -s, -sh, -ch, -x, -z -> +es
+      if str.matches(".*[sxz]$") || str.endsWith("sh") || str.endsWith("ch") then str + "es"
```
LPTK: Feels like these regexps should be compiled outside the function and used here.

hkmc2/shared/src/test/mlscript/llir/Split.mls
```diff
@@ -52,4 +54,4 @@ fun map_sum(f, n) = let it = map(f, dec(n)) in fold(0, it)
 map_sum(x => x, 200)
 //│ 
 //│ Interpreted:
-//│ 20100
+//│ /!!!\ Uncaught error: java.lang.StackOverflowError
```
LPTK: Do you know what's causing this? Where is it from?

hkmc2/shared/src/test/mlscript/rp/JoinPatterns.mls
```diff
@@ -25,6 +25,14 @@ pattern SimpleJoin = ("abc" | "xyz") ~ ("def" | "")
 :todo
```
LPTK: This command is misused.

hkmc2/shared/src/test/mlscript/llir/Split.mls
chengluyu: I don't know the cause. Strangely, there is no problem when I run `~hkmc2DiffTests/Test/run`, but an error occurs here when I run `hkmc2AllTests/test`.

hkmc2/shared/src/main/scala/hkmc2/utils/utils.scala
chengluyu: Oh, that was indeed a bit silly. Fixed in b392c4a.

hkmc2/shared/src/test/mlscript/llir/Split.mls
LPTK: You only ansered one half of my message.

hkmc2/shared/src/main/scala/hkmc2/utils/utils.scala
```diff
@@ -116,26 +116,36 @@ extension (n: Int)
...
-      else if str.matches(".*[^aeiou]y$") then str.dropRight(1) + "ies"
-      // -f -> -ves (with some exceptions)
-      else if str.endsWith("f") && !ves.contains(str) then str.dropRight(1) + "ves"
-      // -fe -> -ves
-      else if str.endsWith("fe") then str.dropRight(2) + "ves"
-      // -o -> -es (with some exceptions)
-      else if str.endsWith("o") && !o.contains(str) then str.dropRight(1) + "es"
-      else str + "s"
+  def pluralize: String = English.irregularPlurals.get(str).getOrElse:
+    str.lift(str.size - 1) match
```
LPTK: Allocating an option every time just to deal with the empty string case seems a bit silly. Just use an `if`-`then`-`else`?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
chengluyu: The code here is related to the naive translation of higher-order patterns and will be fixed later this PR.

LPTK:
@chengluyu What's missing for this to be merged?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
chengluyu: I also added a wildcard case below.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
chengluyu: I plan to address this issue in a new PR.

hkmc2/shared/src/main/scala/hkmc2/semantics/ups/Pattern.scala
```diff
@@ -0,0 +1,339 @@
...
+    case Not(pattern) => Not(pattern.simplify)
+    case Rename(pattern, name) => Rename(pattern.simplify, name)
+    case Extract(pattern, term) => pattern.simplify match
+      case `Never` => Never // Lift up `Never`. Let the caller reduce it.
+      case simplified => Extract(simplified, term)
+    case Synonym(sym) => this
+  
+  /** Expand the pattern by replacing any top-level synonym with its body.
+   *  @return the expanded pattern named "ExPat" in the paper. */
+  def expand(alreadyExpanded: Set[Instantiation] = Set())(using Context, Raise): ExPat = map:
```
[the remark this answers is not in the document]
LPTK: In fact, please remove all default arguments from core logic function, as usual.

hkmc2/shared/src/main/scala/hkmc2/Message.scala
```diff
@@ -31,6 +31,7 @@ object Message:
   final case class Code(ty: TypeLike) extends Bit
   implicit def fromType(ty: TypeLike): Message = Message(Code(ty)::Nil)
   implicit def fromStr(str: Str): Message = Message(Text(str)::Nil)
+  implicit def fromInt(int: Int): Message = Message(Text(int.toString)::Nil)
```
LPTK:
```suggestion
  implicit def fromType(ty: TypeLike): Message = Message(Code(ty) :: Nil)
  implicit def fromStr(str: Str): Message = Message(Text(str) :: Nil)
  implicit def fromInt(int: Int): Message = Message(Text(int.toString) :: Nil)
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ups/Pattern.scala
```diff
@@ -0,0 +1,339 @@
...
+    case Tuple(leading, spread, trailing) =>
+      val leadingText = leading.map(_.showDbg).mkString(", ")
+      val spreadText = spread.map(_.showDbg).mkString(", ")
+      val trailingText = trailing.map(_.showDbg).mkString(", ")
+      List(leadingText, spreadText, trailingText).mkString("[", ", ", "]")
+    case And(Nil) => "\u22A5"
+    case And(pattern :: Nil) => pattern.showDbg
+    case And(patterns) =>
+      patterns.map(_.showDbg).mkString("(", " \u2227 ", ")")
+    case Or(Nil) => "\u22A4"
```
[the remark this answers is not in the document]
chengluyu: Fixed in 6cee1c3.

hkmc2/shared/src/test/mlscript/llir/Split.mls
chengluyu: I don’t know where it is from. But this test now seems to be outdated, So, I marked this comment as resolved.

hkmc2/shared/src/main/scala/hkmc2/semantics/ups/Compiler.scala
chengluyu: Fixed in b873b2c.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
chengluyu: Fixed in 16cce58.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
chengluyu: This check is necessary anymore, because the `NaiveCompiler` has already been moved to`Lowering`, where the symbols are already resolved.

LPTK: Then maybe change the comment?

hkmc2/shared/src/main/scala/hkmc2/semantics/ups/Compiler.scala
```diff
@@ -429,28 +429,51 @@ object Compiler:
...
-        .map(_.reverse.mkString(s"${sym.nme}\u02BF", "_", "\u02BE"))
+        .map(_.reverse.mkString(sym.nme + FAKE_LEFT_PAREN, FAKE_COMMA, FAKE_RIGHT_PAREN))
       case Synonym(Instantiation(symbol, patterns)) =>
         if patterns.isEmpty then S(symbol.nme) else
           patterns.iterator.mapOption(_.shortName(0)).map:
-            // Go's trick: https://news.ycombinator.com/item?id=14276891
-            _.reverse.mkString(s"${symbol.nme}\u1438", "_", "\u1433")
-      case Or(Nil) => S("\uA4D5")
+            _.reverse.mkString(symbol.nme + FAKE_LEFT_ANGLE, FAKE_COMMA, FAKE_RIGHT_ANGLE)
+      case Or(Nil) => S(FAKE_TOP)
```
LPTK: Again, the neutral element of `Or` is bottom, not top. You seem to have mixed them up.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1168,34 +1164,47 @@ extends Importer:
...
+            // 2. extraction bindings, e.g., `value` in `pattern Middle(value) = ...`; and
+            // 3. the rest are reported as invalid parameters.
+            val (patternParams, extractionParams) = ps.fold((Nil, Nil)):
+              _.params.flatMap:
+                // Only `pat` flag is `true`.
+                case p @ Param(flags = FldFlags(false, false, true, false)) => S(p)
+                // All flags are `false`.
+                case p @ Param(flags = FldFlags(false, false, false, false)) => S(p)
+                case Param(flags, sym, _, _) =>
+                  raise(ErrorReport(msg"Unexpected pattern parameter ${sym.name} with flags ${flags.show}" -> sym.toLoc :: Nil))
```
```suggestion
                  raise(ErrorReport(msg"Unexpected pattern parameter ${sym.name} with modifiers: ${flags.show}" -> sym.toLoc :: Nil))
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1168,34 +1164,47 @@ extends Importer:
...
+            log(s"`${patSym.nme}`'s extraction parameters: ${extractionParams.mkString("[", ", ", "]")}")
+            // Empty pattern body is considered as wildcard patterns.
+            val rhs = td.rhs.getOrElse:
+              raise(ErrorReport(msg"Pattern definitions must have a body." -> td.toLoc :: Nil))
+              Tree.Under()
+            // Elaborate the pattern body with the pattern parameters.
+            val pat = pattern(rhs)(using ctx ++ patternParams.iterator.map(p => p.sym.name -> p.sym))
+            // Report all invalid variables we found in the top-level pattern.
+            pat.variables.report
+            // Note that the remaining variables have not been bounded to any
```
```suggestion
            // Note that the remaining variables have not been bound to any
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1168,34 +1164,47 @@ extends Importer:
...
+            // `VarSymbol` yet. Thus, we need to pair them with the extraction
+            // parameters. We only report warnings for unbounded variables
```
```suggestion
            // parameters. We only report warnings for unbound variables
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1168,34 +1164,47 @@ extends Importer:
...
+            // because they are harmless.
+            pat.variables.varMap.foreach: (name, aliases) =>
+              extractionParams.find(_.sym.name == name) match
+                case S(param) => aliases.foreach(_.symbol = param.sym)
+                case N => raise(WarningReport(msg"Useless pattern binding: $name." -> aliases.head.toLoc :: Nil))
+            scoped("ucs:ups")(log(s"elaborated pattern body: ${pat.showDbg}"))
+            scoped("ucs:ups:tree")(log(s"elaborated pattern body: ${pat.showAsTree}"))
+            // `paramsOpt` is set to `N` because we don't want parameters to
```
LPTK, line 1204: If it's always `N`, it shouldn't be a parameter of that class!

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1322,6 +1337,131 @@ extends Importer:
...
+      val (leading, spread, trailing) = ts.foldLeft(z):
+        case (acc @ (_, S(_), _), Spread(`...`, _, _)) =>
+          // Found two spreads in the same tuple pattern. Report an error.
+          raise(ErrorReport(msg"Multiple spread patterns are not supported." -> t.toLoc :: Nil))
+          acc // Do not modify the accumulator and skip this spread.
+        case ((leading, N, trailing), Spread(`...`, _, S(t))) =>
+          // Elaborate the spread pattern and add it to spread element.
+          (leading, S(go(t)), trailing)
+        case ((leading, N, trailing), Spread(`...`, _, N)) =>
+          // Empty spreads results in a wildcard pattern.
```
LPTK: Wait so `P(...)` and `P(_)` do the same thing? This is surprising. The former should match an arbitrary number of arguments.

hkmc2/shared/src/main/scala/hkmc2/semantics/ups/Compiler.scala
chengluyu: Sorry for making the same mistake again!

chengluyu: Fixed in 41f42c17.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
chengluyu: Fixed in 39d804648.

hkmc2/shared/src/main/scala/hkmc2/semantics/ups/Pattern.scala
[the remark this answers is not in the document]
chengluyu: Fixed in b873b2cff.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
chengluyu: Sorry, what I actually meant is: if the `Spread` doesn’t have a `body` (i.e., `Spread(`...`, _, N)`), then _the sub-pattern matched against the spred scrutinee_ will be a wildcard.

```suggestion
          // If `Spread` does not have a `body`, then pattern for the spred
          // scrutinee will be a `Wildcard`.
```

chengluyu:
This PR is ready!

There are still some refactorings that can be done next.

- Rewrite the `Desugarer`. Currently, `semantics.Pattern` is not used by `Desugarer`, which makes it unable to support many useful patterns.
- Some refactoring of `FlatPattern` and `Pattern` can be postponed until the `Desugarer` is rewritten.  
- A lot of the work for checking arguments and pattern arguments is done in `Normalization`. This should be handled in the `NaiveCompiler` instead.

Besides that, there are still some unimplemented features:
- Support tuple patterns in pattern definitions.
- Support locally recursive patterns using `as`.
- Implement compilation of string patterns. Although there are many examples, some string patterns still don't work correctly. https://github.com/hkust-taco/mlscript/blob/513b1ea61f0ce1c27220fdcdb39e8697e1e9e606/hkmc2/shared/src/test/mlscript/rp/regex/Simplification.mls#L18-L28

(I’ve added above contents to #151. I'll add more to #151 when I think of it.)

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: Please add a source code example to illustrate this situation. I don't remember what it means that a "spread does not have a body".

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -341,7 +343,9 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
+        // Do not perform safety check on `MatchResult` and `MatchFailure`.
+        case sym => (sym is State.matchResultClsSymbol) ||
+          (sym is State.matchFailureClsSymbol)
```
LPTK, line 348: If a reference has a `ClassSymbol` as the symbol, it denotes a self reference (ie `this`). So these lines seem to imply you're putting the wrong symbol for these references. They should be `BlockMemberSymbol`s.

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
```diff
@@ -2,98 +2,313 @@ package hkmc2
...
+    case Constructor(target, patternArguments, arguments) =>
+      val targetText = target.symbol.fold(target.showDbg)(_.toString())
+      val patternArgumentsText = if patternArguments.isEmpty then "" else
+        patternArguments.iterator.map(_.showDbg)
+          .map("pattern " + _).mkString("(", ", ", ")")
+      val argumentsText = arguments.fold(""): args =>
+        s"(${args.map(_.showDbg).mkString(", ")})"
+      s"$targetText$patternArgumentsText$argumentsText"
+    case Composition(true, left, right) => s"${left.showDbg} \u2228 ${right.showDbg}"
+    case Composition(false, left, right) => s"${left.showDbg} \u2227 ${right.showDbg}"
```
LPTK: Again, let's just use unicdoe characters directly...

LPTK: And please make the change everywhere. There should be no `\u` left in the diff, unless there's a good reason for it.

hkmc2/shared/src/main/scala/hkmc2/semantics/Split.scala
```diff
@@ -97,26 +97,44 @@ object Split:
...
-          case (n, line) :: tail => (n, (if isTopLevel then "" else "and ") + line) :: tail
+          case (n, line) :: tail => (n, line) :: tail
           case Nil => Nil
         ) ::: split(tail, false, isTopLevel)
         case Split.Let(nme, rhs, tail) =>
           (0, s"let $nme = ${rhs.showDbg}") :: split(tail, false, true)
-        case Split.Else(term) =>
-          (if isFirst then (0, s"then ${term.showDbg}") else (0, s"else ${term.showDbg}")) :: Nil
+        case Split.Else(t) =>
+          // (if isFirst then (0, s"then ${t.showDbg}") else (0, s"else ${t.showDbg}")) :: Nil
```
LPTK: Is this comment useful?

hkmc2/shared/src/main/scala/hkmc2/semantics/Split.scala
```diff
@@ -97,26 +97,44 @@ object Split:
...
+        val Branch(scrutinee, pattern, consequent) = b
+        val lines = split(consequent, true, false)
+        val prefix = s"${scrutinee.sym} is ${pattern.showDbg}"
+        consequent match
+          case Split.Else(_) => (prefix + " then") #: lines
+          case _ => (prefix + " and") #: lines
       val lines = split(s, true, true)
       (if prefix.isEmpty then lines else prefix #: lines).toIndentedString
+  
+  end display
```
LPTK: This `display` stuff looks like a general utility and feels a bit out of place here, no?

Why not move it to `hkmc2/shared/src/main/scala/hkmc2/utils/utils.scala`, for example?

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -304,7 +304,7 @@ sealed trait Statement extends AutoLocated with ProductWithExtraInfo:
     case SynthSel(pre, nme) => s"(${pre.showDbg}.)${nme.name}"
     case DynSel(pre, fld, _) => s"${pre.showDbg}[${fld.showDbg}]"
     case IfLike(kw, body) => s"${kw.name} { ${body.showDbg} }"
-    case Lam(params, body) => s"λ${params.showDbg}. ${body.showDbg}"
+    case Lam(params, body) => s"λ${params.paramSyms.map(_.id.name).mkString(", ")}. ${body.showDbg}"
```
LPTK: Why this change? It's actually wrong (fails to properly show spread params).

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -502,14 +502,28 @@ case class PatternDef(
...
+     */
+    pattern: Pattern,
     annotations: Ls[Annot],
 ) extends ClassLikeDef:
   self =>
   val kind: ClsLikeKind = Pat
   val ext: Opt[New] = N
+  /** Each pattern definition should contain two methods: `unapply` and
+   *  `unapplyStringPrefix`, which are generated in `Lowering`. Hence, there
+   *  is no need to make `body` an parameter. */
```
```suggestion
   *  is no need to make `body` a parameter. */
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -255,6 +287,15 @@ class Normalization(using tl: TL)(using Raise, Ctx, State) extends DesugaringBas
...
     case N => true
   
+  /** Ensure that there are no pattern arguments. */
+  private def ensureArguments(
+      arguments: Opt[Ls[FlatPattern.Argument]]
+  ): Opt[Ls[FlatPattern.Argument.Term]] = arguments.map:
+    _.flatMap:
+      case arg: FlatPattern.Argument.Term => S(arg)
+      case FlatPattern.Argument.Pattern(_, pattern) =>
+        error(msg"Pattern argument `${pattern.showDbg}` cannot be used here." -> pattern.toLoc); N
```
LPTK: `showDbg` is an _**internal debugging tool**_, not somehting for users to see!! Please make sure you do not use it here and elsewhere else in your code (make a pass to check).

hkmc2/shared/src/test/mlscript/HkScratch.mls
```diff
@@ -9,4 +9,3 @@
 // :todo
 
 
-
```
LPTK: 👎

hkmc2/shared/src/test/mlscript/codegen/QQImport.mls
```diff
@@ -1,7 +1,7 @@
...
...nch { class: class Branch }, Split: class Split, Cons: fun Cons { class: class Cons }, Let: fun Let { class: class Let }, Else: fun Else { class: class Else }, End: class End, Keyword: module Keyword { If: If, While: While }, Statement: class Statement, LetDecl: fun LetDecl { class: class LetDecl }, DefineVar: fun DefineVar { class: class DefineVar }, Term: class Term, Lit: fun Lit { class: class Lit }, Builtin: fun Builtin { class: class Builtin }, Ref: fun Ref { class: class Ref }, CSRef: fun CSRef { class: class CSRef }, App: fun App { class: class App }, Sel: fun Sel { class: class Sel }, DynSel: fun DynSel { class: class DynSel }, Tup: fun Tup { class: class Tup }, IfLike: fun IfLike { class: class IfLike }, Lam: fun Lam { class: class Lam }, Blk: fun Blk { class: class Blk }, New: fun New { class: class New }, Region: fun Region { class: class Region }, RegRef: fun RegRef { class: class RegRef }, Assgn: fun Assgn { class: class Assgn }, Deref: fun Deref { class: class Deref }, SetRef: fun SetRef { class: class SetRef }, Ret: fun Ret { class: class Ret }, Throw: fun Throw { class: class Throw }, Try: fun Try { class: class Try }, Context: fun Context { class: class Context } }
```
LPTK: BTW how come such output is not broken into multiple lines? Isn't this using the new rendering function?

hkmc2/shared/src/test/mlscript/llir/Split.mls
```diff
@@ -26,6 +26,8 @@ map_sum(x => x, 200)
 //│ Stopped due to an error during the Llir generation
 
 :intl
+:todo
+// Works fine with `~hkmc2DiffTests/Test/run` but not with `hkmc2AllTests/test`.
```
LPTK: Really? There is no compilation error with `hkmc2AllTests/test`?

hkmc2/shared/src/test/mlscript/rp/Future.mls
```diff
@@ -2,45 +2,57 @@
 
 :todo // Parameterized patterns.
 pattern Rep0[A] = "" | A ~ Rep0[A]
-//│ ╔══[ERROR] Unrecognized pattern (application)
+//│ ╔══[ERROR] Found an unrecognizable pattern.
```
LPTK: The previous error was perfect and better than this one.

hkmc2/shared/src/test/mlscript/rp/Future.mls
```diff
@@ -107,10 +126,26 @@ pattern Lines(pattern L) = case
...
+//│ ║         	^^^^^^^^^^^^^^^^^^^^^
+//│ ║  l.122: 	    "" then []
+//│ ║         	^^^^^^^^^^^^^^
+//│ ║  l.123: 	    "\n" ~ (Lines of L) as t then t
+//│ ║         	^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+//│ ║  l.124: 	  L as h ~ Tail as t then [h, ...t]
+//│ ║         	^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+//│ ║  l.125: 	  
+//│ ║         	^^^
+//│ ║  l.126: 	:todo
```
LPTK: Is this on purpose? (I guess not.)

hkmc2/shared/src/test/mlscript/rp/Future.mls
```diff
@@ -119,38 +154,54 @@ pattern Opt(pattern P) = case
...
 pattern Email(name, domain) = ...
-//│ /!!!\ Uncaught error: scala.MatchError: InfixApp(Case(None,Block(List(InfixApp(App(Ident(P),Tup(List(Ident(value)))),keyword 'then',Bra(Round,App(Ident(Some),Tup(List(Ident(value)))))), InfixApp(StrLit(),keyword 'then',Bra(Round,Ident(None)))))),keyword ':',Ident(todo)) (of class hkmc2.syntax.Tree$InfixApp)
+//│ ╔══[ERROR] Unexpected record property pattern.
+//│ ║  l.152: 	  P(value) then (Some(value))
+//│ ║         	  ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+//│ ║  l.153: 	  ""       then (None)
+//│ ║         	^^^^^^^^^^^^^^^^^^^^^^
+//│ ║  l.154: 	  
+//│ ║         	^^^
+//│ ║  l.155: 	:todo
```
LPTK: 👀

LPTK (commented):
I didn't read all the code or look at all the tests, but I think it looks good enough and we should merge this ASAP, unless @FlandiaYingman finds any problem (hopefully he can make a quick review).

LPTK:
By the way, why is the test folder called `rp`? Why not `ups`?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
```diff
@@ -10,14 +10,16 @@ import Keyword.{`as`, `and`, `or`, `do`, `else`, is, let, `then`, where}
 import collection.mutable.{Buffer, HashMap, SortedSet}
 import Elaborator.{Ctx, Ctxl, State, UnderCtx, ctx}
 import scala.annotation.targetName
-import Pattern.MatchMode
+import FlatPattern.{Argument,MatchMode}
```
FlandiaYingman:
```suggestion
import FlatPattern.{Argument, MatchMode}
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/DesugaringBase.scala
FlandiaYingman, line 1: `Desugarer` and `Normalization` both extend the trait `DesugaringBase`, but some methods in `DesugaringBase` cannot be shared among `Desugarer` and `Normalization`. 

`Desugarer` happens in elaboration, so methods that return a `Resolvable` term, such as `DesugaringBase.app`, should not insert the implicit arguments as they will be resolved in the resolver. (In the future the resolver may raise an internal error if there are already some implicit arguments inserted before its resolution, because it signifies that some terms are resolved multiple times.) `Normalization`, on the other hand, happens in lowering, so it should insert the implicit arguments by `withIArgs`.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
```diff
@@ -177,8 +178,8 @@ class Desugarer(elaborator: Elaborator)(using Ctx, Raise, State, UnderCtx) exten
         tail.foldRight(innermostSplit):
           case ((coda, pat), sequel) => ctx =>
             nominate(ctx, term(coda)(using ctx)):
-              expandMatch(_, pat, sequel)(fallback)
-      expandMatch(scrutSymbol, headPattern, tailSplit)(fallback)
+              expandMatch(_, pat, sequel, Nil)(fallback)
```
FlandiaYingman, line 191: I think we can't just reuse `fallback` here in a `foldRight` because that may create duplicate branches that share the same term. Then the resolver might resolve the same term at different places, resulting in different results for the same term.

FlandiaYingman (approved):
I have some feedback specifically regarding the resolver; LGTM otherwise 😛😵‍💫

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
chengluyu: Fixed in 41ca4cd94.

hkmc2/shared/src/main/scala/hkmc2/semantics/Split.scala
chengluyu: Fixed in 125b12888.

hkmc2/shared/src/main/scala/hkmc2/semantics/Split.scala
chengluyu: This is just for pretty-printing `Split`, so I don't plan to put it in `utils`. I did a simple refactor and turned it into a method on `Split`. Fixed in 125b12888.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
chengluyu: Sorry, this is an outdated and early change. It has been reverted in 52775c532.

hkmc2/shared/src/test/mlscript/llir/Split.mls
chengluyu: There were `StackOverflowError` before merging the main branch. I forgot about this place. 

https://github.com/hkust-taco/mlscript/blob/6ca78ca7b3c1063259deef633332795ee77a551e/hkmc2/shared/src/test/mlscript/llir/Split.mls#L55-L57

This has been reverted in d7087603a.

hkmc2/shared/src/test/mlscript/rp/Future.mls
chengluyu: Fixed in 7ae33a7.

hkmc2/shared/src/test/mlscript/rp/Future.mls
chengluyu: This was caused by a line made of two spaces between two adjacent test blocks, which rendered them to be one single test block. I have fixed this in 5d14750.

hkmc2/shared/src/test/mlscript/rp/Future.mls
chengluyu: Same as above.

hkmc2/shared/src/test/mlscript/rp/Future.mls
```diff
@@ -138,70 +132,68 @@ if input is Lines of Email then
...
 //│ ║         	^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 //│ ║  l.124: 	  L as h ~ Tail as t then [h, ...t]
-//│ ║         	^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-//│ ║  l.125: 	  
-//│ ║         	^^^
-//│ ║  l.126: 	:todo
-//│ ╙──       	^^^^^
+//│ ╙──       	^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+:todo
```
LPTK: Why is this a `:todo`? The expression is clearly ill-formed.

hkmc2/shared/src/test/mlscript/rp/Future.mls
```diff
@@ -138,70 +132,68 @@ if input is Lines of Email then
...
+if input is Lines of Email then
+//│ ╔══[PARSE ERROR] Expected start of expression in this position; found end of input instead
+//│ ║  l.138: 	if input is Lines of Email then
+//│ ╙──       	                               ^
 //│ ╔══[ERROR] Name not found: input
-//│ ║  l.127: 	if input is Lines of Email then
+//│ ║  l.138: 	if input is Lines of Email then
 //│ ╙──       	   ^^^^^
 
 :todo
```
LPTK: Is this really a `:todo`? Do we want to support this? Even if so, it should be
```suggestion
// TODO support this syntax?
:e
```

hkmc2/shared/src/test/mlscript/rp/Future.mls
```diff
@@ -138,70 +132,68 @@ if input is Lines of Email then
...
   P(value) then (Some(value))
   ""       then (None)
-  
+//│ ╔══[ERROR] Unrecognized pattern (case).
+//│ ║  l.148: 	  P(value) then (Some(value))
+//│ ║         	  ^^^^^^^^^^^^^^^^^^^^^^^^^^^
+//│ ║  l.149: 	  ""       then (None)
+//│ ╙──       	^^^^^^^^^^^^^^^^^^^^^^
+
 :todo
```
LPTK: I think this is parsed as
```fs
pattern Email(name, domain) = {
}
```

hkmc2/shared/src/test/mlscript/rp/Future.mls
```diff
@@ -138,70 +132,68 @@ if input is Lines of Email then
...
-//│ ║  l.152: 	  P(value) then (Some(value))
-//│ ║         	  ^^^^^^^^^^^^^^^^^^^^^^^^^^^
-//│ ║  l.153: 	  ""       then (None)
-//│ ║         	^^^^^^^^^^^^^^^^^^^^^^
-//│ ║  l.154: 	  
-//│ ║         	^^^
-//│ ║  l.155: 	:todo
-//│ ╙──       	^^^^^
 
 :todo
```
LPTK: Why is this a `:todo`? Please document what these are for!

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
chengluyu: It means a `Spread` that doesn’t spread anything. It’s just `...` literally.

Note that this is in the elaboration of tuple pattern. The enclosing `foldLeft` is accumulating three parts: the leading patterns, the spread pattern, and the trailing patterns. The comment says that if we found only `...` without a following pattern (i.e., the `body` field of `Spread` is set to `N`) then the spread pattern will be a wildcard, which means that we do not need to further match the middle part of the part.

chengluyu: Improved in 9f4e6d920.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
chengluyu: Fixed in 9f4e6d920.

hkmc2/shared/src/test/mlscript/codegen/QQImport.mls
chengluyu: Fixed in 6f6abdaac. As a result, it doesn't work well with the `:expect` command. We may need a way to solve this problem in a future PR.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
chengluyu: This will be fixed during the refactoring of the `Desugarer`. You can wait for my next PR. Let's merge this PR first.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/DesugaringBase.scala
chengluyu: Fixed in 883f40b94.

hkmc2/shared/src/test/mlscript/rp/Future.mls
chengluyu: This syntax used in this test was outdated. I have adapted it to the latest syntax and add it to its corresponding test file. Solved in 08036b903.

hkmc2/shared/src/test/mlscript/rp/Future.mls
chengluyu: Removed in 08036b903.

hkmc2/shared/src/test/mlscript/rp/Future.mls
chengluyu: Thanks. I marked that this is expected to be reported in 08036b903.

hkmc2/shared/src/test/mlscript/rp/Future.mls
chengluyu: This is an old test and the syntax has been implemented. I have removed it in 08036b903.

hkmc2/shared/src/test/mlscript/HkScratch.mls
chengluyu: Fixed in d7c9b8965.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
chengluyu: Fixed in bc0f600b7ed3d7b78d54d8c1a8f23278e2ebe229.

chengluyu:
Alright, I've fixed most of the issues in this PR. Let's merge it so #325 won’t be blocked.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -294,7 +294,7 @@ class Normalization(using tl: TL)(using Raise, Ctx, State) extends DesugaringBas
     _.flatMap:
       case arg: FlatPattern.Argument.Term => S(arg)
       case FlatPattern.Argument.Pattern(_, pattern) =>
-        error(msg"Pattern argument `${pattern.showDbg}` cannot be used here." -> pattern.toLoc); N
+        error(msg"The pattern argument cannot be used here." -> pattern.toLoc); N
```
LPTK: This is again a badly phrased error.
```suggestion
        error(msg"This `${pattern.describe}` pattern cannot be used as an argument here." -> pattern.toLoc); N
```

(Having a `Pattern#describe` method seems like a good idea for ewrror messages).

LPTK: Also, it seems this error is never exercised (doesn't appear in the diff-tests).

hkmc2/shared/src/test/mlscript/nofib/boyer2.mls
```diff
@@ -325,7 +428,538 @@ let rules =
...
+//│         Node(
+//│           [
+//│             Node(
+//│               [
+//│                 Node(
+//│                   [
+//│                     Empty,
+//│                     [
+//│                       ["a","n","d"],
+//│                       [Conss([Atom(["e","q","u","a","l"]), Conss([Conss([Atom(["a","n","d"]), Conss([Atom(["p"]), Conss([Atom(["q"]), Nill])])]), Conss([Conss([Atom(["i","f"]), Conss([Atom(["p"]), Conss([Conss([Atom(["i","f"]), Conss([Atom(["q"]), Conss([Conss([Atom(["t"]), Nill]), Conss([Conss([Atom(["f"]), Nill]), Nill])])])]), Conss([Conss([Atom(["f"]), Nill]), Nill])])])]), Nill])])])]
```
LPTK: Do you know why this exceeds the column limit?

hkmc2/shared/src/test/mlscript/nofib/boyer2.mls
```diff
@@ -325,7 +428,538 @@ let rules =
...
 
 let lemmas = addlemmalst(makelemmas(rules), Empty)
```
LPTK, line 431:
```suggestion
:silent
let lemmas = addlemmalst(makelemmas(rules), Empty)
```

hkmc2/shared/src/test/mlscript/nofib/eliza.mls
```diff
@@ -246,7 +246,10 @@ let initial =
...
...)],[[["W","H","Y"],["C","A","N","'","T"],["I"]], Lazy(fun)],[[["A","R","E"],["Y","O","U"]], Lazy(fun)],[[["I"],["C","A","N","'","T"]], Lazy(fun)],[[["I"],["A","M"]], Lazy(fun)],[[["I","'","M"]], Lazy(fun)],[[["Y","O","U"]], Lazy(fun)],[[["Y","E","S"]], Lazy(fun)],[[["N","O"]], Lazy(fun)],[[["C","O","M","P","U","T","E","R"]], Lazy(fun)],[[["C","O","M","P","U","T","E","R","S"]], Lazy(fun)],[[["I"],["W","A","N","T"]], Lazy(fun)],[[["W","H","A","T"]], Lazy(fun)],[[["H","O","W"]], Lazy(fun)],[[["W","H","O"]], Lazy(fun)],[[["W","H","E","R","E"]], Lazy(fun)],[[["W","H","E","N"]], Lazy(fun)],[[["N","A","M","E"]], Lazy(fun)],[[["W","H","Y"]], Lazy(fun)],[[["C","A","U","S","E"]], Lazy(fun)],[[["B","E","C","A","U","S","E"]], Lazy(fun)],[[["D","R","E","A","M"]], Lazy(fun)],[[["S","O","R","R","Y"]], Lazy(fun)],[[["H","I"]], Lazy(fun)],[[["D","R","E","A","M","S"]], Lazy(fun)],[[["M","A","Y","B","E"]], Lazy(fun)],[[["H","E","L","L","O"]], Lazy(fun)],[[["A","L","W","A","Y","S"]], Lazy(fun)],[[["Y","O","U","R"]], Lazy(fun)],[[["A","L","I","K","E"]], Lazy(fun)],[[["T","H","I","N","K"]], Lazy(fun)],[[["F","R","I","E","N","D","S"]], Lazy(fun)],[[["F","R","I","E","N","D"]], Lazy(fun)],[[], Lazy(fun)]],
```
LPTK, line 250: Same question: why is it not broken into several lines?

hkmc2/shared/src/test/mlscript/std/RenderingTest.mls
```diff
@@ -243,13 +243,13 @@ tree
 // If the same circular object appears multiple times, it should be rendered as
 // the same reference.
```
LPTK: This no longer seems to be the case?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
LPTK: Sorry, I changed my mind, the previous version (as above this comment) was probrbaly fine for now. This is not a self-reference because you're always using a selection. Still, the symbol logic is very messy and will need to be refactored later.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/DesugaringBase.scala
LPTK: @chengluyu You addressed this, tight? Why not resolve it?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
LPTK: @chengluyu Did you address this? Why not resolve it?

hkmc2/shared/src/test/mlscript/rp/Future.mls
LPTK: @chengluyu You addressed this, tight? Why not resolve it?

hkmc2/shared/src/test/mlscript/rp/Future.mls
LPTK: @chengluyu You addressed this, tight? Why not resolve it?

hkmc2/shared/src/test/mlscript/rp/Future.mls
LPTK: Would be good to mention, in the comment, how this is parsed.

hkmc2/shared/src/test/mlscript/rp/Future.mls
LPTK: @chengluyu You addressed this, tight? Why not resolve it?

LPTK (approved):
Ok, finally LGTM, tho I still have a few questions and minor chnage requests.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
LPTK: Ah never mind, it seems I missed your remark above.

hkmc2/shared/src/test/mlscript/nofib/boyer2.mls
chengluyu: A custom `prettyPrint` method will disable the line breaking settings. Those like `["e","q","u","a","l"]` are from the custom `prettyPrint` method.

LPTK: Ok. We should fix the cursomizable pretty printing in a later PR.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
chengluyu: It’s exercised here:
https://github.com/hkust-taco/mlscript/blob/7a52f7b998695593f0913554e288bdf992923816/hkmc2/shared/src/test/mlscript/rp/syntax/WrongArguments.mls#L1-L12

LPTK: Ah, right.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
chengluyu: Should we keep it as it is now?

hkmc2/shared/src/test/mlscript/std/RenderingTest.mls
chengluyu: No. Caching the rendering output of circular objects only works in single-line mode, because in multi-line mode, the output may vary depending on the position. This change was made because I switched the `render` call in `printRaw` to multi-line mode.

I removed the comment in 4143501c3 and add another test to document the aforementioned behavior in cc7cb3684.

hkmc2/shared/src/test/mlscript/nofib/boyer2.mls
chengluyu: Fixed in df53e8e56.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
chengluyu: Fixed in 62ca88236.

hkmc2/shared/src/test/mlscript/rp/Future.mls
chengluyu: Fixed in 2b722c98b.

chengluyu:
> By the way, why is the test folder called `rp`? Why not `ups`?

At that time, we hadn’t called it ups yet. I just renamed the entire test folder. Let’s merge it!

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
LPTK: No, it was simpler before, so let's revert the change. Both versions are wrong BTW, but these things will be refactored anyway.

LPTK:
> Let’s merge it!

Yes, after you rever the symbol-checkign logic mentioned above.

PR #317, opened 2025-07-05
Eliminate Reusing Mutable Terms

FlandiaYingman:
The current compiler assumes that all `Term` are immutable and is reusing them. This assumption is wrong because of the resolution stage. This PR tries to fix the places where terms are reused.

See also https://github.com/hkust-taco/mlscript/pull/297#discussion_r2060561339.

AFAIK there are still two places where terms are reused and I couldn't fix them.

- During UCS desurgaring, the desugarer reuses the `fallback` split (e.g., in `hkmc2/shared/src/test/mlscript/ucs/syntax/PlainConditionals.mls`).

- Both `Desugarer` and `Normalization` share the same super class `DesugaringBase`. However, desugaring happens before resolution and normalization happens after resolution, so the helper functions that `Desugarer` uses should not initialize the implicit arguments when it creates new terms; while the helper functions that `Normalization` uses should initialize the implicit arguments. Unfortunately they use a same set of helper functions :-( Errors under `hkmc2/shared/src/test/mlscript/rp` are because of this).

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
@@ -203,7 +203,7 @@ object Elaborator:
     val runtimeSymbol = TempSymbol(N, "runtime")
     val termSymbol = TempSymbol(N, "Term")
     val effectSigSymbol = ClassSymbol(DummyTypeDef(syntax.Cls), Ident("EffectSig"))
-    val nonLocalRetHandlerTrm =
+    def nonLocalRetHandlerTrm =
       val id = new Ident("NonLocalReturn")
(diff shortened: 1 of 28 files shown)


LPTK:
Thanks for your WIP! The pattern stuff is completely refactored in https://github.com/hkust-taco/mlscript/pull/316, so it's not worth dealing with here. Maybe we can just disable the check that makes these tests fail for now, and re-enable it later?

FlandiaYingman:
The checks were not there before this PR, so I think we could just wait for Luyu's PR to be merged, and see how we want to change this PR according to that.

FlandiaYingman:
I'm closing this PR as there are too much conflict. Later I'll create a new PR for this change.

PR #318, opened 2025-07-22
Lazy spreads with lazy arrays, and benchmarking sub-project

gok99:
This PR adds:
- Lazy spreads `..`, implemented as LazyArrays, which are `View`s and `Splice`s over regular arrays.
- Improve iterable usage (should eventually use traits), automatically grant array-like methods, that essentially caches a "materialized" array.
- A benchmarking sub-project, taking largely from #303. ~~Custom runtimes are injected with the `runtime` command.~~ `viz.html` for visualizing results written by the benchmarking suite.
  - Note (TODO): benchmarks currently inherit parallel execution, which could be an issue if many more benchmarks than cores are run in parallel.
- Associated test case updates. Some llir tests break due to unsupported select - marked :todo for now.

---

### Motivation

A function like map, should be expressible as follows:

```
fun map(f) = case
  [] then []
  [x, ...xs] then [f(x), ...map(f)(xs)]
```
where `...` is the spread operator that can occur it patterns or terms.

However, spreads on arrays in patterns involve slicing, and in terms involve splicing, both of which are O(n) operations. Hence, this map function which is written naturally, will have the wrong complexity.

To have this work correctly, we could instead perform slicing and splicing lazily. In this case, we will never actually have to slice the array as long as we maintain a constrained view of the array. When we build, we can also delay construction of the array, by maintaining a linear-size structure that we materialize in linear time to an array when accessed. We can perform these lazy operations simply by replacing `...` with `..` both in patterns and terms:

```
fun map(f) = case
  [] then []
  [x, ..xs] then [f(x), ..map(f)(xs)]
```

Lazy arrays make some requirements of correct use:
- Viewing and splicing should be done in disjoint phases - interleaving them will result in repeated materialization, which destroys the asymptotic performance. Where these must be interleaved, FingerTrees (#319) should be used instead.
- While lazy spreads will still work with iterables that are not arrays, they require that the iterable is constant-time indexable for reasonable performance, and don't guarantee that the output iterable is the same as the input iterable.


hkmc2/jvm/src/test/scala/hkmc2/CompileTestRunner.scala
```diff
@@ -20,45 +20,48 @@ class CompileTestRunner
...
+      .filter(_.toIO.isFile)
+      .filter(_.ext in validExt)
       
-      val preludePath = dir/"mlscript"/"decls"/"Prelude.mls"
-      
-      given Config = Config.default
+    lazy val compileTestFiles = allFiles.filter: file =>
+        file.segments.contains("mlscript-compile")
+    
+    // TODO dedup path stuff with DiffTestRunner?
```
LPTK, line 39: Should we address this TODO now?

hkmc2Benchmarks/src/test/bench/viz.html
LPTK, line 1: Please add a disclaimer that this is LLM-generated slop (IIRC).

hkmc2Benchmarks/src/test/bench/mlscript-compile/lazy-spread-runtimes/RuntimeLazyArray.mls
LPTK, line 1: Ok, I'm really not a fan of all that code duplication! All duplicated code is a heavy maintenance burden on future contributors. This one will be particularly harmful as it could easily cause silent behavior divergence.

I'd be happier if you just used a dirty hack in order to do the benchmarking. For example, why not just override the methods/values of interest at runtime in the Runtime object using `set`? This runs on the JS backend, after all :P

(To make sure this doesn't cause perf differences, you would do the overriding for all compared methods, even the default one.)

hkmc2/shared/src/test/mlscript/std/IterTest.mls
LPTK, line 1: A lot of these changes were made when I was demonstrating soem things to you and are not relevant to this test file. Please clean up.

hkmc2/shared/src/test/mlscript/basics/DynamicFields.mls
```diff
@@ -53,12 +53,13 @@ foo.(0)
 :todo
 :re
 foo.[0]
+//│ ═══[RUNTIME ERROR] TypeError: foo.at is not a function
 
 // TODO sanitize
 :todo
```
LPTK: This should be changed to
```suggestion
:re
```
and the TODO removed.

hkmc2/shared/src/test/mlscript-compile/LazyArray.mls
```diff
@@ -0,0 +1,137 @@
...
+import "./Iter.mls"
+
+open Iter
+open LazyArray
+
+fun normIdx(i, len) =
+  if i < 0 then len + i
+  else i
+
+// Assume all underlying `IterableBase`s are constant-time random access
```
LPTK: This is a very confusing comment. We shouldn't make any such assumptions. Are you trying to say what the contract of `LazyArr` should be? Or is it a contract of the `IterableBase` class?

LPTK:
@waterlens Could you please check why the LLIR tests are breaking?

hkmc2/shared/src/test/mlscript-compile/LazyArray.mls
gok99: This is rather poorly worded - It's a contract of `LazyArr` and should be made clear to users of lazy spreads. 

I'll rephrase to make it sound less like the latter.

hkmc2/shared/src/test/mlscript/basics/DynamicFields.mls
gok99: is the :todo in line 53 also resolved?

hkmc2/shared/src/test/mlscript-compile/LazyArray.mls
```diff
@@ -0,0 +1,137 @@
...
+  if i < 0 then len + i
+  else i
+
+// Assume that the underlying iterable of LazyArr is constant-time access
+class LazyArr extends IterableBase with
+
+  fun toString() = "[" + this joined(", ") + "]"
+  fun concat(...args) =
+    let l = this.length + args.reduce((acc, x) => acc + x.length, 0)
+    let a = args.reduce((acc, x) => acc.push(SpliceMarker, x); acc, [SpliceMarker, this])
```
LPTK: Prefer an iterator (avoiding lambdas) and a single iteration.

hkmc2/shared/src/test/mlscript-compile/LazyArray.mls
```diff
@@ -0,0 +1,137 @@
...
+module LazyArray with...
+
+
+fun mk(...args) = args
+
+fun concat(...args) =
+  val len = case
+    [SpliceMarker, e, ...rest] then e.length + len(rest)
+    [a, ...rest] then 1 + len(rest)
+    [] then 0
```
LPTK: Let's avoid creating many temporary arrays.

hkmc2/shared/src/test/mlscript-compile/LazyArray.mls
```diff
@@ -0,0 +1,137 @@
...
+fun materialize = case
+  Array as a then a
+  _ as itr then itr.materialize()
+
+fun slice(beg, fin) = case
+  View(itr, start, end) then View of itr, start + beg, end + fin
+  Splice(bits, l) as s then View of s.materialize(), beg, fin
+  _ as itr then View of itr, beg, fin // assume itr has O(1) at()
+
+val split = SpliceMarker
```
LPTK:
```suggestion
// This is not meant for users; it's referred to by Runtime.mls
// TODO: use `LazyArray.internals.split` when this is implemented
val __split = SpliceMarker
```

hkmc2/shared/src/test/mlscript-compile/Runtime.mls
```diff
@@ -1,10 +1,16 @@
...
+import "./LazyArray.mjs"
+import "./Iter.mjs"
+
+let lar = LazyArray
 
 
 module Runtime with ...
 
 
+val LazyArr = lar
```
LPTK: Looks like this is not really needed (along with `lar` above).

hkmc2/shared/src/test/mlscript-compile/apps/parsing/Lexer.mls
```diff
@@ -226,12 +226,12 @@ fun lex(str: Str, options) =
       None then reverse of acc
       Some(Char.Whitespace) and whitespace(idx) is idx' then
         go(idx', Token.space(idx, idx'))
-      Some("\"") then go(..string(idx + 1))
+      Some("\"") then go(...string(idx + 1))
```
LPTK, line 229: Pls add test cases showing what happens when `..` is misused.

hkmc2/shared/src/test/mlscript/basics/DynamicFields.mls
LPTK: No, it's the same.

hkmc2Benchmarks/src/test/bench/LazySpreads/LazyArrayTest.mls
```diff
@@ -0,0 +1,480 @@
...
+:silent
+suite.add("buildLazy3", () => buildLazy(3))
+suite.add("buildLazy33", () => buildLazy(33))
+suite.add("buildLazy333", () => buildLazy(333))
+suite.add("buildLazy3333", () => buildLazy(3333))
+let res = buildLazy(1000)
+//+ runtime clock time (321984): 3.496362 ms
+
+res |> funnySumEager
+//+ runtime clock time (321984): 21.602414 ms
```
LPTK: Don't forget to remove these.

hkmc2Benchmarks/src/test/bench/LazySpreads/LazyArrayTest.mls
```diff
@@ -0,0 +1,480 @@
...
+//│ = 1001000
+
+suite.add("lazyListFunnySumEager", () => 
+  let res = buildLazy(1000)
+  res |> funnySumEager)
+//│ = [object Object]
+
+let res = buildLazy(1000)
```
LPTK, line 126:
```suggestion
:silent
let res = buildLazy(1000)
```

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -80,7 +80,8 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
   def runtimeVar(using Raise, Scope): Document = getVar(State.runtimeSymbol)
   
   def argument(a: Arg)(using Raise, Scope): Document =
-    if a.spread then doc"...${result(a.value)}" else result(a.value)
+    val spd = if a.eager then "..." else "runtime.Tuple.split, "
```
LPTK: Unhygienic: use `runtimeVar`; below as well.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -495,7 +495,7 @@ enum Value extends Path:
   case Arr(elems: Ls[Arg])
   case Rcd(elems: Ls[RcdArg])
 
-case class Arg(spread: Bool, value: Path)
+case class Arg(spread: Bool, value: Path, eager: Bool = true)
```
LPTK: Make illegal states unrepresentable
```suggestion
case class Arg(spread: Opt[Bool], value: Path)
```

hkmc2Benchmarks/src/test/bench/LazySpreads/LazyArrayTest.mls
```diff
@@ -476,5 +460,5 @@ suite.add("sillyEager15", () =>
 //│ = [object Object]
 
 :silent
-suite |> runSuite
-writeResults(suite, "src/test/bench/LazySpreads/LazyArrayTest.json")
+// suite |> runSuite
```
LPTK: No longer running it?

hkmc2Benchmarks/src/test/bench/LazySpreads/LazyArrayTest.mls
```diff
@@ -1,25 +1,24 @@
...
 import "../mlscript-compile/Benchmark.mls"
+import "../../../../../hkmc2/shared/src/test/mlscript-compile/Runtime.mls"
 
 open LazyArray
 open Benchmark
+open Runtime
+
+set Runtime.Tuple.lazySlice = (xs, i, j) => LazyArray.slice(i, j)(xs)
+set Runtime.Tuple.lazyConcat = (...args) => LazyArray.concat(...args)
+set Runtime.Tuple.split = LazyArray.__split
```
LPTK: Don't forget to do the same for all tests, even if the overriding doesn't change the semantics.

hkmc2Benchmarks/src/test/bench/LazySpreads/LazyArrayTest.mls
gok99: commited accidentally, about to revert :sweat_smile:

hkmc2Benchmarks/src/test/bench/LazySpreads/LazyArrayTest.mls
gok99: This PR only contains LazyArrayTest, as the FingerTree stuff are in a separate PR. I will update the benchmark suite, after they're both merged.

Nevertheless, since the benchmarks are run from lambda functions added to the suite, and this runtime overriding isn't being timed, would we expect it to affect the benchmark numbers? (I might be misunderstanding what you mean)

LPTK: If you compare a version of the benchmarked program that uses a non-overridden Runtime with one that does, the results might be biased. Since overriding module members may have a performance effect.

> This PR only contains LazyArrayTest, as the FingerTree stuff are in a separate PR. I will update the benchmark suite, after they're both merged.

Ok.

hkmc2/shared/src/test/mlscript-compile/LazyArray.mls
```diff
@@ -4,18 +4,52 @@ import "./Iter.mls"
 open Iter
 open LazyArray
 
+// Helper Methods
 fun normIdx(i, len) =
   if i < 0 then len + i
   else i
 
+fun concatH(a, l, ...args) =
```
LPTK: Let's use more descriptive names. A comment could help as well.

hkmc2/shared/src/test/mlscript-compile/LazyArray.mls
```diff
@@ -4,18 +4,52 @@ import "./Iter.mls"
...
+  while (idx < args.length) do
+    let x = args.[idx]
+
+    set a.[init_len + idx * 2] = SpliceMarker
+    set a.[init_len + idx * 2 + 1] = x
+    set l += x.length
+    set idx += 1
+  Splice of a, l
+
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice
```
LPTK, line 26: Why is this relevant?

hkmc2/shared/src/test/mlscript-compile/LazyArray.mls
```diff
@@ -4,18 +4,52 @@ import "./Iter.mls"
...
+fun normIdxSlice(i, len) =
+  if i < 0 and
+    i >= -len
+    then len + i
+    else 0
+  else i
```
LPTK: Oof this formatting hurts my brain.
```suggestion
  if i < 0 and
    i >= -len then len + i
    else 0
  else i
```

hkmc2/shared/src/test/mlscript-compile/LazyArray.mls
```diff
@@ -4,18 +4,52 @@ import "./Iter.mls"
...
+
+fun sliceH(beg, fin, arr) =
```
LPTK: What's this mysterious `H` suffix for?

~~This is not Haskell.~~

hkmc2/shared/src/test/mlscript/std/LazyArrayTest.mls
```diff
@@ -0,0 +1,271 @@
+:js
+// :de
+// :sjs
+// :pt
+// :elt
```
LPTK: Pls clean up.

hkmc2/shared/src/test/mlscript/std/LazyArrayTest.mls
```diff
@@ -0,0 +1,271 @@
...
+
+import "../../mlscript-compile/LazyArray.mls"
+
+open LazyArray
+
+fun equals(a, b) =
```
LPTK: How about we add this to the LazyArray module?

hkmc2/shared/src/test/mlscript/std/LazyArrayTest.mls
```diff
@@ -0,0 +1,271 @@
...
+  if a.length != b.length 
+  then false
+  else
```
LPTK:
```suggestion
  a.length === b.length &&
```

hkmc2/shared/src/test/mlscript-compile/LazyArray.mls
gok99: I'm just following the index semantics of the javascript Array.slice - I'll state this in the comment.

hkmc2/shared/src/test/mlscript-compile/LazyArray.mls
gok99: They're just helpers - I'll rename them `xxxHelper` :sweat_smile:

hkmc2/shared/src/test/mlscript/std/LazyArrayTest.mls
gok99: This is not very specific to LazyArrays, so we could have an equals method in the Iter module, but that would probably use the iterator, instead of `at` that I wanted to use here.

LPTK: For that reason, we should have both, obviously!

LPTK:
It seesm most comments are addressed. Is it ready to merge? Maybe @chengluyu can still do a very quick review.

gok99:
@LPTK Should we disable / limit parallelism of the benchmarks?

hkmc2/shared/src/test/mlscript/std/LazyArrayTest.mls
gok99: Where should they go? Both in Iter.mls? It's not obvious to me why this should belong in LazyArray.

hkmc2/shared/src/test/mlscript/basics/LazySpreads.mls
```diff
@@ -0,0 +1,100 @@
...
+//│ ╔══[ERROR] Lazy spread parameters not allowed.
+//│ ║  l.81: 	fun f(..xs) = xs
+//│ ╙──      	        ^^
+
+:ge
+sum2(..arr)
+//│ ╔══[COMPILATION ERROR] Lazy spreads are not supported in call arguments
+//│ ║  l.87: 	sum2(..arr)
+//│ ╙──      	    ^^^^^^^
+//│ ═══[RUNTIME ERROR] Error: Function expected 1 argument but got 2
```
gok99, line 91: Lazy spread misuse

hkmc2/shared/src/test/mlscript/std/LazyArrayTest.mls
LPTK: The index-based one has nothign to do in `Iter`, where it would be counter-productive (could lead to the wrong complexity!).

The implementation does equate LazyArrat, so why wouldn't it belong there? We might find other places for it too, at which later point we can share the same implementation. This is fine.

LPTK:
> @LPTK Should we disable / limit parallelism of the benchmarks?

It's fine for now. They have error bars and these are small, so I'm not worried.

hkmc2/shared/src/test/mlscript/std/LazyArrayTest.mls
gok99: Sure, I've moved it. It also now uses `!==` instead of `!=`.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
```diff
@@ -505,22 +505,25 @@ class Desugarer(elaborator: Elaborator)(using Ctx, Raise, State, UnderCtx) exten
                     Split.Let(sym, callTupleGet(ref, -1 - lastIndex, sym), wrapInner(split))
                   (wrap, (sym, pat) :: matches)
             val lastMatches = reversedLastMatches.reverse
+            val sliceFn = if kw.isInstanceOf[Keyword.`..`.type] then tupleLazySlice else tupleSlice
```
LPTK: This is bug-prone. Next time these are refactored, it may silently start failign!
```suggestion
            val sliceFn = kw match
              case Keyword.`..` => tupleLazySlice
              case Keyword.`...` => tupleSlice
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
```diff
@@ -505,22 +505,25 @@ class Desugarer(elaborator: Elaborator)(using Ctx, Raise, State, UnderCtx) exten
...
-                  Split.Let(sym, app(tupleSlice, tup(fld(ref), fld(int(lead.length)), fld(int(last.length))), sym), wrapLast(split))
+                  Split.Let(sym, app(sliceFn, tup(fld(ref), fld(int(lead.length)), fld(int(last.length))), sym), wrapLast(split))
                 (wrap, (sym, pat) :: lastMatches)
           case N => (identity: Split => Split, Nil)
         val (wrap, matches) = lead.zipWithIndex.foldRight((wrapRest, restMatches)):
           case ((pat, i), (wrapInner, matches)) =>
             val sym = scrutSymbol.getTupleLeadSubScrutinee(i)
-            val wrap = (split: Split) => Split.Let(sym, Term.SynthSel(ref, Ident(s"$i"))(N), wrapInner(split))
+            val wrap = (split: Split) =>
+              // Split.Let(sym, Term.SynthSel(ref, Ident(s"$i"))(N), wrapInner(split))
```
LPTK, line 529: Why leave this comment?

hkmc2/shared/src/test/mlscript/llir/Split.mls
```diff
@@ -1,4 +1,5 @@
 :llir
+:todo
```
LPTK: Ok, le'ts ignore these LLIR failures now. But we should clarify when they started occurring.
```suggestion
:todo // Started failing after the addition of lazy spreads (#318)
```
Same in other files below.

hkmc2Benchmarks/src/test/bench/mlscript-compile/Benchmark.mls
```diff
@@ -0,0 +1,18 @@
...
+let bm = benchmark
+
+module Benchmark with...
+
+val benchmark = bm
+
+fun mkSuite() = new bm.Suite()
+
+fun runSuite(suite) =
+    let settings = {}
```
LPTK: Please change to 2-space indents (here and anywhere else 4 is used).

hkmc2/jvm/src/test/scala/hkmc2/CompileTestRunner.scala
```diff
@@ -20,45 +20,48 @@ class CompileTestRunner
   
   val pwd = os.pwd
   val workingDir = pwd
-  
-  val dir = workingDir/"hkmc2"/"shared"/"src"/"test"
+
+  val testDir = workingDir/"hkmc2"/"shared"/"src"/"test"  
+  val dirs = testDir :: workingDir/"hkmc2Benchmarks"/"src"/"test" :: Nil
```
LPTK:
```suggestion
  val mainTestDir = workingDir/"hkmc2"/"shared"/"src"/"test"  
  
  // The compilation tests currently include compiling the benchmark instrumentation code.
  val dirs = mainTestDir :: workingDir/"hkmc2Benchmarks"/"src"/"test" :: Nil
```

LPTK (approved):
Ok, it's been long enough. Let's merge this ASAP without waiting for Luyu's PR or Ruqing's fixes.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Desugarer.scala
gok99: It could still be useful to have record of the previous behavior - the breaking LLIR tests for eg, seem to be due to this change

LPTK: Ok, please add your comment above to the code, then.

PR #320, opened 2025-07-30
Rendering improvements

chengluyu:
- Support rendering cyclic objects. Cyclic objects will be displayed as `[Circular *1]`, where `1` is the number of cyclic objects in the current object.
- Support the display of various JavaScript values, for example, symbols, `BigInt`s, `WeakSet`, `WeakMap`, objects with `null` prototype, etc.
- The `render` function now accepts an optional object as its parameter.
  - `indent: Int` specifies how many spaces should each indent have.
  - `breakLength: Int`: The content will be switched to multiline mode when a line exceeds this length.

Minor changes
- In code generation, use `Object.defineProperty` to add `class` field to class constructors to make this field invisible.


hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -307,13 +307,13 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
                 if isModule
                 then doc"(${clsJS});"
                 else doc"const $clsTmp = ${clsJS}; # ${mkThis(owner)}.${sym.nme} = new ${clsTmp
-                  }; # ${mkThis(owner)}.${sym.nme}.class = $clsTmp;"
+                  }; # Object.defineProperty(${mkThis(owner)}.${sym.nme}, 'class', { value: ${clsTmp} });"
```
LPTK: Why do we want the `class` to be invisible?

In any case, this change should be made in a different PR (a micro-PR) to minimize the diff.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -369,10 +369,13 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
           // case _: semantics.ModuleSymbol => doc"=== ${result(pth)}"
           // [invariant:0] If the class represented by `cls` does not exist at
           // runtime, then `pth` is a dummy value and should be discarded.
+          case Elaborator.ctx.builtins.Unit => doc"$sd === runtime.Unit"
```
LPTK: Why is this case needed?

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -35,27 +35,350 @@ fun interleave(sep)(...args) =
...
+// * return `empty` instead.
+fun mkString(xs, f, start, sep, end, empty) =
+  let
+    it = xs.(Symbol.iterator)()
+    res = it.next()
+    buffer = null
+  // Consume the iterator until we find the first non-null value.
+  while res.done is false and buffer is null do
+    if f(res.value) is
+      Str as str do set buffer = start + str
```
LPTK: You are silently ignoring all non-string values? 🤔 

Better throw an exception, no?

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -35,27 +35,350 @@ fun interleave(sep)(...args) =
...
+    set res = it.next()
+  // If the buffer is still null, this means we have drained the iterator.
+  if buffer is null then empty else...
+  while res.done is false do
+    let value = f(res.value)
```
LPTK: It seems this `let` binding is superfluous.

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -35,27 +35,350 @@ fun interleave(sep)(...args) =
...
+    if value is Str as s do
+      set { buffer += sep, buffer += s }
```
LPTK: Is there a more efficient way of building strings in JS (like Java's `StringBuffer`)? This has the wrong complexity.

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -35,27 +35,350 @@ fun interleave(sep)(...args) =
...
+    set res = it.next()
+  set buffer += end
+  buffer
+
+// Do not wrap property names in quotes if they match the following pattern.
+let identifierPattern = new RegExp of
+  """^[\p{ID_Start}_$][\p{ID_Continue}_$]*$""", "u"
```
LPTK: Can you explain (in the doc) what's going on in this regexp? What's the `\p` for? What is `"u"` for?

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -35,27 +35,350 @@ fun interleave(sep)(...args) =
...
+  multiLineStart: "[", multiLineEnd: "]"
+  singleLineSeparator: ", ", multiLineSeparator: ",", empty: "[]"
+
+let symbolsForSet(n: Int) =
+  let headline = "Set(" + n + ")"
+  (
+    singleLineStart: headline + " { ", singleLineEnd: " }"
+    multiLineStart: headline + " {", multiLineEnd: "}"
+    singleLineSeparator: ", ", multiLineSeparator: ",", empty: headline + " {}"
+  )
```
LPTK:
```suggestion
  singleLineStart: headline + " { ", singleLineEnd: " }"
  multiLineStart: headline + " {", multiLineEnd: "}"
  singleLineSeparator: ", ", multiLineSeparator: ",", empty: headline + " {}"
```

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -35,27 +35,350 @@ fun interleave(sep)(...args) =
...
+
+// Do not wrap property names in quotes if they match the following pattern.
+let identifierPattern = new RegExp of
+  """^[\p{ID_Start}_$][\p{ID_Continue}_$]*$""", "u"
+
+fun noNeedForQuotes(id) = identifierPattern.test(id)
+
+let MIN_BREAK_LENGTH = 60
+
+let NULL_PROTO = "[Object: null prototype] "
```
LPTK: Why use this inconsistent printing syntax? We don't want to copy JS, here.

Can we make it
```suggestion
let NULL_PROTO = ""
```

and print objects with the `Object` prototype as `Object { ... }`?

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -35,27 +35,350 @@ fun interleave(sep)(...args) =
...
+    multiLineStart: headline + " {", multiLineEnd: "}"
+    singleLineSeparator: ", ", multiLineSeparator: ",", empty: headline + " {}"
+  )
+
+let symbolsForObject =
+  start: "{", end: "}", entrySeparator: ",", keyValueSeparator: ": ", empty: "{}"
+
+let symbolsForMap(n: Int) =
+  let headline = "Map(" + n + ")"
+  (
```
LPTK: Remove

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -35,27 +35,350 @@ fun interleave(sep)(...args) =
...
+    start: headline + " {", end: "}",
+    entrySeparator: ",", keyValueSeparator: " => ", empty: headline + " {}")
+
+fun render(target, ...args) =
+  let
+    options = if args is Array and args.length > 0 then args.[0] else new Object
```
LPTK: Why allocate a new object here? Can you use `()` or `[]`?

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -35,27 +35,350 @@ fun interleave(sep)(...args) =
...
+    indent = if options.indent is
```
LPTK: Isn't `options.indent` going to crash if it is undefined?

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -35,27 +35,350 @@ fun interleave(sep)(...args) =
...
+      Int as value and value > 0 then " ".repeat(value)
+      true then "  ", else null
```
LPTK:
```suggestion
      true then "  "
      else null
```

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -35,27 +35,350 @@ fun interleave(sep)(...args) =
...
+    // If the joined string is less than or equal `breakLength`, it stays on one
+    // line; otherwise it becomes multiline and each item is prefixed with `\n`
+    // plus `indent`.
+    breakLength = if
+      options.breakLength is Int as value then Math.max(MIN_BREAK_LENGTH, value)
```
LPTK: Why would we ignore `50` and use the completely arbitrary `60` here?

I would expect a `MIN_BREAK_LENGTH` to be the smallest value that's needed for rendering to be well-defined (like `1` or something like that).

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -35,27 +35,350 @@ fun interleave(sep)(...args) =
...
+      indent is null then Number.POSITIVE_INFINITY
+      else MIN_BREAK_LENGTH
```
LPTK: Shouldn't this be called `DEFAULT_BREAK_LENGTH`?

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -35,27 +35,350 @@ fun interleave(sep)(...args) =
...
+    let
+      i = 0
+      length = array.length
+      emptyItemCount = 0
+      // The length of indentations that should be prepended to items in this
+      // array if this array is displayed in multi-lines.
+      itemIndentationLength =
+        if indent is null then 0 else (level + 1) * indent.length
+    // Return `true` if the array has no remaining elements.
+    fun done() =
```
LPTK: Since this is used as a function value and is not recursive, I'd make it
```suggestion
    let done() =
```

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -35,27 +35,350 @@ fun interleave(sep)(...args) =
...
+      result
+    renderSequence of
+      done, next, level, keyLength, startPos, symbolsForSet(theSet.size)
+  fun renderSequence(done, next, level: Int, keyLength: Int, startPos: Int, symbols) =
+    let
+      items = []
+      longestItem = 0
+      hasMultiline = false
+      // Try calculating the length needed to place all items in one line. The
+      // brackets and spaces around items are excluded from the length.
```
LPTK, line 201: Why are they excluded?

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -35,27 +35,350 @@ fun interleave(sep)(...args) =
...
+      null      then "null"
+      Str       then JSON.stringify(arg)
+      BigInt    then arg.toString() + "n"
+      Symbol    and
+        arg.description is undefined then "Symbol()"
+        else "Symbol(" + arg.description + ")"
+      Num       then arg.toString()
+      true      then "true"
+      false     then "false"
+    // Exclude all
```
LPTK: What does this comment mean?

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -35,27 +35,350 @@ fun interleave(sep)(...args) =
...
+        WeakMap then "WeakMap { <items unknown> }"
+        Error then arg.name + ": " + arg.message
+        Function then
+          let desc = Object.getOwnPropertyDescriptor(arg, "prototype")
+          let head = if
+            // ES 2015 classes: `desc.writable` is false
+            desc is Object and desc.writable is false then
+              "[class " + (if arg.name is "" then "(anonymous)" else arg.name) + "]"
+            else
+              "[Function" + (if arg.name is "" then "" else " " + arg.name) + "]"
```
LPTK: Why isn't this
```suggestion
              "[function" + (if arg.name is "" then "" else " " + arg.name) + "]"
```

hkmc2/shared/src/test/mlscript/apps/IterTest.mls
```diff
@@ -11,7 +11,7 @@ open Stack { Cons, Nil }
   filtering of x => x % 2 is 0
   mapping of x => x * 2
   toArray()
-//│ = [4, 8, 12, 16, 20]
+//│ = [ 4, 8, 12, 16, 20 ]
```
LPTK: Do we really want these spaces?

LPTK (commented):
Wow, this is quite some chunky logic!

Why doesn't `CyclicValues.mls` change? That would mean it still crashes, which is unexpected.

The rendering logic is currently quite inefficient. It's based on concatenating strings together (which has the wrong complexity) and on splitting strings after the fact to make them multi-line. This is fine for now, but later we'll want to improve on this.

chengluyu:
> Why doesn't CyclicValues.mls change? That would mean it still crashes, which is unexpected.

```
data class Foo(x)
let foo = Foo(null)
set foo.x = foo
render(foo)
```

The `render` function sees `Foo` has `toString` defined, so it calls `Foo.toString`, which calls `render` itself. The `render` function works with recursive objects that are not instances of MLscript classes.

I don't know how to share `visitedObjects` across different calls of `render` functions. Do you have any suggestions?

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
chengluyu: > Is there a more efficient way of building strings in JS (like Java's `StringBuffer`)?

No. The performance of `+` and `String.concat` is actually very good.

> This has the wrong complexity.

Not really. I don't know the exact complexity, but at the very least, it’s very fast in practice, because strings in V8 are actually ropes, and `+` is optimized very efficiently. Here are two related posts I've read.

- [Javascript Does Not Need a StringBuilder](https://josephmate.github.io/java/javascript/stringbuilder/2020/07/27/javascript-does-not-need-stringbuilder.html): This post tested several methods of building strings and concluded that `+` is still the most efficient.
- [Exploring V8's strings: implementation and optimizations](https://iliazeus.lol/articles/js-string-optimizations-en/): This post reveals how strings are stored internally in V8.

chengluyu: I just realized that this is `mkString` function. It's not used now. I should have removed it.

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
chengluyu: This function (`mkString`) is not used and will be removed.

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
chengluyu: This function (`mkString`) is not used and will be removed.

hkmc2/shared/src/test/mlscript/apps/IterTest.mls
chengluyu: I prefer having clear visual margins. Adding a space can make the boundaries between arrays and objects clearer. For example, `{ y: { x: [ 5 ] } }` vs. `{y: {x: [5]}}`. Also, after adding a space, the single-line and multi-line versions will look more consistent.

```js
[ 0, 1, 2, 3 ]
[
  0,
  1,
  2,
  3
]
```

What do you think?

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
chengluyu: Good idea!

hkmc2/shared/src/test/mlscript/apps/IterTest.mls
chengluyu: Serialization (e.g., `JSON.stringify`) should have no spaces to save space. The purpose of the `Rendering` module is pretty printing, so it should make some changes to improve readability.

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
chengluyu: This comment is outdated. I will review all the comments.

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
chengluyu: This comment is incomplete. I will review all the comments.

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
chengluyu: `[]` is an array, and `()` is `Unit`, but neither seems quite right. In the end, I created a module-level empty object called `defaultOptions` to avoid creating a new object each time.

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
chengluyu: No, there's an `else` below.

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
chengluyu: Fixed in da755d7.

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
chengluyu: Now I think setting a minimum value is stupid. In da755d7, I no longer enforce a minimum value, as long as it's greater than 0.

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
chengluyu: Fixed in da755d7.

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
chengluyu: Maybe it could also be `[fun foo]`?

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -35,74 +35,67 @@ fun interleave(sep)(...args) =
...
-  while res.done is false do
-    let value = f(res.value)
-    if value is Str as s do
-      set { buffer += sep, buffer += s }
-    set res = it.next()
-  set buffer += end
-  buffer
-
-// Do not wrap property names in quotes if they match the following pattern.
+// The following regular expression matches non-empty strings that look like a
```
LPTK: Only _look like_ it? Or does it match them exactly?

LPTK: PS: According to your doc below, it does match them exactly.

LPTK: (Very good doc, BTW.)

hkmc2/shared/src/test/mlscript/basics/LetBindings.mls
```diff
@@ -54,4 +54,13 @@ let x += 1 in
...
+:fixme
+fun foo() =
+  let bar(x: Str): Str = x + x
+  bar("test")
+//│ ╔══[ERROR] Unsupported let binding shape
+//│ ║  l.59: 	  let bar(x: Str): Str = x + x
+//│ ╙──      	      ^^^^^^^^^^^^^^^^^^^^^^^^
+//│ ╔══[ERROR] Name not found: bar
+//│ ║  l.60: 	  bar("test")
+//│ ╙──      	  ^^^
```
LPTK: This test case should probably go in `mlscript/backlog/ToTriage.mls`, as
```suggestion
// To add to `mlscript/basics/LetBindings.mls`
:fixme
fun foo() =
  let bar(x: Str): Str = x + x
  bar("test")
//│ ╔══[ERROR] Unsupported let binding shape
//│ ║  l.59: 	  let bar(x: Str): Str = x + x
//│ ╙──      	      ^^^^^^^^^^^^^^^^^^^^^^^^
//│ ╔══[ERROR] Name not found: bar
//│ ║  l.60: 	  bar("test")
//│ ╙──      	  ^^^
```

hkmc2/shared/src/test/mlscript/std/RenderingTest.mls
```diff
@@ -226,26 +230,26 @@ let tree =
...
 let b = [a, a]
-//│ b = [ <ref *1> { x: [Circular *1], y: 1 }, <ref *1> { x: [Circular *1], y: 1 } ]
+//│ b = [ <ref *1> Object { x: [Circular *1], y: 1 }, <ref *1> Object { x: [Circular *1], y: 1 } ]
 
 b.push(b)
 //│ = 3
 
 b
-//│ = <ref *2> [ <ref *1> { x: [Circular *1], y: 1 }, <ref *1> { x: [Circular *1], y: 1 }, [Circular *2] ]
+//│ = <ref *2> [ <ref *1> Object { x: [Circular *1], y: 1 }, <ref *1> Object { x: [Circular *1], y: 1 }, [Circular *2] ]
```
LPTK: I think we should ideally render this as
```suggestion
//│ = (val ref'2 = [(val ref'1 = Object { x: ref'1, y: 1 }), (val ref'1 = Object { x: ref'1, y: 1 }), ref'2)]
```
(And we should make that valid MLscript syntax later.)

chengluyu: The parentheses in your example do not match. Did you mean:
```
(
  val ref'2 = [
    (
      val ref'1 = Object { x: ref'1, y: 1 }
    ),
    (
      val ref'1 = Object { x: ref'1, y: 1 }
    ),
    ref'2
  ]
)
```

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
chengluyu: Thanks.

hkmc2/shared/src/test/mlscript/apps/IterTest.mls
chengluyu: I made this an option.

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
chengluyu: This is fixed in efc5461.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
chengluyu: Because I want to check if a values is `Unit`. Is there a way to do that without adding this line?

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -288,11 +316,15 @@ fun render(target, ...args) =
...
       push of ...next(items.length is 0, singleLineLength)
     // Complete the single line length.
-    set singleLineLength += 2 // " }"
+    set singleLineLength += symbols.end.length +
+      (if symbols.padding then padding.length else 0)
     let
       layoutMultiLine = () => symbols.start + "\n" + itemIdent + items.join(",\n" + itemIdent)
         + "\n" + parentIndent + symbols.end
-      layoutSingleLine = () => symbols.start + " " + items.join(", ") + " " + symbols.end
+      layoutSingleLine = () =>
```
LPTK:
```suggestion
      layoutSingleLine() =
```

hkmc2/shared/src/test/mlscript-compile/Symbols.mls
```diff
@@ -0,0 +1,7 @@
+
+// Well-known symbols used by MLscript internally.
+object Symbols with ...
+
+// Each class should have a method defined at this symbol. The method should
+// return the class name and the value fields of the class.
+val constructorNames = Symbol()
```
LPTK: This is the one that should be used, but it currently seems unused.

hkmc2/shared/src/test/mlscript/std/RenderingTest.mls
LPTK: Yes, ofc.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
LPTK: What happens if you don't have this line? If you do a normal instance test?

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -22,7 +23,7 @@ abstract class CodeBuilder:
   type Context
   
 
-class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
+class JSBuilder(var internalSymbols: Opt[SeqMap[Str, (Symbol, Str)]] = N)(using TL, State, Ctx) extends CodeBuilder:
```
LPTK: I don't know how many time I'll have to say it. Again, and again, and again, don't put default arguments in core logic!

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -466,7 +481,19 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
         then "./" + os.Path(path).relativeTo(wd).toString
         else path
       doc"""import ${getVar(i._1)} from "${relPath}";"""
-    imps.mkDocument(doc" # ") :/: block(p.main, endSemi = false).stripBreaks :: (
+    val syms = internalSymbols match
```
LPTK: I'm against this overcomplicated mess. I think `program` should always create the bindings in the output. And the symbols should just be stored in `State`, just like `runtimeSymbol`. There's no reason to create new symbols every time and use mutable state like this.

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
LPTK: Ok, so you reverted that idea. What's the current plan for constructor-less objects?

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
LPTK: I'm wondering about the printing of such values. I think we should use a syntax that feels consistent with the rest of the lanfuage. `[function foo]` and `[class Bar] { ... }` make no sense in MLscript.

How about we just make them `fun foo` and `class Bar { ... }`?

hkmc2/shared/src/test/mlscript/backlog/ToTriage.mls
```diff
@@ -413,3 +413,14 @@ w("whoops")
...
 
+fun foo() =
+  let bar(x: Str): Str = x + x
+  bar("test")
+//│ ╔══[ERROR] Unsupported let binding shape
+//│ ║  l.417: 	  let bar(x: Str): Str = x + x
+//│ ╙──       	      ^^^^^^^^^^^^^^^^^^^^^^^^
+//│ ╔══[ERROR] Expected 0 arguments, got 1
+//│ ║  l.418: 	  bar("test")
+//│ ╙──       	     ^^^^^^^^
```
LPTK: You only applied part of my comment, please be more careful.

https://github.com/hkust-taco/mlscript/pull/320#discussion_r2246757955

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
chengluyu: I agree on this. It looks more MLscripty.

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
chengluyu: Yes, I did. You told me to make the rendering output more MLscripty. I think the idea of “objects whose prototype is `null`” only exists in JavaScript. So I no longer plan to distinguish between objects created using object literals (which should be the same as what you call constructor-less objects) and objects with a `null` prototype.

hkmc2/shared/src/test/mlscript-compile/Runtime.mjs
```diff
@@ -11,52 +14,66 @@ let Runtime1;
...
-      constructor(captures) {
-        this.captures = captures;
+    Object.defineProperty(this.MatchResult, 'class', {
+    value: class MatchResult {
+        constructor(captures) {
+          this.captures = captures;
+        }
+        static [constructorName]() { return "MatchResult"; }
+        static [fieldNames]() { return ["captures"]; }
+        static [definitionKind]() { return "class"; }
```
LPTK: As discussed, to avoid needless allocations and lookups, this should be something like
```suggestion
        static [constructorInfo] = return ["class", "MatchResult", "captures"]
```
Is there any reason to make it a function?

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
chengluyu: If I comment this line, it would be like

```
fun f(x) = if x is Unit then 1 else 0
//│ JS (unsanitized):
//│ let f; f = function f(x) { if (x instanceof globalThis.Unit) { return 1 } else { return 0 } };

f of ()
//│ JS (unsanitized):
//│ f(runtime.Unit)
//│ FAILURE LOCATION: mkQuery (JSBackendDiffMaker.scala:155)
//│ ═══[RUNTIME ERROR] Error: Access to required field 'Unit' yielded 'undefined'
```

The condition should be `x instanceof runtime.Unit.class` or `x === runtime.Unit`.

OK. I realized I can use `x === ()` instead. But there should be alternative ways to check if a values is `Unit`.

hkmc2/shared/src/test/mlscript-compile/Runtime.mjs
chengluyu: It should be
```suggestion
        static [constructorInfo] = ["class", "MatchResult", ["captures"]];
```
because the class can be parameter-list-less and we need a way to know if a parameter list is there or not.

hkmc2/shared/src/test/mlscript/backlog/ToTriage.mls
chengluyu: Fixed in 5eff8ff.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
chengluyu: Fixed in 04bacf5.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
chengluyu: > I think `program` should always create the bindings in the output.

Here's the thing: since `JSBackendDiffMaker` uses `program` to generate the output of the `:sjs` command, I chose to output it only once to avoid affecting too many test files.

> And the symbols should just be stored in `State`

So is it okay for me to do this? I think these `Symbol` things are unique to JavaScript. Other backends don't need them, so I didn't put them into `State`.

> There's no reason to create new symbols every time and use mutable state like this.

No, I don't create new symbols every time. In diff tests, the symbols are created by `JSBackendDiffMaker`, and each test file is created once. In the compiler tests, symbols are also created only once for each file.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
LPTK: Why not just `x is ()`?

LPTK: "Unit" is an implementation detail...

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
LPTK: Alternatively, how about we make
```
{ a: 1, b: 2 }
```
create a `null`-prototype object, and
```
new Object with { val a = 1, val b = 2 }
```
create a proper instance of `Object`.

It would make sense that plain record literals don't get polluted with prototype crap. Any downside to this?

LPTK: (If we do this, then we'll want to display `Object` instances as before your latest change.)

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
LPTK: Yes it's fine to put them there.

Just separate the part to initialize the symbols form the rest of `program`'s logic, and use the appropriate thing in DiffTest. It's a tiny refactoring, and that's a million times better than adding some pointless mutable state.

hkmc2/shared/src/test/mlscript-compile/Runtime.mjs
LPTK: Ok, fine.

hkmc2/shared/src/test/mlscript/std/RenderingTest.mls
chengluyu: Implemented in bc4c3c9.

chengluyu:
Following the discussion in https://github.com/hkust-taco/mlscript/pull/320#discussion_r2249940929, I don't know how to achieve the goal after debugging for a while.

At first, I create the symbol for JavaScript `Symbol` in `State`.
```scala
  class State:
    val suid = new Uid.Symbol.State
    given State = this
    val globalThisSymbol = TopLevelSymbol("globalThis")
    // In JavaScript, `import` can be used for getting current file path, as `import.meta`
    val importSymbol = new VarSymbol(syntax.Tree.Ident("import"))
    val runtimeSymbol = TempSymbol(N, "runtime") // <-------------- HERE
    val definitionMetadataSymbol = TempSymbol(N, "definitionMetadata")
```

At this point, my goal is to generate statements like `const definitionMetadata = Symbol.for("definitionMetadata")` in the beginning of `program`’s output. I tried a few ways, but I don’t know how to make the name `Symbol` hygienic.

### Approach 1

At first, I tried this.

```scala
  def program(p: Program, exprt: Opt[BlockMemberSymbol], wd: os.Path)(using Raise, Scope): Document =
    reserveNames(p)
    // [... a few of lines code are omitted]
    val syms = doc"""const ${scope.lookup_!(State.definitionMetadataSymbol)} = ${result(Value.Ref(ctx.builtins.Symbol))}.for("mlscript.definitionMetadata");"""
    imps.mkDocument(doc" # ") :/: syms :/: block(p.main, endSemi = false).stripBreaks :: (
      exprt match
        case S(sym) => doc"\nlet ${sym.nme} = ${scope.lookup_!(sym)}; export default ${sym.nme};\n"
        case N => doc""
      )
```

It reported a compilation error.

```
[info] ═══[INTERNAL ERROR] `this` not in scope: class:Symbol
[error] Unexpected error in watcher: java.lang.Exception: Internal Error: Program reached an unexpected state.
[error] java.lang.Exception: Internal Error: Program reached an unexpected state.
[error]         at mlscript.utils.package$.lastWords(package.scala:230)
[error]         at mlscript.utils.package$.die(package.scala:229)
[error]         at hkmc2.utils.Scope.thisError(Scope.scala:44)
[error]         at hkmc2.utils.Scope.getParent$1$$anonfun$1$$anonfun$1(Scope.scala:54)
[error]         at scala.Option.fold(Option.scala:263)
```

### Approach 2

Then I tried to synthesize the variable definitions in `MLsCompiler`. 

```scala
     val blk = new semantics.Term.Blk(
        semantics.Import(State.runtimeSymbol, runtimeFile.toString) ::
          semantics.Import(State.termSymbol, termFile.toString) ::
          // Generate `definitionMetadata = Symbol.for("mlscript.definitionMetadata")`.
          semantics.LetDecl(State.definitionMetadataSymbol, Nil) ::
          semantics.DefineVar(State.definitionMetadataSymbol, {
            import syntax.Tree, semantics.{FlowSymbol, Term}
            val symbolRef = newCtx.builtins.Symbol.ref().withIArgs(Nil)
            val flowSymbol = FlowSymbol("definitionMetadataSymbol")
            Term.App(
              Term.SynthSel(symbolRef, Tree.Ident("for"))(N).withIArgs(Nil),
              Term.Lit(Tree.StrLit("mlscript.definitionMetadata"))
            )(Tree.App(Tree.Dummy, Tree.Dummy), N, flowSymbol).withIArgs(Nil)
          }) :: blk0.stats,
        blk0.res
      )
```

The same problem happened.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
chengluyu: > Why not just `x is ()`?

`()` is not elaborated as any pattern now.

```
() is ()
//│ FAILURE LOCATION: expandMatch (Desugarer.scala:615)
//│ ╔══[ERROR] Unrecognized pattern (unit)
//│ ║  l.11: 	() is ()
//│ ╙──      	      ^^
//│ = false
```

chengluyu: Actually, I'm not sure if using `()` as the unit pattern is misleading or not. If `(a: x)` is a record pattern, then `()` can be understood as an empty record pattern, which should match any objects.

LPTK:
It's because you're trying to use a `Value.Ref(ctx.builtins.Symbol)`, where `builtins.Symbol` is a `ClassSymbol`. References to inner symbols like this are currently treated as self-reference (hence "`this` not in scope").

Now, using the appropriate member symbol instead (which you can get with `assumeBuiltin` instead) does not work either, because currently we don't actually put these symbols in the Scope. Maybe we should. Currently a user-written `Symbol` is lowered to `globalThis.Symbol`. So you can use the same approach, here.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
LPTK: Ok, and since we're not supposed to match this thing anyway, it doesn't matter.

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
chengluyu: The problem is the efficiency of the creation of objects with a `null` prototype. I made a benchmark: https://measurethat.net/Benchmarks/Show/34989/0/various-ways-of-object-creation

The results show that even the fastest way to create an object with a `null` prototype is 10x slower than the regular method of creating an object with the same properties.

If it's just one or two objects, it doesn't matter. But what if every object is 10x slower to create? Our program could become 10 times slower as a result, considering that object creation is everywhere in JavaScript.

hkmc2/shared/src/main/scala/hkmc2/MLsCompiler.scala
```diff
@@ -86,14 +85,25 @@ class MLsCompiler(preludeFile: os.Path, mkOutput: ((Str => Unit) => Unit) => Uni
...
       resolver.traverseBlock(blk0)(using Resolver.ICtx.empty)
       val blk = new semantics.Term.Blk(
-        semantics.Import(State.runtimeSymbol, runtimeFile.toString) :: semantics.Import(State.termSymbol, termFile.toString) :: blk0.stats,
-        blk0.res
+        semantics.Import(State.runtimeSymbol, runtimeFile.toString) ::
+          semantics.Import(State.termSymbol, termFile.toString) ::
+          // Generate `definitionMetadata = Symbol.for("mlscript.definitionMetadata")`.
+          semantics.LetDecl(State.definitionMetadataSymbol, Nil) ::
+          semantics.DefineVar(State.definitionMetadataSymbol, {
+            import syntax.{Fun, Tree}, Tree.{Dummy, DummyTup, Ident, StrLit}, semantics.{FlowSymbol, Term, PlainFld}
```
LPTK: Why not just `import syntax.*, syntax.Tree.*`?

hkmc2/shared/src/main/scala/hkmc2/MLsCompiler.scala
```diff
@@ -86,14 +85,25 @@ class MLsCompiler(preludeFile: os.Path, mkOutput: ((Str => Unit) => Unit) => Uni
...
+            )(Tree.App(Dummy, Dummy), S(ctx.builtins.Symbol.`for`), FlowSymbol("definitionMetadata")).withIArgs(Nil)
+          }) :: blk0.stats,
+      blk0.res
       )
       val low = ltl.givenIn:
         new codegen.Lowering()
           with codegen.LoweringSelSanityChecks
       val jsb = ltl.givenIn:
-        codegen.js.JSBuilder(N)
+        new codegen.js.JSBuilder
```
LPTK: Is this change necessary?

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -480,19 +465,8 @@ class JSBuilder(var internalSymbols: Opt[SeqMap[Str, (Symbol, Str)]])(using TL,
...
-      case N =>
-        // Otherwise, we need to create them and declare them.
-        val symbols = createInternalSymbols
-        internalSymbols = S(symbols)
-        symbols.map:
-          case (key, (_, name)) =>
-            doc"const $name = Symbol.for(${s"mlscript.$key".escaped});"
-        .toList
-    (imps ::: syms).mkDocument(doc" # ") :/: block(p.main, endSemi = false).stripBreaks :: (
+    // val syms = doc"""const ${scope.lookup_!(State.definitionMetadataSymbol)} = ${result(Value.Ref(ctx.builtins.Symbol))}.for("mlscript.definitionMetadata");"""
```
LPTK: Leftover comment.

hkmc2/shared/src/main/scala/hkmc2/utils/Scope.scala
```diff
@@ -53,6 +53,8 @@ class Scope
       else thisError(thisSym)
     )
     curThis match
+    case S(S(sym: TopLevelSymbol)) if sym === State.globalThisSymbol =>
+      "globalThis" // `this` refers to `undefined` in strict mode
```
LPTK:
```suggestion
      // `this` at the top level evaluates to `undefined` in strict mode.
      // We need this because generate some code that uses `this`/`globalThis`,
      // for example the symbol loading code.
      "globalThis"
```

hkmc2/shared/src/test/mlscript/std/RenderingTest.mls
```diff
@@ -296,14 +296,28 @@ renderAsStr of
...
+
+// Parameter-less classes should be rendered as `class`.
+renderAsStr of Phantom
+//│ = "class Phantom"
+
 data class Point(x: Num, y: Num)
 
+// While classes with parameters should be rendered as `fun`.
+renderAsStr of Point
+//│ = "fun Point"
```
LPTK: Would it make sense to show
```suggestion
//│ = "fun Point { class: class Point }"
```
which makes it clear what the desugaring is?

hkmc2/shared/src/test/mlscript/std/RenderingTest.mls
```diff
@@ -316,3 +330,118 @@ open Stack
...
+  val y = 0
+
+renderAsStr of new Origin
+//│ = "Origin {x: 0, y: 0}"
+
+class Point(x: Num, y: Num) with
+  val x = x
+  val y = y
+
+// Should we display the members?
```
LPTK:
```suggestion
// Note that we do not display members defined in the body.
// Some classes define lots of members, so doing so could be problematic.
```

hkmc2/shared/src/test/mlscript/std/RenderingTest.mls
```diff
@@ -316,3 +330,118 @@ open Stack
...
+renderAsStr of new Point(0, 1)
+//│ = "Point()"
```
LPTK: Feels like we should render this as
```suggestion
//│ = "Point(_, _)"
```
to match the way this would be matched as a patterm.

(In general, I think the syntax pretty-printed stuff mostly corresponds to the syntax of corresponding patterns.)

LPTK: Hmm if that's really the case, it feels like we might want to display
```fs
//│ = "(val ref'1 = SelfRef {a: ref'1, b: 1})"
```
as
```fs
//│ = "SelfRef {a: ref'1, b: 1} as ref'1"
```
instead, corresponding to a pattern or type (this should be both a valid pattern and a valid type).

This also has the advantage of showing the more important stuff first (the object's constructor).

What do you think?

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
LPTK: What, that's crazy! Thanks for benchmarking it.

I'm guessing the null-prototype objects are not treated as class instances by the JIT (which is how JITs represent objects efficiently).

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -14,6 +14,8 @@ import hkmc2.codegen.Value.Lam
 
 import Scope.scope
 import hkmc2.syntax.Tree.UnitLit
+import collection.immutable.SeqMap
```
LPTK: Unusued?

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -302,17 +304,12 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
                     doc" # ${mtdPrefix}get ${td.sym.nme}() ${ braced(body(bod, endSemi = true)) }"
                 .mkDocument(" ")
               }${
-                if mtds.exists(_.sym.nme == "toString")
```
LPTK, line 305: I thing we should still generate a `toString` method if one isn't set, and that method should call the rendering method in Rendering.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
LPTK: ?

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -619,7 +619,7 @@ object JSBuilder:
         then c.toString
         else f"\\u${c.toInt}%04X"
     }.mkString
-  
+
```
LPTK, line 652: 👍

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -1,7 +1,8 @@
-
+import "./RuntimeJS.mjs"
```
LPTK, line 1:
```suggestion
import "./RuntimeJS.mjs"

```

hkmc2/shared/src/test/mlscript-compile/Rendering.mls
```diff
@@ -1,7 +1,8 @@
...
 
 module Rendering with ...
 
+let definitionMetadataSymbol = RuntimeJS.symbols.definitionMetadata
```
LPTK, line 7: You removed a padding empty line.
```suggestion

let definitionMetadataSymbol = RuntimeJS.symbols.definitionMetadata
```

hkmc2/shared/src/test/mlscript/backlog/NewWith.mls
```diff
@@ -78,13 +78,13 @@ new Foo(123) with { fun foo = 27 }
 :re
 new (Foo(123)) with { fun foo = 27 }
 //│ > Foo 123 124
-//│ ═══[RUNTIME ERROR] TypeError: Class extends value [object Object] is not a constructor or null
+//│ ═══[RUNTIME ERROR] TypeError: Class extends value #<Foo> is not a constructor or null
```
LPTK: Do you know where this weird hash-angles syntax `#<Foo>` comes from?

hkmc2/shared/src/test/mlscript/basics/LetBindings.mls
```diff
@@ -54,4 +54,3 @@ let x += 1 in
 //│ > 1
 //│ = 1
 
-
```
LPTK: 👀

LPTK (commented):
Very nice contribution, thanks! I just have a few comments.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
chengluyu: *Unused

hkmc2/shared/src/test/mlscript/backlog/NewWith.mls
chengluyu: It’s from Node.js. This is my first time seeing it too.

```
Welcome to Node.js v24.3.0.
Type ".help" for more information.
> class Hello {}
undefined
> class Greetings extends (new Hello) {}
Uncaught TypeError: Class extends value #<Hello> is not a constructor or null
>
```

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -619,7 +617,7 @@ object JSBuilder:
         then c.toString
         else f"\\u${c.toInt}%04X"
     }.mkString
-    
+
```
LPTK: Hey!! I was not sarcastic about my liking that change! Let's keep the indent.

hkmc2/shared/src/main/scala/hkmc2/utils/Scope.scala
```diff
@@ -54,7 +54,10 @@ class Scope
     )
     curThis match
     case S(S(sym: TopLevelSymbol)) if sym === State.globalThisSymbol =>
-      "globalThis" // `this` refers to `undefined` in strict mode
+      // `this` at the top level evaluates to `undefined` in strict mode.
+      // We need this because generate some code that uses `this`/`globalThis`,
```
LPTK:
```suggestion
      // We need this because we generate some code that uses `this`/`globalThis`,
```

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
chengluyu: The `Rendering` module calls `toString` and use it output as the result if it is defined, which has higher precedence than checking `definitionMetadata`. 🤔 Hence, the change will cause stack overflow.

LPTK: This doesn't seem like a real problem. Can't we pass a flag to the render function that says not to call `toString`?

The reason we want to define `toString` is to make MLscript definitions "feel" like they work consistently, incl. when inserted into string interpolators etc.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
chengluyu: First of all, I changed my mind that I made the `class` property on class constructor functions enumerable. But the `class` property on patterns, objects, and modules should remain non-enumerable, as we don't want to see `pattern Foo {class: class Foo}` or `object Test {class: class Test}`, which looks confusing.

A reason to use `defineProperty` is that it prevents `class` property from being overwritten and deleted.

```
> let foo = {}
undefined
> Object.defineProperty(foo, "hello", { value: 0 })
{}
> foo.hello
0
> foo.hello = 1
1
> foo.hello
0
> delete foo.hello
false
> foo.hello
0
```

It is slower than direct assignments. But the number of classes is fixed.

I know that `with` can be used to create new anonymous classes, but people usually don't create so many new classes, right? I believe the cost of `defineProperty` is less than the cost of class creation.

LPTK: Ok, it somewhat makes sense.

Relatedly, we may also want to `Object.freeze` these class/module/object values anyway, no? I'm working on mutability changes in another branch, and considering doing this. Then, there should no longer be a concern about deleting/overriding stuff.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
chengluyu: The problem is that we can’t tell whether a `toString` is an automatically generated version that calls the `render` function, or a user-defined version. Hence, we should not call its `toString` in `render` at all, otherwise it might cause stack overflow because two `render` calls don’t share the `visitedObjects` map.

> Can't we pass a flag to the render function that says not to call `toString`?

No, in this case, it should not be called by default.

> The reason we want to define `toString` is to make MLscript definitions "feel" like they work consistently, incl. when inserted into string interpolators etc.

I agree. We can just don’t call `toString` in `Rendering` module, as it also hinders the print width limitation in pretty-printing.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
chengluyu: > Relatedly, we may also want to `Object.freeze` these class/module/object values anyway, no?

That would be nice. We should do it.

hkmc2/shared/src/test/mlscript/std/RenderingTest.mls
chengluyu: Both `Point(_, _)` syntax and `SelfRef {a: ref'1, b: 1} as ref'1` are done.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
LPTK: Ok, not calling `toString` when rendering seems to be the more reasonable approach for now.

LPTK (approved):
Awesome, LGTM now, thanks!

PR #321, opened 2025-08-04
Enforce that only Arrays (and array-like constant time indexables) are used with LazyArrays

gok99:
In #318, the following requirement was made of users of lazy spreads:
> While lazy spreads will still work with iterables that are not arrays, they require that the iterable is constant-time indexable for reasonable performance, and don't guarantee that the output iterable is the same as the input iterable.

However, since not adhering to this requirement could silently result in reasonable-looking programs having bad asymptotic complexity, let's enforce at runtime that only Arrays and some common array-like constant time indexables are used with LazyArrays.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
@@ -383,6 +383,7 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
           case Elaborator.ctx.builtins.Num => doc"typeof $sd === 'number'"
           case Elaborator.ctx.builtins.Bool => doc"typeof $sd === 'boolean'"
           case Elaborator.ctx.builtins.Int => doc"globalThis.Number.isInteger($sd)"
+          case Elaborator.ctx.builtins.TypedArray => doc"globalThis.ArrayBuffer.isView($sd) && !($sd instanceof globalThis.DataView)"
           case _ => doc"$sd instanceof ${result(pth)}"
         case Case.Tup(len, inf) => doc"$runtimeVar.Tuple.isArrayLike($sd) && $sd.length ${if inf then ">=" else "==="} ${len}"
         case Case.Field(n, safe = false) =>
--- a/hkmc2/shared/src/test/mlscript-compile/LazyArray.mls
+++ b/hkmc2/shared/src/test/mlscript-compile/LazyArray.mls
@@ -9,6 +9,10 @@ fun normIdx(i, len) =
   if i < 0 then len + i
   else i
 
+fun enforceArray = case
+  Array | LazyArr as a then a
+  ow then throw Error("Expected an Array, got: " + ow.toString())
+
 fun concatHelper(a, l, ...args) =
   let idx = 0
   let init_len = a.length
@@ -17,7 +21,7 @@ fun concatHelper(a, l, ...args) =
     let x = args.[idx]
 
     set a.[init_len + idx * 2] = SpliceMarker
-    set a.[init_len + idx * 2 + 1] = x
+    set a.[init_len + idx * 2 + 1] = enforceArray(x)
     set l += x.length
     set idx += 1
   Splice of a, l
@@ -44,7 +48,7 @@ fun sliceHelper(beg, fin, arr) =
   if arr is
     View(itr, start, end) then View of itr, start + beg, end + (len - fin)
     Splice(bits, l) as s then View of s.materialize().slice(beg, fin), 0, 0
-    _ as itr then View of itr, beg, (arr.length - fin)
+    _ as itr then View of enforceArray(itr), beg, (arr.length - fin)
 
 // Marker base class for View and Splice
 class LazyArr extends IterableBase with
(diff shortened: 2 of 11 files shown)


hkmc2/shared/src/test/mlscript-compile/apps/parsing/Keywords.mls
```diff
@@ -143,7 +143,7 @@ fun extended = all
 
 pattern Letter = "a" ..= "z" | "A" ..= "Z"
 
-fun hasLetter(s) = [..s].some((ch, _, _) => ch is Letter)
+fun hasLetter(s) = [...s].some((ch, _, _) => ch is Letter)
```
LPTK: Should we also allow strings as well as arrays? They are also known to be efficiently indexable. In fact, we could include all such standard indexable collections, including other JS types like TypedArray implementations (which you can [apparently](https://stackoverflow.com/a/29651223/1518588) test with `ArrayBuffer.isView`).

LPTK (commented):
Great, thanks! Just a minor request to generalize it a bit.

hkmc2/shared/src/test/mlscript-compile/apps/parsing/Keywords.mls
gok99: Sure, I think it makes sense to allow also for strings and `ArrayBuffer.isView`s  - I'm unable to find many other standard examples...

However, the highlighted example here really has nothing to benefit from laziness since it traverses the result immediately, and probably will be harmed by the overheads - so it should just be using an eager spread.

LPTK: In fact, the highlighted example is dumb 😁

It converts the string to an array just to test the existence of something, which is incredibly inefficient. It should use `Iter.some` from our standard library instead.

hkmc2/shared/src/test/mlscript-compile/LazyArray.mls
```diff
@@ -10,7 +10,8 @@ fun normIdx(i, len) =
   else i
 
 fun enforceArray = case
-  Array | LazyArr as a then a
+  Array | String | LazyArr as a then a
+  a and typeof(a) === "string" || globalThis.ArrayBuffer.isView(a) then a
```
LPTK:
```suggestion
  Array | String | LazyArr | Str as a then a
  a ArrayBuffer.isView(a) then a
```
where `ArrayBuffer` should be added to `Prelude.mls`.

LPTK: Even better: we can also add `TypedArray` to the Prelude and treat it as a virtual class – see https://github.com/hkust-taco/mlscript/blob/87509a4bf3eab3c6604df6539cee675adac5ddbf/hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala#L378-L386

hkmc2/shared/src/test/mlscript/std/LazyArrayTest.mls
```diff
@@ -257,15 +257,24 @@ s.slice(0, 0)
...
 
+slice(1, 1)("hello")
+//│ = [e, l, l]
+
 val mm = MutMap.empty
 //│ mm = MutMap(Map{})
 
 mm |> MutMap.insert of "a", 1
 
+val uint8Arr = new globalThis.Uint8Array([1, 2, 3, 4, 5])
```
LPTK: `Uint8Array` and family should also be added to Prelude.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -383,6 +383,7 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
           case Elaborator.ctx.builtins.Num => doc"typeof $sd === 'number'"
           case Elaborator.ctx.builtins.Bool => doc"typeof $sd === 'boolean'"
           case Elaborator.ctx.builtins.Int => doc"globalThis.Number.isInteger($sd)"
+          case Elaborator.ctx.builtins.TypedArray => doc"globalThis.ArrayBuffer.isView($sd) && !($sd instanceof globalThis.DataView)"
```
gok99, line 386: > Even better: we can also add TypedArray to the Prelude and treat it as a virtual class

Do you mean something like this? (`TypedArray` is not an exposed type)

I realized that `DataView`s are not iterables which could cause issues when used with `LazyArray`, so actually `TypedArray`s alone are what we want to allow.

hkmc2/shared/src/test/mlscript/llir/nofib/awards.mls
```diff
@@ -88,7 +88,7 @@ fun competitors(i) =
 
 fun testAwards_nofib(n) =
   map(x => print(findallawards(competitors(intMod(x, 100)))), enumFromTo(1, n))
-//│ ═══[COMPILATION ERROR] Name not found: member:findallawards
+//│ ═══[COMPILATION ERROR] Name not found: member:print
```
gok99: ~~I have no clue what causes these changes...~~ Oh there were some breaking tests I didn't notice

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
LPTK: Yes, that's exactly what I meant!

LPTK (approved):
Great, thanks!

PR #322, opened 2025-08-04
Auxiliary Parameter List Changes & Some Documentation

CAG2Mark:
Changed the way auxiliary parameters are handled to conform with the intended behavior.
- Classes without a main parameter list, i.e. `paramsOpt` is `None`, will no longer have a companion function, nor will they have a `.class` field, regardless of whether they have an auxiliary parameter list.
- The codegen will no longer assign auxiliary parameters automatically. This is now done in the ctor explicitly, just like with normal fields.
- Documented the semantics of `paramsOpt`, `auxParams`, `privateFields` and `publicFields`.


hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -360,6 +360,36 @@ final case class ValDefn(
...
+  If there is no such parameter list, for example `class A`, then paramsOpt will be None.
+  
+  auxParams are the secondary parameter lists, and in the future, will be defined using the syntax
+  
+  class A with
+    constructor(plist1)(plist2) = ...
+  
+  with the difference being that they are not printed in the class's toString function. If paramsOpt is None,
+  the class won't have a `.class` field and must be instantiated using `new`. Otherwise, it can be instantiated
+  using either a function call or `new A.class`. The first parameter list will always be passed to `paramsOpt`,
```
```suggestion
  using either a function call or `new A` or even `new A.class` (the latter is the recommended way when done in JS). The first parameter list will always be passed to `paramsOpt`,
```

LPTK (approved):
Otherwise LGTM, thanks!

PR #323, opened 2025-08-06
Make immutability the default and add opt-in `mut` modifier

LPTK:
WIP

* Modules and objects are frozen upon creation
* Instances of parameterized classes are frozen upon invocation
* `new` creates a frozen object; `new mut` does not
* `[...]` creates a frozen array; `mut [...]` does not
* `{ ... }` creates a frozen record; `mut { ... }` does not
* `val` creates an immutable field unless the object is instantiaed with `new mut` – then, it becomes mutable
* `mut val` creates accessors around a private mutable field; such fields can always be mutated

Some of the lifter tests fail because the lifter logic is not fully adapted to the changes.


chengluyu:
I made a very simple benchmark of performance of freezing arrays and objects.


Test case name | Result
-- | --
Freeze arrays | Freeze arrays x 30,344,254 ops/sec ±1.83% (64 runs sampled)
Arrays | Arrays x 88,782,406 ops/sec ±4.67% (54 runs sampled)
Freeze objects | Freeze objects x 30,461,158 ops/sec ±1.57% (65 runs sampled)
Objects | Objects x 88,484,189 ops/sec ±5.51% (57 runs sampled)

https://measurethat.net/Benchmarks/Show/35002/0/freezing-objects-and-arrays

LPTK:
@chengluyu Thanks, but your micro-benchmark is probably only quantifying the overhead of a single function call, not testing anything of interest. It is worth noting that we are indeed calling an additional function, but then again, we're already calling functions all over the place.

The good thing is that we can easily disable freezing with a compiler flag, which will allow us to make a real benchmark.

hkmc2/shared/src/test/mlscript/std/RenderingTest.mls
```diff
@@ -387,8 +387,10 @@ pattern Null = null
 renderAsStr of Null
 //│ = "pattern Null"
 
+// FIXME?
+:re
 renderAsStr of Null.class
-//│ = "pattern Null"
+//│ ═══[RUNTIME ERROR] Error: Access to required field 'class' yielded 'undefined'
```
LPTK, line 393: @chengluyu No idea what happened, here.

hkmc2/shared/src/test/mlscript-compile/Predef.mjs
```diff
@@ -107,7 +107,7 @@ let Predef1;
   static notImplemented(msg) {
     let tmp;
     tmp = "Not implemented: " + msg;
-    throw globalThis.Error(tmp);
+    throw globalThis.Error(tmp)
```
chengluyu, line 110: Just curious, why did the `;` disappear here?

hkmc2/shared/src/test/mlscript/HkScratch.mls
```diff
@@ -8,3 +8,5 @@
 // :d
 // :todo
 
+
+
```
chengluyu, line 12: Is this change intentional?

hkmc2/shared/src/test/mlscript/ucs/syntax/Empty.mls
```diff
@@ -5,7 +5,7 @@
 :re
 if {}
 //│ JS (unsanitized):
-//│ throw new globalThis.Error("match error");
+//│ throw globalThis.Object.freeze(new globalThis.Error("match error"))
```
chengluyu, line 8: Do we really even need to freeze instances of built-in classes?

chengluyu (approved):
I have a few minor questions. Other than that, LGTM.

hkmc2/shared/src/test/mlscript/ucs/syntax/Empty.mls
LPTK: Sure, why not.

hkmc2/shared/src/test/mlscript/HkScratch.mls
LPTK: Nope, but I actually prefer it this way 🤓

hkmc2/shared/src/test/mlscript-compile/Predef.mjs
LPTK: Because of this change: https://github.com/hkust-taco/mlscript/pull/323/files#diff-beca2e9c53b86930c65d8ed61aadcfdffd9ab663dba40fd0f7296d1387146cd7L431-R454

The extra semicolon was due to a previous omission.

PR #325, opened 2025-08-10
Parameter Elaboration Refactorization

FlandiaYingman:
This PR refactored the parameter "elaboration" logic, moving the logic to Tree where it becomes more appropriate. This PR also fixed the problem that `clsParams` is not correctly parsed when the `using` modifier is specified.


FlandiaYingman:
For the duplicated resolution error problem: that's because we (I) mistakenly resolve `Foo` once and the synthesized `Foo.class` another time, resulting in two identical error messages. 

`Foo` has a symbol of `member:Foo` and `Foo.class` has the symbol `class:Foo`. I believe we shouldn't resolve the terms that have symbol `class:Foo`. Before I make the change, could you remind me what is the difference between `member:` and `class:`? 

```fs
:elt
x => x.Foo#T
//│ FAILURE LOCATION: f (Elaborator.scala:506)
//│ ╔══[ERROR] Class 'Foo' does not contain member 'T'.
//│ ║  l.19: 	x => x.Foo#T
//│ ╙──      	           ^
//│ Elaborated tree:
//│ Blk:
//│   stats = Nil
//│   res = Lam:
//│     params = ParamList:
//│       flags = ()
//│       params = Ls of 
//│         Param:
//│           flags = ()
//│           sym = x
//│           sign = N
//│           modulefulness = Modulefulness of N
//│       restParam = N
//│     body = SelProj:
//│       prefix = Ref:
//│         sym = x
//│         iargsLs = N
//│       cls = SynthSel{class:Foo}:
//│         prefix = Ref:
//│           sym = member:Foo
//│           iargsLs = N
//│         nme = Ident of "class"
//│         iargsLs = N
//│       proj = Ident of "T"
//│ FAILURE LOCATION: resolveArg (Resolver.scala:772)
//│ ╔══[ERROR] Cannot query instance of type T for call: 
//│ ║  l.19: 	x => x.Foo#T
//│ ║        	      ^^^^
//│ ╟── Required by contextual parameter declaration: 
//│ ║  l.15: 	class Foo(using T) with
//│ ║        	                ^
//│ ╙── Missing instance: Expected: T; Available: ‹none available›
//│ FAILURE LOCATION: resolveArg (Resolver.scala:772)
//│ ╔══[ERROR] Cannot query instance of type T for call: 
//│ ║  l.19: 	x => x.Foo#T
//│ ║        	      ^^^^
//│ ╟── Required by contextual parameter declaration: 
//│ ║  l.15: 	class Foo(using T) with
//│ ║        	                ^
//│ ╙── Missing instance: Expected: T; Available: ‹none available›
//│ = fun
```

LPTK:
> Before I make the change, could you remind me what is the difference between `member:` and `class:`?

`member:` is a `BlockMemberSymbol`. It denotes a name as seen form the outside, which could denote an overloaded definition. `class:` means specifically the symbol of the class in question.

LPTK:
Thanks for the PR BTW. I noticed that @chengluyu also made changes to this part of the code, so there might be annoying conflicts. Which PR should we merge first? (Ideally, @FlandiaYingman should have based this PR on @chengluyu's branch.)

FlandiaYingman:
Fortunately enough Luyu's PR didn't change much about the param thing. I believe it's just a minor conflict. I could rebase onto Luyu's branch and try to resolve the conflicts, it shouldn't take much effort.

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
```diff
@@ -239,35 +244,97 @@ enum Tree extends AutoLocated:
...
+    case td @ Tree.TermDef(m, head, N) =>
+      (td.extension.isEmpty && td.withPart.isEmpty && m == modifier) || head.isModified(modifier)
+    case Modified(m, _, body) =>
+      modifier == m || body.isModified(modifier)
+    case _ =>
+      false
 
 object Tree:
+  // A parameter yet to be elaborated.
+  case class TreeParam(
```
LPTK: It should be called `ParamTree`, not `TreeParam`.

But actually, since it's already in `Tree`, why not just call it `Param`?

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
```diff
@@ -453,9 +520,11 @@ trait TypeDefImpl(using State) extends TypeOrTermDef:
   
   lazy val clsParams: Ls[semantics.TermSymbol] =
     this.paramLists.headOption.fold(Nil): tup =>
-      tup.fields.iterator.flatMap(_.asParam(false).toOption).map:
-        case (S(spd), id, _) => ??? // spreads are not allowed in class parameters
-        case (N, id, _) => semantics.TermSymbol(ParamBind, symbol.asClsLike, id)
+      val pts = tup.fields
+      val inUsing = pts.headOption.map(_.isModified(Ins)).exists(identity)
```
LPTK:
```suggestion
      val inUsing = pts.headOption.exists(_.isModified(Ins))
```

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
```diff
@@ -453,9 +520,11 @@ trait TypeDefImpl(using State) extends TypeOrTermDef:
...
+      pts.flatMap(_.asParam(inUsing = inUsing, inDataClass = false).toOption).map:
```
LPTK: How do you know `inDataClass = false`?

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
```diff
@@ -239,35 +244,97 @@ enum Tree extends AutoLocated:
...
+        R(TreeParam(flags, new Ident("_").withLocOf(und), N, S(SpreadKind.Lazy), modifiers))
+      // fun f(..._)
+      case Spread(Keyword.`...`, _, S(und: Under)) => 
+        R(TreeParam(flags, new Ident("_").withLocOf(und), N, S(SpreadKind.Eager), modifiers))
+      // fun f(..)
+      case Spread(Keyword.`..`, kwLoc, N) =>
+        R(TreeParam(flags, new Ident("_").withLoc(kwLoc), N, S(SpreadKind.Lazy), modifiers))
+      // fun f(...)
+      case Spread(Keyword.`...`, kwLoc, N) =>
+        R(TreeParam(flags, new Ident("_").withLoc(kwLoc), N, S(SpreadKind.Eager), modifiers))
```
LPTK: There is too much duplication. This is what extractors (objects with an `unapply` method) are for.

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
```diff
@@ -239,35 +244,97 @@ enum Tree extends AutoLocated:
...
+    @tailrec
+    def go(t: Tree, flags: FldFlags, modifiers: Set[DeclKind]): Diagnostic \/ TreeParam = t match
+      // * Base Cases.
+      
+      // fun f(_)
+      case und: Under => 
+        R(TreeParam(flags, new Ident("_").withLocOf(und), N, N, modifiers))
+
+      // In `using` clauses, identifiers and type applications are
+      // understood as type names for unnamed contextual parameters:
```
LPTK: I think that instead of special-casing identifiers and type applications, you should special case `InfixApp(_, Keyword.`:`, _)`. You shouldn't make such strict assumptions about the shape of a type tree. We can very well have type trees of other shapes, like `OpApp`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1302,19 +1289,20 @@ extends Importer:
...
               raise(ErrorReport(msg"Keyword `using` must occur before all parameters." -> hd.toLoc :: Nil))
-            isSpd match
-            case S(eagerSpd) =>
-              if !eagerSpd then raise(ErrorReport(msg"Lazy spread parameters not allowed." -> hd.toLoc :: Nil))
-              if tl.isEmpty then (ParamList(flags, acc.reverse, S(p)), newCtx)
+            val newCtx = ctx + (p.sym.name -> p.sym)
+            val newFlags = flags.copy(ctx = inUsing)
+            spd match
+            case S(spd) =>
+              if spd == SpreadKind.Lazy then
```
LPTK: `==` is deprecated.

```suggestion
              if spd is SpreadKind.Lazy then
```

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
FlandiaYingman: It is impossible to know in the `TypeDef` unless we insert specifically the modifier information:

```fs
data class F
//│ Parsed tree:
//│ Modified:
//│   modifier = keyword 'data'
//│   modLoc = N
//│   body = TypeDef:
//│     k = Cls
//│     head = Ident of "F"
//│     rhs = N
```

Fortunately we are only using the spread information (to reject wrong trees) or the identifiers (to create a new symbol from it). Whether the data class flag is true or false doesn't affect the resolved identifier.

```scala
      pts.flatMap(_.asParam(inUsing = inUsing, inDataClass = false).toOption).map:
        case TreeParam(spd = S(_)) => lastWords("spreads are not allowed in class parameters")
        case TreeParam(ident = id) => semantics.TermSymbol(ParamBind, symbol.asClsLike, id)
```

FlandiaYingman: Alternatively we could extract the data class param parsing logic to the elaborator, so that the parsing function `asParam` would not require the `inDataClass` flag: the elaborator sets the `val` flag in `Param` when it rebuilds `Param` from the `TreeParam` (or `ParamTree`).

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
FlandiaYingman: The elaborator has an import statement `import Tree.*` so that I specifically gave it a prefix. 

But after seeing your comment I found that there is a funny syntax in Scala `import Tree.{Param => _, _}`, which imports everything from `Tree` except for `Tree.Param`. Which one do you prefer, using this funny import or rename to `ParamTree`?

FlandiaYingman: I think better name it `TreeParam` and move it outside the Tree, because not only in the elaborator there is `import Tree.*` but also some other places.

FlandiaYingman: > I think better name it `TreeParam` and move it outside the Tree, because not only in the elaborator there is `import Tree.*` but also some other places.

I made this change. Let me know if you like it

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
FlandiaYingman: > Alternatively we could extract the data class param parsing logic to the elaborator, so that the parsing function `asParam` would not require the `inDataClass` flag: the elaborator sets the `val` flag in `Param` when it rebuilds `Param` from the `TreeParam` (or `ParamTree`).

I made this change and let me know if you like it

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
FlandiaYingman: I understand what you meant by 

> You shouldn't make such strict assumptions about the shape of a type tree. We can very well have type trees of other shapes, like OpApp.

But I don't understand about special-casing InfixApp. I did the changes, could you check if that's what you meant?

FlandiaYingman:
BTW I am waiting for Luyu's branch to be merged so that I could merge the main branch into this branch; or should I just merge his branch into this branch directly?

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
```diff
@@ -239,33 +248,69 @@ enum Tree extends AutoLocated:
...
+      // fun f(_)
+      case und: Under => 
+        R(ParamTree(flags, new Ident("_").withLocOf(und), N, N, modifiers))
+      // fun f(a)
+      case id: Ident if !inUsing =>
+        R(ParamTree(flags, id, N, N, modifiers))
+      // fun f(a: A)
+      case InfixApp(id: Ident, Keyword.`:`, sign) =>
+        R(ParamTree(flags, id, S(sign), N, modifiers))
+      // fun f(..a) / fun f(...a)
```
LPTK:
```suggestion
      // fun f(..a) | fun f(...a)
```

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
LPTK: Ok, this looks fine to me (I see you used `ParamTree` instead of `TreeParam` – the latter makes no sense).

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
LPTK: Yes, it's much better, now!

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
LPTK: Yes, I think it's better, now. The `InfixApp` case is tried before we consider `using` parameters (and this consideration does not make any assumption on the shape of the type). That's what I meant.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1265,27 +1265,16 @@ extends Importer:
...
+  def param(t: Tree, inUsing: Bool, inDataClass: Bool): Ctxl[Diagnostic \/ (Param, Opt[SpreadKind])] =
+    t.desugared.asParam(inUsing).map: pt =>
+      log(s"Elaborating ParamTree: ${pt}")
+      val ParamTree(flags, id, sign, spd, modifiers) = pt
+      val flg = flags.copy(isVal = flags.isVal || inDataClass)
+      val sym = VarSymbol(id)
+      val sig = sign.map(term(_))
+      val p = Param(flg, sym, sig, Modulefulness.ofSign(sig)(Mod in modifiers))
+      sym.decl = S(p)
+      (p, spd)
```
```suggestion
    t.desugared.asParam(inUsing).map:
      case pt @ ParamTree(flags, id, sign, spd, modifiers) =>
        log(s"Elaborating ParamTree: ${pt}")
        val flg = flags.copy(isVal = flags.isVal || inDataClass)
        val sym = VarSymbol(id)
        val sig = sign.map(term(_))
        val p = Param(flg, sym, sig, Modulefulness.ofSign(sig)(Mod in modifiers))
        sym.decl = S(p)
        (p, spd)
```

LPTK (approved):

LPTK:
> BTW I am waiting for Luyu's branch to be merged so that I could merge the main branch into this branch; or should I just merge his branch into this branch directly?

Merging Luyu's branch now into this one is perfectly fine.

FlandiaYingman:
Surprisingly there is only one conflict

PR #327, opened 2025-08-12
Add syntax for dynamic selection `x!a`, instantiation `new! c`, and `not` operator

LPTK:
The eventual goal is to statically resolve every selection and instantiation,
based on a pass of lightweight flow analysis.
This PR also cleans up the lexer (removing compatibility with Scala 2).
We also add a new tree form for prefix keyword applications `PrefixApp` (instead of abusing `Modified`).
This commit also corrects some embarrassing Copilot-generated slop in Lowering.scala.


hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
```diff
@@ -74,9 +74,11 @@ enum Tree extends AutoLocated:
   case SynthSel(prefix: Tree, name: Ident)
   case Sel(prefix: Tree, name: Ident)
   case MemberProj(cls: Tree, name: Ident)
+  case PrefixApp(kw: Keyword.Prefix, kwLoc: Opt[Loc], rhs: Tree)
```
FlandiaYingman, line 86: Should we make `Keyword` a `case` in `Tree`? Because a `Tree` naturally has a `Loc`, and it's annoying to `withLoc` many times in the elaborator. 🤔

hkmc2/shared/src/test/mlscript/backlog/ToTriage.mls
```diff
@@ -456,3 +456,9 @@ data class Foo(mut x)
 //│ ╙──       	                   ^
 
 // ——— ——— ———
+
+// FIXME should report an error
+fun Foo
+class Foo
```
FlandiaYingman, line 462: Wow, why `class Foo` is considered an error

FlandiaYingman: Oh nevermind it's name conflict

FlandiaYingman:
Wow amazing, but what are the differences between `PrefixApp` and `Modified` in the parser? Are they parsed differently?

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
LPTK: Hmm that's a good idea. Tho we'd need to give it a sligthly annoying type parameter `case Kwrd[K <: Keyword](kw: K)` to keep track of the refined type of keyword (`Keyword.Prefix`, etc.).

LPTK:
There's functionally no difference, but it always bothered me that the parser produces an inadequate tree for things like `do t`. Doing so made some error messages weird.

PR #330, opened 2025-08-22
Misc fixes and improvements

ychenfo:
- More efficient `inScope` function using `Set.contains` instead of `Iterator.contains`.
- `Block.flattened` now flattens class methods
- The paramter of `DummyTypeDef` is properly used
- Remove a small piece of dead code in `JSBuilder.scala`

The output of `ParamClasses.mls` changes. This is because the set `Scope.existingNames` and the values of `Scope.bindings` may not in sync: `Scope.allocateName` may get called more than once for the same symbol. One possible source of multiple `allocateName` calls is [here](https://github.com/hkust-taco/mlscript/blob/6677f788b57f069471bc546ad28971e7b8367e46/hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala#L237-L238) and [here](https://github.com/hkust-taco/mlscript/blob/6677f788b57f069471bc546ad28971e7b8367e46/hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala#L370-L376), where the symbols in `paramsOpt` get `allocateName`d twice.

I am not sure whether there are more places where `allocateName` get called twice or how to fix this now in this PR since the logic seems to be quite complicated...

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/MLsCompiler.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/MLsCompiler.scala
@@ -98,8 +98,7 @@ class MLsCompiler(preludeFile: os.Path, mkOutput: ((Str => Unit) => Unit) => Uni
         utils.Scope.empty
       // * This line serves for `import.meta.url`, which retrieves directory and file names of mjs files.
       // * Having `module id"import" with ...` in `prelude.mls` will generate `globalThis.import` that is undefined.
-      baseScp.bindings += Elaborator.State.importSymbol -> "import"
-      baseScp.existingNames += "import"
+      baseScp.addToBindings(Elaborator.State.importSymbol, "import")
       val nestedScp = baseScp.nest
       val nme = file.baseName
       val exportedSymbol = parsed.definedSymbols.find(_._1 === nme).map(_._2)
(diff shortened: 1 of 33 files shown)


hkmc2/shared/src/main/scala/hkmc2/MLsCompiler.scala
```diff
@@ -99,6 +99,7 @@ class MLsCompiler(preludeFile: os.Path, mkOutput: ((Str => Unit) => Unit) => Uni
       // * This line serves for `import.meta.url`, which retrieves directory and file names of mjs files.
       // * Having `module id"import" with ...` in `prelude.mls` will generate `globalThis.import` that is undefined.
       baseScp.bindings += Elaborator.State.importSymbol -> "import"
+      baseScp.existingNames += "import"
```
LPTK: Instead of havign to remember to do that anywhere `bindings` is extended, the proper appeoach is to make `bindings` private and only allow adding bindings throug a methods that maintains the invariant.

LPTK (approved):
Great, thanks! I'll look into the "multiple `allocateName` calls" problem tomorrow.

hkmc2/shared/src/main/scala/hkmc2/MLsCompiler.scala
ychenfo: Thanks! I have pushed the changes to use a method to maintain the invariant.

PR #331, opened 2025-08-23
Implement companion overloading

LPTK:
This PR adds some preliminary amount of overloading, for example between a class and its companion module.

The PR ended up being quite the yak-shave, as the required changes affected many parts of the compiler, with cascading required changes. It required contributions from four different people!

@FlandiaYingman attempted an ill-fated symbol refactoring, but bailed out mid-way and will make another separate attempt later.

---

- [x] Remove the `describe` and `expand` hacks
- [x] @LPTK Improve lowering error message when symbol is not found
- [ ] [now [moot](https://github.com/hkust-taco/mlscript/pull/331/commits/9c608c771e1501655a3a5964f4261e6c8cfa83c8)] @CAG2Mark Improve the `FIXME: DisambBlockMemberSymbol workaround` comments, esp. in Lifer
- [x] @FlandiaYingman Remove `Unexpanded ` noise


hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -544,7 +544,9 @@ extends Importer:
       sym match
       // * Enforcing [invariant:1]
       case S(ms: BlockMemberSymbol)
-        if !inAppPrefix && ms.isParameterizedMethod && !preTrm.symbol.exists(_.isModuleful) =>
+        // FIXME: move the check to resolver becuase preTrm's symbol may
+        // not be resolved yet.
+        if !inAppPrefix && ms.isParameterizedMethod && preTrm.symbol.exists(_.existsNonModuleful) =>
```
LPTK: 👍

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -28,9 +28,24 @@ abstract class Symbol(using State) extends Located:
     res
   def refsNumber: Int = directRefs.size
   
-  def isModuleful: Bool =
-    // * Modules overloaded with classes, objects, or functions are
-    asMod.nonEmpty && asCls.isEmpty && asObj.isEmpty && !hasTrmDef
+  def existsNonModuleful: Bool = this match
+    case mod: ModuleSymbol => !(mod.tree.k is Mod)
+    case mem: BlockMemberSymbol => 
+      mem.trees.isEmpty ||
```
LPTK: Maybe add a comment on why this is needed? @FlandiaYingman

hkmc2/shared/src/test/mlscript/HkScratch100.mls
```diff
@@ -0,0 +1,70 @@
...
+class ClsMod with
+  val whoami = "class"
+module ClsMod with
+  val whoami = "module"
+
+type TypMod = Int
+module TypMod with
+  val whoami = "module"
+
+:todo
```
LPTK: Seems like this should be `:fixme`. @FlandiaYingman

hkmc2/shared/src/test/mlscript/HkScratch100.mls
```diff
@@ -0,0 +1,70 @@
...
+
+fun f: ClsMod
+
+fun f: module ClsMod
+
+fun f: TypMod
+
+fun f: module TypMod
+
+:todo
```
LPTK: A `:todo` should specify what's going wrong in a comment. What error are you expecting? @FlandiaYingman

LPTK: Also, it should be something like
```suggestion
:breakme
:e // blah
```

hkmc2/shared/src/test/mlscript/HkScratch100.mls
```diff
@@ -0,0 +1,70 @@
...
+assertModule(TypMod).whoami
+//│ = "module"
+
+assertModule(TrmMod).whoami
+//│ = "module"
+
+
+fun assertNonModule(m) = m
+
+// FIXME: should be resolver error
```
LPTK:
```suggestion
:breakme // FIXME: should be resolver error
:e
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
FlandiaYingman: OK!

hkmc2/shared/src/test/mlscript/HkScratch100.mls
FlandiaYingman: OK, fixed! But what's the semantic difference between them

hkmc2/shared/src/test/mlscript/HkScratch100.mls
FlandiaYingman: OK!

hkmc2/shared/src/test/mlscript/HkScratch100.mls
FlandiaYingman: OK, added!

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
FlandiaYingman: BTW this comes from a limitation of `BlockMemberSymbol` itself that we might want to fix it in the future: currently all overloaded definitions share the same `BlockMemberSymbol`, but the `defn` field that lies in the `BlockMemberSymbol` is just the bottommost definition only, so one cannot reliably read the `defn` field.

Currently we fish the symbol or definition from the `trees` field of `BlockMemberSymbol` in order to find an overloaded symbol (or definition), but this approach does not work if the definition does not have a tree at all!

```fs
data class Foo(x) with
  // x internally has a val definition, but it does not have a corresponding tree!
  fun f = x
```

I think we could make `BlockMemberSymbol` not extending `MemberSymbol[Definition]`, and add a `var defns: Ls[Definition]` (or maybe a Map?) for filtering definitions...

LPTK: `BlockMemberSymbol` is very loosely defined (based on a plain list of trees) because it's constructed very early, before elaboration.

The idea was to have definitions have a much more specific structure, where illegal states (ie illegal overloadings) are not representable. So storing a list of `defns` would defeat that – it would be pretty poor design. Instead, we should store a definition that precisely corresponds to the overloaded structure. This is currently somewhat (but not fully) represented in `Definition`, but the elaborator probably doesn't store the right info.

hkmc2/shared/src/test/mlscript/HkScratch100.mls
LPTK: I'd recommend against using `:todo`, actually. It is not specific enough. If you end up implementing the feature that needs to be done, `:todo` will remain without complaining, instead of failing.

hkmc2/shared/src/test/mlscript-compile/Str.mls
```diff
@@ -1,5 +1,10 @@
 
-object Str with ...
+// TODO:
+// The `object Str` in this file overrides the default `class Str`,
+// rather than overloading it. As a workaround, declare an overloaded
+// `class Str` again.
+declare class Str
```
LPTK: Wow that's vile

hkmc2/shared/src/test/mlscript/basics/GenericClasses.mls
```diff
@@ -41,7 +41,7 @@ class Bar(x: Foo[Int])
 //│ ╔══[ERROR] Wrong number of type arguments
 //│ ║  l.39: 	class Foo[A](x: Bar[A])
 //│ ╙──      	                ^^^^^^
-//│ ╔══[ERROR] Expected 0 type arguments, got 1
+//│ ╔══[ERROR] Expected no type arguments, got 1
```
LPTK, line 44: The most important change in this PR.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -831,18 +832,93 @@ class Resolver(tl: TraceLogger)
...
+      traverseType(_, 
         expect = if p.modulefulness.modified
           then Module(S(msg"Module parameter must have a module type."))
           else NonModule(S(msg"Non-module parameter must have a non-module type.")),
       )
   
+  def traverseType(t: Term, expect: Expect, inTyAppPrefix: Bool = false)(using ictx: ICtx): Unit =
+  trace(s"Traversing type ${t}, expecting ${expect}"):
+    def checkTypeArity(sym: FieldSymbol): Unit =
+      if inTyAppPrefix && !t.isInstanceOf[Term.TyApp] then
```
LPTK: Why would you abort the checks in this case? Can you documemt the reason for the `!t.isInstanceOf[Term.TyApp]` check? @FlandiaYingman

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -831,18 +832,93 @@ class Resolver(tl: TraceLogger)
...
+          val targsNum = targs.map(_.length)
+          if tparamsNum != targsNum then
+            val tparamsMsg = tparamsNum.getOrElse("no").toString()
+            val targsMsg = targsNum.getOrElse("none").toString()
+            raise:
+              ErrorReport:
+                msg"Expected ${tparamsMsg} type arguments, " +
+                msg"got ${targsMsg}" -> t.toLoc :: Nil
+    
+    def check(body: => Unit): Unit =
```
LPTK: I still don't understand why you use this convoluted way of writing code. Why don't you just put the checking code after the traversing code, instead of adding a method with a by-name parameter? @FlandiaYingman

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: Looks like a place that I forgot to clean up after changing the approach.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: Oh, right LOL. I should have removed them. It was a legacy from the resolver prototype.

hkmc2/shared/src/test/mlscript/basics/GenericClasses.mls
FlandiaYingman: Amazing!

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
LPTK: Ok, pls do remove them.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
LPTK: Ok, pls fix.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
LPTK: @FlandiaYingman So, will you remove them?

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -44,79 +50,83 @@ sealed trait ResolvableImpl:
...
+      s"${desc} without a resolvable definition"
   
   def instantiate = expansion match
     case S(S(t)) => t
-    case S(N) => t
-    case N => lastWords(s"missing expansion for term ${t}")
+    case S(N) => this
+    // case N => lastWords(s"missing expansion for term ${this}")
+    case N =>
+      // FIXME: instantiating a term without an expansion should be an
```
LPTK: @FlandiaYingman Why? I thought `instantiate` would be what everyone would call by default before inspecting a term. (Remember this whole mutable business is just an optimization over just rewriting an immutable tree.)

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -44,79 +50,83 @@ sealed trait ResolvableImpl:
...
+      // error
+      this
 
   /** This method is only supposed to be called by Resolver. */
-  private[semantics] def expand(expansionFn: Opt[Term => Term]): this.type =
-    val newExpansion = expansionFn.map(_(t.duplicate.resolve))
+  private[semantics] def expand(expansionFn: Opt[this.type => Term]): this.type =
```
LPTK: @FlandiaYingman Why does `expand` take a function?

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -44,79 +50,83 @@ sealed trait ResolvableImpl:
...
   def resolve: this.type = expand(N)
+  def dontResolve: this.type = this // TODO rm
   
   def hasExpansion = expansion.isDefined
   
-  def defn: Opt[Definition] = t.resolvedSymbol match
-    case S(sym: MemberSymbol[?]) => sym.defn
+  def defn: Opt[Definition] = resolvedSymbol match
+    case S(sym: MemberSymbol[?]) => sym.defn.map(_.asPrincipalDefn)
+    // TODO: @Harry check if this case is really needed.
```
LPTK: @FlandiaYingman 👀

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -282,23 +282,58 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
+      term(lhs, inStmtPos = inStmtPos)(k)
+    case st.Tup(fs) =>
+      args(fs)(args => k(Value.Arr(mut = false, args)))
+    case Mut(st.Tup(fs)) =>
+      args(fs)(args => k(Value.Arr(mut = true, args)))
+    case st.CtxTup(fs) =>
+      // Q: is this ever supposed to be reached?
+      args(fs)(args => k(Value.Arr(mut = false, args)))
+    case ref @ st.Ref(sym) =>
+      // FIXME: DisambBlockMemberSymbol workaround
```
LPTK: @FlandiaYingman Could you pls be more specific about what should be fixed, here?

hkmc2/shared/src/test/mlscript-compile/Str.mls
LPTK: Top-quality:
```fs
import "../../mlscript-compile/Str.mls"

:sjs
"a" is Str
//│ JS (unsanitized):
//│ let scrut; scrut = "a"; if (scrut instanceof globalThis.Str) { true } else { false }
//│ FAILURE LOCATION: mkQuery (JSBackendDiffMaker.scala:154)
//│ ═══[RUNTIME ERROR] Error: Access to required field 'Str' yielded 'undefined'
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
FlandiaYingman: I remember that comment was added before we agree on the change of expanding mechanism 😗 I'll do a cleanup for the resolver to remove all these misleading comments.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
FlandiaYingman: That's a safety measure that I initially took to prevent from forgetting `duplicate`ing the term. Now it seems that it doesn't add much safety... I think we should remove it and accept `Term` instead.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
FlandiaYingman: OK I'll check it.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
FlandiaYingman: Ideally after resolution, all `BlockMemberSymbol` should be resolved to `DisambBlockMemberSymbol`, so we should simply remove all cases about `BlockMemberSymbol` in the stages after resolution (i.e., `Lowering` and `Lifter`).

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: Yes!

hkmc2/shared/src/test/mlscript-compile/Str.mls
FlandiaYingman: Argh, how can we fix it?

FlandiaYingman: How about we special-case the MLscript types `Str` `Num` etc. that will never be overridden by another definition, but will be overloaded with any later definition seen?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
LPTK: Ok, I meant please add that explanation to the comment! ;^)

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
LPTK: Ok, pls make the change.

hkmc2/shared/src/test/mlscript-compile/Str.mls
LPTK: Actually how about we just don't mess with `Str`? 😬 

I think adding this module was a mistake. Could easily rename it to `StrOps` or just integrate the contents into `Predef`.

At some point we'll want to change the way `declare` works so we can write something like `declare class std.Str` – so we know this is a proper forward declaration and everything can actually fall back into place.

LPTK:
Ok, the PR is ready. @FlandiaYingman @CAG2Mark please look into the agreed parts (and check your TODOs here: https://github.com/hkust-taco/mlscript/pull/331#issue-3348322909), and let me know if you see anything suspicious (also feel free to do a full review if you want).

hkmc2/shared/src/test/mlscript-compile/Str.mls
FlandiaYingman: Okay, let's rename it to StrOps!

FlandiaYingman:
Aye, aye, Captain

CAG2Mark:
Is there any reason `DisambBlockMemberSymbol` doesn't extend `BlockMemberSymbol`? I feel like it would make a lot of things cleaner.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -58,19 +58,35 @@ sealed trait ResolvableImpl:
...
-
-  /** This method is only supposed to be called by Resolver. */
-  private[semantics] def expand(expansionFn: Opt[this.type => Term]): this.type =
-    val newExpansion = expansionFn.map(_(duplicate))
-    expansion = S(newExpansion)
+    case N => this
+
+  /** 
+   * Expanding a term to another, which can be later retrieved by the
+   * `instantiate` method.
```
LPTK, line 74: BTW this should probably be called `expanded`, not `instantiate`. There is no "instantiation" going on, here.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -838,7 +838,7 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
         val newCtx = ctx
           .replCapturePaths(newCapturePaths)
           .replLocalPaths(newLocalsPaths)
-          .addIsymPaths(newIsymPaths)
+          .addIsymPaths(newIsymPaths.asInstanceOf) // FIXME
```
LPTK: 👀

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -57,7 +57,11 @@ import Subst.subst
 class Lowering()(using Config, TL, Raise, State, Ctx):
   
   extension (t: Term)
-    def instantiated = t match
+    def instantiated =
```
```suggestion
    def expanded =
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -373,19 +377,20 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
           return k(Value.Lam(paramLists.head, bodyBlock))
       case sym: (BlockMemberSymbol | DisambBlockMemberSymbol[?]) =>
         // TODO: Ideally, all lowering-facing symbols should be disambiguiated...
-        val (bs, defn) = sym match
-          case bs: BlockMemberSymbol => bs -> bs.defn
-          case dbs: DisambBlockMemberSymbol[?] => dbs.bsym -> dbs.sym.defn
+        val (dbs, defn) = sym match
+          // FIXME: BMS legacy
```
LPTK: 👀 

(Just adding comments so we can keep track of what needs to be addressed in the PR before it can be merged.)

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -219,9 +219,9 @@ object Elaborator:
       val id = new Ident("NonLocalReturn")
       val sym = ClassSymbol(DummyTypeDef(syntax.Cls), id)
       Term.Sel(runtimeSymbol.ref(), id)(S(sym))
-    val nonLocalRet =
-      val id = new Ident("ret")
-      BlockMemberSymbol(id.name, Nil, true)
+    // val nonLocalRet =
```
LPTK: 👀

LPTK:
Ok @FlandiaYingman the code of this PR was already cleaned up. I don't think it's a good idea to push to this branch commits with tons of comments and WIP changes that fail the tests. Better make keep these chnages separate for now and make a (possibly draft) PR to this branch.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -201,15 +206,33 @@ class BlockMemberSymbol(val nme: Str, val trees: Ls[TypeOrTermDef], val nameIsMe
   
   def subst(using sub: SymbolSubst): BlockMemberSymbol = sub.mapBlockMemberSym(this)
 
-  def disamb[Defn <: Definition](sym: DisambSymbol[Defn]): DisambBlockMemberSymbol[Defn] =
+    // def isPrincipalOverload: Bool = this match
+    // case cls: ModuleOrObjectDef => cls.companion.isEmpty
+    // case _ => true
+
+  def asPrincipal: DefinitionSymbol[?] =
```
LPTK: I thought we said we don't need this after all?

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
```diff
@@ -222,7 +225,7 @@ enum Tree extends AutoLocated:
   
   def showDbg: Str = toString // TODO
   
-  lazy val desugared: Tree = this match
+  def desugared(using State): Tree = this match
```
LPTK: WOW you can't just do that! The desugared tree is often accessed many time and shoudl NOT be recomputed each time!

LPTK:
@FlandiaYingman I'm afrait your revert revert needs to be revereted, as it does not pass the CI.

PR #326, opened 2025-08-10
Eta Expansion

FlandiaYingman:
This is a really messy WIP PR that implements Eta Expansion limitedly. As eta-expansion requires expanding a term into another, and it happens *possibly multiple times* on the same term, I have to be very careful about the number of times that a term is resolved, which means that this PR probably has to be based on #317, which in turn has to be based on #316 😛.

PR #332, opened 2025-08-23
Fix eta-expansion redundancy

FlandiaYingman:


hkmc2/shared/src/test/mlscript/ctx/EtaExpansion.mls
```diff
@@ -28,30 +28,80 @@ using Num = 0.42
...
+:sjs
 let f3 = baz
+//│ JS (unsanitized):
+//│ let f3, f31;
+//│ f31 = function f3(x) {
+//│   let lambda1;
+//│   lambda1 = (undefined, function (y) {
+//│     let tmp, tmp1, tmp2;
+//│     tmp = baz(x);
+//│     tmp1 = runtime.safeCall(tmp(instance$Ident$_Int$_));
```
LPTK, line 82: I wonder why there are all these `safeCall`s (these are for unknown, possibly JS calls that could return undefined). Don't the calls have symbols? Or we could adapt the way calls are recognized to be calls to MLscript functions.

LPTK (approved):
Thanks! I just have one comment.

hkmc2/shared/src/test/mlscript/ctx/EtaExpansion.mls
FlandiaYingman: I checked the logic in `Lowering`. The logic simply looks at the LHS of the App to see if it has a MLs function term definition. Since `baz(x)` does not have a term definition nor a symbol (only `baz` has), it is not recognized as a MLs call. 

We can fix this behavior by inserting `CallableDefinition` to terms, which was created by the Resolver for internal representation of the applied state (the definition of a "callable" after it being applied a number of times) of a MLs function. Also the current implementation does not handle constructor calls properly (because constructors have class definitions, but not term definitions):

```fs
:sjs
Foo(1, 2)
//│ JS (unsanitized):
//│ let tmp4; tmp4 = Foo7(1, 2); runtime.safeCall(tmp4(instance$Ident$_Int$_))
//│ = Foo0 { x: 1, y: 2, z: 123 }
```

which should also be compiled without `safeCall`! We could also fix it by with the `CallableDefinition` thing.

I can fix it together with the refactorization of `CallableDefinition`, as planned here
https://github.com/hkust-taco/mlscript/pull/326#discussion_r2279390253

LPTK: Great, that sounds like a good plan! Let's merge this PR now and pls open another PR for the chnages described above.

PR #334, opened 2025-09-06
Properly Resolve Symbols for Resolvable Terms

FlandiaYingman:


hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -799,21 +809,42 @@ class Resolver(tl: TraceLogger)
     case _ =>
     
     t match
-    case t @ Apps(base: Resolvable, ass) => base.termDefn match
-      case S(lhsDefn) if lhsDefn.params.length == ass.length =>
+    case t @ Apps(base: Resolvable, ass) => 
+      base.decl match
+      // TODO: Handle class / object annotations.
+      case S(lhsDefn: TermDefinition) if lhsDefn.params.length == ass.length =>
```
LPTK: Pls you stop using `==`. I keep finding after the fact and fixing uses of `==` instead of `===`.

hkmc2/shared/src/test/mlscript/ctx/SymbolResolution.mls
```diff
@@ -72,23 +72,23 @@ M.that.that
...
-//│         prefix = Sel{module:N}:
-//│           prefix = Sel{module:M}:
-//│             prefix = Sel{module:N}:
-//│               prefix = Ref{member:M} of member:M
+//│   res = Sel:
+//│     prefix = Sel:
+//│       prefix = Sel:
+//│         prefix = Sel:
+//│           prefix = Sel:
+//│             prefix = Sel{sym=member:that,typ=module:N}:
```
LPTK, line 90: Hmm why did the other symbols disappear? 🤔

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: ??? Why is there `===` in Scala? Are we changing Scala to ScalaScript?

FlandiaYingman: Oh nvm, I saw there's a shorthand.

But what's the reason?

hkmc2/shared/src/test/mlscript/ctx/SymbolResolution.mls
FlandiaYingman: This is a top-quality bug, which will be fixed tomorrow

FlandiaYingman: (Notice that the tests failed)

FlandiaYingman: Fixed :)

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: Fixed

FlandiaYingman:
Now the resolver distinguishes between symbols of *definitions* and symbols of the evaluation result. I used to mix them up.

```fs
module M
fun f: module M = M

:rt
f
//│ Resolved tree:
//│ Blk:
//│   stats = Nil
//│   res = Ref{sym=member:f,typ=module:M} of member:f
//│ = class M
```

Now `f` has a resolved symbol of `member:f` but has a resolved *type symbol* of `module:M`.

FlandiaYingman:
Now resolved symbols are encoded into expansions. But the change is a bit messy because I am fishing the expansion instead of naturally having them in the beginning (another Taj Mahal!). Also, the GitHub shows the diff wrongly because I indented a huge block.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
LPTK: `==` is loosely typed and super error prone. Will not warn you when you refactor the code and suddenly start comparing things of unrelated types.

LPTK:
> Now `f` has a resolved symbol of `member:f` but has a resolved _type symbol_ of `module:M`.

Wait, by "type symbol", do you mean the type signature? What happens if the sigature is something like `f: (module M) -> module M` (which I guess should happen for a declaration of the form `fun f(m: module M): module M)`?

> But the change is a bit messy because I am fishing the expansion instead of naturally having them in the beginning

So... is there a reason you're doing things the ugly way? Is it only temporary?

FlandiaYingman:
> Wait, by "type symbol", do you mean the type signature? 

Yes. It is inferred from the type signature. The purpose of this "type symbol" is mainly for resolving selections, so the "type symbol" is there only if the signature is a class-like thing. It is not really a "type" thing that can represent lambda and other complex types. 

(I think "type symbol" is not really a good name...)

> What happens if the signature is something like `f: (module M) -> module M` (which I guess should happen for a declaration of the form `fun f(m: module M): module M)`?

The "type symbol" of `f` does not exist (resolved to `None`). Yet, the symbol of `f(...)` is correctly resolved to `module:M`.

> So... is there a reason you're doing things the ugly way? Is it only temporary?

Yes, it is temporary. When I was writing the reason of doing things in _the ugly way_, I suddenly figured out how to write it properly. So I think I'll change it :)

LPTK:
Is this "type symbol" thing mostly just for the moduleness checking?

I don't see why we should try to special-case things in this weird way, instead of just having a proper way of handling type signatures generally, from which all relevant info can be recovered. After all, we need proper type signatures for other things, too (implicit resolution, flow analysis...).

FlandiaYingman:
> Is this "type symbol" thing mostly just for the moduleness checking?

No! It's essentially for resolving symbols of selections. It can also check the "statically known class" property of an expression. 

> I don't see why we should try to special-case things in this weird way, instead of just having a proper way of handling type signatures generally, from which all relevant info can be recovered.

We can also store definitions/declarations instead of symbols. After all, there just should be a mechanism to properly resolve selections (and to "propagate" selection chains). Alternatively, we can also add more kinds of symbols so that we can encode complex types into this "type symbol" field. 

Rather than saying that I'm special casing the things, I'd say it's just an incomplete feature that could only work on modules. In the future we can generalize it to other things as well.

LPTK:
I am not at all sure we are on the same page regarding the use of "symbols".

A symbol is a _symbolic_ handle representing an entity in the code. It is meant to be lightweight, efficiently comparable through an object identity check. A symbol also possibly stores a reference to its associated definition, which is a full representation of that entity that the symbol denotes.

Symbols are not meant to stand in for types. Types are used to represent semantic information that is known at compile time, partially through user annotations, and partially through inference/propagation. Many types _rely_ on symbols – eg, the type `MyClass[A]` is a type application based on a reference to the `MyClass` class symbol and a type argument that is itself a reference to an `A` type parameter identified by a corresponding symbol.

When looking at an expression, if we want to know what kind of things it computes/returns, we should really look at its _type_, not at its "symbol". Currently, we store symbols in selections, and make do with that. But it feels like implementing a system based on proper types is sorely needed (as well as reimplementing the full type system, of course, but that one can come later).

FlandiaYingman:
> Many types _rely_ on symbols – eg, the type `MyClass[A]` is a type application based on a reference to the `MyClass` class symbol and a type argument that is itself a reference to an `A` type parameter identified by a corresponding symbol.

> When looking at an expression, if we want to know what kind of things it computes/returns, we should really look at its type, not at its "symbol". 

I see. I think this was the reason that we were not on the same page. I tried to use solely symbols to emulate types, but the proper way should be having a way of manipulating *types* properly.

Currently we have a very naive type representation in the resolver, which was for the resolver to resolve implicit arguments and to match corresponding types. Do you want me to extend it, so that we can have a *naive and proper* type system for this refactorization?

> Currently, we store symbols in selections, and make do with that.

Did you mean the `sym` field in selections?

https://github.com/hkust-taco/mlscript/blob/d2453b960f9951a989dc0dc0e8b46467b908caab/hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala#L178

Based on my understanding, this `sym` field is not directly related to types. It should be the symbol of the entity of the selection, but not the type of the selection. e.g.,

```fs
module M with
  val foo: Int = 42

M.foo
// The `sym` field of this selection should be the entity of this selection, 
// but not the type of this selection.
// The `sym` should be a term symbol of `foo`, 
// but the type of this term should be `Int` (from the type annotation)
```

LPTK:
Yes, let's use a proper representation of types for all these things, and not just implicit resolution.

I suggest you take inspiration from the types in `bbml` and its `GeneralType` data type:
https://github.com/hkust-taco/mlscript/blob/d2453b960f9951a989dc0dc0e8b46467b908caab/hkmc2/shared/src/main/scala/hkmc2/bbml/types.scala#L13

(((
You could also look at the `SimpleType` data type of the old compiler, but that one is much more complicated:
https://github.com/hkust-taco/mlscript/blob/d2453b960f9951a989dc0dc0e8b46467b908caab/shared/src/main/scala/mlscript/TyperDatatypes.scala#L160

The immediately relevant subclasses for you would be `ExtrType`, `ComposedType`, `NegType`, `FunctionType`, `TypeRef` (but instead of `defn: TypeName`, store a symbol), and `ClassTag` (but for the latter, don't store the parents, that was bottom-level design; and as the `id`, store the class symbol).
)))

> Based on my understanding, this `sym` field is not directly related to types. It should be the symbol of the entity of the selection, but not the type of the selection. e.g.,

Yes, true.

hkmc2/shared/src/main/scala/hkmc2/semantics/Type.scala
```diff
@@ -0,0 +1,80 @@
...
+sealed trait TypeArgImpl:
+  self: Type =>
+end TypeArgImpl
+
+type TypeArg = Type & TypeArgImpl
+
+enum Type extends TypeArgImpl:
+  case Error
+  case Top
+  case Bot
```
FlandiaYingman: I use two different cases `Top` and `Bot` for `ExtrType`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Type.scala
```diff
@@ -0,0 +1,80 @@
...
+  case Ref(sym: TypeSymbol | VarSymbol)
+  case App(base: Type, args: Ls[TypeArg])
```
FlandiaYingman: I use two different cases `Ref` and `App`, rather than using a case `Ref(base: TypeSymbol | Var, args: Ls[TypeArgs])`. This was to make refactorization easier and I'll change it later in this PR.

hkmc2/shared/src/main/scala/hkmc2/semantics/Type.scala
```diff
@@ -0,0 +1,80 @@
...
+  case Fun(args: Ls[Type], ret: Type, eff: Opt[Type])
+  case Neg(t: Type)
+  case Union(lhs: Type, rhs: Type)
+  case Inter(lhs: Type, rhs: Type)
```
FlandiaYingman: I use two different cases `Union` and `Inter` for `ComposedType`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -46,53 +46,65 @@ object Resolver:
...
+          && (lhs.eff, rhs.eff).match
+            case (N, N) => true
+            case (S(le), S(re)) => le <= re // covariant
+            case _ => false
+        case (lhs: Type.Neg, rhs) => ???
+        case (lhs, rhs: Type.Neg) => ???
+        case (lhs, rhs: Type.Union) => lhs <= rhs.lhs || lhs <= rhs.rhs
+        case (lhs, rhs: Type.Inter) => lhs <= rhs.lhs && lhs <= rhs.rhs
+        case (lhs: Type.Wildcard, rhs: Type.Wildcard) => ???
+        case (lhs, rhs) => lhs === rhs
```
FlandiaYingman: Here I wrote a simple subtyping logic to replace the old resolver type comparison logic. This could also be useful later when we really implement subtyping in instance resolution.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -997,47 +973,31 @@ class Resolver(tl: TraceLogger)
...
-            N
-      
+      case Term.TyApp(con, args) =>
+        Type.App(resolveType(con), args.map(resolveType(_)))
+
       // Complex types are not supported.
       // TODO: Handle complex types.
-      case _: (Term.FunTy | Term.WildcardTy | Term.CompType | Term.Neg | Term.Forall | Term.Tup) => N
+      case _: (Term.FunTy | Term.WildcardTy | Term.CompType | Term.Neg | Term.Forall | Term.Tup) =>
+        Type.Error
```
FlandiaYingman: We can also handle "complex" types now and I believe it won't be hard. I can try it later in this PR.

hkmc2/shared/src/test/mlscript/basics/BadTypeClasses.mls
```diff
@@ -74,7 +74,7 @@ M.foo()
 :e
 fun f[T](using t: T) = t
 f
-//│ ╔══[ERROR] Cannot query instance of type ‹unspecified› (type parameter T) for call: 
+//│ ╔══[ERROR] Cannot query instance of type ⊤ (type parameter T) for call:
```
FlandiaYingman: I use top type to replace the `‹unspecified›` type query in instance resolution.

hkmc2/shared/src/main/scala/hkmc2/semantics/Type.scala
```diff
@@ -0,0 +1,67 @@
...
+package semantics
+
+import mlscript.utils.*, shorthands.*
+
+object Type:
+end Type
+
+sealed trait TypeArgImpl:
+  self: Type =>
+end TypeArgImpl
```
FlandiaYingman: Could you explain a bit on how does the `TypeArg` trait work? You said via discord that it was for `in`/`out` type arguments in Kotlin, but it seems that in the places (BBML) you showed me, the `in`/`out` was mainly handled by `Wildcard`. So I am not sure what to add here.

FlandiaYingman (commented):
I just created a very naive representation of types. In some places I used a different design, which are commented below, because I think that would be easier to write things by that design. 

I am going to conclude other things in this PR tomorrow.

FlandiaYingman:
Oh wait, what happened to BBML? Why does it refuse to compile? It works on my side 🥺😥

LPTK:
> Oh wait, what happened to BBML? Why does it refuse to compile? It works on my side 🥺😥

Try cleaning up and compiling again. You might have introduced some name clashes.

hkmc2/shared/src/main/scala/hkmc2/semantics/Type.scala
LPTK: A type argument can be either a single specific type like `Int` or a wildcard, ie a range, like `in Nat out Int`, similar to Scala's `? >: Nat <: Int` and Java's `? super Nat extends Int`.

FlandiaYingman: Oh I see, so is it the case that wildcard isn't a type, but can be used a type argument?

LPTK: Yes. (Tho not sure why you ask – the answer is right there in the code.)

hkmc2/shared/src/test/mlscript/basics/BadTypeClasses.mls
FlandiaYingman: Oh wait, yes, `‹unspecified›` is not the same as top type!

hkmc2/shared/src/main/scala/hkmc2/semantics/Type.scala
FlandiaYingman: I figured it out after reading your comment twice and the code three times 🫣

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: Oh no, it doesn't warn me when I tried to compare `Option[Int]` and `Int`... 😥

image

hkmc2/shared/src/test/mlscript/basics/BadTypeClasses.mls
```diff
@@ -19,31 +19,37 @@ M.f
...
 
-:e
 using 42 = 42
-//│ ╔══[ERROR] Expected a type symbol, got integer literal
-//│ ║  l.23: 	using 42 = 42
-//│ ╙──      	      ^^
+
+:todo // support literal type
+use[42]
+//│ ╔══[ERROR] Cannot query instance of type ‹error› (type parameter T) for call:
```
LPTK: Well, silently producing `‹error›` types is a pretty bad idea. Why aren't you reporting an error a the same time?

FlandiaYingman: OK. Now I distinguish error types and not-yet implemented types. Producing error types reports a compilation error with proper message (e.g., "Term ... does not represent a type"). Producing not-yet implemented types is silent, but trying to use it - either adding it into the context or resolving an instance of the type - reports an internal error.

hkmc2/shared/src/test/mlscript/ctx/TypeResolution.mls
```diff
@@ -0,0 +1,153 @@
...
+new Foo
+//│ Resolved tree:
+//│ Blk:
+//│   stats = Nil
+//│   res = New:
+//│     cls = Ref{sym=member:Foo} of member:Foo
+//│     args = Nil
+//│     rft = N
+
+// * dynamic new-expression shouldn't be resolvable because of dynamic nature
```
LPTK: Dynamic doesn't mean we can _never_ resolvew; it means it's possible we can't in some cases.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1580,12 +1585,10 @@ extends Importer:
   class VarianceTraverser(var changed: Bool = true) extends Traverser:
     override def traverseType(pol: Pol)(trm: Term): Unit = trm match
       case Term.TyApp(lhs, targs) =>
-        lhs.symbol.flatMap(sym => sym.asTpe orElse sym.asMod orElse sym.asObj) match
```
FlandiaYingman, line 1583: I changed `asTpe` so that it also tries `asMod` and `asObj`. So I
removed them here.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1580,12 +1585,10 @@ extends Importer:
...
+        lhs.symbol.flatMap(sym => sym.asTpe) match
           case S(sym: ClassSymbol) =>
             sym.defn match
             case S(td: ClassDef) =>
-              if td.tparams.sizeCompare(targs) =/= 0 then
-                raise(ErrorReport(msg"Wrong number of type arguments" -> trm.toLoc :: Nil)) // TODO BE
```
FlandiaYingman, line 1588: Because of the change above, the checks here started to report false
errors (which was due to the change of the order of `asTpe`, `asMod`,
etc.) 

I think it's not appropriate to do the checks here because it doesn't
have enough information to distinguish overloaded definitions, and we already
have the same check in the resolver. So I just removed them.

hkmc2/shared/src/test/mlscript/basics/BadModuleUses.mls
```diff
@@ -17,7 +17,7 @@ module Example with
 
 :e
 let m = Example.foo()
-//│ ╔══[ERROR] Unexpected moduleful application denoting module definition 'M'.
+//│ ╔══[ERROR] Unexpected moduleful application without resolved definition.
```
FlandiaYingman: The error message here used to be misleading: `Example.foo()` itself
does not "denote" a module defintiion `M`. It's the evaluation result of
`Example.foo()` that denotes a module definition. 

Now it correctly reports that `Example.foo()` has no definition, but I
think it could be improved a bit. Shall we include the type information
in the error message?

LPTK: No, you have it backwards.

Applications don't _have_ definitions. Saying "application without resolved definition" does not make sense.

An application, just like any another term, might _denote_ a module, meaning that it evaluates to such module. `Predef.print` _denotes_ a function defined in `Predef`. Its evaluation result is a runtime representation of the function. We don't normally say this value denote the function, we say it _represents_ it.

hkmc2/shared/src/test/mlscript/basics/BadTypeClasses.mls
```diff
@@ -19,37 +19,31 @@ M.f
 //│ ║        	              ^^^
 //│ ╙── Missing instance: Expected: Int; Available: ‹none available›
 
-using 42 = 42
```
LPTK:
```suggestion
```

hkmc2/shared/src/main/scala/hkmc2/typing/Type.scala
```diff
@@ -0,0 +1,111 @@
+package hkmc2
+package typing
+
+import mlscript.utils.*, shorthands.*
+import semantics.{TypeSymbol, VarSymbol}
+
+object Type:
+end Type
+
```
LPTK: Top-quality definition.
```suggestion
```

hkmc2/shared/src/test/mlscript/OverloadedModulesInSignatures.mls
```diff
@@ -33,7 +33,7 @@ fun f: module TypMod = TypMod
 
 :e
 fun f: TrmMod = 42
-//│ ╔══[ERROR] Expected a non-module type; found reference denoting module definition 'TrmMod'.
+//│ ╔══[ERROR] Expected a non-module type; found reference.
```
LPTK: This message doesn't make much sense :/

LPTK: Anyway, let's stop spending time on module checks

FlandiaYingman: Yeah I see. I am fixing it

FlandiaYingman: OK this is because that the error occurs during type resolution, so at the time the type is not ready. I changed the old message back for errors occurring during type resolution.

FlandiaYingman: > Anyway, let's stop spending time on module checks

Oops, I didn't see this message until I refreshed the page 🫣

FlandiaYingman:
Oh, wait, don't merge it. I have to add the tests

hkmc2/shared/src/test/mlscript/ctx/TypeResolution.mls
```diff
@@ -31,6 +32,57 @@ foo
...
+new Foo2
+//│ Resolved tree:
+//│ Blk:
+//│   stats = Nil
+//│   res = New:
+//│     cls = Ref{sym=member:Foo2} of member:Foo2
+//│     args = Nil
+//│     rft = N
+
+// * dynamic new-expression shouldn't be resolvable because of dynamic nature
```
LPTK: 🤔

FlandiaYingman: Oh, I thought your previous comments there were just to remind me of the possibility of resolution. What I meant by this comment is that we won't make `DynNew` a `Resolvable` because it's dynamic - all mechanism related to `DynNew` should not depend upon the potentially resolvable information!

Do you want me to remove this comment

LPTK: There are zero reason not to resolve the symbol when we have the context to do so, no? Why would we refrain from it? Resolving it could help with type inference, error reporting, optimization, etc.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -46,53 +44,71 @@ object Resolver:
...
     def withTypeArg(param: VarSymbol, arg: Type): ICtx =
       copy(tEnv = tEnv + (param -> arg))
+    
+    extension (t: Type)
+      private def key: Type = t match
+        case Type.Ref(base, _) => Type.Ref(base, Nil)
+        case _ => t
+    extension (lhs: Type)
+      private def =:= (rhs: Type): Bool = lhs <= rhs && rhs <= lhs
+      private def <= (rhs: Type): Bool = (lhs, rhs) match
```
LPTK: Why choose `=:=` and `<=` instead of `<:<`?

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -85,15 +85,26 @@ abstract class Symbol(using State) extends Located:
...
   
   def asBlkMember: Opt[BlockMemberSymbol] = this match
     case mem: BlockMemberSymbol => S(mem)
     case mem: MemberSymbol[?] => mem.defn match
       case S(defn: TypeLikeDef) => S(defn.bsym)
       case S(defn: TermDefinition) => S(defn.sym)
       case N => N
-  
+
+  def asPrincipal = asCls orElse
```
```suggestion
  def asPrincipal =
    asCls orElse
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -85,15 +85,26 @@ abstract class Symbol(using State) extends Located:
...
+    .orElse[TypeSymbol](asAls)
   
   def asBlkMember: Opt[BlockMemberSymbol] = this match
     case mem: BlockMemberSymbol => S(mem)
     case mem: MemberSymbol[?] => mem.defn match
       case S(defn: TypeLikeDef) => S(defn.bsym)
       case S(defn: TermDefinition) => S(defn.sym)
       case N => N
-  
+
```
LPTK, line 101: It would be good to bring back my masterful comment that you lost when you summarily deleted my own version of `asPrincipal`!

LPTK (approved):
LGTM, but pls fix the remaining comments.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: IDK. Is it a convention to use `<:<`? I can change it

LPTK: Yes.

FlandiaYingman:
Done!

hkmc2/shared/src/test/mlscript/ctx/TypeResolution.mls
```diff
@@ -13,7 +13,7 @@ new Foo
 //│     args = Nil
 //│     rft = N
 
-// * dynamic new-expression shouldn't be resolvable because of dynamic nature
+// * TODO: new-expression should also be resolvable for its type
```
LPTK:
```suggestion
// * TODO: dynamic new-expression should also be resolvable for its type
```

FlandiaYingman: Masterful comment!

LPTK: Such a low effort handling of my masterful review comment...

Screenshot 2025-09-15 at 13 45 03

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -85,15 +85,35 @@ abstract class Symbol(using State) extends Located:
...
+  /** 
+   * Get the symbol as the "representative" symbol of a set of
+   * overloaded definitions, or the sole definition, if it is not
+   * overloaded. 
+   *
+   * We should consider the ordering terms > classes/objects/types >
+   * modules, for this purpose.
+   * */
```
LPTK: I don't know why you mangled my masterful comment by ruining its beautiful layout with an arbitrarily low line length restruction.
```suggestion
  /** Get the symbol corresponding to the "representative" of a set of overloaded definitions,
    * or the sole definition, if it is not overloaded.
    * We should consider the ordering terms > classes/objects/types > modules, for this purpose. */
```
FYI we consider that there's no need to wrap lines less than 100 characters. And 100+ɛ is okay.

FlandiaYingman: I have a plugin that automatically wraps the lines. 🫣

hkmc2/shared/src/test/mlscript/ctx/TypeResolution.mls
FlandiaYingman: 🫣

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
FlandiaYingman: Now changing it to 100+ɛ

PR #336, opened 2025-09-16
Unify patterns in UCS and UPS

chengluyu:
### The Updated UCS and UPS Compilation Pipelines

- `Term`: `If` now contains `SimpleSplit`, which represents splits using nested patterns. Nested patterns are represented by `Pattern`, which is also used by pattern definitions.
- `Elaborator`: Since the elaboration of split is implemented by several mutually recursive methods, I placed them in a trait called `SplitElaborator` to control the length of `Elaborator` and maintain modularity.
    - At this step, we report semantic errors that do not require context, such as mixing `do` and `then`, as well as some obvious unreachable cases, like having more branches after an `else`.
- `Lowering` is now responsible for most of the work.
    - `SimpleSplit`s in UCS expressions are expanded to `Split`s using `SplitCompiler`, then normalized by `Normalization`, and finally lowered to `Block`.
    - The naïve compilation (generation of `unapply` and `unapplyStringPrefix` is now done by calling `SplitCompiler` for each `PatternDef` in `Lowering`.
    - The efficient compilation (done by `Instantiator` and `Compiler`) is invoked by `SplitCompiler` when it sees `@compile` annotated on patterns.
    - At this step, we perform all the checks on arguments and parameters.


hkmc2/shared/src/test/mlscript/codegen/ClassMatching.mls
```diff
@@ -120,17 +120,28 @@ fun f(x) = if x is
...
+//│           break split_root$
+//│         } else {
+//│           break split_1$
+//│         }
+//│       } else if (x3 instanceof None1.class) {
+//│         tmp5 = "ok";
+//│         break split_root$
+//│       } else { break split_1$ }
 //│     }
-//│   } else if (x3 instanceof None1.class) { return "ok" } else { return Predef.print("oops") }
```
LPTK, line 133: It looks like the new approach turns some things that were tail calls into non-tail calls.

I wonder if we could preserve them in general (without duplicating the default branches). Seems difficult.

We probably want to add this to the UCS tracking issue (with a link to this message).

hkmc2/shared/src/test/mlscript/codegen/ClassMatching.mls
```diff
@@ -186,17 +197,25 @@ fun f(x) = print of if x is
...
+//│       } else if (x3 instanceof None1.class) {
+//│         tmp9 = "ok";
+//│         break split_root$
+//│       } else { break split_1$ }
 //│     }
-//│   } else if (x3 instanceof None1.class) {
-//│     tmp9 = "ok";
-//│   } else { tmp9 = "oops"; }
+//│     tmp9 = "oops";
+//│     break split_root$;
```
LPTK: Would it be possible to avoid generating this `break` when it doesn't do anything?

hkmc2/shared/src/test/mlscript/ucs/general/LogicalConnectives.mls
```diff
@@ -25,19 +25,31 @@ fun test(x) =
...
+//│       scrut6 = test(42);
+//│       if (scrut6 === true) {
+//│         tmp4 = true;
+//│         break split_root$3
+//│       } else {
+//│         break split_1$3
+//│       }
+//│     } else { break split_1$3 }
+//│   }
+//│   tmp4 = false;
```
LPTK, line 40: As discussed, it would be good to avoid the labels and breaks when the default branch is small (esp. when it's a simple value).

hkmc2/shared/src/test/mlscript/ucs/hygiene/CrossBranchCapture.mls
```diff
@@ -44,12 +44,30 @@ fun process(e) =
...
+//│ ╔══[ERROR] Duplicate pattern variable.
+//│ ║  l.43: 	    Pair(Numb(n), Vec(n))  then n
+//│ ║        	                      ^
+//│ ╟── The previous definition is as follows.
+//│ ║  l.43: 	    Pair(Numb(n), Vec(n))  then n
+//│ ╙──      	              ^
 
 
-// * FIXME should warn, be rejected, or compare both values for equality
+:e
```
LPTK, line 62:
```suggestion
// * TODO: we might want to accept this and compare the two values for equality
:e
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1431,9 +1431,19 @@ extends Importer with ucs.SplitElaborator:
      *  variables we found in `p`. */
     def arrow(lhs: Tree, rhs: Tree): Ctxl[Pattern] =
       val pattern = go(lhs)
-      val termCtx = ctx ++ pattern.variables.allocate
+      val variables = pattern.variables.allocate
+      // - `contextEntries` will be added to the context
```
LPTK, line 1482: TODO: document why symbols are duplicated here; add to future minor improvements in tracking issue.

hkmc2/shared/src/test/mlscript/ucs/hygiene/CrossBranchCapture.mls
chengluyu: FIxed in 0fb7066cd76bbc568ee4dbd17526de86e405bb71.

hkmc2/shared/src/test/mlscript/codegen/ClassMatching.mls
chengluyu: Fixed in 3ca6d5578ffc65212521249af282644f470b462b.

hkmc2/shared/src/test/mlscript/ucs/examples/BinarySearchTree.mls
```diff
@@ -34,7 +34,7 @@ fun insert(t, v) = if t is
   Node(v', l, r) and
     v < v' then Node(v', insert(l, v), r)
     v > v' then Node(v', l, insert(r, v))
-    _ then t
```
LPTK, line 37: Does this no longer work? Any reason for changing it?

hkmc2/shared/src/test/mlscript/ucs/syntax/Else.mls
```diff
@@ -119,3 +119,22 @@ else
...
+
+:w
+fun foo(x) = if x is
+  Box(0) then "The box contains zero."
+  Box(1) then "The box contains one."
+  else "I can't tell you what's inside the box."
+  else "I don't want to tell you what's inside the box."
+  else "I shouldn't tell you what's inside the box."
+//│ ╔══[WARNING] This else clause makes the following branches unreachable.
+//│ ║  l.129: 	  else "I can't tell you what's inside the box."
```
LPTK, line 133: BTW now there are locations attached to keywords, so we should be able to report a more appropriate error location here.

hkmc2/shared/src/test/mlscript/ups/specialization/SimpleLiterals.mls
```diff
@@ -22,6 +22,12 @@ fun checkZeroOneTwo(x) = if x is
 //│ ╔══[ERROR] Unrecognized pattern split (annotated).
 //│ ║  l.14: 	  @compile Two then 2
 //│ ╙──      	   ^^^^^^^^^^^^^^^^^^
+//│ ╔══[WARNING] This else clause makes the following branches unreachable.
+//│ ╟── This branch is unreachable.
```
LPTK: 🤔

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
chengluyu: Resolved in 28336d78a7cc23eb1504d7ffb96c115a0fa45961.

hkmc2/shared/src/test/mlscript/ucs/syntax/Else.mls
chengluyu: Resolved in b80d4bcd3b231729e084ce4cda1fb579608191cc.

hkmc2/shared/src/test/mlscript/ups/specialization/SimpleLiterals.mls
chengluyu: Fixed in c58b4fd94a659509aae892d50ef2fdcf6d61ec55.

hkmc2/shared/src/test/mlscript/ucs/examples/BinarySearchTree.mls
chengluyu: Well, I now regret this change.

hkmc2/shared/src/test/mlscript/ucs/general/LogicalConnectives.mls
chengluyu: Good suggestion. I’ll add it into the backlog.

hkmc2/shared/src/test/mlscript/codegen/ClassMatching.mls
chengluyu: These non-tail calls are added to prevent the control flow from falling through into the outer branches. I think there is a way to preserve the tail calls. I gave it another shot today but the logic is convoluted and I don’t have time to work on it. I’ll add it to the backlog issue.

hkmc2/shared/src/main/scala/hkmc2/semantics/SimpleSplit.scala
```diff
@@ -1,32 +1,36 @@
...
+  case Else(default: Term)(val kw: Opt[Keywrd[`else`.type | `then`.type | `do`.type]])
   case End
   
   inline def ~:(head: SimpleSplit.Head): Cons = Cons(head, this)
   
   def ~~:(front: SimpleSplit): SimpleSplit =
     front match
       case Cons(head, tail) => Cons(head, tail ~~: this)
-      case Else(default) => Else(default)
+      case `else`: Else => `else`
```
LPTK: I'm very uncomfortable with this. It seems to be a Scala 3 quirk. Normally `` `ident` `` in a pattern is treated as a reference to an externally-defined identifier!

hkmc2/shared/src/main/scala/hkmc2/semantics/SimpleSplit.scala
```diff
@@ -1,32 +1,36 @@
...
   
-  lazy val hasElse: Bool = this match
-    case Cons(_, tail) => tail.hasElse
-    case Else(_) => true
-    case End => false
-  
   protected def children: List[Located] = this match
     case Cons(branch, tail) => List(branch, tail)
-    case Else(default) => List(default)
+    case `else` @ Else(default) => `else`.kw match
```
LPTK: Idem.

hkmc2/shared/src/main/scala/hkmc2/semantics/SimpleSplit.scala
chengluyu: Fixed in 69ceb31ecfee23c5c17c90a961efebb72df4b30d.

hkmc2/shared/src/main/scala/hkmc2/semantics/SimpleSplit.scala
chengluyu: Idem.

hkmc2/shared/src/test/mlscript/ucs/examples/BinarySearchTree.mls
chengluyu: Reverted in 55da3fe188a9d236f6f5816102382dc7300c8dd3.

chengluyu:
It’s been a while. I need some time for self review and to record the issues that should be added to the backlog. Then I will ping you and let you know which parts need your review. @LPTK

chengluyu (commented):
I reviewed all the changes. I will then push some fixes for the issues I found.

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
```diff
@@ -336,6 +336,13 @@ enum Pattern extends AutoLocated:
     case Annotated(_, _) => "annotated pattern"
     case Guarded(_, _) => "guarded pattern"
   
+  // Do we really need this method? At first, I wanted to implement `mkClone`
+  // for `SynthIf`, and that eventually led me here. But I feel it may not be
+  // necessary. `SynthIf` might not be copied very often. In addition, the
+  // fields in `FlatPattern` may still need to be adapted according to the new
+  // UCS desugaring logic.
+  def mkClone(using State): Pattern = ???
```
LPTK: So... do we need it? 👀

hkmc2/shared/src/main/scala/hkmc2/semantics/Split.scala
```diff
@@ -20,16 +26,29 @@ enum Split extends AutoLocated with ProductWithTail:
...
+  
+  private var _duplicated: Bool = false
+  
+  def setDuplicated: this.type =
+    if this != End then _duplicated = true
+    this
   
   def duplicate: Split =
-    val copy = this match
+    (this match
```
LPTK, line 46: Why not keeo the `val`?

hkmc2/shared/src/test/mlscript/ucs/examples/Permutations.mls
```diff
@@ -1,5 +1,4 @@
 :js
-
```
LPTK: 👎

hkmc2/shared/src/test/mlscript/ucs/patterns/Compilation.mls
```diff
@@ -0,0 +1,36 @@
+:js
+
+open annotations
+
+:e
+fun trimStart(str) =
+  if str is (@compile ((" " | "\t") ~ rest)) then trimStart(rest) else str
+//│ ╔══[ERROR] Unrecognized pattern (unit).
```
LPTK, line 9: Where is this unit from?

hkmc2/shared/src/test/mlscript/ucs/patterns/Compilation.mls
```diff
@@ -0,0 +1,36 @@
...
+//│ ║  l.7: 	  if str is (@compile ((" " | "\t") ~ rest)) then trimStart(rest) else str
+//│ ╙──     	            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+//│ ╔══[ERROR] Name not found: rest
+//│ ║  l.7: 	  if str is (@compile ((" " | "\t") ~ rest)) then trimStart(rest) else str
+//│ ╙──     	                                                            ^^^^
+
+:pe
+:e
+:w
+// Note: `compile ((" " | "\t") ~ rest)` is parsed as an annotation.
```
LPTK, line 19: Is the `then` parsed as part of the annotation? 🤯

chengluyu:
I just finished the final major refactor of this PR. The latest commit successfully declutters the `Normalization` class.

For historical reasons, `Normalization` not only checked the number of pattern parameters and reported errors, but also performed the efficient and backtracking-free pattern compilation. This made the code look very bloated.  

Now, we check the number of arguments in the pattern and call efficient pattern compilation in the process of expanding splits with nested patterns into splits with flattened patterns (handled by `SplitCompiler`). Since `SplitCompiler` is also responsible for the naïve compilation of patterns, this refactoring not only reduces duplicate code but also makes the error messages about patterns in UPS and UCS more consistent.

Later today I will do the final minor wrap-up for this PR.

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
chengluyu: This method is no longer needed after decluttering, as fixed in 8ce2d07bcb315a14ebaebaea71cb3fe879d89f58.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/FlatPattern.scala
```diff
@@ -3,85 +3,61 @@ package semantics
...
-import FlatPattern.*
-import hkmc2.codegen.Block
-
-// TODO TODO: Renamed to `ResolvedPattern`.
-/** Flat patterns for pattern matching */
+/**
+  * Flattened patterns used in splits for `Normalization` and `Lowering`. All
+  * cases of patterns declared hereby can be matched in constant time and fixed
+  * number of comparisons. Non-trivial patterns (e.g., unions, intersections,
+  * transformations, etc) have been compiled to and represented by `Split`.
```
LPTK:
```suggestion
  * transformations, etc) have been compiled to `Split`.
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1433,7 +1433,7 @@ extends Importer with ucs.SplitElaborator:
      *  determine whether this identifier refers to a pattern, a class, an
      *  object, or creates a new binding.
      * 
-     *  TODO TODO: This routine is insufficient to look up definitions defined
+     *  FIXME: This routine is insufficient to look up definitions defined
```
LPTK, line 1453: So... this is still not fixed?

hkmc2/shared/src/test/mlscript/ups/examples/Record.mls
```diff
@@ -41,12 +41,22 @@ fun bmi'(record) = if record is MatchOrNone[RoundedBMI] as value then value
...
 
 bmi of height: 1.75
 //│ = None
 
+:todo // Make it happen!
+fun bmi'(record) = if record is @compile MatchOrNone(RoundedBMI) as value then value
+//│ ╔══[ERROR] Pattern chaining is not supported in pattern compilation.
+//│ ║  l.20: 	pattern RoundedBMI = CalculateBMI as Rounded
+//│ ╙──      	                     ^^^^^^^^^^^^^^^^^^^^^^^
+//│ ╔══[COMPILATION ERROR] No definition found in scope for member 'value'
```
LPTK: I wonder why it says that `value` is a member. Shouldn't it be a local variable?

hkmc2/shared/src/test/mlscript/ucs/patterns/Compilation.mls
chengluyu:
```
if str is @compile ((" " | "\t") ~ rest) then trimStart(rest) else str
```

The pattern is parsed as part of the annotation. The parser wants to parse a term next, but found `then`.

hkmc2/shared/src/test/mlscript/ucs/patterns/Compilation.mls
chengluyu: I guess it is because the pattern was parsed as part of the annotation, and nothing followed it until the closing parenthesis. As a result, the annotation was not applied to anything. Therefore, the parenthesis is treated as an empty one, which means it becomes unit.

hkmc2/shared/src/main/scala/hkmc2/semantics/Split.scala
chengluyu:
```scala
  def duplicate: Split =
    (this match
      case End => End).setDuplicated
```
It's the return value.

hkmc2/shared/src/test/mlscript/ups/JoinPatterns.mls
```diff
@@ -22,16 +22,18 @@ pattern SimpleJoin = ("abc" | "xyz") ~ ("def" | "")
 "abcxyzdef" is SimpleJoin
 //│ = false
 
-pattern Exponential = ("a" | "b") ~ ("c" | "d") ~ ("e" | "f") // ~ ("g" | "h") ~ ("i" | "j") ~ ("k" | "l") ~ ("m" | "n") ~ ("o" | "p") ~ ("q" | "r") ~ ("s" | "t") ~ ("u" | "v") ~ ("w" | "x") ~ ("y" | "z")
+:todo
```
LPTK, line 25: 👀

hkmc2/shared/src/test/mlscript/ups/examples/EvaluationContext.mls
```diff
@@ -65,17 +65,20 @@ App(Var("a"), Var("a_1")).freshVar("a")
...
-  Abs(Var(name'), body) and
-    name === name' then term
-    let target' = if target.freeVars.has(name')
-      then subst(target, name', target.freshVar(name'))
-      else target
-    else Abs(Var(name'), subst(body, name, target'))
-  App(lhs, rhs) then App(subst(lhs, name, target), subst(rhs, name, target))
+fun (<~>) alphaEquivalent(lhs, rhs) =
+  let varMap = new Map()
+  pattern Var' = Var(n) => if varMap.has(n) then varMap.get(n) else n
```
LPTK: Neat local pattern!

hkmc2/shared/src/test/mlscript/ups/examples/EvaluationContext.mls
```diff
@@ -65,17 +65,20 @@ App(Var("a"), Var("a_1")).freshVar("a")
...
+  fun go(lhs, rhs) = if lhs is
+    Var(n) and rhs is Var' as n' then n === n'
+    App(l, r) and r is App(l', r') then go(l, l') and go(r, r')
+    Abs(Var(n), b) and rhs is Abs(Var(n'), b') then
+      let no = if varMap.has(n) then Some(varMap.get(n)) else None
```
LPTK: Don't we have a map stdlib module that handles such logic?

hkmc2/shared/src/test/mlscript/ups/examples/EvaluationContext.mls
```diff
@@ -161,10 +188,31 @@ subst(Abs(Var("a"), App(Var("b"), Var("a"))), "b", Var("a"))
 
 pattern Value = Abs
 
-// This pattern matches a reducible term and performs the substitution.
-pattern Redex = App(Abs(Var(x), m), Value as n) => subst(m, x, n)
-
-fun redex(term) = if term is Redex as term' then term'
+object ByValue with
```
LPTK, line 194:
```suggestion
module ByValue with
```

hkmc2/shared/src/test/mlscript/ups/syntax/MixedParameters.mls
```diff
@@ -0,0 +1,89 @@
...
+
+fun f(x) = if x is S(a, 0, b) then [a, b]
+
+f([0, 0])
+//│ = [0, 0]
+
+:fixme
+// Order actually matters.
+fun f(x) = if x is S(Bit, a, b) then [a, b]
+//│ ╔══[COMPILATION ERROR] No definition found in scope for member 'a'
```
LPTK, line 59: I'm confused by the error. Shouldn't it say that `a` is not bound in the `then` clause?

LPTK (commented):
Some minor preliminary comments.

hkmc2/shared/src/test/mlscript/ups/JoinPatterns.mls
chengluyu: The code generated by this pattern will grow exponentially. It should be addressed in a future optimized string patterns PR.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
chengluyu: No. I just haven’t done it yet, but it shouldn’t be difficult.

hkmc2/shared/src/test/mlscript/ups/examples/EvaluationContext.mls
chengluyu: This probably won’t work, because later I pass` ByValue` as a parameter to other functions. Reference `module` in this way will cause an error.

hkmc2/shared/src/test/mlscript/ups/syntax/MixedParameters.mls
chengluyu: Yes, the error message here is not clear. I mentioned it in the test block above.

The situation here is a bit tricky. The key is that we cannot distinguish between "subpatterns" and "pattern parameters" during elaboration. Only after name resolution can we decide so by checking `PatternDef`. In this case, `Bit` and `b` are subpatterns, `a` is a pattern argument.

The difference between them is that variables in pattern parameters cannot be used in the consequent branch, while those in subpatterns can. But we create symbols for them during elaboration and assume they can be accessed in the consequent branch.

Perhaps this is another reason to separate pattern parameters from extraction parameters.

hkmc2/shared/src/test/mlscript/ups/examples/EvaluationContext.mls
LPTK: Ah, right.

hkmc2/shared/src/test/mlscript/ups/JoinPatterns.mls
LPTK: Why not say as much in a comment?

LPTK: By the way, I did not expect this code to grow exponentially even without compilation 🤔 How come?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: Pls put it in the tracking issue or fix it now.

hkmc2/shared/src/test/mlscript/ups/syntax/MixedParameters.mls
LPTK: Yeah, that's a very good reason: we should probably write `S[Bit](a, b)`. It can be done in a follow-up PR.

LPTK: (Pls add to tracking issue.)

hkmc2/shared/src/test/mlscript/ups/examples/Record.mls
chengluyu: 🤦‍♂️ I debugged for a while and found that it’s a precedence issue. The precedence of `as` is higher than that of annotation, so what `@compile` actually annotates is `MatchOrNone(RoundedBMI) as value` instead of `MatchOrNone(RoundedBMI)`.

Currently, the pattern compilation cannot yet make the bound variables directly accessible within the branch. I’ll make a note of this as a limitation as well.

hkmc2/shared/src/test/mlscript/ups/JoinPatterns.mls
chengluyu: > By the way, I did not expect this code to grow exponentially even without compilation 🤔 How come?

Because it concatenates many disjunctions. For example, `("a" | "b") ~ ("c" | "d") ~ ("e" | "f") ~ ("g" | "h") ~ ("i" | "j") ~ ("k" | "l")` has six disjunctions in a row, resulting in 2^6 = 64 branches.

The branch deduplication only works for the innermost consequents.

hkmc2/shared/src/test/mlscript/ups/examples/Record.mls
```diff
@@ -49,13 +49,10 @@ bmi of height: 1.75
 //│ = None
 
 :todo // Make it happen!
-fun bmi'(record) = if record is @compile MatchOrNone(RoundedBMI) as value then value
+fun bmi'(record) = if record is (@compile MatchOrNone(RoundedBMI)) as value then value
```
LPTK, line 52: Whoops, looks like we should tighten the precedence of annotations.

LPTK: Please add the original test case to the `ToTriage.mls` file with the comment above.

hkmc2/shared/src/test/mlscript/ups/examples/EvaluationContext.mls
chengluyu: I used it in 571f84ed6b9475bd6c425a000e55ecc06ce7dcd4

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -1092,6 +959,11 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
       (using Subst): (List[ParamList], Block) =
     (paramLists, returnedTerm(bodyTerm))
   
+  /** Same as the other overload, but expect the body to be a `Split`. */
+  def setupFunctionDef(paramLists: List[ParamList], bodySplit: Split, name: Option[Str])
```
LPTK: This is used only once, has a trivial definition, and is not overridden. Why do we need it?

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -371,8 +377,9 @@ sealed trait Statement extends AutoLocated, ProductWithExtraInfo:
       case DynSel(o, f, _) => "dynamic selection"
       case Tup(fields) => "tuple literal"
       case CtxTup(fields) => "contextual tuple literal"
-      case IfLike(Keyword.`if`, body) => "`if` expression"
-      case IfLike(Keyword.`while`, body) => "`while` expression"
+      case IfLike(Keyword.`if`, _) => "`if` expression"
+      case IfLike(Keyword.`while`, _) => "`while` expression"
+      case SynthIf(split) => "`if` expression"
```
```suggestion
      case SynthIf(split) => "synthetic `if` expression"
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -371,8 +377,9 @@ sealed trait Statement extends AutoLocated, ProductWithExtraInfo:
       case DynSel(o, f, _) => "dynamic selection"
       case Tup(fields) => "tuple literal"
       case CtxTup(fields) => "contextual tuple literal"
-      case IfLike(Keyword.`if`, body) => "`if` expression"
-      case IfLike(Keyword.`while`, body) => "`while` expression"
+      case IfLike(Keyword.`if`, _) => "`if` expression"
+      case IfLike(Keyword.`while`, _) => "`while` expression"
```
LPTK: Why these changes?

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -673,6 +683,14 @@ case class Import(sym: Symbol, str: Str, file: os.Path) extends Statement
 
 sealed abstract class Declaration:
   val sym: Symbol
+  
+  /** Whether this declares a class, a pattern, an object, or a pattern
+    * parameter. Only they can be at the constructor position in patterns. */
```
```suggestion
    * parameter. Only these can be in patterns constructor position. */
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1443,46 +1433,95 @@ extends Importer:
...
   
   def pattern(t: Tree): Ctxl[Pattern] =
-    import ucs.Desugarer.{Ctor, unapply}, Keyword.*, Pattern.*, InvalidReason.*
-    import ups.NaiveCompiler.isInvalidStringBounds, ucs.extractors.to
+    import ucs.{Ctor, unapply, error}, ucs.extractors.*, Keyword.*, Pattern.*, InvalidReason.*
     given TraceLogger = tl
+    /** String range bounds must be single characters. */
+    def isInvalidStringBounds(lo: StrLit, hi: StrLit)(using Raise): Bool =
+      val ds = collection.mutable.Buffer.empty[(Message, Option[Loc])]
+      if lo.value.length != 1 then
```
LPTK: Don't use unsafe equality operators (here and elsewhere).

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1443,46 +1433,95 @@ extends Importer:
...
      *  variables we found in `p`. */
     def arrow(lhs: Tree, rhs: Tree): Ctxl[Pattern] =
       val pattern = go(lhs)
-      val termCtx = ctx ++ pattern.variables.allocate
+      // The symbol allocated here will be bound in `split` to the values
+      // destructed from the scrutinee.
+      val variables = pattern.variables.allocate
+      // The `VarSymbol` created here is used as the parameters of the
+      // lambda expression generated by extraction (such as `p => t`).
+      // To avoid repeating `t`, we generate a lambda function before the
```
```suggestion
      // To avoid duplicating `t`, we generate a lambda function before the
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1443,46 +1433,95 @@ extends Importer:
...
+      // branch starts, which will be called if the `split` succeeds.
+      // - `contextEntries` will be added to the context
+      // - `correspondence` is mapping from symbols representing variables
+      //   in the pattern to symbols for parameters to be used in the `term`.
+      val (contextEntries, correspondence) = variables.iterator.map:
+        case (name, symbol) =>
+          // We create a symbol specifically for `Param` for each variable to
+          // avoid repetitively declare symbols in `Scope` during code
```
```suggestion
          // avoid redundantly redeclaring symbols in `Scope` during code
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1443,46 +1433,95 @@ extends Importer:
...
+      Transform(pattern, correspondence, term(rhs)(using ctx ++ contextEntries))
     /** Elaborate tuple patterns like `[p1, p2, ...ps, pn]`. */
     def tuple(ts: Ls[Tree]): Ctxl[Pattern.Tuple] =
-      // We are accumulating three components: the leading patterns, the spread
-      // pattern, and the trailing patterns.
-      val z = (Ls[Pattern](), N: Opt[Pattern], Ls[Pattern]())
-      val (leading, spread, trailing) = ts.foldLeft(z):
-        case (acc @ (_, S(_), _), Spread(Keywrd(`...`), _)) =>
-          // Found two `...p` in the same tuple pattern. Report an error.
+      // We are accumulating two components: the leading patterns, the spred
```
```suggestion
      // We are accumulating two components: the leading patterns, the spread
```

hkmc2/shared/src/test/mlscript/ups/MatchResult.mls
LPTK, line 1: I think `MatchResult` should be renamed to `MatchSuccess`. But this file should still be name `MatchResult` (the name of the type, as in `type MatchResult[A, F] = MatchSuccess[A] | MatchFailure[F]`)

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
```diff
@@ -50,13 +43,23 @@ object Pattern:
...
       varMap.iterator.map: (name, aliases) =>
-        tl.scoped("ucs:translation"):
-          tl.log(s"allocating symbols for variable: ${name}")
-        val symbol = VarSymbol(Ident(name)) // Location is lost.
+        // We need to assign the same symbol to the same name. But I realize
+        // that the following cases will break this constraint: `(x where x > 0)
+        // | (x where x < 0)`. In both alternatives, `x` needs to be allocated
+        // in advanced, but at the level of the disjunction, `x` needs to be
+        // assigned to the same variable. This represents an edge case which
+        // needs to be fixed later.
```
LPTK, line 51: Please add a test case reproducing the problem (if easily feasibly) and add to the tracking issue.

hkmc2/shared/src/main/scala/hkmc2/semantics/SimpleSplit.scala
```diff
@@ -0,0 +1,162 @@
+package hkmc2
+package semantics
+
+import mlscript.utils.*, shorthands.*, syntax.*, Tree.{BoolLit, Keywrd}
+import Keyword.{`do`, `else`, `then`}, utils.TL, Elaborator.{Ctx, State}
+
+enum SimpleSplit extends AutoLocated with ProductWithTail:
```
LPTK, line 10: Add a line documenting what this is?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/FlatPattern.scala
```diff
@@ -3,60 +3,61 @@ package semantics
...
+import syntax.*, Tree.Ident, Elaborator.State
 
-import FlatPattern.*
-import hkmc2.codegen.Block
-
-/** Flat patterns for pattern matching */
+/**
+  * Flattened patterns used in splits for `Normalization` and `Lowering`. All
+  * cases of patterns declared hereby can be matched in constant time and fixed
+  * number of comparisons. Non-trivial patterns (e.g., unions, intersections,
```
LPTK: Seems redundant
```suggestion
  * cases of patterns declared hereby can be matched in constant time.
  * Non-trivial patterns (e.g., unions, intersections,
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ups/Instantiator.scala
```diff
@@ -25,7 +25,14 @@ class Instantiator(using tl: TL)(using Ctx, State, Raise):
   
   /** We instantiate patterns in a breadth-first manner. The queue contains all
    *  patterns that need to be instantiated. */
-  val thingsToDo: MutQueue[Instantiation] = MutQueue.empty
+  val worklist: MutQueue[Instantiation] = MutQueue.empty
+  
+  /** Instantiate an anonymous pattern. */
+  def apply(pattern: SP): (Pat, Context) = scoped("ucs:instantiation"):
+    // TODO: We should not pass an empty map to the `instantiate` method if the
```
LPTK: So will we miscompile things if this happens? Should this be tracked?

hkmc2/shared/src/test/mlscript/codegen/PartialApps.mls
```diff
@@ -245,10 +245,28 @@ _ - _ of 1, 2
 
 // `_` here is interpreted as a UCS default case...
 :sjs
+:w
 let f = if _ then 1 else 0
+//│ ╔══[WARNING] This else clause makes the following branches unreachable.
```
LPTK: I guess the message should read:
```suggestion
//│ ╔══[WARNING] This catch-all clause makes the following branches unreachable.
```

hkmc2/shared/src/test/mlscript/llir/ControlFlow.mls
```diff
@@ -259,6 +259,7 @@ f(A(1))
 
 :sllir
 :intl
+:fixme // Started failing since split deduplication have been introduced.
```
LPTK:
```suggestion
:fixme // Started failing since split deduplication was introduced.
```

hkmc2/shared/src/test/mlscript/ucs/hygiene/CrossBranchCapture.mls
```diff
@@ -44,12 +44,31 @@ fun process(e) =
 //│ ╔══[ERROR] Name not found: n
 //│ ║  l.41: 	    Pair(Vec(xs), Vec(ys)) then n
 //│ ╙──      	                                ^
+//│ ╔══[ERROR] Duplicate pattern variable.
```
LPTK:
```suggestion
//│ ╔══[ERROR] Duplicated pattern variable.
```

hkmc2/shared/src/test/mlscript/ucs/patterns/Compilation.mls
```diff
@@ -0,0 +1,36 @@
+:js
+
+open annotations
+
+:e
```
LPTK, line 6:
```suggestion
// * Note that `(@compile (...))` is currently parsed as `(@compile(...) ())`...
:e
```

hkmc2/shared/src/test/mlscript/ucs/patterns/Compilation.mls
LPTK: Can we add a test with `@compile { (" " | "\t") ~ rest }`? Does it behave correctly? (It should.)

hkmc2/shared/src/test/mlscript/ucs/patterns/ConjunctionPattern.mls
```diff
@@ -8,6 +9,9 @@ object A
 object B
 data class C(a)
 
+// Note: It is normalization that caused the following tests to generate
```
LPTK: Is this referring to a past incorrect behavior? Where is the problem, here?

package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "name": "mlscript",
```
LPTK, line 2: Do you know what caused this change?

LPTK (approved):
Impressive work, thanks!

Let's finally get this merged ASAP (event hough I have not reviewed everything in detail due to lack of time).

package.json
chengluyu: I changed it in this commit: [Keep name field in the lockfile stable](https://github.com/hkust-taco/mlscript/pull/336/commits/09316dd9c94ed43ee286c13cff9f77e2d5e861bc). The reason is that `npm` copies the `name` field from `package.json` to the lockfile and will use the project folder's name if it's missing.

People might clone this repo to folders with different names and then run `npm`. So, it's necessary to keep this field to make the lockfile stable.

hkmc2/shared/src/test/mlscript/ucs/hygiene/CrossBranchCapture.mls
chengluyu: Manually fixed in 2dde44510ef5b335764f98c51622e676f9cc3088.

hkmc2/shared/src/test/mlscript/ucs/patterns/Compilation.mls
chengluyu: Manually fixed in 2dde44510ef5b335764f98c51622e676f9cc3088.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
chengluyu: Manually fixed in 2dde44510ef5b335764f98c51622e676f9cc3088.

hkmc2/shared/src/test/mlscript/ups/MatchResult.mls
chengluyu: Done in 64dad63faae11c2e3dd2e1b92f7eed6d345faedf.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
chengluyu: Manually fixed in 2dde44510ef5b335764f98c51622e676f9cc3088.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
chengluyu: Manually fixed in 2dde44510ef5b335764f98c51622e676f9cc3088.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
chengluyu: Fixed in ed8744a419a5151b24d0c91bd7c50d777ef0fb1f.

hkmc2/shared/src/main/scala/hkmc2/semantics/Pattern.scala
chengluyu: I added it to the tracking issue.

hkmc2/shared/src/main/scala/hkmc2/semantics/SimpleSplit.scala
chengluyu: Done in 679d8b07e34c42bf7872d320256b6d5d05e19f2e.

hkmc2/shared/src/test/mlscript/codegen/PartialApps.mls
chengluyu: Fixed in 62a116ccc79af2a674e6a571324826d9b087b394.

hkmc2/shared/src/main/scala/hkmc2/semantics/ups/Instantiator.scala
chengluyu: It's oudated and removed in 20ff59c2ffe94badd1032455cb74cdd8819b99eb.

hkmc2/shared/src/test/mlscript/ucs/patterns/ConjunctionPattern.mls
chengluyu: Now. The current behavior is incorrect. The normalization assumes all classes are mutually exclusive, causing patterns like `A & B` to be optimized away instead of properly testing both classes.


I clarify the comment in caabab8f6edc1c74438a5e50acb6d69d55a84af3.

hkmc2/shared/src/test/mlscript/ucs/patterns/Compilation.mls
chengluyu: Yes. It does. I added the test in commit 8f51cd104932d0d37c26f91cb51b9f154b69825f.

hkmc2/shared/src/test/mlscript/ucs/patterns/ConjunctionPattern.mls
```diff
@@ -5,26 +5,30 @@ import "../../../mlscript-compile/Stack.mls"
...
 
 
-object A
-object B
-data class C(a)
+abstract class Q
+abstract class A extends Q
+abstract class B extends Q
+
+// Note: The following tests expose a normalization issue where conjunction
```
LPTK, line 12: Wow that's a huge miscompilation issue; should be front and center in the tracking issue!

package.json
LPTK: Ah, great! I've had this problem but didn't know it could be fixed this way 👍

hkmc2/shared/src/test/mlscript/ucs/patterns/ConjunctionPattern.mls
LPTK: Wow that's a huge miscompilation issue; should be front and center in the tracking issue!

[comment moved from the wrong place]

hkmc2/shared/src/test/mlscript/ucs/patterns/ConjunctionPattern.mls
chengluyu: It has always been working like this because we don't have typing or inheritance here. We could support this when we have more type information.

hkmc2/shared/src/test/mlscript/ups/JoinPatterns.mls
LPTK: Ouch.

hkmc2/shared/src/test/mlscript/ucs/patterns/ConjunctionPattern.mls
LPTK: There's no need for full type information. You should just make sure to reason about the inheritance hierarchy. This is a much simpler form of reasoning. BTW, compilation does not require type checking.

LPTK: PS: the fact this huge issue has been around for a long time doesn't make it less huge.

chengluyu: Sure. I added it to the tracking issue.

hkmc2/shared/src/test/mlscript-compile/Runtime.mjs
```diff
@@ -832,17 +838,22 @@ globalThis.Object.freeze(class Runtime {
...
-        } else {
-          tmp = runtime.Unit;
+      split_root$: {
+        split_1$: {
+          scrut = prevHandlerFrame.nextHandler !== null;
+          if (scrut === true) {
+            scrut1 = prevHandlerFrame.nextHandler.handler !== cur.handler;
+            if (scrut1 === true) {
+              prevHandlerFrame = prevHandlerFrame.nextHandler;
+              tmp = runtime.Unit;
```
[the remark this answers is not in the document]
LPTK: Wow, thanks Copilot!!!

PR #337, opened 2025-09-17
Blocks No Longer `extends AutoLocated`

ychenfo:


LPTK (approved):
LGTM, and thanks for carefully updating the comment!

PR #338, opened 2025-09-24
`Value.Lam`, `Value.Arr` and `Value.Rcd` now extend `Result`

ychenfo:


hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -550,6 +550,13 @@ case class Call(fun: Path, args: Ls[Arg])(val isMlsFun: Bool, val mayRaiseEffect
 
 case class Instantiate(mut: Bool, cls: Path, args: Ls[Arg]) extends Result
 
+case class LamRes(params: ParamList, body: Block) extends Result
+
+case class ArrRes(mut: Bool, elems: Ls[Arg]) extends Result
+
+case class RcdRes(mut: Bool, elems: Ls[RcdArg]) extends Result
+
```
ychenfo, line 558: I am not sure if these names are good enough...

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -152,7 +152,7 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
       reportAnnotations(decl, annotations)
       blockImpl(DefineVar(sym, Term.Lit(Tree.UnitLit(false))) :: stats, res)(k)
     case DefineVar(sym, rhs) :: stats =>
-      subTerm(rhs): r =>
+      term(rhs): r =>
         Assign(sym, r, blockImpl(stats, res)(k))
```
ychenfo, line 156: Since the `rhs` of `Assign` accepts a `Result`, I think we can just use `term` here? This avoids `subTerm` creating unnecessary temp variables for `Results`.

For example, before this change the program `let x = 1 :: Nil` compiles to:
```
let x, tmp; tmp = Cons(1, Nil); x = tmp;
```

After the change it compiles to
```
let x; x = Cons(1, Nil);
```

hkmc2/shared/src/test/mlscript/basics/ObjectExtensions.mls
```diff
@@ -48,6 +48,5 @@ object Bar extends! Foo!class
 //│ ║  l.46: 	object Bar extends! Foo!class
 //│ ║        	                  ^
 //│ ╙── The 'new' keyword requires a statically known class; use the 'new!' operator for dynamic instantiation.
-//│ ═══[RUNTIME ERROR] TypeError: Class extends value (...args) => {   runtime.checkArgs("", 1, true, args.length);   let arg = args[0];   return ! arg } is not a constructor or null
```
ychenfo, line 51: Before this PR it reports error because the program compiles to
```
...
class Bar4 extends ((arg) => {
  return ! arg
})
...
```

Now it's
```
...
class Bar4 extends (lambda)
...
```
And I guess that's why this runtime error no longer exists after this PR.

hkmc2/shared/src/test/mlscript/bbml/bbGetters.mls
```diff
@@ -89,16 +89,23 @@ fun test2() =
...
 //│   funny = function funny() {
-//│     let lambda;
+//│     let lambda, lambda1;
 //│     lambda = (undefined, function (caseScrut) {
+//│       if (caseScrut === 0) {
+//│         return 0
+//│       } else {
+//│         throw globalThis.Object.freeze(new globalThis.Error("match error"))
+//│       }
+//│     });
```
ychenfo, line 101: When lambda is `Value.Lam`, the whole lambda body is ignored [here](https://github.com/hkust-taco/mlscript/blob/7b4e86d4513d3dc1bdbbb93378ce036e03ca5e96/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala#L128), but now only the variable `lambda` becomes unused.

hkmc2/shared/src/test/mlscript/codegen/IfThenElse.mls
```diff
@@ -9,14 +9,11 @@ if true then 1 else 0
 :sjs
 let f = x => if x then print("ok") else print("ko")
 //│ JS (unsanitized):
-//│ let f, tmp, lambda;
+//│ let f, lambda;
 //│ lambda = (undefined, function (x) {
-//│   if (x === true) {
-//│     return Predef.print("ok")
-//│   } else { return Predef.print("ko") }
+//│   if (x === true) { return Predef.print("ok") } else { return Predef.print("ko") }
```
LPTK, line 14: Weird, do you have any idea why there's a change here?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -947,11 +965,17 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
     val b = rec(as)
-    val args = if fsr.isEmpty then asr else Arg(N, Value.Rcd(mut = false, fsr.reverse)) :: asr
-    Begin(
-      b,
-      k(args.reverse)
-    )
+    if fsr.isEmpty then
+      Begin(b, k(asr.reverse))
+    else
+      val arrSym = new TempSymbol(N, "rcd")
```
LPTK: Shouldn't this be
```suggestion
      val rcdSym = new TempSymbol(N, "rcd")
```
?

LPTK (approved):
Thanks, this LGTM. But why use an ugly `Res` postfix in `Lam`/`Arr`/`Rcd`?

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
LPTK: Ah, see my concurrent comment below 😛 

Can we drop the "Res" postfixes?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
ychenfo: Ah, thanks for the catch!

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
LPTK: Yes, nice change!

hkmc2/shared/src/test/mlscript/basics/ObjectExtensions.mls
LPTK: Does it really still read `extends (lambda)`? Couldn't it just be `extends lambda` now?

hkmc2/shared/src/test/mlscript/codegen/IfThenElse.mls
ychenfo: Since the previous program has another `tmp` variable, maybe that changes how [`mkString(100)`](https://github.com/hkust-taco/mlscript/blob/f3b0969dbdc0f4a4872462be1141907346e0725d/hkmc2DiffTests/src/test/scala/hkmc2/JSBackendDiffMaker.scala#L87) formats the output?

hkmc2/shared/src/test/mlscript/basics/ObjectExtensions.mls
ychenfo: Ah yes, it's indeed `extends lambda` in the `:sjs` output. In the comment above I deleted the `(arg) => ...` and manually typed in `lambda` without deleting the brackets😅

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
ychenfo: Yeah I can't decide which is better.. If we use shorthand `Lam` is also used in `Term.Lam`. If we use full name `Array` is scala built-in. Maybe I should just override the scala builtin array and use the full names `Lambda`, `Array` and `Record`?

hkmc2/shared/src/test/mlscript/codegen/IfThenElse.mls
LPTK: It's quite surprising to me. There's probably something funky in the rendering logic there.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
LPTK: We could also just use `Tuple` instead of `Array`. In my mind, a tuple is an array literal, which is what this is.

ychenfo: Thanks for the suggestion! I have updated the name to be `Lambda`, `Tuple` and `Record`.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
ychenfo: Fixed in 6075d309a1dd7914223337f28358a4f3da026f05

PR #339, opened 2025-09-24
Use existing TermSymbol for ValDefn

Derppening:
This ensures that class fields and `ValDefn`s in the constructor body refer refer to the same symbol.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
@@ -174,7 +174,7 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
             subTerm_nonTail(bod)(r =>
               // Assign(td.sym, r,
               //   term(st.Blk(stats, res))(k)))
-              Define(ValDefn.mk(td.owner, knd, td.sym, r),
+              Define(ValDefn(td.tsym, td.sym, r),
                 blockImpl(stats, res)(k)))
           case syntax.Fun =>
             val (paramLists, bodyBlock) = setupFunctionOrByNameDef(td.params, bod, S(td.sym.nme))
@@ -184,7 +184,7 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
             // Implicit instances are not parameterized for now.
             assert(td.params.isEmpty)
             subTerm(bod)(r =>
-              Define(ValDefn.mk(td.owner, syntax.ImmutVal, td.sym, r),
+              Define(ValDefn(td.tsym, td.sym, r),
                 blockImpl(stats, res)(k)))
           case syntax.LetBind | syntax.ParamBind | syntax.HandlerBind => fail:
             ErrorReport(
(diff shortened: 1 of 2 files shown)


LPTK (approved):
LGTM, thx

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -184,7 +184,7 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
             // Implicit instances are not parameterized for now.
             assert(td.params.isEmpty)
             subTerm(bod)(r =>
-              Define(ValDefn.mk(td.owner, syntax.ImmutVal, td.sym, r),
+              Define(ValDefn(td.tsym, td.sym, r),
```
FlandiaYingman, line 187: This part changed the semantics. I am not sure if that's a problem, so just to note it here.

Implicit instances used to be compiled into `ValDefn` of kind `ImmutVal`. Now it has been changed to `ValDefn` of kind `Ins`. 

But in the JS Backend, all `ValDefn` are blindly compiled to the same thing regardless of their kind, so it doesn't change the current behavior.

```diff
using Int = 42
//│ Elab: { using member:instance$Ident(Int)‹536›: member:Int‹127›#666 = 42; }
//│ Lowered:
//│ Program:
//│   imports = Nil
//│   main = Define:
//│     defn = ValDefn:
- //│       tsym = .instance$Ident(Int)(value)‹542›
+ //│       tsym = .instance$Ident(Int)(implicit instance)‹542›
//│       sym = member:instance$Ident(Int)‹536›
//│       rhs = Lit of IntLit of 42
//│     rest = Assign: \
//│     lhs = $block$res‹541›
//│     rhs = Lit of UnitLit of false
//│     rest = End of ""
```
###### (I added some code to show the kind)

FlandiaYingman (approved):
LGTM

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
LPTK: I think the correct approach is to not expose the Ins kind to the lowered tree, which shouldn't care about what was implicit in the  source program (te whole point of lowering is to forget irrelevant details of the source program). So @Derppening I think you might want to revert that change. Was the chnage needed? (I guess it may be related to that assertion we discussed on Discord.)

FlandiaYingman: > I think the correct approach is to not expose the Ins kind to the lowered tree, which shouldn't care about what was implicit in the source program (te whole point of lowering is to forget irrelevant details of the source program).

If that's the case, I think we should keep David's change. Here we are not explicitly exposing the kind, but just that the kind is embedded in the TermSymbol.

But we might want to remove the following line because it exposes the "irrelevant details" (and I checked it is not used at all)!

```diff
 final case class ValDefn(
     tsym: TermSymbol,
     sym: BlockMemberSymbol,
     rhs: Path,
 ) extends Defn:
   val innerSym = S(tsym)
-  val k = tsym.k
   val owner: Opt[InnerSymbol] = tsym.owner
```

LPTK: Ah yes, I agree with your change and with keeping David's change.

FlandiaYingman: OK I just fixed it.

PR #342, opened 2025-09-27
Lifter optimization

CAG2Mark:
Avoid creating multiple closures for the same function, i.e.
```
fun f(x) =
  fun g() = x
  [g, g]
```
will allocate only one closure for `g`.

There are still some cases where closures are unnecessarily allocated, but these require further changes that require a lot more changes to the code and some extra analysis.

(Also fixed the `FlowSymbol` weirdness that we discussed in private.)


hkmc2/shared/src/test/mlscript/lifter/FunInFun.mls
```diff
@@ -625,3 +629,205 @@ fun f(es, vs) =
 //│ });
 //│ f23 = function f(es, vs) { let v$_; v$_ = es; return A$(false, v$_) };
 
+// try to avoid creating multiple closures
+
+:sjs
+fun f(x) =
+  fun g = x
```
LPTK: TODO:
```suggestion
  fun g() = x
```

hkmc2/shared/src/test/mlscript/lifter/FunInFun.mls
```diff
@@ -625,3 +629,205 @@ fun f(es, vs) =
...
+//│   capture = new f$capture19(x1);
+//│   g$here = runtime.safeCall(g23(capture));
+//│   a1 = g$here;
+//│   tmp10 = capture.x0$ + 1;
+//│   capture.x0$ = tmp10;
+//│   g$here = runtime.safeCall(g23(capture));
+//│   return globalThis.Object.freeze([ a1, g$here ])
+//│ };
+
+// Can be avoided in `h` by passing the initialized closure, but this is more complicated
```
LPTK, line 457: TODO: document that rewriting `h() = g` to `h() = a` automatically would resolve the issue.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -83,20 +83,22 @@ object Lifter:
   object AccessInfo:
     val empty = AccessInfo(Set.empty, Set.empty, Set.empty)
 
+  type LocalVarSymbol = VarSymbol | TempSymbol
+  
   def getVars(d: Defn)(using state: State): Set[Local] = d match
     case f: FunDefn =>
       (f.body.definedVars ++ f.params.flatMap(_.paramSyms)).collect:
-        case s: FlowSymbol if !(s is state.runtimeSymbol) => s
+        case s: LocalVarSymbol if !(s is state.runtimeSymbol) => s
```
LPTK: TODO: remove the `runtimeSymbol` special-case here and later.

CAG2Mark: OK I double checked. It seems I added this because `runtimeSymbol` is actually a `TempSymbol` and we definitely do not want to capture that.

But since `runtime` is never assigned to, it is never caught by the lifter. However, I've made a small change to `UsedVarAnalyzer` to allow us to explicitly specify any symbols we want to ignore.

CAG2Mark:
OK, done

LPTK:
@CAG2Mark Ah, there's a conflict now. After you fix it we can merge this.

CAG2Mark:
:(

CAG2Mark:
:)

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -649,6 +678,41 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
...
+                    (dflt2 is dflt) && (rst2 is rst)
+                  then b else Match(scrut2, arms2, dflt2, rst2)
+            
+        case Label(lbl, bod, rst) =>
+          val lbl2 = applyLocal(lbl)
+          val bod2 = applySubBlockAndReset(bod)
+          val rst2 = applySubBlock(rst)
+          if (lbl2 is lbl) && (bod2 is bod) && (rst2 is rst) then b else Label(lbl2, bod2, rst2)
+        case Begin(sub, rst) =>
+          val sub2 = applySubBlockAndReset(sub)
```
LPTK: Do we need to reset, here? `Begin` represents a sequence; what's defined in `sub` should be available in `rst`.

hkmc2/shared/src/test/mlscript/lifter/FunInFun.mls
LPTK, line 1: I think this file has way too much JS output, especially with the newly added cases. Please remove or comment out all `:sjs` commands except for just 2 or 3 uses showing the most interesting/difficult cases. These should be enough to catch regressions if something stops working in the transformation, which is what diff-testing is all about.

LPTK: Doing so in the other `lifter/` files would also be nice.

LPTK (approved):
Great, thanks! Please address my comments and then we can merge.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: Oh, ok

CAG2Mark:
Done

PR #343, opened 2025-09-27
`Term.Resolved`

FlandiaYingman:
This is a draft PR that adds `Term.Resolved` to indicate that a term's symbol is resolved. This information can be used later in the lowering stage. For example, append `.class` to a ref to an overloaded module definition.


FlandiaYingman:
I am facing some problems now:

- Now only `Ref` can expand to be appended with `.class`. But we should also do that to selections. (This is the reason that the current test cases fail.) However, this is particularly difficult for selections because there is already an existing symbol mechanism for selections. I tried a bit for a few hours, but I guess I am going into another refactorization hell 🤥...

FlandiaYingman:
OK I managed to wrap selections with `Term.Resolved`. The next step is to remove the existing symbols in selections, and replace them with the symbols in `Term.Resolved`.

FlandiaYingman:
Hello @CAG2Mark and @chengluyu, I need your help on this PR 🤧🥺, particularly regarding the lifter and the UCS/UPS normalization.

### Summary on this PR

In this PR, we try to disambiguate `Term.Ref`'s and `Term.Sel`'s `BlockMemberSymbol` during resolution (disambiguation means to infer which overloaded definition does the term refer to in the specific context), by wrapping the terms that have to be disambiguated in a term `Term.Resolved(t: Term, sym: DefinitionSymbol[?])`. (`DefinitionSymbol` is the same as `MemberSymbol`, except that `BlockMemberSymbol` is not a `DefinitionSymbol` but a `MemberSymbol`).

A direct consequence is that, we now have enough information in the lowering/codegen stage to defer the process of adding the `Sel(t, new Ident(".class")` thing for a `Ref` or `Sel` referring to a overloaded term. Originally it was in the elaborator, and then with one of my previous PR it was moved to the resolver. Now in this PR, it will be moved to the JS builder, which is also the right place, as this was just a JS trick to compile overloaded definitions.

I changed `Value.Ref` to encode the disambiguation information (to be used later during code generation):

```diff
- case Ref(l: Local)
+ case Ref(l: Local, disamb: Opt[DefinitionSymbol[?]])
```

and `Select` to encode the disambiguated symbol information (we want to be as specific as possible):

```diff
- case class Select(qual: Path, name: Tree.Ident)(val symbol: Opt[FieldSymbol])
+ case class Select(qual: Path, name: Tree.Ident)(val symbol_SelectSymbol: Opt[DefinitionSymbol[?]])
```

Then, the JS builder automatically, based on the information encoded, decides whether or not to append the `.class` suffix to the generated code for a `Ref` or `Select`. The exact logic of appending `.class` is here:

```scala
    case Value.Ref(l, disamb) => l match
      case l: BlockMemberSymbol if {
        // TODO @Harry: refactor(ize) duplicated logic
        l.hasLiftedClass && 
        disamb.flatMap(d => d.asMod orElse d.asCls).exists(_ isnt ctx.builtins.Array)
      } =>
        doc"${getVar(l, l.toLoc)}.class"
      case _ =>
        getVar(l, r.toLoc)
```

```scala
    case s @ Select(qual, id) => 
      val dotClass = s.symbol_SelectSymbol match
        case S(ds) if {
          val v = ds.asBlkMember.exists(_.hasLiftedClass) && 
          (ds.asMod orElse ds.asCls).exists(_ isnt ctx.builtins.Array)
          v
        } => doc".class"
        case _ => doc""
      val name = id.name
      doc"${result(qual)}${
        if isValidFieldName(name)
        then doc".$name"
        else name.toIntOption match
          case S(index) => doc"[$index]"
          case N => doc"[${makeStringLiteral(name)}]"
      }${dotClass}"
```

### What's the Problem

In the lifter and the UCS/UPS normalization, there is a bunch of customized logic of appending `.class` to a term. With this PR, they now append duplicate `.class` suffixes. 

Also, seems like the compiler largely transforms the terms in the lifter and the UCS/UPS normalization process. Some encoded information is stripped out, but I couldn't find out the reason 😥.

### UCS/UPS Normalization

I am taking `EvenOddTree.mls` as an example.

Looking at the resolved tree for the patterns:

```fs
:rt
pattern OddTree =
  A |
  Node(EvenTree, A, EvenTree) | Node(OddTree, A, OddTree) |
  Node(EvenTree, B, OddTree) | Node(OddTree, B, EvenTree)
pattern EvenTree =
  B |
  Node(EvenTree, A, OddTree) | Node(OddTree, A, EvenTree) |
  Node(EvenTree, B, EvenTree) | Node(OddTree, B, OddTree)
//│ Resolved tree:
//| ...
//│             left = Composition:
//│               polarity = true
//│               left = Constructor:
//│                 target = Resolved{sym=object:A,typ=object:A}:
//│                   t = Ref{sym=member:A} of member:A
//│                   sym = object:A
//│                 patternArguments = Nil
//│                 arguments = N
//│               right = Constructor:
//│                 target = Resolved{sym=term:class:Node.Node}:
//│                   t = Ref{sym=member:Node} of member:Node
//│                   sym = term:class:Node.Node
```

The references to `A` and `Node` are resolved correctly to `object:A` and `term:class:Node.Node` (which is the constructor of `Node`). But after lowering, the lowered tree becomes:

```fs
:lot
B is @compile EvenTree
//| Lowered tree:
//| ...
//│               path = Ref:
//│                 l = member:A
//│                 disamb = N
//| ...
//│               path = Select{sym=class:Node}:
//│                 qual = Ref:
//│                   l = member:Node
//│                   disamb = N
//│                 name = Ident of "class"
```

while we expect it to be

```fs
:lot
B is @compile EvenTree
//| Lowered tree:
//| ...
//│               path = Ref:
//│                 l = member:A
//│                 disamb = object:A
//| ...
//│               Ref:
//│                   l = member:Node
//│                   disamb = term:class:Node.Node
```

Would you provide some help on this case? Thanks Luyu 😉

### Lifter

Taking `lifter/ClassWithCompanion.mls` as an example:

The `C.empty` in

```fs
fun foo(x) =
  class C(y) with
    fun get = [x, y]
  module C with
    val empty = C(123)
  C.empty
```

should be compiled to JS `C.class.empty`. The resolved tree indicates that it is correctly resolved that `C` actually refers to a module `C`:

```fs
//│         res = Resolved{sym=term:module:C.empty}:
//│           t = Sel{sym=member:empty}:
//│             prefix = Resolved{sym=module:C,typ=module:C}:
//│               t = Ref{sym=member:C} of member:C
//│               sym = module:C
//│             nme = Ident of "empty"
//│           sym = term:module:C.empty
```

Afte lowering, it is expected to be

```
//│         rhs = Select{sym=term:module:C.empty}:
//│           qual = Ref{disamb=module:C}:
//│             l = member:C
//│             disamb = S of module:C
//│           name = Ident of "empty"
```

However, if `:lifter` is specified, the inner reference magically loses its disambiguation information:

```
//│         rhs = Select{sym=term:module:C.empty}:
//│           qual = Ref:
//│             l = member:C
//│             disamb = N
//│           name = Ident of "empty"
```

Would you look into it? Thanks Mark 💓

CAG2Mark:
Oh cool, something to do in tomorrow's conference when I am bored

CAG2Mark:
You can try this commit @FlandiaYingman

https://github.com/CAG2Mark/mlscript/commit/f24b45474c25d8fea15f6de2c843e4fd9b9dbd9c

chengluyu:
@FlandiaYingman 🥹 To be honest, I still don’t quite understand what you want to do. Let’s set a time on Discord to discuss it in person.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -42,17 +42,17 @@ class Normalization(using tl: TL)(using Raise, Ctx, State) extends TermSynthesiz
...
-        case S(mem: BlockMemberSymbol) =>
-          // If the class is declaration-only, we do not need to select the
-          // class.
-          if !mem.hasLiftedClass || mem.defn.exists(_.hasDeclareModifier.isDefined) then
-            lhs.constructor
-          else
-            Term.SynthSel(lhs.constructor, Tree.Ident("class"))(mem.clsTree.orElse(mem.modOrObjTree).map(_.symbol), N).resolve
-        case _ => lhs.constructor
-      lhs.copy(constructor)(lhs.tree, lhs.output)
+      // val constructor = lhs.constructor.symbol match
```
LPTK: 👀

FlandiaYingman:
Finally, now only 15 cases fail. 

- [ ] Lifter (I applied Mark's patch, but a new problem arrives!)
- [x] LLIR
- [x] Handler
- [x] WASM

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -113,9 +113,9 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
       else errExpr(msg"Illegal reference to builtin symbol '${l.nme}'")
     case Value.Ref(l, disamb) => l match
       case l: BlockMemberSymbol if {
-        // TODO @Harry: refactor(ize) duplicated logic
+        // println(disamb.flatMap(d => d.asModOrObj orElse d.asCls))
```
LPTK: Strange comment substitution 👀

hkmc2/shared/src/test/mlscript/codegen/MergeMatchArms.mls
```diff
@@ -19,13 +19,13 @@ if a is
   C then 3
   D then 4
 //│ JS (unsanitized):
-//│ if (a instanceof A1.class) {
+//│ if (a === A1) {
```
LPTK: This logic is incorrect. In MLscript, currently, object can be extended further.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -591,7 +591,7 @@ class HandlerLowering(paths: HandlerPaths, opt: EffectHandlers)(using TL, Raise,
         f.owner match
         case None => S(Call(f.sym.asPath, params)(true, true))
         case Some(owner) => 
-          S(Call(Select(owner.asPath, Tree.Ident(f.sym.nme))(S(f.innerSym.asInstanceOf)), params)(true, true))
+          S(Call(Select(owner.asPath, Tree.Ident(f.sym.nme))(f.innerSym.asInstanceOf), params)(true, true))
           // TODO:                                               ^^^^^^^^
```
LPTK: This TODO is now misaligned!

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -844,25 +844,57 @@ class Resolver(tl: TraceLogger)
...
-                t.expand(S(t.withSym(bms)))
-            log(s"Resolved symbol for ${t}: ${bms}")
-          case N => 
-            t.expand(S(t.withSym(ErrorSymbol(id.name, Tree.Dummy))))
-            raise: 
-              ErrorReport(
-                msg"${mdef.kind.desc.capitalize} '${mdef.sym.nme}' " +
-                msg"does not contain member '${id.name}'" -> t.toLoc :: Nil,
-                extraInfo = S(mdef))
+                // TODO @Harry: Appropriately resolve all selections on classes.
```
LPTK, line 903: 👀 

(Just commenting here to keep track of the TODO.)

hkmc2/shared/src/test/mlscript/ctx/Summon.mls
```diff
@@ -62,7 +62,6 @@ fun f(using Int) = use[Num]
 
 data class Some[T](value: T)
 
-
```
LPTK, line 65: 👎

hkmc2/shared/src/test/mlscript/codegen/MergeMatchArms.mls
FlandiaYingman: Ah okay, will change it

FlandiaYingman:
OK the reason why WASM failed is because I forgot to `npm install`. 🤡

FlandiaYingman:
Now only 4 test cases fail!

- [x] Lifter (asking Mark)
- [x] Object Buffer

FlandiaYingman:
Now only 2 test cases fail!

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -670,20 +671,34 @@ final case class HandlerTermDefinition(
...
+  
+  def extractMembers(blk: Term.Blk): Ls[ErrorReport] \/ Map[Str, BlockMemberSymbol] =
+    val (errs, mems) = blk.stats.collect:
+      case td: TermDefinition => td.sym -> td
+      case td: ClassLikeDef => td.bsym -> td
+      case td: TypeDef => td.bsym -> td
+    .groupBy(_._1.nme)
+    .partitionMap: (nme, syms) =>
+      if syms.map(_._1).distinct.tail.nonEmpty then L:
+        (msg"Found duplicate members with the same name '${nme}'." -> N) ::
```
```suggestion
        (msg"Duplicate definition of member named '${nme}'." -> N) ::
```

hkmc2/shared/src/test/mlscript/codegen/ThisCallVariations.mls
```diff
@@ -53,9 +53,17 @@ id(Example) .!. oops(2)
...
 data class Example2(val a) with
   val a = 1
   fun f(inc) = Example2(a + inc)
+//│ ╔══[ERROR] Found duplicate members with the same name 'a'.
+//│ ╟── Defined at: 
+//│ ║  l.57: 	data class Example2(val a) with
+//│ ║        	                        ^
+//│ ╟── Defined at: 
+//│ ║  l.58: 	  val a = 1
+//│ ╙──      	          ^
```
LPTK: Strange location 🤔

hkmc2/shared/src/test/mlscript/ctx/TypeResolution.mls
```diff
@@ -231,3 +235,81 @@ fun f(o: O) = o
 //│       annotations = Nil
 //│       companion = N
 //│   res = Lit of UnitLit of false
+
+// The implicit `this` should be resolved correctly...
```
LPTK: 👀

FlandiaYingman:
Hoooooooooray!!! Thanks @CAG2Mark and @chengluyu 💞 for the help

I guess this PR is almost ready. I'll complete the cleanup tomorrow (today).

In this PR, the symbols in `Term.Sel` and `Term.SynthSel` were not removed, and the symbol resolution in the elaborator was also not completely moved to the resolver. I'll work on them when I get my courage back to touch on the symbols :<

But the good news (maybe) is that the two symbol resolution mechanism are completely separated. i.e., the lowering stage solely relies on the symbol resolved by the resolver. This shall facilitate the WASM backend and other backends (if any).

hkmc2/shared/src/test/mlscript/ctx/TypeResolution.mls
FlandiaYingman: What?

FlandiaYingman:
It's ready for review~

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -313,6 +313,79 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
       subTerm_nonTail(arg): ar =>
         k(Arg(spread = S(true), ar) :: Nil)
   
+  def ref(ref: st.Ref, disamb: Opt[DefinitionSymbol[?]], inStmtPos: Bool)(k: Result => Block)(using Subst): Block =
+    def warnStmt = if inStmtPos then
+      raise:
+        WarningReport(msg"Pure expression in statement position" -> ref.toLoc :: Nil, S(ref))
+    
+    val st.Ref(sym) = ref
```
LPTK: Why not
```suggestion
    val sym = ref.symbol
```
?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -453,51 +463,58 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
+        case sym: sem.TermSymbol => sym.k is syntax.Fun
         // Do not perform safety check on `MatchResult` and `MatchFailure`.
         case sym => (sym is State.matchResultClsSymbol) ||
           (sym is State.matchFailureClsSymbol)
       def conclude(fr: Path) = lowerCall(fr, isMlsFun, arg, t.toLoc)(k)
       // * We have to instantiate `f` again because, if `f` is a Sel, the `term`
       // * function is not called again with f. See below `Sel` and `SelProj` cases.
       f.instantiated match
-      case t if t.resolvedSym.exists(_ is ctx.builtins.js.bitand) =>
+      case t if t.resolvedSym.flatMap(_.asBlkMember).exists(_ is ctx.builtins.js.bitand) =>
```
LPTK: Let's just define a lazy `resolvedBlkMemberSym` field and avoid all this horribly inefficient flatmapping and existing.

hkmc2/shared/src/test/mlscript/ctx/TypeResolution.mls
LPTK: The comment's meaning is unclear.

LPTK (commented):
A couple preliminary comments.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -560,19 +560,42 @@ case class Record(mut: Bool, elems: Ls[RcdArg]) extends Result
...
   def selSN(id: Str): Path = selN(new Tree.Ident(id))
   def asArg = Arg(spread = N, this)
 
-case class Select(qual: Path, name: Tree.Ident)(val symbol: Opt[FieldSymbol]) extends Path with ProductWithExtraInfo:
-  def extraInfo: Str = symbol.mkString
+/**
+ * @param symbol The symbol representing the definition that the selection refers to, if known.
+ */
+case class Select(qual: Path, name: Tree.Ident)(val symbol: Opt[DefinitionSymbol[?]]) extends Path with ProductWithExtraInfo:
+  def extraInfo: Str =
```
LPTK: I can't make sense of this implementation choice. It almost seems like it's been voluntarily obfuscated. You put an option in a singleton list and then collect it back?! And use `mkString` with a separator for a collection of at most one element?!

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -591,7 +591,7 @@ class HandlerLowering(paths: HandlerPaths, opt: EffectHandlers)(using TL, Raise,
         f.owner match
         case None => S(Call(f.sym.asPath, params)(true, true))
         case Some(owner) => 
-          S(Call(Select(owner.asPath, Tree.Ident(f.sym.nme))(S(f.sym)), params)(true, true))
+          S(Call(Select(owner.asPath, Tree.Ident(f.sym.nme))(N), params)(true, true))
```
LPTK, line 594: I wonder why `f.sym` was there 👀

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -234,15 +234,18 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
...
+      case PubField(isym, sym) => Select(isym.asPath, Tree.Ident(sym.nme))(N)
       
     def asArg = read.asArg
     
     def assign(value: Result, rest: Block) = this match
       case Sym(l) => Assign(l, value, rest)
       case PubField(isym, sym) => AssignField(isym.asPath, Tree.Ident(sym.nme), value, rest)(S(sym))
-    
-    
+
```
LPTK: There were two empty lines; now there is only one.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -1024,6 +1021,7 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
             val isMutSym = VarSymbol(Tree.Ident("isMut"))
             
             var curSym = TempSymbol(None, "tmp")
+            // TODO: This special case Select(..., "class") is redundant after this PR.
```
LPTK: Why not remove it, then?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -1185,9 +1185,10 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
       case c: ClsLikeDefn => c.copy(owner = N)
       case d => d
 
+    // TODO: The special case Select(..., "class") is redundant after the PR.
```
LPTK: Why not remove it, then?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -1288,7 +1290,11 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
                 case Some(bms) =>
                   val nestedIn = analyzer.defnsMap(bms)
                   nestedIn match
-                    case cls: ClsLikeDefn => S(c.sym -> TermSymbol(syntax.ImmutVal, S(cls.isym), Tree.Ident(c.sym.nme + "$")))
+                    // These will be the names of the objects/modules after being lifted
+                    // We should use the nested object/module's **original name** if nested inside a class,
+                    // so they can be accesed directly by name from te outside.
```
```suggestion
                    // so they can be accesed directly by name from the outside.
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -1288,7 +1290,11 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
...
+                    // For example, if a class C has an object M, (new C).M as a dynamic selection works
```
LPTK, line 1292: I don't understand this comment. Why would we access lifted definitions from the outside? Can't make sense of that. Wouldn't the behavior change when lifting is on and off? Lifting is supposed to preserve semantics.

Besides, we probably won't want to make nested classes accessible through dynamic selection.

hkmc2/shared/src/test/mlscript/codegen/SanityChecks.mls
```diff
@@ -97,10 +97,14 @@ f(2)
 
 :expect NaN
 :ssjs
+:e
 :noSanityCheck
 data class Cls(x, y) with
   fun f(z, p) = x + y + z + p
 Cls(1, 2).f(3)
```
LPTK: To keep the original intent of the test file:
```suggestion
id(Cls(1, 2)).f(3)
```
(same for the cases below)

hkmc2/shared/src/test/mlscript/backlog/Lifter.mls
```diff
@@ -40,9 +40,13 @@ f(2)
 data class A(x) with
   data class B(y) with
     fun getB() = x + y
-  fun getA() = this.B(2).getB()
+  fun getA() = 
+    // NOTE: The Term.Resolved PR made it so `this.B` gets resolved correctly.
+    // Explicitly binding `t = this` reproduces the previous error
```
LPTK, line 45: 👍

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/TermSynthesizer.scala
LPTK, line 1: @chengluyu should be able to help with resolving conflicts related to this file.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
LPTK, line 1: @chengluyu should be able to help with resolving conflicts related to this file.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -946,6 +996,11 @@ object Apps:
     case t => S(t, Nil)
 
 
+object ResolvedOpt:
```
LPTK: It seems this definition is unused.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -189,6 +203,9 @@ enum Term extends Statement:
   case UnitVal()
   case Missing // Placeholder terms that were not elaborated due to the "lightweight" elaboration mode `Mode.Light`
   case Lit(lit: Literal)
+  /** A term that wraps another term, indicating that the symbol of the inner term is resolved. */
```
```suggestion
  /** A term that wraps another term, indicating that the symbol of the inner term is resolved.
    * This is mainly used to disambiguate overloaded definitions. */
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -293,11 +305,33 @@ sealed trait ClassLikeSymbol extends IdentifiedSymbol:
   def subst(using sub: SymbolSubst): ClassLikeSymbol
 
 
+/**
+ * A symbol that the entity it represents has a definition.
```
```suggestion
 * A symbol for entities with a definition.
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -100,24 +103,27 @@ abstract class Symbol(using State) extends Located:
...
+  def asPrincipal: Opt[TermSymbol | ClassSymbol | ModuleOrObjectSymbol | TypeAliasSymbol | PatternSymbol] =
+    val asTrm: Opt[TermSymbol | ClassSymbol | ModuleOrObjectSymbol | TypeAliasSymbol | PatternSymbol] = this.asTrm
+    asTrm orElse
     asCls orElse
     asObj orElse
     asAls orElse
     asPat orElse
     asMod
-
+
```
LPTK, line 117: Best part of the PR 👍

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -197,6 +203,12 @@ class BuiltinSymbol
 class BlockMemberSymbol(val nme: Str, val trees: Ls[TypeOrTermDef], val nameIsMeaningful: Bool = true)(using State)
     extends MemberSymbol[Definition]:
   
+  // * This is a hack for that `TermDef` currently doesn't have a symbol. 
+  // * So, the symbol is from the `TermDefinition`. 
+  // * To prevent the `TermDefinition` from being overridden by other definitions,
```
LPTK: Didn't we conclude that `BlockMemberSymbol` should not even get its `defn` set? So there should be no overriding potential. We should try to make `BlockMemberSymbol` *not* have a `defn`, if possible, to avoid confusion.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -245,27 +245,59 @@ object Elaborator:
...
       val ps = PlainParamList(
         Param(flag, VarSymbol(Ident("output")), N, Modulefulness(N)(false)) ::
         Param(flag, VarSymbol(Ident("bindings")), N, Modulefulness(N)(false)) ::
         Nil)
       cs.defn = S(ClassDef.Parameterized(N, syntax.Cls, cs, BlockMemberSymbol(cs.name, Nil),
         Nil, ps, Nil, N, ObjBody(Blk(Nil, Term.Lit(UnitLit(false)))), N, Nil))
-      cs
-    val matchFailureClsSymbol =
+      ts.defn = S:
+        TermDefinition(
```
LPTK: Why not store `ts` here (and in other places `TermDefinition`s are created), instead of using the hacks with `tdefn`?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1108,6 +1140,8 @@ extends Importer:
                         Fun, mtdSym, tsym, PlainParamList(Param(FldFlags.empty, valueSym, N, Modulefulness.none) :: Nil) :: Nil,
                         N, N, S(valueSym.ref(Ident("value"))), FlowSymbol(s"‹result of non-local return›"), TermDefFlags.empty, Modulefulness.none, Nil, N)
                       tsym.defn = S(td)
+                      mtdSym.defn = S(td)
```
LPTK: Why define `mtdSym.defn`? Who uses it? Same for the other assignments below.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -1103,17 +1129,17 @@ trait LoweringSelSanityChecks(using Config, TL, Raise, State)
     extends Lowering:
   
   private val instrument: Bool = config.sanityChecks.isDefined
-
```
LPTK, line 976: Worst part of the PR 👎

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -275,25 +279,13 @@ class Resolver(tl: TraceLogger)
           ictx
       
       traverseStmts(rest)(using newICtx)
-
```
LPTK, line 278: Don't delete an extra line?

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -789,25 +840,79 @@ class Resolver(tl: TraceLogger)
...
+                disambSym(prefer, sign)(bms) match
+                  case S(ds) =>
+                    t.expand(S(t.withSym(bms).resolved(ds)))
+                  case N =>
+                    log(s"Unable to disambiguate ${bms}")
+                    t.expand(S(t.withSym(bms)))
+                log(s"Resolved symbol for ${t}: ${bms}")
+              case N =>
+                // TODO @Harry: Appropriately resolve all selections on classes.
+                // The type annotations in the current code bases are largely incomplete,
```
LPTK: Not sure I understand which type annotations this is referring to.

LPTK (commented):
Nice work, thanks!

But I have some questions on the use of `defn` for `BlockMemberSymbol`.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
CAG2Mark: It probably should be there but BlockMemberSymbol is not a DefinitionSymbol

LPTK: @FlandiaYingman shouldn't we try to get the correct `TermSymbol` here?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
FlandiaYingman: Did you mean:

```scala
      // ...
      lazy val resolvedBms = t.resolvedSym.flatMap(_.asBlkMember)
      // * We have to instantiate `f` again because, if `f` is a Sel, the `term`
      // * function is not called again with f. See below `Sel` and `SelProj` cases.
      f.instantiated match
      case t if resolvedBms.exists(_ is ctx.builtins.js.bitand) =>
        conclude(Value.Ref(State.runtimeSymbol).selN(Tree.Ident("bitand")))
```

?

FlandiaYingman: ok done

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
FlandiaYingman: Oops.

FlandiaYingman: Deobfuscated!

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
FlandiaYingman: I forgot the exact intent that I wrote this, but I think it was because the definition of `FunDefn` that its `innerSym` is specifically set as a constant `N`. 

```scala
final case class FunDefn(
    owner: Opt[InnerSymbol],
    sym: BlockMemberSymbol,
    params: Ls[ParamList],
    body: Block,
) extends Defn:
  val innerSym = N
```

I think it'd be of course better if we can insert the correct symbol here, but `FunDefn` is used in too many places, so that I think I should first focus on getting the main goal of this PR ready first.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
FlandiaYingman: OK (I think it's introduced by Mark's patch 😈)

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
FlandiaYingman: cc @CAG2Mark

hkmc2/shared/src/test/mlscript/codegen/SanityChecks.mls
FlandiaYingman: Applied; and I just realized that I introduced the same problem for this test file in one of my previous PRs; fixed them as well.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
FlandiaYingman: Yes, but `defn` of `BlockMemberSymbol` is already used for a long time everywhere in the compiler. Changing it in this PR would be difficult because I was also changing other things as well.

Probably a better way than not setting `BlockMemberSymbol.defn` is to crash the compiler whenever `BlockMemberSymbol.defn` is read or written, by

```
class BlockMemberSymbol(...) extends MemberSymbol[...]:

  override def defn = die
  override def defn_=(d: ...) = die
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: Not sure if I understand. Did you mean that we could directly store the `TermSymbol` instead of a `TermDefinition`, so that we won't have to create a dummy `TermDefinition` for situations like this?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: Oh yes, `mtdSym` is a `BlockMemberSymbol` that I shouldn't set its `defn`. Just removed and nothing happens, so I guess it's not used anywhere.

> Same for the other assignments below.

Other assignments below of `BlockMemberSymbol.defn` were not introduced by this PR. IDK how they are used, but they are indeed used somewhere that I couldn't remove them (otherwise the lowering creates an error).

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: The extra line there was a mistake. There shouldn't be two empty lines there because `traverseStmt` and `traverse` are closely related functions 🫣

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
FlandiaYingman: Oops, I just checked the error again. It's somehow incomplete but not _largely_ incomplete.

Sometimes in MLscript programs we call JS's (Node's) built-in functions directly, e.g.,

```
Class 'Str' does not contain member 'length'.
Class 'RangeIterator' does not contain member 'toArray'.
```

which are not appropriately defined in MLscript.

Other errors involving selection on classes include inherited members etc. Let's fix this in the next PR.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -336,11 +264,11 @@ class Normalization(lowering: Lowering)(using tl: TL)(using Raise, Ctx, State) e
               mkMatch(mkArgs(clsParams.iterator.zip(args).toList))
             // Select the constructor's `.class` field.
             lazy val ctorTerm = ctor.symbol match
-              case S(mem: BlockMemberSymbol) =>
-                // If the class is declaration-only, we do not need to
-                // select the class.
-                if !mem.hasLiftedClass || mem.defn.exists(_.hasDeclareModifier.isDefined) then ctor
-                else Term.SynthSel(ctor, Tree.Ident("class"))(mem.clsTree.orElse(mem.modOrObjTree).map(_.symbol), N).resolve
+              // case S(mem: BlockMemberSymbol) =>
```
LPTK: Can we remove this code?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: Oh, I thought we wanted this, given that there is a test for it in the backlog. The current semantics allow for dynamic selection of nested definitions.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -344,11 +345,14 @@ class Normalization(lowering: Lowering)(using tl: TL)(using Raise, Ctx, State) e
     lazy val rootBreakLabel = new TempSymbol(N, "split_root$")
     lazy val breakRoot = (r: Result) => Assign(l, r, Break(rootBreakLabel))
     lazy val assignResult = (r: Result) => Assign(l, r, End())
+    val loopCont = if config.rewriteWhileLoops
+      then Return(Call(Value.Ref(f, N), Nil)(true, true), false)
+      else Continue(loopLabel)
```
FlandiaYingman, line 342: Here I saw the symbol `f` is a (dummy) `BlockMemberSymbol` denoting a `while` loop label (`new BlockMemberSymbol("while", Nil, false)`). 

Is it necessary for it to be a `BlockMemberSymbol`? It seems like there is no meaningful disambiguation for such symbol `f` like here, so my opinion is to change it to another symbol type.

LPTK: This symbol is not for the label, but for the inserted function, and `FunDefn` requires a `BlockMemberSymbol`. Are you suggesting changing the definition of `FunDefn`?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
FlandiaYingman: IIRC this was the changes from the main branch. Probably you were looking at the wrong commit?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1600,7 +1597,9 @@ extends Importer with ucs.SplitElaborator:
         case p @ Pun(false, _) :: Nil => record(p)
         case p @ InfixApp(_: Ident, Keywrd(Keyword.`:`), _) :: Nil => record(p)
         case lhs :: Nil => arrow(lhs, rhs)
-        case _ :: _ | Nil => ??? // TODO: When is this case reached?
+        // TODO: When is this case reached?
+        // * Answer: pattern p = () => Unit
+        case _ :: _ | Nil => ???
```
```suggestion
        case _ :: _ | Nil => ??? // TODO: this case reached by, eg, `pattern p = () => Unit`
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -201,6 +211,9 @@ class BuiltinSymbol
 class BlockMemberSymbol(val nme: Str, val trees: Ls[TypeOrTermDef], val nameIsMeaningful: Bool = true)(using State)
     extends MemberSymbol[Definition]:
   
+  def defn: Option[Definition] = ???
+  def defn_=(d: Option[Definition]): Unit = ???
```
LPTK: TODO: let's not forget to remove these

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -250,6 +266,9 @@ class TermSymbol(val k: TermDefKind, val owner: Opt[InnerSymbol], val id: Tree.I
     with NamedSymbol:
   def nme: Str = id.name
   def name: Str = nme
+  
+  var defn: Opt[TermDefinition] = N
```
LPTK: Can't we move these to `DefinitionSymbol`?

hkmc2/shared/src/test/mlscript-compile/Predef.mjs
```diff
@@ -114,10 +114,10 @@ globalThis.Object.freeze(class Predef {
   static notImplemented(msg) {
     let tmp;
     tmp = "Not implemented: " + msg;
-    throw globalThis.Error(tmp)
+    throw runtime.safeCall(globalThis.Error(tmp))
```
LPTK: This looks like a regression.

FlandiaYingman: Yup, so that's why the commit is _preliminary_!

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
LPTK: Amazing! Change approved.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
LPTK: Note that for `Str`, we just need to add a `length` method in `declare class Str`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
FlandiaYingman: Moved!

hkmc2/shared/src/test/mlscript-compile/Predef.mjs
FlandiaYingman: *Re*moved!

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -298,24 +290,22 @@ type BaseTypeSymbol = ClassSymbol | ModuleOrObjectSymbol
...
-type FieldSymbol = MemberSymbol[?]
-
 /**
   * ErrorSymbol is a placeholder symbol denoting error (during symbol
   * resolution in the elaborator / resolver). This helps prevent the
   * same error from throwing multiple times.
   */
-case class ErrorSymbol(val nme: Str, tree: Tree)(using State) extends MemberSymbol[Nothing]:
+case class ErrorSymbol(val nme: Str, tree: Tree)(using State) extends MemberSymbol:
   var defn: Opt[Nothing] = N
```
LPTK: I guess we can remove this?

FlandiaYingman: Yes!

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
FlandiaYingman: Ideally we should, but I guess it'd be another PR

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: Many tests are just there to guard us against regressions and unexpected changes in behaviors.

Which specific test are you referring to?

Also, can you give a concrete example of what this comment is meant to describe?

LPTK (approved):
Ok, let's merge this now and address the rest in follow-up PRs. Nice progress towards cleaning up the logic!

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: OK, the test I was referring to was https://github.com/hkust-taco/mlscript/blob/hkmc2/hkmc2/shared/src/test/mlscript/backlog/Lifter.mls (line 24). But I seem to have remembered incorrectly as that test was for classes and not objects, and it was for `this.B` selections specifically, not arbitrary `someInstanceOfA.B`. It also seems this PR now correctly adds the selection symbol to `this.B` selections -- these symbols were missing before this PR.

I don't remember exactly why I made that change, but it was probably a workaround for when the lifter was not adapted to the new symbol system correctly and/or when selection symbols were not propagated (which they now are). For example, the selection symbol for selecting `this.M` in
```fs
data class A(x) with
  object M with
    fun getB() = x
  fun getA() = M.getB()
```
was not transformed correctly, and the generated code would have `this.M.getB()` instead of `this.M$.getB()`, the latter of which the lifter would assign the lifted instance of `M` to. But this is now fixed by reading the selection symbol correctly.

The highlighted change in the lifter can be safely reverted without breaking any existing tests and the comment can be considered void.

For your information, this code actually does work without the lifter:
```fs
class A with
  object M
(new A).M
```

LPTK: Great! Please fix it and open a quick PR when you can.

```fs
class A with
  object M
(new A).M
```

I know it does. Does the lifter currently break it?

CAG2Mark: It used to, but what I did in the lifter patch present in tihs PR un-breaks it.

PR #345, opened 2025-10-04
Attach annotation to symbols

ychenfo:


hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -569,7 +569,8 @@ sealed trait Statement extends AutoLocated, ProductWithExtraInfo:
       s"type ${sym}${tparams.mkStringOr(", ", "[", "]")} = ${rhs.fold("")(x => x.showDbg)}"
     case Missing => "missing"
 
-final case class LetDecl(sym: LocalSymbol, annotations: Ls[Annot]) extends Statement
+final case class LetDecl(sym: LocalSymbol, annotations: Ls[Annot]) extends Statement:
+  sym.annotations :::= annotations
```
LPTK: That's a bad idea. It means every time we reconstruct the term (it could be reconstructed for various reasons), we're going to re-add the annotations to the symbol again. Instead, adding the annotations should be done when the first `LetDecl` is created. Also, we probably don't need the `annotations` field on `LetDecl` anymore, right?

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -754,7 +759,8 @@ case class PatternDef(
   val paramsOpt: Opt[ParamList] = N
   val auxParams: Ls[ParamList] = Nil
   val companion: Opt[CompanionSymbol] = N // TODO support
-
```
LPTK, line 758: 👎

hkmc2/shared/src/test/mlscript/syntax/annotations/Declarations.mls
```diff
@@ -5,230 +5,86 @@ object tailrec
 tailrec
 //│ = tailrec
 
-:w
+
```
LPTK: If you remove a command, don't add an empty line instead. Only remove the command's line.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -642,6 +643,8 @@ final case class TermDefinition(
   def extraAnnotations: Ls[Annot] = annotations.filter:
     case Annot.Modifier(Keyword.`declare` | Keyword.`abstract`) => false
     case _ => true
+  sym.annotations :::= annotations
```
LPTK: This `BlockMemberSymbol` is shared among all overloaded version of the denoted entity. But only the relevant overload (here, the term) should get the annotation.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -642,6 +643,8 @@ final case class TermDefinition(
...
+  tsym.annotations :::= annotations
```
LPTK: Same remark as above. We don't want simple case class constructors to have such surprising side effects.

LPTK: And again we can probably remove the `annotations` field.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
ychenfo: Thanks for the comment, I see. I have moved the mutation to Elaborator, and removed the `annotations` field on `LetDecl` and `TypeLikeDef`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
ychenfo: This should be fixed in the latest commits.

hkmc2/shared/src/test/mlscript/syntax/annotations/Declarations.mls
ychenfo: Indeed, I added back the warnings reported during lowering because I feel it makes more sense this way: the annotations are still not taken into consideration during lowering; the elaborator now does not ignore them because it responsibly passes the annotation information to the lowering stage.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
ychenfo: I see, now `BlockMemberSymbol`s will not get annotations in `Elaborator.scala`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
ychenfo: The `annotations` field in `TypeLikeDef` is also removed.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1025,7 +1026,8 @@ extends Importer:
           case N =>
             if tups.nonEmpty then
               raise(ErrorReport(msg"Expected a right-hand side for let bindings with parameters" -> hd.toLoc :: Nil))
-            LetDecl(sym, annotations).mkLocWith(kw) :: acc
+            sym.annotations = annotations
```
LPTK, line 1029: Why do this only here instead of where the symbol is created, just above?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1113,10 +1115,11 @@ extends Importer:
               
               val tsym = TermSymbol(k, owner, id) // TODO?
               val tdf = TermDefinition(k, sym, tsym, pss, tps, s, body, r, 
-                TermDefFlags.empty.copy(isMethod = isMethod), mfn, annotations, N)
+                TermDefFlags.empty.copy(isMethod = isMethod), mfn, N)
               tsym.defn = S(tdf)
               sym.defn = S(tdf)
-              
+              tsym.annotations = annotations
+              sym.annotations = annotations
```
LPTK, line 1122: You should document that this is only a temporary measure while the situation with term symbols is messed up.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -352,7 +352,9 @@ sealed trait Statement extends AutoLocated, ProductWithExtraInfo:
     case t: Term => lastWords(s"overridden implementation")
     case d: Definition => ???
     case imp: Import => Import(imp.sym, imp.str, imp.file)
-    case LetDecl(sym, annotations) => LetDecl(sym, annotations.map(_.mkClone))
+    case LetDecl(sym) =>
+      sym.annotations = sym.annotations.map(_.mkClone)
```
LPTK, line 356: This is obviously nonsense.

LPTK:
As discussed in the meeting, we prefer to not actually do it this way.

PR #346, opened 2025-10-06
Object buffer

AnsonYeung:


hkmc2/shared/src/test/mlscript/objbuf/Mutation.mls
```diff
@@ -0,0 +1,101 @@
...
+
+:fixme
+:sjs
+@bufferable
+class A(x) with
+  mut val z = x
+  fun f(y) =
+    set z += 1
+    z
+//│ /!!!\ Uncaught error: hkmc2.InternalError: `this` not in scope: class:A
```
AnsonYeung: Looks like this uses `AssignField` which doesn't contain any symbol for replacement...

So it's probably needed to either detect such usage of the form that assign to the field of a class symbol and keep track of all the fields with the name to match on the field name which doesn't feel right to be done in lowering

hkmc2/shared/src/test/mlscript-compile/ObjectBuffer.mls
```diff
@@ -18,11 +18,13 @@ module ObjectBuffer with
       buf.[1] = -1
 
     fun mkNew(cls) =
+      if cls.size == 0 do return 0
```
LPTK: Please use `===` and `!==` here and below.

hkmc2/shared/src/test/mlscript/objbuf/Mutation.mls
LPTK: Sorry I can't understand what you wrote.

> Looks like this uses `AssignField` which doesn't contain any symbol for replacement...

What do you mean?

> So it's probably needed to either [...]

Your "either" is not paired wiht an "or".

AnsonYeung: Nevermind there is a symbol in the auxillary parameter list

AnsonYeung:
Ready for review now

hkmc2/shared/src/test/mlscript/objbuf/Basics.mls
```diff
@@ -0,0 +1,223 @@
...
+
+// UAF
+A.f(buf, obj1)(1)
+//│ = 11
+
+A.f(buf, obj3)(1)
+//│ = 11
+
+
+:sjs
```
LPTK: Does this show anything interesting that's not shown above? Otherwise, let's remove it.

hkmc2/shared/src/test/mlscript/objbuf/Basics.mls
```diff
@@ -0,0 +1,223 @@
...
+//│ = 4
+
+buf.del(Oops, obj1)
+
+:expect 4
+A.f(buf, obj2)(1)
+//│ = 4
+
+
+:sjs
```
LPTK: Does this show anything interesting that's not shown above? Otherwise, let's remove it.

hkmc2/shared/src/test/mlscript/objbuf/Mutation.mls
```diff
@@ -0,0 +1,150 @@
+:js
+
+import "../../mlscript-compile/ObjectBuffer.mls"
+open ObjectBuffer
+open annotations
+
+
+:sjs
```
LPTK, line 8: This `:sjs` seems strictly more interesting/useful than the ones in `Basics.mls`, so we don't need the ones in `Basics.mls`, right?

hkmc2/shared/src/test/mlscript/objbuf/Mutation.mls
```diff
@@ -0,0 +1,150 @@
...
+
+// UAF
+A.class.f(buf, obj1)(1)
+//│ = 11
+
+A.class.f(buf, obj3)(1)
+//│ = 12
+
+
+:sjs
```
LPTK: Again, there's not need to add so much generated code to git. We already have a case like this above. Maybe consolidate both things in a single example with `:sjs`?

hkmc2/shared/src/main/scala/hkmc2/codegen/StackSafeTransform.scala
```diff
@@ -116,7 +116,8 @@ class StackSafeTransform(depthLimit: Int, paths: HandlerPaths, doUnwindMap: Map[
         rewriteBlk(preCtor, L(BlockMemberSymbol("TODO", Nil)), 1), // TODO: preCtor is not translated in handler lowering
         if isTopLevel && (defn.k is syntax.Mod) then transformTopLevel(ctor) else rewriteBlk(ctor, R(isym), 1),
         mod.map(rewriteObjBody(_, isTopLevel)),
-      )
+        bufferable,
+    )
```
```suggestion
      )
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -298,6 +298,7 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
       sortedVars.iterator.foldLeft[Block](End()):
         case (acc, (_, _, vd)) => Define(vd, acc),
       N,
+      N, // TODO: capture class is likely bufferable
```
LPTK: Even if it is, how would that help? We can't allocate these on buffers, as there is no way to automaticaly manage their lifetime. Capture class instances can leave the current scope and live indefinitely.

hkmc2/shared/src/main/scala/hkmc2/codegen/BufferableTransform.scala
```diff
@@ -0,0 +1,98 @@
...
+import semantics.Elaborator.State
+import hkmc2.Message.MessageContext
+import hkmc2.syntax.Tree.DummyTypeDef
+
+class BufferableTransform()(using Ctx, State, Raise):
+  def transform(blk: Block): Block =
+    val transformer = new BlockTransformer(SymbolSubst()):
+      override def applyDefn(defn: Defn)(k: Defn => Block): Block = defn match
+        case cls: ClsLikeDefn =>
+          cls.bufferable.flatMap(_.optionIf(cls.k is syntax.Cls)).fold(super.applyDefn(defn)(k)): bufferable =>
```
LPTK:
```suggestion
          cls.bufferable.filter(_.k is syntax.Cls).fold(super.applyDefn(defn)(k)): bufferable =>
```

LPTK (approved):
Nice change, thanks, but please address the comments.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
AnsonYeung: Right, I forgot to consider that.

LPTK: Then we can probably delete the comment.

AnsonYeung:
I've addressed all the comments

LPTK (approved):

PR #351, opened 2025-11-10
Proper handling of local variables capturing inside a loop body

AnsonYeung:
Previously, variables inside a loop body are treated the same as that of the function body, which leads to bad semantics when a function captures a local variable defined within a loop body.

As mentioned in https://github.com/hkust-taco/mlscript/pull/293#issuecomment-3496516112, this PR lowers while loop into tail recursive function and rely on the scoping of the function for the correct semantics. This is done during lowering as we already know the location of inserting Continue statements and we just need to convert these into tail calls.

`runtime.LoopEnd` is introduced to handle the case when a return statement occur inside the loop body. This value is returned when the loop exits without early return and is tested to determine whether to return from the outer function.


LPTK:
Nice! Please first resolve the conflicts and then request my review.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -572,9 +572,9 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
     
     
     case iftrm: st.IfLike => ucs.Normalization(this)(iftrm)(k)
-
```
LPTK, line 575: 👎

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -1073,7 +1073,7 @@ trait LoweringTraceLog(instrument: Bool)(using TL, Raise, State)
       TempSymbol(N) -> pureCall(traceLogFn, Arg(N, Value.Ref(retMsgSym)) :: Nil)
     ) |>:
       Ret(Value.Ref(resSym))
-    )
+    )(using LoweringCtx.func)
```
LPTK: Shouldn't you copy the current context instead of creating a fresh new one?

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -490,8 +490,6 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
     case Label(lbl, loop, bod, rst) =>
       scope.allocateName(lbl)
       
-      // [fixme:0] TODO check scope and allocate local variables here (see: https://github.com/hkust-taco/mlscript/pull/293#issuecomment-2792229849)
```
LPTK, line 493: This is not actually fixed, as `while` loops aren't always rewritten yet...

hkmc2/shared/src/test/mlscript/lifter/Loops.mls
```diff
@@ -27,13 +27,11 @@ fun foo() =
     set i += 1
     fs.push of () => x
 
-// * Note that this also fails without lifting, as we need to fix [fixme:0]
```
LPTK, line 30: Pls add a `:dontRewriteWhile` option to reproduce the problem that still exists, instead of trying to sweep it under the rug ;^)

LPTK (changes requested):
Good work, but let's not claim victory too fast.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
AnsonYeung: In the latest commit I've added this back, but it shouldn't be an issue after loop rewriting. If a loop does not contain any local function, loop local variables can be treated as function locals since there is no capturing.

AnsonYeung (commented):
I've addressed the comments.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
LPTK: Yes, that's right, but it's better to keep the doc/tests in sync with what's actually implemented at the moment.

hkmc2/shared/src/test/mlscript/lifter/Loops.mls
```diff
@@ -43,29 +45,47 @@ fun foo() =
...
+//│ while$1 = function while$(foo$capture6) {
+//│   let scrut, tmp2, lambda$here;
+//│   scrut = true;
+//│   if (scrut === true) {
+//│     tmp2 = foo$capture6.x$capture$0 + 1;
+//│     foo$capture6.x$capture$0 = tmp2;
+//│     lambda$here = runtime.safeCall(lambda2(foo$capture6));
+//│     return lambda$here
+//│   } else {
+//│     foo$capture6.tmp$capture$1 = runtime.Unit;
```
LPTK, line 66: I just noticed this crazy capture of a temporary variable that just stores the unit value. It's pretty crazy. I wonder why Unit is bound to a temporary variable in the first place.

LPTK (approved):
LGTM

PR #352, opened 2025-11-17
Annotation improvements

CAG2Mark:
In this PR:
- Improve annotation system (supporting multiple annotation on terms, improved reporting system)
- There were some places where annotations were not reported, which I've fixed
- Adds the `@tailrec` and `@tailcall` annotations (but these are not yet implemented)

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
@@ -219,7 +219,7 @@ sealed abstract class Block extends Product:
           val newBody = d.body.flattened
           if newBody is d.body
           then d
-          else d.copy(body = newBody)
+          else d.copy(body = newBody)(d.isTailRec)
         case v: ValDefn => v
         case c: ClsLikeDefn =>
           val newPreCtor = c.preCtor.flattened
@@ -227,7 +227,7 @@ sealed abstract class Block extends Product:
           val newMethods = c.methods.mapConserve:
             case f@FunDefn(owner, sym, params, body) =>
               val newBody = body.flattened
-              if newBody is body then f else f.copy(body = newBody)
+              if newBody is body then f else f.copy(body = newBody)(f.isTailRec)
           if (newPreCtor is c.preCtor) && (newCtor is c.ctor) && (newMethods is c.methods)
           then c
           else c.copy(preCtor = newPreCtor, ctor = newCtor, methods = newMethods)
@@ -339,6 +339,8 @@ final case class FunDefn(
     sym: BlockMemberSymbol,
     params: Ls[ParamList],
     body: Block,
+  )(
+    val isTailRec: Bool,
 ) extends Defn:
   val innerSym = N
 
@@ -552,7 +554,7 @@ type Local = Symbol
  * regardless of whether the check for effect is inserted or not.
  * Note that the check for effect is inserted during HandlerLowering and setting this to true
  * after handler is lowered does not have any effect on the code generation. */
-case class Call(fun: Path, args: Ls[Arg])(val isMlsFun: Bool, val mayRaiseEffects: Bool) extends Result
+case class Call(fun: Path, args: Ls[Arg])(val isMlsFun: Bool, val mayRaiseEffects: Bool, val isTailCall: Bool) extends Result
 
 case class Instantiate(mut: Bool, cls: Path, args: Ls[Arg]) extends Result
 
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
@@ -227,7 +227,7 @@ sealed abstract class Block extends Product:
           val newMethods = c.methods.mapConserve:
             case f@FunDefn(owner, sym, params, body) =>
               val newBody = body.flattened
-              if newBody is body then f else f.copy(body = newBody)(f.isTailRec)
+              if newBody is body then f else f.copy(body = newBody)(isTailRec = f.isTailRec)
           if (newPreCtor is c.preCtor) && (newCtor is c.ctor) && (newMethods is c.methods)
           then c
           else c.copy(preCtor = newPreCtor, ctor = newCtor, methods = newMethods)
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
@@ -334,7 +334,7 @@ sealed abstract class Defn:
         -- auxParams.flatMap(_.paramSyms)
   
 
-// NOTE: Setting isTailRec to false does not affect whether function is optimized.
+// NOTE: Setting isTailRec to false does not affect whether the function is optimized.
 // It only affects whether a warning is thrown if the function is not actually tailrec.
 final case class FunDefn(
     owner: Opt[InnerSymbol],
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
@@ -219,7 +219,7 @@ sealed abstract class Block extends Product:
           val newBody = d.body.flattened
           if newBody is d.body
           then d
-          else d.copy(body = newBody)(d.isTailRec)
+          else d.copy(body = newBody)(isTailRec = d.isTailRec)
         case v: ValDefn => v
         case c: ClsLikeDefn =>
           val newPreCtor = c.preCtor.flattened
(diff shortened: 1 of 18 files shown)


hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -219,15 +219,15 @@ sealed abstract class Block extends Product:
           val newBody = d.body.flattened
           if newBody is d.body
           then d
-          else d.copy(body = newBody)
+          else d.copy(body = newBody)(d.isTailRec)
```
LPTK: To avoid Boolean confusion in case someone adds another parameter later:
```suggestion
          else d.copy(body = newBody)(isTailRec = d.isTailRec)
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -552,7 +554,7 @@ type Local = Symbol
  * regardless of whether the check for effect is inserted or not.
  * Note that the check for effect is inserted during HandlerLowering and setting this to true
  * after handler is lowered does not have any effect on the code generation. */
-case class Call(fun: Path, args: Ls[Arg])(val isMlsFun: Bool, val mayRaiseEffects: Bool) extends Result
+case class Call(fun: Path, args: Ls[Arg])(val isMlsFun: Bool, val mayRaiseEffects: Bool, val isTailCall: Bool) extends Result
```
LPTK: I would name this
```suggestion
case class Call(fun: Path, args: Ls[Arg])(val isMlsFun: Bool, val mayRaiseEffects: Bool, val explicitTailCall: Bool) extends Result
```
as the call might not be in actual tail position (then, we'll either emit an error or transform the code to make it tail).

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -396,16 +396,18 @@ class Normalization(lowering: Lowering)(using tl: TL)(using Raise, Ctx, State) e
           val isReturned = TempSymbol(N)
           val loopEnd: Path =
             Select(Value.Ref(State.runtimeSymbol), Tree.Ident("LoopEnd"))(S(State.loopEndSymbol))
+          // NOTE: Setting isTailRec to false does not affect whether function is optimized.
+          // It only affects whether a warning is thrown if the function is not actually tailrec.
```
LPTK: This comment probably belongs in the code documentation of `Call`, no?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -994,12 +1001,42 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
-      case annot => raise:
-        WarningReport(msg"This annotation has no effect." -> annot.toLoc :: Nil)
+      case annot => warn(annot)
 
+  def reportAnnotations(receiver: Term, annotations: Ls[Annot]): Unit =
+    def warn(annot: Annot, msg: Opt[Message] = N) =
+      val message = msg match
+        case None => msg"Such annotations are not supported on ${receiver.describe} terms."
+        case Some(value) => value
+
```
```suggestion
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -994,12 +1001,42 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
+    def warn(annot: Annot, msg: Opt[Message] = N) =
+      val message = msg match
+        case None => msg"Such annotations are not supported on ${receiver.describe} terms."
+        case Some(value) => value
+      
+      raise:
+        WarningReport(
+          msg"This annotation has no effect." -> annot.toLoc ::
+          message -> receiver.toLoc :: Nil)
+
```
LPTK, line 1026:
```suggestion
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -994,12 +1001,42 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
+        
       case Annot.Modifier(syntax.Keyword("staged")) => ()
-      case annot => raise:
-        WarningReport(msg"This annotation has no effect." -> annot.toLoc :: Nil)
+      case annot => warn(annot)
 
+  def reportAnnotations(receiver: Term, annotations: Ls[Annot]): Unit =
+    def warn(annot: Annot, msg: Opt[Message] = N) =
+      val message = msg match
+        case None => msg"Such annotations are not supported on ${receiver.describe} terms."
```
```suggestion
        case None => msg"This kind of annotations is not supported on ${receiver.describe} terms."
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -96,13 +98,14 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
       isMlsFun = true,
+      isTailCall = false,
       args.headOption,
       N, // TODO: location?
     )(c => Return(c, implct = true))
   
   // * Used to work around Scala's @tailrec annotation for those few calls that are not in tail position.
-  final def term_nonTail(t: st, inStmtPos: Bool = false)(k: Result => Block)(using LoweringCtx): Block =
-    term(t: st, inStmtPos: Bool)(k)
+  final def term_nonTail(t: st, annots: List[Annot] = Nil, inStmtPos: Bool = false)(k: Result => Block)(using LoweringCtx): Block =
```
LPTK: 1. Default arguments in core logic must be avoided. There are very few exceptions to this (`inStmtPos` is one). Doing this with `annots` would almost certainly lead to bugs down the line.

2. Instead of adding an `annots` parameter to this very general function, why not locally handle all annotations found on a term in the same place where you define `val insted = t.instantiated`?

LPTK (commented):
Thanks for the work!

It would be better to merge the changes related to annotations first, as a small PR, and open new PRs for the other changes.

> - [ ] Tail recursion optimization on tail recursive functions
> - [ ] Tail recursion optimization for mutually tail-recursive functions

Does it make sense to terat these two cases separately?

CAG2Mark:
> Thanks for the work!
> 
> It would be better to merge the changes related to annotations first, as a small PR, and open new PRs for the other changes.
> 
> > * [ ]  Tail recursion optimization on tail recursive functions
> > * [ ]  Tail recursion optimization for mutually tail-recursive functions
> 
> Does it make sense to terat these two cases separately?

True

LPTK:
> True

Just rename the PR and open the follow-up ones later ;)

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
CAG2Mark: I've rewritten it as "This annotation is not supported on ..." which I think is clearer

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -414,7 +414,13 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
         WarningReport(msg"Pure expression in statement position" -> t.toLoc :: Nil, S(t))
     
-    val insted = t.instantiated
+    @tailrec
+    def getAnnots(t: st, acc: List[Annot]): (List[Annot], st) = t match
+      case st.Annotated(annot, trm) => getAnnots(trm, annot :: acc)
+      case _ => (acc, t)
+    val (annots, trm) = getAnnots(t, Nil)
+    
+    val insted = trm.instantiated
```
LPTK: It should match on `insted`, rather than `trm`. Basically, one should never pattern match on terms before "isntantiating" them.
```suggestion
    def getAnnots(t: st, acc: List[Annot]): (List[Annot], st) = t match
      case st.Annotated(annot, trm) => getAnnots(trm.instantiated, annot :: acc)
      case _ => (acc, t)
    val insted = trm.instantiated
    val (annots, trm) = getAnnots(insted, Nil)
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -334,11 +334,15 @@ sealed abstract class Defn:
         -- auxParams.flatMap(_.paramSyms)
   
 
+// NOTE: Setting isTailRec to false does not affect whether function is optimized.
```
```suggestion
// NOTE: Setting isTailRec to false does not affect whether the function is optimized.
```

LPTK (approved):
Any reason this is marked as Draft? It LGTM.

CAG2Mark:
> Any reason this is marked as Draft? It LGTM.

Actually I wanted to add another change that adds a TermSymbol parameter to FunDefn, but I guess I should make that another PR

CAG2Mark:
OK, all done

(and I may have pinged two GitHub users called "tailrec" and "tailcall" when editing my PR description, oops...)

CAG2Mark:
~~Hold on, I noticed one thing~~ nevermind, I misread

PR #354, opened 2025-11-21
Leading dot selection resolution

thornoar:
We use flow analysis to resolve selections of the form `.method`, where the target is unknown, based on the type into which this selection flows. Examples include

```scala
class A
module A with
  fun a = new A
fun f(x : A) = x
f(.a) // <- .a flows into type `A`, hence we find the module A and resolve to A.a
// = A
```

```scala
class B with
  fun b = new B
module B with
  fun c = (x) => 5
fun g(x: B) = x
g(.c(4).b.b.b) // Chained selections store a reference to the original leading dot selection
// = B
```

See the `flows/LeadingDotAccesses.mls` file for more examples.

This is a draft PR, I have the following remaining tasks in mind:
- Miscellaneous `TODO`s and `???`s that should be properly implemented
- Testing
- Reporting only one error if the first selection in a chain fails to be resolved


hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -696,7 +696,7 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
           if usesResTmp then k(Value.Ref(l))
           else k(unit) // * it seems this currently never happens
         )
-
```
LPTK, line 699: 👎

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -564,6 +564,8 @@ extends Importer:
       val preTrm = subterm(pre)
       val sym = resolveField(nme, preTrm.symbol, nme)
       Term.SynthSel(preTrm, nme)(sym, N)
+    case Sel(Empty(), nme) =>
+      Term.Sel(Term.LeadingDotTarget, nme)(N, N, S(ctx))
```
LPTK: If we're going to add a new term, it should be `Term.LeadingDotSel(nme)`, to reduce the amount of illegal states that are representable.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -203,6 +204,27 @@ class BuiltinSymbol
 
   def subst(using sub: SymbolSubst): BuiltinSymbol = sub.mapBuiltInSym(this)
 
+  def signature : semantics.flow.Producer =
```
LPTK: Why not make this a `lazy val`?

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -203,6 +204,27 @@ class BuiltinSymbol
...
+    val nullaryType : Type = Top
+    val typ =
+      Union(
+      Union(
+      if (binary) then binaryType else Bot,
+      if (unary) then unaryType else Bot,
+      ),
+      if (nullary) then nullaryType else Bot,
+      )
+      // (binary, unary, nullary) match
```
LPTK: I actually prefer this version.

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/Constraint.scala
```diff
@@ -23,10 +23,11 @@ case class Constraint(lhs: Producer, rhs: Consumer):
 enum Producer:
   case Flow(sym: FlowSymbol)
   case Fun(lhs: Consumer, rhs: Producer, captures: Ls[(Producer, Consumer)])
-  case Tup(elems: Ls[Opt[SpreadKind] -> Producer])
+  case Tup(elems: Ls[(Opt[SpreadKind], Producer)])
```
LPTK: Isn't the `->` version easier to read?

hkmc2/shared/src/test/mlscript/flows/SelExpansion.mls
```diff
@@ -258,6 +259,43 @@ foo.test
...
+//│     flow:test³  ->  ⋅test¹
+//│ foo = A
+
+:e
+:re
+foo.test
+//│ ╔══[ERROR] Unresolved selection:
+//│ ║  l.286: 	foo.test
+//│ ║         	^^^^^^^^
+//│ ╙── Type `((a‹804›) -> A)` does not contain member 'test'
```
LPTK: I wonder why we get a function type for `foo`, here. Seems like a bug.

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
```diff
@@ -367,8 +438,7 @@ class FlowAnalysis(using tl: TraceLogger)(using Raise, State, Ctx):
           dig(c.lhs, c.rhs, Vector.empty)
           
         if fuel === 0 then
-          raise(ErrorReport(
-            msg"Could not solve all constraints within $MAX_FUEL iterations." -> N :: Nil))
+          raise(ErrorReport(msg"Could not solve all constraints within $MAX_FUEL iterations." -> N :: Nil))
```
LPTK: We normally try to keep things under 100 columns.

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
```diff
@@ -204,38 +219,98 @@ class FlowAnalysis(using tl: TraceLogger)(using Raise, State, Ctx):
...
+        // * An error should alsoready be reported in this case
+      case targets => raise:
+        ErrorReport:
+          msg"Ambiguous selection with multiple apparent targets" -> sel.toLoc
+          :: targets.map:
+            case ObjectMember(sym) => msg"object member ${sym.nme}" -> sym.toLoc
+            case CompanionMember(_, sym) => msg"companion member ${sym.nme}" -> sym.toLoc
   
-  def solveConstraints() =
+  def findConsumerSymbols(cons: Consumer): Ls[Symbol] =
```
LPTK: Why are we trying to extract consumer symbols from things like tuples and function left/right-hand sides? 🤔

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
```diff
@@ -204,38 +219,98 @@ class FlowAnalysis(using tl: TraceLogger)(using Raise, State, Ctx):
...
+      case Nil =>
+        // FIXME: actually allow that in dead code (use floodfill constraints from exported members to detect)
+        if !sel.isErroneous then raise:
           ErrorReport:
-            msg"Ambiguous selection with multiple apparent targets" -> sel.toLoc
-            :: targets.map:
-              case ObjectMember(sym) => msg"object member ${sym.nme}" -> sym.toLoc
-              case CompanionMember(_, sym) => msg"companion member ${sym.nme}" -> sym.toLoc
+            msg"Cannot resolve selection" -> sel.toLoc :: Nil
+        // * An error should alsoready be reported in this case
```
LPTK: "alsoready" kind of makes sense, but is unfortunately not a word.

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
```diff
@@ -163,10 +177,11 @@ class FlowAnalysis(using tl: TraceLogger)(using Raise, State, Ctx):
...
         case f: Fld => N -> typeProd(f.term))
     
-    case Error =>
-      P.Ctor(Extr(false), Nil)(t)
+    case Error => P.Ctor(Extr(false), Nil)(t)
+
+    case Missing => P.Unknown(Missing)
     
-    // case _ => P.Flow(FlowSymbol("TODO"))
+    case _ => P.Flow(FlowSymbol("TODO"))
```
LPTK: Pls don't forget to revert this.

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
```diff
@@ -177,13 +192,13 @@ class FlowAnalysis(using tl: TraceLogger)(using Raise, State, Ctx):
       Type.Error
   
   def typeParam(p: Param): C =
+    val fs = FlowSymbol(p.sym.name)
```
LPTK: Not a fan of this. We should access flow symbols that are reachable from the parameter, so that these can be retrieved later.

LPTK (commented):
Nice! But I have a bunch of comments. Also, it seems the CI is failing.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
thornoar: what are you referring to? The pattern match or the `if` block? if it's the latter, do you mean that we should assign the signature only based on the `nullary` flag?

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/Constraint.scala
thornoar: Personally, `->` looks to me like a function type, so `(,)` is much more readable. But the call is entirely yours of course, I will follow the project style

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
thornoar: that's your comment, not mine :)

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
thornoar: This is the sort of generalization I was talking about, and how resolutions like `.b(new B)` work, i.e. by solving the constraint `_?_.b <: ((B) => <something>)` and traversing the consumer of this constraint to find the type `B` and its module, where `b` can be found. The `findConsumerSymbols` function grabs at any symbols found in the consumer and returns them to be inspected for companion modules. Of course, this raises the problem of conflicting methods, which is why you can toggle this off, and we may remove this feature entirely.

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
thornoar: Makes sense. What led me to this was the inability to define a flow symbol directly in the `Param` class, due to a lack of contextual `State` in that scope. Would it be better to add an `Option[FlowSymbol]` field in `Param`, and set this field to `S(FlowSymbol(p.sym.name))` in the `typeParam` function?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
thornoar: That makes sense, and I've implemented it, but there arises a problem with `Lowering` again. If I understand correctly, leading dot selections should not be lowered at all, but the `term` function in `Lowering.scala` still gets called, and when the term was `Sel(LeadingDotTarget, nme)`, this was fine since the `term` function handles the `Sel` case. However, on `LeadingDotSel`... what should it do..?

FlandiaYingman: You just elaborate (expand) it so to make sure no `LeadingDotSel` appears after the flow analysis stage!

FlandiaYingman: and probably you would want to add a case in the term function, so that if a `LeadingDotSel` attempts to get itself lowered, the compiler crashes by calling `lastWords` (internal error)

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
LPTK: I just think making unions with `Bot` is very inelegant when the problem's simplicity doesn't really call for it. If you really want to do se, you should at least use a smart constructor like `Union.mk` so that no actual `Bot` union is constructed.

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/Constraint.scala
LPTK: It doesn't look like a function type in Scala ;\^)

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
LPTK: Oh so that's why it seems to make sense to me 🤡

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
LPTK: Again, as I already said on Discord, why don't you just check of current sel-expansion is implemented? All you have to do is copy that (it "expands" the term (I don't think the terminology, which is from Harry, is very good – we should call it `rewrite` (mutably, in place))).

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
LPTK: Yeah this seems like an anti-feature. It will just become super hard for users to predict when things will successfully resolve or not. We don't want another C++-style cryptic/unintelligible language design, things that ought to work ought to be obvious ;^)

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
LPTK: When `Param` is constructed, a scope is available, isn't it? You can just add `(using State)` to `Param`.

hkmc2/shared/src/test/mlscript/flows/SelExpansion.mls
```diff
@@ -13,18 +13,21 @@ class Foo with
   val a = 123
 module Foo with
   fun foo(x: Foo) = x.a
+//│ FAILURE: Unexpected type error
+//│ FAILURE LOCATION: expandTerms (FlowAnalysis.scala:244)
+//│ ╔══[ERROR] Cannot resolve selection
+//│ ║  l.15: 	  fun foo(x: Foo) = x.a
```
LPTK: Wow unrelated features are getting broken in the process apparently :S

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
thornoar: true

hkmc2/shared/src/test/mlscript/flows/LeadingDotAccesses.mls
```diff
@@ -107,28 +59,3 @@ f(.a)
 //│     at REPLServer.onLine (node:repl:925:10)
 //│     at REPLServer.emit (node:events:519:28)
 //│     at REPLServer.emit (node:domain:489:12)
-
-:flow
-:df
-.b(new B)
```
LPTK: It's also useful to keep negative test cases that are expected fail.

hkmc2/shared/src/test/mlscript/flows/LeadingDotAccesses.mls
```diff
@@ -1,99 +1,51 @@
...
-//│ | New flow: ((x‹615›) -> 3) ~> flow:b‹616›
 
 :flow
 :df
 fun f(x : A) = 4
 f(.a)
-//│ Typing producer: { fun member:f(x: member:A) = 4; member:f(_?_.Ident(a)) }
+//│ FAILURE: Unexpected parse error
+//│ FAILURE LOCATION: simpleExprImpl (Parser.scala:697)
+//│ ╔══[PARSE ERROR] Expected an expression; found static selector instead
```
LPTK: Wait this doesn't even parse anymore? 👀

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
FlandiaYingman: Then let's expand "expand" to "rewrite"!

hkmc2/shared/src/test/mlscript/basics/NewlineOps.mls
```diff
@@ -161,17 +161,6 @@ let x
     .a
 //│ ═══[RUNTIME ERROR] Error: Access to required field 'a' yielded 'undefined'
 
-:pe
```
LPTK, line 164: You should never remove test cases just because they inconvenience you! Negative test cases are just as useful as positive ones...

thornoar: That's true, I just thought that this particular test was no longer relevant, since the parsing pattern that it exemplified is no longer there... I changed the flag from `pe` to `parseOnly` for now

UPD: I learned the proper error flags now

hkmc2/shared/src/test/mlscript/flows/LeadingDotAccesses.mls
```diff
@@ -6,156 +6,264 @@ class A with
...
+//│ | Solving: flow:b‹627› <: ⋅b‹662›   (Flow, Flow)   []
+//│ | New flow: flow:b‹627› ~> ⋅b‹662›
+//│ Handling constraint: _?_.a <: {b: ⋅b‹663›} (from _?_.a)
+//│ | Solving: _?_.a <: {b: ⋅b‹663›}   (LeadingDotSel, Sel)   []
+//│ | Found a pre-existing expansion member:A.a‹member:a› for _?_.a
+//│ | Solving: A <: {b: ⋅b‹663›}   (Ctor, Sel)   []
 //│ | Found immediate member member:b
-//│ | Solving: flow:b‹636› <: ⋅b‹645›   (Flow, Flow)   []
-//│ | New flow: flow:b‹636› ~> ⋅b‹645›
+//│ | Solving: flow:b‹627› <: ⋅b‹663›   (Flow, Flow)   []
```
LPTK: I found that these IDs are sometimes unsatble, and I'm not sure why. Better avoid committing code with `:df`.

hkmc2DiffTests/src/test/scala/hkmc2/MLsDiffMaker.scala
```diff
@@ -294,6 +294,7 @@ abstract class MLsDiffMaker extends DiffMaker:
       val flo = floan.typeProd(trm)
       floan.solveConstraints()
       floan.expandTerms()
+      floan.expandLeadingDotSels()
```
LPTK: This should be rolled into the general `expandTerms` function. There's no sense in separating them at this level.

hkmc2/shared/src/test/mlscript/flows/LeadingDotAccesses.mls
thornoar: It does now

hkmc2/shared/src/test/mlscript/flows/SelExpansion.mls
thornoar: Doesn't that make sense, given that the last term in the `foo` block is `(new A).test`, and `test` is a function?

LPTK: No, `(new A).test` does not return a function. It rewrites to `A.test(new A)`, which returns an instance of `A`, which is not a function!

LPTK: But it looks like this may not have anything to do with your changes. Was it akready behaving like this before?

thornoar: Yeah, it behaves the same way on the main `mlscript/hkmc2-flow` branch

thornoar: Here (lines 414-417 on my branch, 340-343 in the original) we might be pushing the wrong constraint. `lhs` ends up being the flow symbol of the static method which inherits a function type, and so this function type passes to the selection result
```
val lhs = memb match
  case memb: BlockMemberSymbol => P.Flow(memb.flow)
  case _ => TODO(memb)
toSolve.push(Constraint(lhs, sel.res))
```

LPTK: Yeah, that's probably it. Good catch! It should instead constraint something like `Constraint(lhs, FunctionType(receiver, sel.res))`

hkmc2/shared/src/test/mlscript/flows/SelExpansion.mls
```diff
@@ -12,7 +12,7 @@
 class Foo with
   val a = 123
 module Foo with
-  fun foo(x: Foo) = x.a
+  fun foo(x: Foo) = x.a
```
LPTK: 👀

hkmc2/shared/src/test/mlscript/flows/LeadingDotAccesses.mls
```diff
@@ -0,0 +1,119 @@
...
+  fun a = new C
+
+:sf
+:ge
+:e
+fun g(x : C) = x
+let d = .a
+f(d)
+g(d)
+//│ ╔══[ERROR] Ambiguous selection with multiple apparent targets
```
LPTK: There should be the location of `.a` attached to the main error message.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -157,6 +157,10 @@ object FlowSymbol:
   def sel(nme: Str)(using State) =
     // FlowSymbol(s"⋅$nme")
     FlowSymbol(s"⋅$nme")
+
+  def lds(using State) =
+    // FlowSymbol(s"⋅$nme")
+    FlowSymbol("lds")
```
LPTK: Maybe
```suggestion
    FlowSymbol(s"Ɛ⋅$nme")
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -93,6 +94,7 @@ sealed trait ResolvableImpl:
       case t: Term.TyApp => t.copy()(S(typ))
       case t: Term.Sel => t.copy()(t.sym, S(typ), t.originalCtx)(using t.state)
       case t: Term.SynthSel => t.copy()(t.sym, S(typ))
+      case _ => lastWords(s"Cannot attach a type to leading dot selection ${this.showDbg}")
```
LPTK: Why not?

LPTK: Hmm it's true this should never be needed. But then it's be better to move the Type stuff to a separate `TypeableImpl` and make LDS not inherit it.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -196,6 +198,30 @@ object Resolvable:
...
+  def hasLDS: Boolean = this.originalSel.isDefined
+
+  def withLDS(sel: Opt[Term.LeadingDotSel]) =
+    this.originalSel = sel
+    this
+
+  def inheritLDS(other: Term) = other match
+    case trm: LeadingDotRefImpl =>
+      this.withLDS(trm.originalSel)
+      trm.originalSel = N
```
LPTK: _Wow_ that's cursed!

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -196,6 +198,30 @@ object Resolvable:
...
 
+trait LeadingDotSelImpl(using State):
+  self: Term.LeadingDotSel =>
+  val resSym: FlowSymbol = FlowSymbol("lds")
+  var resolvedTargets: Ls[flow.SelectionTarget.CompanionMember] = Nil // * filled during flow analysis
+
+trait LeadingDotRefImpl:
+  self: Term =>
+
+  var originalSel: Opt[Term.LeadingDotSel] = N
```
LPTK: I don't think this is the best way of doing things. Why not apply the suggestion I made to just syntactically match terms that are immediate successions of selections and applications and attach the LDS producer to the output flow symbol of that chain?

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -1006,6 +1044,8 @@ extends Declaration, AutoLocated:
   // * it is not meant to be maintained afterwards (so it does not need to be copied around).
   var fldSym: Opt[FieldSymbol] = N
   
+  var flow: FlowSymbol = sym
```
LPTK: Why is there a mutable `flow` here?

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
```diff
@@ -17,12 +17,11 @@ import Producer as P
 import Consumer as C
 
 
-
```
LPTK, line 20: 👎

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
```diff
@@ -32,14 +31,23 @@ enum SelectionTarget:
   case CompanionMember(comp: Term, sym: FieldSymbol)
 
 
+// case class LeadingDotSelTarget(cs: Opt[ClassSymbol], trm: Term, sym: FieldSymbol)
```
LPTK: Obsolete comment?

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
thornoar: Oh, thank you for pointing it out, I forgot to change it to `val`

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
thornoar: I tried moving the `withTyp` method into a separate trait, but that messes up the `Resolver` because it can no longer call that method on instances of `ResolvableImpl`, and to fix it would mean that `Term.LeadingDotSel` is no longer an instance of `Resolvable`, and that causes the lowering to complain again... I can try to deal with this mess, but I can't promise anything. Would it be better to just add a `val typ: Opt[Type]` to `LeadingDotSel` and properly implement `withTyp`?

hkmc2/shared/src/test/mlscript/flows/LeadingDotAccesses.mls
thornoar: For some reason `Term.LeadingDotSel` doesn't infer a proper `Loc`... but if I put `sel.nme.toLoc` instead of `sel.toLoc`, that works

hkmc2/shared/src/test/mlscript/flows/SelExpansion.mls
thornoar: idk where it came from, lol)

hkmc2/shared/src/test/mlscript/flows/LeadingDotAccesses.mls
LPTK: Ok, a few things were missing, I've fixed the locations: https://github.com/hkust-taco/mlscript/pull/354/commits/1da4e7bd7bec2b1a66c5444d3c5a48e8e3e6d7f6

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
LPTK: I've made the refactoring for you: https://github.com/hkust-taco/mlscript/pull/354/commits/3dae1f3b9da6e01e8739312a42b9d0abd5e03ed9

LPTK: On second thought, I don't like the added complexity, which I find unnecessary. The reason is that we'll soon want to move away from resolving types in the resolver anyways (since implicit resolution will be moved to flow analysis Cc @FlandiaYingman), so the change is probably counter-productive. Let's keep this code as it is for now. Reverted: https://github.com/hkust-taco/mlscript/pull/354/commits/f772776ec830cda9d3ef457932b2d1dfecc90231

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -276,7 +276,7 @@ enum Term extends Statement:
     derivedClsSym: ClassSymbol, defs: Ls[HandlerTermDefinition], body: Term)
   case LeadingDotSel(nme: Tree.Ident)(
       val originalCtx: Opt[Elaborator.Ctx]
-      // var reachedType: Boolean
+      // val typ: Opt[Type]
```
LPTK: I think we can remove this.

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
```diff
@@ -33,33 +33,49 @@ enum SelectionTarget:
...
     
     def constrain(lhs: P, rhs: C): Unit = collectedConstraints += ((src = t, c = Constraint(lhs, rhs)))
     
+    def checkLDS(sub: Term)(res: Producer => Producer): Producer = 
+      t.ldsRoot match
+      case S(lds) if !insideSelAppChain =>
+        val sym = FlowSymbol("bind")
+        log("Constraining leading dot selection at the top level")
+        constrain(P.LeadingDotSel(lds), C.Flow(sym))
+        constrain(res(typeProd(sub, true)), C.Flow(sym))
```
LPTK:
```suggestion
        constrain(res(typeProd(sub, insideSelAppChain = true)), C.Flow(sym))
```

hkmc2/shared/src/test/mlscript/flows/SelExpansion.mls
```diff
@@ -27,6 +27,7 @@ module Foo with
 //│     flow:a⁰  ->  ⋅a⁰
 //│     flow:foo⁰  <~  ((type Foo) -> ⋅a⁰)
 
+
```
LPTK: No need for this extra line,

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
```diff
@@ -113,42 +124,53 @@ class FlowAnalysis(using tl: TraceLogger)(using Raise, State, Ctx):
...
-      case S(sym) => ???
-      case N =>
-        val sym = sel.resSym
-        constrain(pre1, C.Sel(nme, C.Flow(sym))(sel))
-        P.Flow(sym)
+      log(s"Selection ${sel.showDbg} ${sel.typ}")
+      checkLDS(pre): pre_t =>
+        sel.resolvedSym match
+        case S(sym: BlockMemberSymbol) => P.Flow(sym.flow)
+        case S(_) => P.Unknown(sel)
```
LPTK: You should not just return `Unknown` when you're not sure what to do. This will make it unnecessarily hard for us to later investigate why some things are not working. At the very least, log a message such as "unhandled symbol reference: ...".

LPTK: The same comment applies to the other uses of `Unknown` below.

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
LPTK: Dude... You marked this as "resolved" and then forgot to address it. The whole point of having these discussion items is to keep track of what's left to do in the PR. Please don't prematurely mark things as resolved.

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
```diff
@@ -177,13 +198,15 @@ class FlowAnalysis(using tl: TraceLogger)(using Raise, State, Ctx):
       Type.Error
   
   def typeParam(p: Param): C =
+  trace[C](s"Typing param: ${p.showDbg}", post = res => s": ${res.showDbg}"):
     p.signType match
     case S(typ) =>
-      val fs = p.sym.asInstanceOf[FlowSymbol]/*FIXME*/
-      fs.producers += ConcreteProd(Vector.empty, P.Typ(typ))
+      log(p.flow.showDbg)
```
LPTK: I don't think this log call is very informative.

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
```diff
@@ -177,13 +198,15 @@ class FlowAnalysis(using tl: TraceLogger)(using Raise, State, Ctx):
...
+      p.flow.producers += ConcreteProd(Vector.empty, P.Typ(typ))
       C.Typ(typ)
     case N =>
-      C.Flow(p.sym.asInstanceOf[FlowSymbol]/*FIXME*/)
+      log("nothing")
```
LPTK: I don't think this log call is very informative.

LPTK:
@FlandiaYingman I think reviewing this PR migth help you get a good grasp of the new flow analysis component.

FlandiaYingman:
> @FlandiaYingman I think reviewing this PR migth help you get a good grasp of the new flow analysis component.

Aye aye

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
thornoar: Oh, sorry for that, I forgot to remove it after debugging. My bad, won't happen again.

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
thornoar: Same as above

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
thornoar: That's fair, I'll add proper logging (since I am indeed not sure what to do otherwise)

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
LPTK: It's not the end of the world if it happens again, but yeah you should make sure it doesn't happen too often ;^)

hkmc2/shared/src/test/mlscript/basics/NewlineSels.mls
```diff
@@ -38,15 +38,12 @@ rcd
...
-//│ ╙──      	^^
-//│ ╔══[PARSE ERROR] Expected end of input; found static selector instead
-//│ ║  l.40: 	.a
-//│ ╙──      	^^
 //│ ╔══[WARNING] Pure expression in statement position
 //│ ║  l.39: 	rcd
 //│ ╙──      	^^^
+//│ ╔══[COMPILATION ERROR] Unexpected term form in expression position (leading dot selection)
+//│ ║  l.40: 	.a
+//│ ╙──      	^^
```
FlandiaYingman, line 46: I saw this error message is raised during lowering, but I guess it's not so appropriate to do that there?

If the resolver or flow analyzer fails to resolve a leading-dot selection, shouldn't it fail immediately and report the reason why it fails, rather than silently keep the term as-is and output a vague (to me) error during lowering?

hkmc2/shared/src/test/mlscript/flows/LeadingDotAccesses.mls
```diff
@@ -0,0 +1,151 @@
...
+//│ ║  l.125: 	f(.w)
+//│ ╙──       	  ^^
+//│ Flowed:
+//│   f⁰(_?_.w)‹app¹³›
+//│ where
+//│     app¹³  <~  type A
+//│ ╔══[COMPILATION ERROR] Unexpected term form in expression position (leading dot selection)
+//│ ║  l.125: 	f(.w)
+//│ ╙──       	  ^^
+//│ ═══[RUNTIME ERROR] Error: Function 'f' expected 1 argument but got 0
```
FlandiaYingman, line 129: This is similar to the previous comment. I think we won't want to raise an error `Unexpected term form in expression position (leading dot selection)` that is directly caused by another error `Cannot resolve leading dot selection`. This would create duplicate errors and unmeaningful errors to users.

The probably right way is to expand a term to `Term.Error` to silence its lowering error, given that we have already raised an error earlier regarding the same thing.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -256,6 +266,12 @@ enum Term extends Statement:
       case N => this
     case _ => this
   
+  lazy val ldsRoot: Opt[LeadingDotSel] = this match
+    case Sel(prefix, nme) => prefix.ldsRoot
+    case App(lhs, rhs) => lhs.ldsRoot
+    case sel: LeadingDotSel => S(sel)
+    case _ => N
```
FlandiaYingman, line 280: Probably adding some comments to this field would be helpful!

FlandiaYingman (commented):
Wow, amazing PR 🫣🫡

hkmc2/shared/src/test/mlscript/basics/NewlineSels.mls
LPTK: Good point. It should behave the same in this respect as normal selections. But that's something we're about to change anyways, so no need to do it right now.

hkmc2/shared/src/test/mlscript/flows/LeadingDotAccesses.mls
LPTK: Also a fine possibility. Turn them into runtime errors.

hkmc2/shared/src/test/mlscript/basics/NewlineSels.mls
thornoar: though currently the resolver doesn't resolve leading-dot selections, and the flow analyzer isn't even run in this example, so the term is passed straight to lowering. We could substitute this error with a better "unresolved LDS" error by doing what you suggested below and running flow analysis in the example, although I'm not sure that the `flow` flag is appropriate in this test file, idk what guidelines you follow in terms of which tests use which flags

FlandiaYingman: Wow so now flow analysis is an opt-in feature? Then I guess probably we want to raise an error saying that the leading dot selection is not resolved, and to resolve it we need the flow flag.

LPTK: Ah that's rigth, flow analysis is not run here. It's all good then, carry on, nothing to see here.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -588,6 +595,7 @@ sealed trait Statement extends AutoLocated, ProductWithExtraInfo:
       case imp: Import =>
         doc"import ${"\""}.../${imp.file.lastOpt.getOrElse("")}${"\""} as ${imp.sym.showName}"
       case LeadingDotSel(name) => doc"${this.showDbg}"
+      case Error => doc"error"
```
```suggestion
      case Error => doc"‹error›"
```

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
```diff
@@ -70,18 +82,17 @@ class FlowAnalysis(using tl: TraceLogger)(using Raise, State, Ctx):
         case stmt: DefineVar =>
           val rhs = typeProd(stmt.rhs)
           stmt.sym match
-          case sym: FlowSymbol =>
-            constrain(rhs, C.Flow(sym))
+          case sym: FlowSymbol => constrain(rhs, C.Flow(sym))
+          case _ => ()
```
LPTK: Is this the correct thing to do, or is it just an error-prone kludge (and maintenance burden) for a case that's not yet handled?

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
```diff
@@ -100,7 +111,7 @@ class FlowAnalysis(using tl: TraceLogger)(using Raise, State, Ctx):
                     ),
                   Nil,
                 )
-            case N => P.Unknown(cd)
+            case N => P.Ctor(cd.sym, Nil)(cd.body.blk)
```
LPTK: This is wrong. A class definition does not produce an instance fo the class (which `P.Ctor(cd.sym, ...)` represents).

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
```diff
@@ -70,18 +82,17 @@ class FlowAnalysis(using tl: TraceLogger)(using Raise, State, Ctx):
...
+          val sign_ty = t.sign.map(typeProd(_, insideSelAppChain)) // TODO use sign_ty
           val ps = t.params.map(typeParamList)
           t.body.foreach: bod =>
             val bod_ty = typeProd(bod)
             val fun_ty = ps.foldRight(bod_ty): (pl, acc) =>
               P.Fun(C.Tup(pl, N), acc, Nil)
             constrain(fun_ty, C.Flow(t.sym.flow))
-        case t: Term =>
-          typeProd(t)
+        case t: Term => typeProd(t)
```
LPTK: You have a lot of useless changes like that. In bit pattern matches, I find this form less readable. Please revert all of them.

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
thornoar: I probably made this change in when I was only starting on the project, without much understanding of the consequences. Thank you for pointing it out, I'll revert it for now

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
thornoar: Duly noted and reverted

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
```diff
@@ -158,15 +158,15 @@ class FlowAnalysis(using tl: TraceLogger)(using Raise, State, Ctx):
     case nw @ New(cls, args, rft) =>
       rft match
       case N =>
-        cls.resolvedSym.flatMap(_.asCls) match
```
LPTK: Well for this one your change was right! This code had no business not being indented (it was not done on purpose, and I'm surprised Scala accepted it).

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
```diff
@@ -285,11 +351,13 @@ class FlowAnalysis(using tl: TraceLogger)(using Raise, State, Ctx):
             case (P.Fun(pl, pr, _), C.Fun(cl, cr)) =>
               dig(cl, pl, path) // FIXME path
               dig(pr, cr, path) // FIXME path
-            case (P.Ctor(sym1, args1), C.Ctor(sym2, args2))
-            if (sym1 is sym2) && args1.size === args2.size // TODO generalize
-              =>
-              args1.zip(args2).foreach: (a1, a2) =>
+            case (P.Ctor(sym1, args1), C.Ctor(sym2, args2)) =>
+              if (sym1 is sym2) && args1.size === args2.size // TODO generalize
```
LPTK: This change is not correct. Two symbols can be different without there being a symbol mismatch (think of inheritance). We'll need to handle the subtyping constraints correctly later. Until then, let's not raise an incorrect error.

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
thornoar: yeah, it felt weird to me as well, the child cases having the same level of indentation as the parent case, but it compiled just fine, and at this point I didn't question it...

hkmc2/shared/src/main/scala/hkmc2/semantics/flow/FlowAnalysis.scala
thornoar: Good point, thank you

LPTK (approved):
Nice work, thanks!

PR #355, opened 2025-11-24
Instrumentation into Block IR

ChingLongTin:
Call implementation will be after the algorithms for instrumenting functions are completed.

1. allocating `optionNme`
This changed some diff tests (`OpenWildcard.mls`, `Imports.mls`, `ImportMLs.mls`) that imported Option themselves.
If I [allocate `optionNme` after checking if `stageCode` is enabled](https://github.com/hkust-taco/mlscript/commit/aae0086c576f3af053909ef64b357de11d27bb0c), it fails to link to the file.
Is there some way to avoid changing the diff tests, or is this fine?

2. implementing I-Inst
The current shape for Class `C(n:s)` includes the field names (`n`), but the field names are not available in `Instantiation`.
The field names are used in `sel`. If the class is staged using I-Cls, then we can use the staged `ClsLikeDefn` to retrieve the symbol names, but when do we do when it isn't staged?
```
class A(a)
staged module B
  class C(a)
  fun f() = 
    new C(1) // ok, C is staged
    new A(1) // not ok, we don't know field names of A after instrumentation
```


LPTK:
> * allocating `optionNme`
>   This changed some diff tests (`OpenWildcard.mls`, `Imports.mls`, `ImportMLs.mls`) that imported Option themselves.
>   If I [allocate `optionNme` after checking if `stageCode` is enabled](https://github.com/hkust-taco/mlscript/commit/aae0086c576f3af053909ef64b357de11d27bb0c), it fails to link to the file.
>   Is there some way to avoid changing the diff tests, or is this fine?
> *

Just follow the pattern of `Runtime`: you can name the symbols
```scala
    val shapeSetSymbol = TempSymbol(N, "shapeSet")
    val optionSymbol = TempSymbol(N, "option")
```
to avoid unnecessarily reserving the name `Option`.

LPTK:
> 2\. The current shape for Class `C(n:s)` includes the field names (`n`), but the field names are not available in `Instantiation`.
>     The field names are used in `sel`. If the class is staged using I-Cls, then we can use the staged `ClsLikeDefn` to retrieve the symbol names, but when do we do when it isn't staged?

So, in principle it should be possible to obtain information about a class from its symbol. Could we make the symbols referring to classes point to the class' `ClsLikeDefn`, even if that class is not staged (in which case we can probably store less information in the `ClsLikeDefn`)?

hkmc2/shared/src/test/mlscript-compile/Block.mls
```diff
@@ -29,7 +29,7 @@ class Case with
   constructor
     Lit(val lit: Literal)
     Cls(val cls: Symbol, val path: Path)
-    Tup(val len: Int) // TODO: remove inf?
```
LPTK: FYI `inf = true` is suppossed to indicate that the `len` is a lower bound and that the tuple could be bigger. It's what you get from a pattern like `[0, 1, ...xs, 8, 9]`.

LPTK:
Do you really need `showDefn`? I think the default pretty-printing is probably better, as it lays out large values on multiple lines.

ChingLongTin:
> Do you really need `showDefn`? I think the default pretty-printing is probably better, as it lays out large values on multiple lines.

The `show` functions are currently a placeholder for later, when we need to convert the mlscript Block constructs back into mlscript in [this branch](https://github.com/ChingLongTin/mlscript/tree/fix-pretty-print). For now, it just has the effect of ensuring the expression is well-typed.

ChingLongTin:
> Could we make the symbols referring to classes point to the class' `ClsLikeDefn`?

I have added a `ClassSymbol` class, so that both staged and unstaged functions can access the parameter names after instrumentation. So for example in `Functions.mls`,
```
:js
:staging
:sjs
class Outside(a)
staged module ClassInstrumentation with
  class NoArg
  fun inst1() = new Outside(1)
  fun inst2() = new NoArg
```
In `inst1` and `inst2`, there will be `ClassSymbol("Outside", Some([Symbol("a"), Symbol("b")]))` and `ClassSymbol("NoArg", None))` to access the parameter names.

hkmc2/shared/src/test/mlscript-compile/Block.mls
ChingLongTin: Do we add it back? In the formailzation the match pattern `[]^n` seems to indicate we only handle cases where `n` is exact.

LPTK: It's better to stick closest to the real IR. The formalization is slightly "idealized" to simplify the presentation.

hkmc2/shared/src/test/mlscript-compile/Shape.mls
```diff
@@ -28,31 +28,32 @@ class Shape with
...
 fun zipMrg[A](a: Array[A], b: Array[A]): Array[A] =
   a.map((a, i, _) => mrg2(a, b.at(i)))
 
-fun mrg2(s1: Shape, s2: Shape) = if s1
-  == s2 then s1
-  is
-    Lit(l) and s2 is Class(sym, params)
+fun mrg2(s1: Shape, s2: Shape) =
+  // we want different objects with same parameters to be equal
+  if show(s1) == show(s2) then s1
```
LPTK: This is a huge temporary hack, and deserves at least a big
```suggestion
  if show(s1) == show(s2) then s1 // FIXME: hack!
```

LPTK: An interesting side project you could undertake is to add generation of equality functions for `data class`es.

ChingLongTin:
From the meeting, the logic for shape will now be implemented in mlsript, so I am reducing the scope of Instrumentation.scala to just handle conversion from Block into Block IR. I will mark this pr as ready when that is done.

hkmc2DiffTests/src/test/scala/hkmc2/MLsDiffMaker.scala
```diff
@@ -167,7 +167,9 @@ abstract class MLsDiffMaker extends DiffMaker:
     if stageCode.isSet then
       given Config = mkConfig
       processTrees(
-        PrefixApp(Keywrd(`import`), StrLit(blockFile.toString)) :: Nil)
+        PrefixApp(Keywrd(`import`), StrLit(blockFile.toString))
+        :: PrefixApp(Keywrd(`import`), StrLit(optionFile.toString))
```
LPTK: Such imports are for things meant to be accessed by _users_, as they will lead to the introduction of a new name in scope (here, `Option`). But this won't work that well for users as you're importing the `mjs` instead of the `mls`, so we won't get MLscript symbols.

I think the import you want here is closer to the way `Runtime` is imported, which is never made available. Check out how that's done.

For the record, it looks like the `termFile` import above added by @NeilKleistGao is equally wrong.

ChingLongTin: I think you mean like in `JSBackendDiffMaker.scala:58` using `importRuntimeModule`, which is already replicated for both `Block` and `Option` in line 63-64.

I removed the import statement for `Option` in `MLsDiffMaker`, and kept the one for `Block`, since the former isn't available, while the latter is.

LPTK: Why is Block being handled differently than Option and Term? And again, if you want this to behave well, you should import the `mls`, not `mjs`.

ChingLongTin: ah, so what you meant was to just never make Block and Option available to the user. If we need the MLscript symbols, we would need to directly import the mls file in the mls file instead, instead of in the diff maker's `init`.

LPTK: Yes, more or less. Things that are imported by default are defined in `Predef`. There is no need to import more things by default. And if we did want to import these additional things, that should be done correctly by importing the `.mls` and not the `.mjs`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -310,6 +310,7 @@ object Elaborator:
     def init(using State): Ctx = Ctx.empty.copy(env = Map(
       "globalThis" -> globalThisSymbol,
       "Term" -> termSymbol,
+      "Block" -> blockSymbol,
```
LPTK: Again, I don't think this should be here and neither should `"Term"`. This is the context meant for the user. For internal uses, one should use an internal symbol (like `runtimeSymbol`), not a plain name. Cc @NeilKleistGao.

NeilKleistGao: I think the `block` and the `shape` module would not be used explicitly by users, but users do need to write code like `Term.codegen`. if we remove it from the context, then users should import the `term` module explicitly.

LPTK: If some functions like this should be accessible by users, they should be in the `Predef`. The `Predef` could either re-export `Term` (but it's a bit weird, since most things in it are not meant to be used directly by users) or define some submodule, such as `module meta with { fun codegen... }`.

hkmc2/shared/src/main/scala/hkmc2/codegen/Instrumentation.scala
```diff
@@ -4,12 +4,326 @@ package codegen
...
+      val companion = c.companion.get
+      val (stagedMethods, debugPrintCode) = companion.methods
+        .map(impl.transformFunDefn(sym, _))
+        .unzip
+
+      val unit = Select(Value.Ref(State.runtimeSymbol), Tree.Ident("Unit"))(N)
+      val debugBlock = debugPrintCode.foldRight(Return(unit, true))(concat)
+      def debugCont(rest: Block) =
+        Begin(debugBlock, rest)
+      val newCtor = impl.transformBlock(companion.ctor)(using new HashMap())(_ => debugCont(End()))
```
NeilKleistGao: So far, if we introduce things from #387, it will fail, since we do not consider the mutable features that #387 is using. We probably need to only handle classes inside modules and leave other things untouched.

ChingLongTin: I'm not sure how to bring the classes defined in the module to the next stage, The class `C` depends on the variable `x` defined within the module, so the `let` statement needs to be staged as well.
Will the classes never depend on variables that use mutable features?
```
module A with
  let x = 1
  fun get() = x
  class C() with
    fun get() = x
  fun add() =
    x = x + 1
```

LPTK: Mutation should probably not be allowed in staged modules, I think.

BTW, it should be `set x = x + 1` or equivalently `set x += 1`. OTOH `x = x + 1` is an only an initialization statement (and wll be rejected here once we have type checking).

hkmc2/shared/src/test/mlscript-compile/Block.mls
```diff
@@ -0,0 +1,189 @@
...
+    Define(val defn: Defn, val rest: Block)
+    // TODO: [fyp] handle Scoped, Label, Break nodes
+    Scoped(val symbols: Array[Symbol], val rest: Block)
+    Label(val labelSymbol: Symbol, val loop: Bool, val body: Block, val rest: Block)
+    Break(val labelSymbol: Symbol)
+    End()
+  
+fun concat(b1: Block, b2: Block) = if b1 is
+  Match(scrut, arms, dflt, rest) then Match(scrut, arms, dflt, concat(rest, b2))
+  Return(res, implct) then b2 // discard return?
```
NeilKleistGao: `b1` should not be discarded. When we have an early-return, we should discard the rest of the body, i.e., `b2`.

hkmc2/shared/src/main/scala/hkmc2/codegen/Instrumentation.scala
```diff
@@ -4,12 +4,326 @@ package codegen
...
+  // transformation helpers
+
+  def transformSymbol(sym: Symbol, symName: Str = "sym")(k: Path => Block): Block =
+    sym match
+    case clsSym: ClassSymbol =>
+      transformParamsOpt(clsSym.defn.get.paramsOpt): paramsOpt =>
+        blockCtor("ClassSymbol", Ls(toValue(sym.nme), paramsOpt), symName)(k)
+    case t: TermSymbol if t.defn.exists(_.sym.asCls.isDefined) =>
+      transformSymbol(t.defn.get.sym.asCls.get, symName)(k)
+    case _ => blockCtor("Symbol", Ls(toValue(sym.nme)), symName)(k)
```
NeilKleistGao: Notice that the default name of `TempSymbol` is always `tmp`, and those symbols will be renamed with a number suffix during codegen. We probably also need to use `Scope` to avoid name clashes here.

hkmc2/shared/src/main/scala/hkmc2/codegen/Instrumentation.scala
ChingLongTin: From our meeting on Tuesday, the mutable feature was only used to ensure no name clash happens for the generated class, but since the generated symbol name can't be spelled by the user, we don't need the mutable feature in the constructor. Therefore, we are able to stage the constructor normally.

LPTK:
Somehow there is a compilation error...

hkmc2/shared/src/main/scala/hkmc2/codegen/Instrumentation.scala
```diff
@@ -4,12 +4,346 @@ package codegen
...
+
+// it should be possible to cache some common constructions (End, Option) into the context
+// this avoids having to rebuild the same shapes everytime they are needed
+
+// transform Block to Block IR so that it can be instrumented in mlscript
+class InstrumentationImpl(using State, Raise):
+  type ArgWrappable = Path | Symbol
+  type Context = HashMap[Path, Path]
+  // TODO: there could be a fresh scope per function body, instead of a single one for the entire program
+  var scope = Scope.empty
```
NeilKleistGao:
```suggestion
  val scope = Scope.empty
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Instrumentation.scala
```diff
@@ -4,12 +4,346 @@ package codegen
...
+            if s.symbol.isDefined
+            then transformSymbol(s.symbol.get)
+            else blockCtor("Symbol", Ls(toValue(name)))
+          sym: sym =>
+            blockCtor("Select", Ls(x, sym), "sel")(k)
+      case DynSelect(qual, fld, arrayIdx) =>
+        transformPath(qual): x =>
+          transformPath(fld): y =>
+            blockCtor("DynSelect", Ls(x, y, toValue(arrayIdx)), "dynsel")(k)
+      case _ => ??? // not supported
```
NeilKleistGao: Better to throw a compilation error to say it is not supported.

hkmc2/shared/src/main/scala/hkmc2/codegen/Instrumentation.scala
```diff
@@ -4,12 +4,346 @@ package codegen
...
+      transformSymbol(labelSymbol): labelSymbol =>
+        transformBlock(body): (body, ctx) =>
+          transformBlock(rest)(using ctx): (rest, ctx) =>
+            blockCtor("Label", Ls(labelSymbol, toValue(loop), body, rest))(k(_, ctx))
+    case Break(labelSymbol) =>
+      transformSymbol(labelSymbol): labelSymbol =>
+        blockCtor("Break", Ls(labelSymbol))(k(_, ctx))
+    case _ => ??? // not supported
+
+  def transformFunDefn2(f: FunDefn)(using Context)(k: Path => Block): Block =
```
NeilKleistGao: What's the difference between `transformFunDefn` and `transformFunDefn2`?

hkmc2/shared/src/main/scala/hkmc2/codegen/Instrumentation.scala
```diff
@@ -4,12 +4,346 @@ package codegen
...
+    (newFun, debug)
+
+// TODO: rename as InstrumentationTransformer?
+class Instrumentation(using State, Raise) extends BlockTransformer(new SymbolSubst()):
+  val impl = new InstrumentationImpl
+
+  def concat(b1: Block, b2: Block): Block =
+    b1.mapTail {
+      case _: End => b2
+      case _ => ???
```
NeilKleistGao:
```suggestion
      case _ => lastWords("reason why this branch should be unreachable.")
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -269,13 +269,28 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
+                // val newDefn = ClassDef.Plain(mod.owner, syntax.Cls, clsSymb,
+                //   mod.bsym,
+                //   Nil,
+                //   N,
+                //   ObjBody(Blk(Nil, UnitVal())),
+                //   S(mod.sym),
+                //   stagedAnnots
+                // )
+                // clsSymb.defn = S(newDefn)
+                // newDefn
```
NeilKleistGao: Remove the comments if they are no longer useful.

hkmc2/shared/src/test/mlscript-compile/Shape.mls
```diff
@@ -0,0 +1,120 @@
...
+    [Lit(l), Class(sym, params)]
+      and isPrimitiveTypeOf(sym, l)
+      then Class(sym, params)
+    [Class(sym1, ps), Class(sym2, s2)]
+      and sym1.name == sym2.name
+      then Class(sym1, ps.map(p => [p.0, zipMrg(p.1, s2)]))
+    [Arr(s1, false), Arr(s2, false)]
+      and s1.length == s2.length
+      then Arr(zipMrg(s1, s2), false)
+    else Dyn()
```
NeilKleistGao: I think you can remove it (and other functions that will no longer be used).

hkmc2/shared/src/test/mlscript-compile/Shape.mls
```diff
@@ -0,0 +1,120 @@
...
+    ["Bool", b] and b is Bool then true
+    else false
+
+module Shape with...
+
+class Shape with
+  constructor
+    Dyn()
+    Lit(val l: Literal)
+    Arr(val shapes: Array[Shape], val inf: Bool)
```
NeilKleistGao: It seems that we forgot to discuss whether we really need spread parameters and infinite arrays in this project. How do you think? @LPTK

hkmc2/shared/src/test/mlscript/block-staging/Functions.mls
```diff
@@ -0,0 +1,114 @@
...
+//│ > FunDefn(Symbol("ctor$"), ([]), End, true)
+//│ > FunDefn(Symbol("call"), ([]), Assign(Symbol("$_no$_symbol$_"), Call(Ref(Symbol("+")), [Lit(1), Lit(1)]), Assign(Symbol("$_no$_symbol$_"), Call(Select(Ref(Symbol("Nonstaged")), Symbol("f")), []), Return(Call(Select(Ref(Symbol("Staged")), Symbol("f_gen")), []), false))), true)
+
+:w
+staged module Arguments with
+  fun f(x) =
+    x = 1
+    x
+  fun g(x)(y, z)() = z
+//│ ╔══[WARNING] Multiple parameter lists are not supported in shape propagation yet.
```
NeilKleistGao, line 71: Functions with multiple parameter lists can be desugared by the first-class function transformer (Enabled by `:ftc`. Probably after #404 is merged).

hkmc2/shared/src/main/scala/hkmc2/codegen/Instrumentation.scala
```diff
@@ -4,12 +4,346 @@ package codegen
...
+  def fnConcat(p1: Path, p2: Path, symName: String = "concat")(k: Path => Block): Block =
+    blockCall("concat", Ls(p1, p2), symName)(k)
+
+  // transformation helpers
+
+  def transformSymbol(sym: Symbol, symName: Str = "sym")(k: Path => Block): Block =
+    sym match
+    case clsSym: ClassSymbol =>
+      val name = scope.allocateOrGetName(sym)
+      transformParamsOpt(clsSym.defn.get.paramsOpt): paramsOpt =>
```
NeilKleistGao: Some class symbols (e.g., generated by another pass) may have no `defn`, which will lead to a crash.

NeilKleistGao (changes requested):

hkmc2/shared/src/test/mlscript-compile/Shape.mls
LPTK: There is no such thing as an infinite array, so not sure what you mean.

Spread parameters are important, but not on the critical path. Let's try to do without spreads for now, and add them later or if they become critical for progress.

hkmc2/shared/src/main/scala/hkmc2/codegen/Instrumentation.scala
ChingLongTin: `transformFunDefn2` is a continuation that gets reused to stage class functions. If the classes are lifted out of the module, then it can just be inlined into `transformFunDefn`.

LPTK: It's fine to leave it out, but consider using a more descriptive name.

hkmc2/shared/src/main/scala/hkmc2/codegen/Instrumentation.scala
ChingLongTin: Since the parameters are necessary to reconstruct the class instance during specialization, this case throws a compiler error for now.

hkmc2/shared/src/test/mlscript-compile/Shape.mls
NeilKleistGao: Then we shall remove this `in` parameter, I guess, and mark spread parameters not supported for now.

hkmc2/shared/src/main/scala/hkmc2/codegen/Instrumentation.scala
NeilKleistGao: I see. Then it is enough to raise a compilation error here. I will set a dummy definition with parameters in #404 so that our code can cooperate.

Is there any other information that you need to retrieve from the `defn` field?

ChingLongTin: From our discussions last time, we need `paramsOpt` and `auxParams`.

hkmc2/shared/src/test/mlscript-compile/Shape.mls
ChingLongTin: Spread operators are also what causes `Block.Tup` to have `inf = true`, as discussed here: https://github.com/hkust-taco/mlscript/pull/355/#discussion_r2563723032

Are we also disallowing pattern matching with spread parameters, so both (1) and (2) are not allowed?
```
if [1, ..[1, 2]] is // (1: spread in Arg)
  [1, ...x] then x // (2: spread in Tup)
  else 0
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Instrumentation.scala
NeilKleistGao: OK. It seems that dummy definitions do not help either. I think a proper way is to have another pass before starting instrumentation to collect these locally generated classes (represented in block IR). If there is no definition for the class symbol, you try to retrieve it from the collected data. These generated classes are not visible outside the module, so when we access a class via the path, it should always be possible to find the defined `defn` field.

hkmc2/shared/src/test/mlscript-compile/Shape.mls
LPTK: Yes, it's probably fine to reject them _for now_. Again, ideally, we'll try to revisit this later, though.

LPTK: I've added this as a draft issue in the Dynamic Staging project https://github.com/orgs/hkust-taco/projects/17/views/1?pane=issue&itemId=162809949

ChingLongTin:
There's a diff test change on `backlog/NamedArgs.mls`, `codegen/BasicTerms.mls`, `codegen/Do.mls`, `ucs/syntax/NestedOpSplits.mls`, from the printing of the `Predef` symbol. For example in the first file we have the following diff:
```
diff --git a/hkmc2/shared/src/test/mlscript/backlog/NamedArgs.mls b/hkmc2/shared/src/test/mlscript/backlog/NamedArgs.mls
index e5e0724..4c99118 100644
--- a/hkmc2/shared/src/test/mlscript/backlog/NamedArgs.mls
+++ b/hkmc2/shared/src/test/mlscript/backlog/NamedArgs.mls
@@ -37,7 +37,7 @@ id(a: 0)
 //│   stats = Nil
 //│   res = App:
 //│     lhs = SynthSel{sym=member:id⁰}:
-//│       prefix = Ref{sym=member:Predef⁰} of member:Predef⁰
+//│       prefix = Ref{sym=member:Predef¹} of member:Predef¹
 //│       nme = Ident of "id"
 //│     rhs = Tup of Ls of 
 //│       Fld:
```

LPTK:
@ChingLongTin, indeed, there was some flakiness that should be fixed now.

hkmc2/shared/src/main/scala/hkmc2/codegen/Instrumentation.scala
LPTK: @ChingLongTin are you going to follow @NeilKleistGao's suggestion in this PR? Or in a follow-up one?

ChingLongTin: I will do it in a follow-up one.

hkmc2/shared/src/test/mlscript/block-staging/Functions.mls
```diff
@@ -0,0 +1,135 @@
...
+//│ > FunDefn(Symbol("match2"), ([]), Scoped([Symbol("tmp1"), Symbol("element1$"), Symbol("element0$"), Symbol("a")], Label(Symbol("split_root$"), false, Label(Symbol("split_1$"), false, Match(Ref(Symbol("x1")), [Tup(0) -> Assign(Symbol("tmp1"), Lit(1), Break(Symbol("split_root$"))), Tup(2) -> Assign(Symbol("element0$"), Call(Select(Select(Ref(Symbol("runtime")), Symbol("Tuple")), Symbol("get")), [Ref(Symbol("x1")), Lit(0)]), Assign(Symbol("element1$"), Call(Select(Select(Ref(Symbol("runtime")), Symbol("Tuple")), Symbol("get")), [Ref(Symbol("x1")), Lit(1)]), Match(Ref(Symbol("element0$")), [Lit(1) -> Match(Ref(Symbol("element1$")), [Lit(2) -> Assign(Symbol("tmp1"), Lit(2), Break(Symbol("split_root$")))], Assign(Symbol("a"), Ref(Symbol("element0$")), Break(Symbol("split_1$"))), End)], Assign(Symbol("a"), Ref(Symbol("element0$")), Break(Symbol("split_1$"))), End)))], Assign(Symbol("tmp1"), Lit(0), Break(Symbol("split_root$"))), End), Assign(Symbol("tmp1"), Lit(3), End)NaN, Return(Ref(Symbol("tmp1")), false)NaN), true)
```
LPTK: I just noticed these suspicious `NaN` occurrences. Is a wrong concatenation being done?

LPTK (changes requested):

hkmc2/shared/src/test/mlscript/block-staging/Functions.mls
LPTK: (The problem is still there after the merge.)

ChingLongTin: oops, i fixed it now.

LPTK (approved):
Ok, LGTM, but please remember to address the two remaining comments in follow-up PRs.

PR #356, opened 2025-11-27
Add `Scoped` blocks

ychenfo:
This PR introduces `Scoped` blocks. ~~It is still a draft because there is a lot of clean up to be done.~~

Currently, this PR:
- Only symbols that are known to be declared in the elaborated tree and symbols generated during lowering are put in the `Scoped` block
- Does not update later passes (except for `LambdaRewriter`) to properly maintain the `Scoped` blocks
- Does not generate nested `Scoped` blocks for if branches
- Also fixed the scope problem in `preCtor`/`ctor` JS code generation


hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -304,7 +304,7 @@ case class Label(label: Local, loop: Bool, body: Block, rest: Block) extends Blo
 case class Break(label: Local) extends BlockTail
 case class Continue(label: Local) extends BlockTail
 
-case class Scoped(syms: collection.Set[Local], body: Block) extends BlockTail
+case class Scoped(syms: collection.Set[Local], body: Block, topLevel: Bool) extends BlockTail
```
LPTK: What's the semantics of `topLevel`?

LPTK:
Let's try to avoid producing code with `{{`. If we're already at the top-level of a block, there's no need for the extra curly braces. This can be achieved by calling a small wrapper `def returningBlock` instead of calling `def returningTerm` when we build the JS of a block, or something like that.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -479,7 +479,7 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
       t :: e :: returningTerm(rest, endSemi)
     
     case Begin(sub, thn) =>
-      doc"${returningTerm(sub, endSemi = true)}${returningTerm(thn, endSemi)}"
+        doc"${returningTerm(sub, endSemi = true)}${returningTerm(thn, endSemi)}"
```
LPTK: Why?

LPTK:
> [WIP: Try to remove redundant braces (not clean up yet)](https://github.com/hkust-taco/mlscript/pull/356/commits/64385a29c6236b5452a5ddedc37724ada2b34141)

Ok, this clearly looks like the wrong approach.

We may want to do it _the other way around_: never output braces unless we're producing the code of a nested "rest part" (like the `rest` of an `Assign` or `Begin`, for example).

NeilKleistGao:
> > [WIP: Try to remove redundant braces (not clean up yet)](https://github.com/hkust-taco/mlscript/pull/356/commits/64385a29c6236b5452a5ddedc37724ada2b34141)
> 
> Ok, this clearly looks like the wrong approach.
> 
> We may want to do it _the other way around_: never output braces unless we're producing the code of a nested "rest part" (like the `rest` of an `Assign` or `Begin`, for example).

It seems that after we have the smart constructors, we do not need to add braces for `Scoped` explicitly. I believe the latest commit removes the unnecessary braces correctly.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -399,6 +399,99 @@ sealed trait Statement extends AutoLocated, ProductWithExtraInfo:
     case RcdField(field, rhs) => RcdField(field.mkClone, rhs.mkClone)
     case RcdSpread(rcd) => RcdSpread(rcd.mkClone)
     case DefineVar(sym, rhs) => DefineVar(sym, rhs.mkClone)
+
+  lazy val definedSyms: Set[Symbol] = this match
```
LPTK: This is wrong because it ignores the fact we may have rewritten some subterms (through `expand`). I don't see why you need it at all.

LPTK:
> It seems that after we have the smart constructors, we do not need to add braces for `Scoped` explicitly. I believe the latest commit removes the unnecessary braces correctly.

You can't just produce wrong code for some given IR just because in our current tests and uses we happen to use the smart constructors! Obviously, the code generator should work for **_all_** valid IR representations, not just those that are currently reachable from our tests. If you want to make nested scopes invalid, that should be expressed as an invariant that's actually enforced (and at the very least causes crashes when it's violated, rather than miscompiling). But I am not sure we actually want to do that: again, theoretically it can be useful to have the ability of generating nested scopes (eg to reduce the space used by function activation frames).

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
ychenfo: Thanks, I was not aware of the `expand` thing. Does it make sense to change this `lazy val` to `def`? Or do you think we should also use the lowering ctx to collect the symbols in this `definedSyms` for term?

Indeed the lowering process traverses terms anyway and this traversal of terms in `definedSyms` can be removed; currently I do this because we feel it is less error‑prone and cleaner to collect the defined symbols that can be known from elaborated terms in one centralized place, rather than sprinkling that logic throughout the lowering code... Or is there another way that I am missing?

LPTK: So you abandoned my suggestion to use the lowering context? I don't know how that is "sprinkling logic" instead of using a centralized place. `Scope` is a middle IR concept and `Lowering` is the place where the corresponding logic should be. _That's_ where we determine exactly what code needs to be let-bound in later code generation. Note all bound symbols from terms will survive lowering/erasure, and on the other hand the terms don't know all symbols that the lowering will need. So I really fail to see how your argument makes sense.

ychenfo: Oh, I didn't mean to abandon the lowering context... I meant that without `term.definedSyms`, to add the bound symbols from terms (which will all survive the lowering), we need to insert even more calls to `loweringCtx.collectScopedSym(s: Symbol*)` at many different places in the current lowering code (which indeed is already done for e.g. temp symbols that lowering generates, at line 406, 480, 576, 613, 664, 667, 671, 681... in `Lowering.scala`).

LPTK: > which will all survive the lowering

Are you saying that `T` will not be part of `s.definedSyms` when `s` is the statement `type T = Int`? That is counter-intuitive, and, as you see, misaligned concerns/levels of reasoning.

> we need to insert even more calls to `loweringCtx.collectScopedSym(s: Symbol*)`

First, don't make this a vararg. Varargs are expensive at runtime and I don't think adding more than one at a time is the common case.

> at many different places in the current lowering code

In my book this is strictly better than collecting some of the symbols in lowering and some of the symbols in a completely different part of the codebase (the frontend, which should not be concerned with lowering/codegen). What you currently do _is exactly_ "sprinkling logic" instead of doing things in "one centralized place". This one centralized place should be Lowering, which should be the source of truth about how terms translate into IR.

ychenfo: Oh now I see, I will remove this and update the lowering pass. Thanks for the explanation.

hkmc2/shared/src/test/mlscript/basics/CompanionModules_Classes.mls
```diff
@@ -107,7 +107,6 @@ module Foo with
 //│   }
 //│   constructor() {}
 //│   static {
-//│     let tmp2;
```
LPTK, line 110: It seems that `tmp2` was wrongly moved out of this scope.

hkmc2/shared/src/test/mlscript/codegen/Scoped.mls
```diff
@@ -0,0 +1,277 @@
...
+//│ let f;
+//│ f = function f() {
+//│   let scrut, a, b;
+//│   scrut = true;
+//│   if (scrut === true) { a = 4; return a + 4 } else { b = 5; return b + 9 }
+//│ };
+
+
+
+// `let a` is declared in the correct place,
```
LPTK: Again, it's not "the correct place". There are several possible places where it would make sense, and this one is not the best (to be imprived on later).

hkmc2/shared/src/test/mlscript/basics/CompanionModules_Classes.mls
LPTK: This will actually lead to incorrect semantics, as it makes constructors non-reentrant.

ychenfo: Thanks for the catch! This has been fixed.

hkmc2/shared/src/test/mlscript/codegen/Scoped.mls
ychenfo: Thanks I see, I have updated this comment

hkmc2/shared/src/test/mlscript-compile/Predef.mjs
```diff
@@ -124,27 +124,26 @@ globalThis.Object.freeze(class Predef {
   } 
   static foldr(f) {
     return (first, ...rest) => {
-      let len, scrut, i, init, tmp; /** scoped **/
+      let len, scrut, i, init, scrut1, tmp, tmp1, tmp2, tmp3; /** scoped **/
```
LPTK: This looks like a regression. We should at least insert a scope around the body of `while (true) { ... }` while generating the code. Can we do that? Then, this means we can disable while loop rewriting (https://github.com/hkust-taco/mlscript/pull/351) by default.

ychenfo: `:effectHandlers` floats out definitions inside while loops, so having `Scoped` blocks for while loop bodies leads to variables captured by the floated out definitions to be unbound. I tried to modify the handler so that it does not float defintions out across `Scoped` blocks (such that we can have `Scoped` blocks for while loop bodies). But in this setting, there are still some test failures if I disable while loop rewriting by default... (though everything seems to work if while loop rewriting is still on by default).

LPTK: As discussed in the group meeting, we are currently generating code that fails to output bindings when in sanitized mode.

After fixing that, maybe

> I tried to modify the handler so that it does not float defintions out across Scoped blocks (such that we can have Scoped blocks for while loop bodies). But in this setting, there are still some test failures if I disable while loop rewriting by default...

will start working.

(If not, just selectively enable loop rewriting when handler lowering is on.)

ychenfo: Thanks for the conclusion of the group meeting! Unfortunately, another problem that I didn't get to mention in the group meeting still exists after the code is fixed to generate correct bindings in the sanitized mode.

The problem is related to `Cont$func$start` generated by `:effectHandler`:
```
:effectHandlers
:sjs
fun start(x, f) =
  while x do
    f(() => x)
```

<details>

<summary>This program is currently compiled to (in line 69 `lambda` is unbound):</summary>

```
047: //│ let start; /** scoped **/
048: //│ start = function start(x, f) {
049: //│   let tmp; /** scoped **/
050: //│   let Cont$func$start$1, doUnwind;
051: //│   globalThis.Object.freeze(class Cont$func$start$ extends runtime.FunctionContFrame.class {
052: //│     static {
053: //│       Cont$func$start$1 = this
054: //│     }
055: //│     constructor(pc) {
056: //│       let tmp1;
057: //│       tmp1 = super(null);
058: //│       this.pc = pc;
059: //│     }
060: //│     resume(value$) {
061: //│       if (this.pc === 1) {
062: //│         tmp = value$;
063: //│       }
064: //│       contLoop: while (true) {
065: //│         if (this.pc === 2) {
066: //│           return tmp
067: //│         } else if (this.pc === 3) {
068: //│           if (x === true) {
069: //│             tmp = runtime.safeCall(f(lambda));
070: //│             if (tmp instanceof runtime.EffectSig.class) {
071: //│               return this.doUnwind(tmp, 1)
072: //│             }
073: //│             this.pc = 1;
074: //│             continue contLoop
075: //│           } else {
076: //│             tmp = runtime.Unit;
077: //│             this.pc = 2;
078: //│             continue contLoop
079: //│           }
080: //│           this.pc = 2;
081: //│           continue contLoop
082: //│         } else if (this.pc === 1) {
083: //│           this.pc = 3;
084: //│           continue contLoop
085: //│         }
086: //│         break;
087: //│       }
088: //│     }
089: //│     toString() { return runtime.render(this); }
090: //│     static [definitionMetadata] = ["class", "Cont$func$start$"]; 
091: //│   });
092: //│   doUnwind = function doUnwind(res, pc) {
093: //│     res.contTrace.last.next = new Cont$func$start$1(pc);
094: //│     res.contTrace.last = res.contTrace.last.next;
095: //│     return res
096: //│   };
097: //│   tmp1: while (true) {
098: //│     let lambda; /** scoped **/
099: //│     lambda = (undefined, function () {
100: //│       return x
101: //│     });
102: //│     if (x === true) {
103: //│       tmp = runtime.safeCall(f(lambda));
104: //│       if (tmp instanceof runtime.EffectSig.class) {
105: //│         return doUnwind(tmp, 1)
106: //│       }
107: //│       continue tmp1
108: //│     } else { tmp = runtime.Unit; }
109: //│     break;
110: //│   }
111: //│   return tmp
112: //│ };
```
</details>

When loop rewriting is switched on, it works fine, so I guess now I can just selectively enable loop rewriting when handler lowering is on?

LPTK: Ok, sure, let's do that for now (but pls document this quirk at the place where you implement the selective switch).

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -544,6 +544,16 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
               source = Diagnostic.Source.Compilation)
         conclude(Value.Ref(ctx.builtins.debug.getLocals, N).withLocOf(f))
+      case t if instantiatedResolvedBms.exists(_ is ctx.builtins.scope.locally) =>
+        arg match
+          case Tup(Fld(_, Lam(ParamList(_, Nil, N), body), N) :: Nil) =>
+            inScopedBlock(true)(block(Nil, R(body))(k))
+          case _ =>
+            return fail:
+              ErrorReport(
+                msg"locally requires a lambda with no parameter." ->
```
LPTK: Why would it require a lambda at all? Just embed the whole `t` argument as a sub-block.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -304,7 +304,9 @@ case class Label(label: Local, loop: Bool, body: Block, rest: Block) extends Blo
 case class Break(label: Local) extends BlockTail
 case class Continue(label: Local) extends BlockTail
 
-case class Scoped(syms: collection.Set[Local], body: Block) extends BlockTail
+
+// set dontFlatten to true to force a nested Scoped
+case class Scoped(syms: collection.Set[Local], body: Block)(val dontFlatten: Bool) extends BlockTail
```
LPTK: It's not acceptable to pollute the definition of `Scoped` (making it harder to use) just for this use case. It should be enough to make the lowering just construct a `Begin` block with a body that's not flattened. You know, you can avoid the smart constructor by using `new Begin`.

hkmc2DiffTests/src/test/scala/hkmc2/MLsDiffMaker.scala
```diff
@@ -99,7 +99,7 @@ abstract class MLsDiffMaker extends DiffMaker:
       liftDefns = Opt.when(liftDefns.isSet)(LiftDefns()),
       stageCode = stageCode.isSet,
       target = if wasm.isSet then CompilationTarget.Wasm else CompilationTarget.JS,
-      rewriteWhileLoops = !dontRewriteWhile.isSet,
+      rewriteWhileLoops = false //!dontRewriteWhile.isSet,
```
LPTK: Then, it would make more sense to change the flag to `rewriteWhile` and actually use its value. Of course we don't want to leave a flag that's silently ignored!

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -579,7 +579,9 @@ class HandlerLowering(paths: HandlerPaths, opt: EffectHandlers)(using TL, Raise,
   private def thirdPass(b: Block): Block =
     // to ensure the fun and class references in the continuation class are properly scoped,
     // we move all function defns to the top level of the handler block
-    val (blk, defns) = b.floatOutDefns()
+    // NOTE: this is to prevent floating things out too aggressively
+    // such that the captured variables become unbound.
+    val (blk, defns) = b.floatOutDefnsUntilScoped()
```
LPTK: These two comments seem contradictory. What does "too aggressively" mean, exactly? Will it not break the semantics of handler lowering if "too aggressively" happens to not be enough for the handler lowering's original purpose.

Do you actually understand how these things interact and can you actually prove that this interaction is correct?

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -595,7 +595,7 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
     val (scopedSyms, unscopedMain) = p.main match
       case Scoped(syms, body) /* if exprt.isDefined */ => (syms, body)
       case _ => (Set.empty, p.main)
-    imps.mkDocument(doc" # ") :/: (genLetDecls(scopedSyms.iterator.map(l => l -> scope.allocateName(l)), true) :: block(unscopedMain, endSemi = false)).stripBreaks :: (
+    imps.mkDocument(doc" # ") :/: (genLetDecls(scopedSyms.toArray.sortBy(_.uid).iterator.map(l => l -> scope.allocateName(l)), true) :: block(unscopedMain, endSemi = false)).stripBreaks :: (
```
LPTK: line is way too long

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -605,8 +605,13 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
     reserveNames(p)
     lazy val imps = p.imports.map: i =>
       doc"""${getVar(i._1, N)} = await import("${i._2.toString}").then(m => m.default ?? m);"""
-    blockPreamble(p.imports.map(_._1).toSeq ++ p.main.definedVarsNoScoped.toSeq) ->
-      (imps.mkDocument(doc" # ") :/: returningTerm(p.main, endSemi = false).stripBreaks)
+    p.main match
+      case Scoped(syms, body) =>
```
LPTK: no indent needed

hkmc2DiffTests/src/test/scala/hkmc2/JSBackendDiffMaker.scala
```diff
@@ -141,19 +131,9 @@ abstract class JSBackendDiffMaker extends MLsDiffMaker:
...
-        val vars = toplvlDefinedVars.toArray.sortBy(_.uid).map(l => l -> baseScp.allocateName(l))
-        (if vars.isEmpty then doc"" else
-        doc" # let " :: vars.map: (_, nme) =>
-          nme
-        .toList.mkDocument(", ")
-        :: doc";").stripBreaks.mkString(100)
       val (pre, js) = nestedScp.givenIn:
         jsb.worksheet(le)
-      val preStr = pre.stripBreaks.mkString(100) + varsFromScopedStr
+      val preStr = pre.stripBreaks.mkString(100) //+ varsFromScopedStr
```
LPTK: obsolete comment?

hkmc2DiffTests/src/test/scala/hkmc2/JSBackendDiffMaker.scala
```diff
@@ -109,16 +109,16 @@ abstract class JSBackendDiffMaker extends MLsDiffMaker:
...
       val lowered0 = low.program(blk)
       val le = lowered0.copy(main = lowered0.main.mapTail:
-        case e: End =>
-          Assign(resSym, Value.Lit(syntax.Tree.UnitLit(false)), e)
-        case Return(res, implct) =>
-          assert(implct)
-          Assign(resSym, res, Return(Value.Lit(syntax.Tree.UnitLit(false)), true))
-        case tl: (Throw | Break | Continue) => tl
-      )
+          case e: End =>
```
LPTK: why change this indentation?

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -582,24 +605,44 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
...
+      case _ =>
+        blockPreamble(p.imports.map(_._1).toSeq ++ p.main.definedVarsNoScoped.toSeq) ->
+          (imps.mkDocument(doc" # ") :/: returningTerm(p.main, endSemi = false).stripBreaks)
+
+  def genLetDecls(vars: Iterator[(Symbol, Str)], isScoped: Bool): Document =
+    if vars.isEmpty then doc"" else
+      doc" # let " :: vars.map: (_, nme) =>
+        nme
+      .toList.mkDocument(", ")
+      :: (if isScoped then doc"; /** scoped **/" else doc";")
```
LPTK: Please remove this (you can add it when debugging is on, but not by defautl!).

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -582,24 +605,44 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
...
+    case Scoped(syms, body) => 
+      val vars = syms.filter(scope.lookup(_).isEmpty).toArray.sortBy(_.uid).iterator.map(l =>
+        l -> scope.allocateName(l))
+      genLetDecls(vars, true) :: k(body)
+    case _ => k(blk)
+  
   
   def block(t: Block, endSemi: Bool)(using Raise, Scope): Document =
-    blockPreamble(t.definedVars) :: returningTerm(t, endSemi)
+    // println(s"$t :::::::: ${t.definedVars}")
```
LPTK: obsolete comment?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
NeilKleistGao: Then, is it acceptable to change the Elaborator to allow the following program:
```fs
scope.locally of
  let x = 42
  x
//│ FAILURE LOCATION: subterm (Elaborator.scala:383)
//│ ╔══[ERROR] Expected a body for let bindings in expression position
//│ ║  l.5: 	  let x = 42
//│ ╙──     	      ^^^^^^
//│ FAILURE LOCATION: subterm (Elaborator.scala:456)
//│ ╔══[ERROR] Name not found: x
//│ ║  l.6: 	  x
//│ ╙──     	  ^
//│ FAILURE LOCATION: term (Lowering.scala:556)
//│ ╔══[COMPILATION ERROR] locally requires a lambda with no parameter.
//│ ║  l.4: 	scope.locally of
//│ ╙──     	^^^^^^^^^^^^^
```
so that in the lowering we can get an application with a single `Blk` parameter?

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
ychenfo: Indeed I have found another example program that breaks when I was trying to justify the current implementation... The problem is that some nested Scoped blocks were still generated for *nested* `if`s, and this combined with `floatOutDefnsUntilScoped` causes the bodies of some classes (which are inserted by effect handler lowering on the top level of functions, and contain references to the variables for inner `if`s) to contain unbound variables, because the definitions are not floated out enough.

A simplified program that causes the trouble:
```
fun foo(h) =
  h.perform()
  if true then
    if true then
      fun g() = 2
      g()
handle h = Effect with
  fun perform()(k) = k(())
foo(h)
```

<details>

<summary> This was incorrectly compiled to (`g_not_found` in line 315): </summary>

```
286: let foo, tmp; /** scoped **/
287: let handleBlock$;
288: foo = function foo(h) {
289:   let scrut, tmp1, Cont$func$foo$1, doUnwind; /** scoped **/
290:   let res, res1;
291:   globalThis.Object.freeze(class Cont$func$foo$ extends runtime.FunctionContFrame.class {
292:     static {
293:       Cont$func$foo$1 = this
294:     }
295:     constructor(pc) {
296:       let tmp2;
297:       tmp2 = super(null);
298:       this.pc = pc;
299:     }
300:     resume(value$) {
301:       let scrut1;
302:       if (this.pc === 1) {
303:         tmp1 = value$;
304:       } else if (this.pc === 3) {
305:         res1 = value$;
306:       } else if (this.pc === 2) {
307:         res = value$;
308:       }
309:       contLoop: while (true) {
310:         if (this.pc === 1) {
311:           scrut = true;
312:           if (scrut === true) {
313:             scrut1 = true;
314:             if (scrut1 === true) {
315:               return g_not_found()
316:             } else {
317:               res = globalThis.Object.freeze(new globalThis.Error("match error"));
318:               if (res instanceof runtime.EffectSig.class) {
319:                 return this.doUnwind(res, 2)
320:               }
321:               this.pc = 2;
322:               continue contLoop
323:             }
324:             this.pc = 4;
325:             continue contLoop
326:           } else {
327:             res1 = globalThis.Object.freeze(new globalThis.Error("match error"));
328:             if (res1 instanceof runtime.EffectSig.class) {
329:               return this.doUnwind(res1, 3)
330:             }
331:             this.pc = 3;
332:             continue contLoop
333:           }
334:           this.pc = 4;
335:           continue contLoop
336:         } else if (this.pc === 4) {
337:           break contLoop
338:         } else if (this.pc === 3) {
339:           throw res1
340:         } else if (this.pc === 2) {
341:           throw res
342:         }
343:         break;
344:       }
345:     }
346:     toString() { return runtime.render(this); }
347:     static [definitionMetadata] = ["class", "Cont$func$foo$"]; 
348:   });
349:   doUnwind = function doUnwind(res2, pc) {
350:     res2.contTrace.last.next = new Cont$func$foo$1(pc);
351:     res2.contTrace.last = res2.contTrace.last.next;
352:     return res2
353:   };
354:   tmp1 = runtime.safeCall(h.perform());
355:   if (tmp1 instanceof runtime.EffectSig.class) {
356:     return doUnwind(tmp1, 1)
357:   }
358:   scrut = true;
359:   if (scrut === true) {
360:     let g, scrut1; /** scoped **/
361:     g = function g() {
362:       return 2
363:     };
364:     scrut1 = true;
365:     if (scrut1 === true) {
366:       return g()
367:     } else {
368:       res = globalThis.Object.freeze(new globalThis.Error("match error"));
369:       if (res instanceof runtime.EffectSig.class) {
370:         return doUnwind(res, 2)
371:       }
372:       throw res
373:     }
374:   } else {
375:     res1 = globalThis.Object.freeze(new globalThis.Error("match error"));
376:     if (res1 instanceof runtime.EffectSig.class) {
377:       return doUnwind(res1, 3)
378:     }
379:     throw res1
380:   }
381: };
382: handleBlock$ = function handleBlock$() {
383:   let h, res, Cont$handleBlock$h$3, doUnwind, Handler$h$3;
384:   globalThis.Object.freeze(class Handler$h$2 extends Effect1 {
385:     static {
386:       Handler$h$3 = this
387:     }
388:     constructor() {
389:       let tmp1;
390:       tmp1 = super();
391:     }
392:     perform() {
393:       let hdlrFun;
394:       hdlrFun = function hdlrFun(k) {
395:         return runtime.safeCall(k(runtime.Unit))
396:       };
397:       return runtime.mkEffect(this, hdlrFun)
398:     }
399:     toString() { return runtime.render(this); }
400:     static [definitionMetadata] = ["class", "Handler$h$"]; 
401:   });
402:   h = new Handler$h$3();
403:   globalThis.Object.freeze(class Cont$handleBlock$h$2 extends runtime.FunctionContFrame.class {
404:     static {
405:       Cont$handleBlock$h$3 = this
406:     }
407:     constructor(pc) {
408:       let tmp1;
409:       tmp1 = super(null);
410:       this.pc = pc;
411:     }
412:     resume(value$) {
413:       if (this.pc === 5) {
414:         res = value$;
415:       }
416:       contLoop: while (true) {
417:         if (this.pc === 5) {
418:           return res
419:         }
420:         break;
421:       }
422:     }
423:     toString() { return runtime.render(this); }
424:     static [definitionMetadata] = ["class", "Cont$handleBlock$h$"]; 
425:   });
426:   doUnwind = function doUnwind(res1, pc) {
427:     res1.contTrace.last.next = new Cont$handleBlock$h$3(pc);
428:     return runtime.handleBlockImpl(res1, h)
429:   };
430:   res = foo(h);
431:   if (res instanceof runtime.EffectSig.class) {
432:     return doUnwind(res, 5)
433:   }
434:   return res
435: };
436: tmp = handleBlock$();
437: if (tmp instanceof runtime.EffectSig.class) { tmp = runtime.topLevelEffect(tmp, false); }
438: tmp
```

</details>



What I currently think to be a reasonable fix is:
- For `if`s, no nested Scoped blocks should be generated at all, even for nested `if`s (previously nested scoped blocks were generated for nested `if`s (not branches) because I thought it comes for free when implementing `Scoped` bodies for `while`s, but I didn't realize this intricate interaction with handler lowering).
- For `while`s, nested Scoped blocks should still be generated:
  - when handler lowering is off (which means that loops are not rewritten to functions by default), this is crucial for the correct semantics
  - when handler lowering is on (which means loops are always rewritten to functions), how the "while bodies" contain nested `Scoped` blocks matches with the top level scopes of while functions' bodies.
- No `floatOutUntilScoped`

LPTK: Ok, this is a really unsatisfying kludge, but let's just use that approach for now. We'll simplify it later when the new handler lowering is merged.

hkmc2DiffTests/src/test/scala/hkmc2/JSBackendDiffMaker.scala
ychenfo: Fixed in https://github.com/hkust-taco/mlscript/pull/356/commits/08d4467433684ced96ead7987e15e50665d4305c

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
ychenfo: Thank I see. I will remember to remove this after this PR is no longer a draft.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
ychenfo: Fixed in https://github.com/hkust-taco/mlscript/pull/356/commits/08d4467433684ced96ead7987e15e50665d4305c

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
ychenfo: Fixed in https://github.com/hkust-taco/mlscript/pull/356/commits/1ac0ae9c17125e03ae345aaba5dc4cc586fc8e7c with tests and more updated comments on https://github.com/hkust-taco/mlscript/pull/356/commits/63fe1c41e2f509c99ccb9560bcd109117e145bfa

hkmc2DiffTests/src/test/scala/hkmc2/MLsDiffMaker.scala
ychenfo: Fixed in https://github.com/hkust-taco/mlscript/pull/356/commits/43fdc78b728c47211c2eaa437970ff8bf53aefac

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
ychenfo: Fixed in https://github.com/hkust-taco/mlscript/pull/356/commits/a64ac1f6240e4f66db3616ce0ab7a6c172d12db8 and https://github.com/hkust-taco/mlscript/pull/356/commits/ab74823c15aad6122a5491e296c56fe37e68ce0b

hkmc2/shared/src/test/mlscript/codegen/ScopedBlocks.mls
```diff
@@ -377,66 +377,3 @@ module M with
...
-fun f(x) =
-  while x do
-    while x do
-      let z = 4
-      handle h = Effect with
-        fun perform(x)(k) =
-          print(x)
-          k(x + z)
-      print(h.perform(32))
-  else 0
```
LPTK: Why remove all these tests? Are they duplicated?

LPTK: Did you just essentially move them to [ScopedBlocksAndHandlers.mls](https://github.com/hkust-taco/mlscript/pull/356/files/e060b40e420bbf0ffe2660d03875153c3f4a8bb9..30f3850d728a12621d4b214278b9202a58b6374c#diff-e1b7825f28c115f66e836c47fb2aa4204424d292ab190649ff223e1591be2cef)? In which case, it's fine.

hkmc2/shared/src/test/mlscript/codegen/ScopedBlocksAndHandlers.mls
```diff
@@ -0,0 +1,263 @@
...
+
+
+
+abstract class Effect with
+  fun perform()
+
+
+
+:effectHandlers
+:sjs
```
LPTK: I think this is too much output code.

LPTK:
```suggestion
```

hkmc2DiffTests/src/test/scala/hkmc2/JSBackendDiffMaker.scala
LPTK: Please just mark the conversation as resolved yourself when the fix is done and obvious, and no need to comment.

hkmc2DiffTests/src/test/scala/hkmc2/MLsDiffMaker.scala
LPTK: It looks like the command is never tested.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
LPTK: Why not?

LPTK: I guess you're asking about supporting `let` statements in argument/expression position.

We can do this later. For now, just use parens.

```fs
scope.locally of (
  let x = 42
  x
)
```

LPTK: (Or `let`/`in`.)

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -151,9 +159,11 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
             blockImpl(stats, L((mut, RcdArg(S(l), r) :: flds)))(k)
     case (decl @ LetDecl(sym, annotations)) :: (stats @ ((_: DefineVar) :: _)) =>
       reportAnnotations(decl, annotations)
+      if sym.asTrm.fold(true)(_.owner.isEmpty) then loweringCtx.collectScopedSym(sym)
```
```suggestion
      if sym.asTrm.forall(_.owner.isEmpty) then loweringCtx.collectScopedSym(sym)
```
Same below.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -1104,42 +1156,55 @@ trait LoweringTraceLog(instrument: Bool)(using TL, Raise, State)
...
+      val retMsgSym = TempSymbol(N, dbgNme = "traceLogRetMsg")
+      val psInspectedSyms = params.params.map(p => TempSymbol(N, dbgNme = s"traceLogParam_${p.sym.nme}") -> p.sym)
+      val resInspectedSym = TempSymbol(N, dbgNme = "traceLogResInspected")
+      
+      loweringCtx.collectScopedSym(enterMsgSym)
+      loweringCtx.collectScopedSym(prevIndentLvlSym)
+      loweringCtx.collectScopedSym(resSym)
+      loweringCtx.collectScopedSym(retMsgSym)
+      loweringCtx.collectScopedSym(resInspectedSym)
+      for (s, _) <- psInspectedSyms do loweringCtx.collectScopedSym(s)
```
LPTK: I would prefer that you add an overload called `collectScopedSyms` that does have a vararg, for occasional uses.

hkmc2DiffTests/src/test/scala/hkmc2/MLsDiffMaker.scala
ychenfo: Tests added [here](https://github.com/hkust-taco/mlscript/pull/356/commits/70ed84a84cd079a3a2ad9937cf91818d3eff45d3#diff-b2a55c83b18544e26b05f237542cc7ad4dc9c4850233abecd8ebf00f09d2b202R380-R427)

LPTK: https://github.com/hkust-taco/mlscript/pull/356#discussion_r2591266178

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -29,6 +29,30 @@ sealed abstract class Block extends Product:
     case _: End => true
     case _ => false
   
+  // this variation of `definedVars` excludes the `syms` in `Scoped` blocks
+  // so that `Scoped` blocks' vars can be seen in the output js code;
+  // otherwise `blockPreamble` allocates all the `definedVars` such that
+  // we cannot see and test vars generated because of the existence of the `Scoped` blocks
```
LPTK: This sentence needs to be rephrase as it's currently incomprehensible.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -530,6 +550,32 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
               t.toLoc :: Nil,
               source = Diagnostic.Source.Compilation)
         conclude(Value.Ref(ctx.builtins.debug.getLocals, N).withLocOf(f))
+      case t if instantiatedResolvedBms.exists(_ is ctx.builtins.scope.locally) =>
+        arg match
+          case Tup(flds) =>
```
LPTK: What's the point of supporting multiple arguments in this way? This complicated logic looks completely unnecessary. Just suppport exactly one argument.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -572,34 +591,54 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
...
   def worksheet(p: Program)(using Raise, Scope): (Document, Document) =
     reserveNames(p)
     lazy val imps = p.imports.map: i =>
       doc"""${getVar(i._1, N)} = await import("${i._2.toString}").then(m => m.default ?? m);"""
-    blockPreamble(p.imports.map(_._1).toSeq ++ p.main.definedVars.toSeq) ->
+    p.main match
+    case Scoped(syms, body) =>
+      blockPreamble(p.imports.map(_._1).toSeq ++ syms.toSeq) ->
+      (imps.mkDocument(doc" # ") :/: block(body, endSemi = false).stripBreaks)
+    case _ =>
```
LPTK: Does it actually happen that we get a non-empty `definedVarsNoScoped` but there is no scope surrounding it? Sounds like if it does, it should be fixed. But we could fix it later (then, add a comment).

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -572,34 +591,54 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
...
+  def genLetDecls(vars: Iterator[(Symbol, Str)], isScoped: Bool): Document =
     if vars.isEmpty then doc"" else
       doc" # let " :: vars.map: (_, nme) =>
         nme
       .toList.mkDocument(", ")
       :: doc";"
   
+  def blockPreamble(ss: Iterable[Symbol])(using Raise, Scope): Document =
+    // TODO document: mutable var assnts require the lookup
+    val vars = ss.filter(scope.lookup(_).isEmpty).toArray.sortBy(_.uid).iterator.map(l =>
```
LPTK, line 662: With this PR, the ugly `lookup` should no longer be needed. I saw that removing it currently crashes the compilation tests. Any idea why?

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -572,34 +591,54 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
...
+      l -> scope.allocateName(l))
+    genLetDecls(vars, false)
+
+  // Only handle non-nested Scoped nodes: we output the bindings, but do not add another brace
+  def nonNestedScoped(blk: Block)(k: Block => Document)(using Raise, Scope): Document = blk match
+    case Scoped(syms, body) => 
+      val vars = syms.filter(scope.lookup(_).isEmpty).toArray.sortBy(_.uid).iterator.map(l =>
```
LPTK: We should call `blockPreamble` here, instead of duplicating the code.

LPTK (changes requested):
Ok, I think this is almost ready to merge. Thanks for your sustained work through this complicated change!

The nice thing is that this change also lets us avoid the needless assignment to `undefined` when writing `let x` without an initializer.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
LPTK: (Note: the comment is _not_ outdated.)

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
NeilKleistGao: I also removed an unused parameter.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
NeilKleistGao: Both `let`/`in` and the parentheses form are supported now.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
LPTK: (Note: the comment is also _not_ outdated.)

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
NeilKleistGao: I think this is mainly because in `blockk` function, we have `val pre = blockPreamble(t.definedVarsNoScoped)`. Then `t.definedVarsNoScoped` will also cover symbols that have been collected by the outer `Scoped` object. Without the `filtering` will fail the assertion when we try to duplicate those symbols.

LPTK: Why do we need to use `t.definedVarsNoScoped`, here? We should _already_ know what is supposed to be in scope, thanks to `Scoped`.

Do you have a concrete example of this?

NeilKleistGao: > Why do we need to use `t.definedVarsNoScoped`, here? We should _already_ know what is supposed to be in scope, thanks to `Scoped`.
> 
> Do you have a concrete example of this?

One example is the `let done() = ...` in `mlscript-compile/Render.mls`. If we do not call `blockPreamble` to allocate names, then this `done` will not be found later. I guess it is related to lambda lifting. Should we leave a comment here and handle it in another PR?

LPTK: > Should we leave a comment here and handle it in another PR?

Ok, let's do that.

Lambda lifting should be updated to deal with this later.

NeilKleistGao: I've added a comment to track it.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
NeilKleistGao: This is also related to lambda lifting. Some generated variables, like `lambda`, can only be handled by `definedVarsNoScoped`. I've left a comment as well.

hkmc2/shared/src/test/mlscript/codegen/RandomStuff.mls
```diff
@@ -68,7 +68,7 @@ do
 //│ JS (unsanitized):
 //│ let f, f1;
 //│ f1 = 1;
-//│ f = function f() { loopLabel: while (true) { continue loopLabel; break; } };
+//│ f = function f() { loopLabel: while (true) { let tmp; continue loopLabel; break; } };
```
LPTK, line 17: Where does this useless `tmp` come from? Why is it collected in the first place?

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -29,10 +29,15 @@ sealed abstract class Block extends Product:
...
     case _ => false
   
-  /** This variation of `definedVars` excludes the `syms` in `Scoped` blocks
-    * so that `Scoped` blocks' vars can be seen in the output js code;
-    * otherwise `blockPreamble` allocates all the `definedVars` such that
-    * we cannot see and test vars generated because of the existence of the `Scoped` blocks */
+  /** This variation of `definedVars` excludes the `syms` in `Scoped` blocks.
+    * It is used in JSBuilder now: names are allocated in JSBuilder for these `definedVarsNoScoped` symbols.
+    * This is needed now because
+    *   - there are symbols that are not collected in the `Scoped` blocks (due to later passes),
```
```suggestion
    *   - there are symbols that are not collected in the `Scoped` blocks (due to some passes that have not yet been adapted to Scoped),
```

hkmc2/shared/src/test/mlscript/codegen/RandomStuff.mls
ychenfo: With `:noTailRec`, `fun f = f` is compiled to the following program, where `tmp` is the result of application.
```
:noTailRec
:sjs
fun f = f
//│ JS (unsanitized):
//│ let f; f = function f() { let tmp; tmp = f(); return tmp };
```

`tmp` is created and collected [here](https://github.com/NeilKleistGao/mlscript/blob/899ee24c95901738aaa106f4f4e5b06be2c0c1bf/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala#L447-L448).

LPTK: Ok, fair enough. This will be better dealt with through a pass of dead-binding elimination.

hkmc2/shared/src/test/mlscript/codegen/Formatting.mls
```diff
@@ -7,7 +7,10 @@
 
 let discard = drop _
 //│ JS (unsanitized):
-//│ let discard, discard1; discard1 = function discard(_0) { return runtime.Unit }; discard = discard1;
+//│ let discard;
```
LPTK: It looks like the `discard` variable is not properly collected.

hkmc2/shared/src/test/mlscript/codegen/Getters.mls
```diff
@@ -174,7 +174,8 @@ fun baz() =
 //│ JS (unsanitized):
 //│ let baz;
 //│ baz = function baz() {
-//│   let w, z, lambda;
+//│   let w, z;
+//│   let lambda;
```
LPTK: Why is this `lambda` bound separately? Is it not properly collected?

hkmc2/shared/src/test/mlscript/codegen/Hygiene.mls
```diff
@@ -60,7 +60,7 @@ let f = () => x
 let x = 2
 f()
 //│ JS (unsanitized):
-//│ let x, f, x1, f1; x = 1; f1 = function f() { return x }; f = f1; x1 = 2; runtime.safeCall(f())
+//│ let x, f, x1; let f1; x = 1; f1 = function f() { return x }; f = f1; x1 = 2; runtime.safeCall(f())
```
LPTK: Why is `f2` bound separately? Is it not properly collected?

hkmc2/shared/src/test/mlscript/codegen/NestedScoped.mls
LPTK, line 1: There is too much `:sjs` output in this file. Please only keep one or two representative cases.

hkmc2/shared/src/test/mlscript/codegen/Quasiquotes.mls
```diff
@@ -76,12 +76,13 @@ x `=>
   print(x)
   x
 //│ JS (unsanitized):
-//│ let x1, tmp6, tmp7, tmp8, arr2;
-//│ tmp6 = globalThis.Object.freeze(new Term.Symbol("x"));
-//│ x1 = globalThis.Object.freeze(new Term.Ref(tmp6));
-//│ tmp7 = Predef.print(x1);
+//│ let tmp6;
```
LPTK, line 79: Why is this bound separately? Is it not properly collected?

hkmc2/shared/src/test/mlscript/codegen/ScopedBlocks.mls
LPTK, line 1: There is too much `:sjs` output in this file. Please only keep one or two representative cases and those where the comments specifically talks about the generated code.

hkmc2/shared/src/test/mlscript/ctx/MissingDefinitions2.mls
```diff
@@ -16,66 +16,38 @@ x
 
 val x
 
-:ge
+// with Scoped blocks, `x` is declared
```
LPTK: This comment will be impossible to understand for someone else. Pls rephrase it.

hkmc2/shared/src/test/mlscript/ctx/MissingDefinitions2.mls
```diff
@@ -16,66 +16,38 @@ x
...
+// :ge
```
LPTK: Why would we keep this commented line?

hkmc2/shared/src/test/mlscript/ctx/MissingDefinitions2.mls
LPTK: Same remark for the blocks below.

hkmc2/shared/src/test/mlscript/ctx/MissingDefinitions2.mls
LPTK: Same remark for the blocks below.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -435,12 +445,13 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
         // * (non-local functions are compiled into getter methods selected on some prefix)
         if td.params.isEmpty then
           val l = new TempSymbol(S(ref))
+          loweringCtx.collectScopedSym(l)
```
LPTK: The `new TempSymbol` + `collectScopedSym` pattern is all over the place. This is a time where you introduce a helper function (call it `registerTempSymbol`) to avoid the needless repetition. Keep it DRY!

LPTK (approved):
Ok, I think this PR is now almost ready to be merged. Nice work on this major change of the IR!

But please address the comments I left first.

hkmc2/shared/src/test/mlscript/codegen/Formatting.mls
ychenfo: One of the `discard` symbols is created by lambda rewriter, so it's indeed not collected now.

hkmc2/shared/src/test/mlscript/codegen/Getters.mls
ychenfo: This `lambda` is also created by the lambda rewriter and not collected now.

hkmc2/shared/src/test/mlscript/codegen/Hygiene.mls
ychenfo: This `f1` is also created by the lambda rewriter and not collected now.

Actually do we still want to generate `Result.Lambda` at all during lowering (since lambda rewriter is always on)? If we don't do so all these un-collected symbols related to lambdas should be gone.

hkmc2/shared/src/test/mlscript/codegen/Quasiquotes.mls
NeilKleistGao: This is because we have not handled temporary variables in quotations. I think I will do it in the next PR (this PR is already huge and complicated).

hkmc2/shared/src/test/mlscript/codegen/Hygiene.mls
LPTK: Clearly, either the "lambda rewriter" should be fixed (which should be very easy; it's distinct from the lifter) or we should just lower directly to a function on the fly and remove lambda rewriting. I am not sure why we currently still have lambdas in values. I think it's a historical artifact.

hkmc2/shared/src/test/mlscript/codegen/Quasiquotes.mls
LPTK: Ok, but FYI this PR is not really huge 👀

hkmc2/shared/src/test/mlscript/codegen/Hygiene.mls
LPTK: Ah, I just noticed `Lambda` is already no longer a value. I guess it also does not need to be a Result...

ychenfo: Oh, so do we want to totally remove the `Lambda` form (since `Lambda`s don't need to be either `Value`s or `Result`s)...?

I guess in this PR it would be easier to only add fixes in `LambdaRewriter.scala`.

LPTK: Ok, sure. Pls do it.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -391,17 +394,26 @@ object Match:
   def apply(scrut: Path, arms: Ls[Case -> Block], dflt: Opt[Block], rest: Block): Block = dflt match
     case S(Match(`scrut`, arms2, dflt2, _: End)) => // TODO: also handle non-End rest (may require a join point)
       // * Currently, this branch does not seem used, because the UCS already does a good job at merging matches
-      Match(scrut, arms ::: arms2, dflt2, rest)
```
LPTK, line 393: I don't think you need to change this line. It's already using the smart constructor. Why duplicate the code?

hkmc2/shared/src/test/mlscript/tailrec/TailRecOpt.mls
```diff
@@ -20,39 +20,35 @@ f(10000, 20000, 0)
...
-//│           id = 0;
-//│           continue loopLabel
-//│         } else {
-//│           scrut1 = param1 > 0;
-//│           if (scrut1 === true) {
-//│             tmp2 = param1 - 1;
-//│             param1 = tmp2;
-//│             param3 = 2;
+//│     switch (id) {
+//│       case 0:{
```
LPTK: There probably shouldn't be braces, here.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
NeilKleistGao: But we would get a duplicate definition of `object Match` then.

LPTK: OMG GitHub is messing up again...

Here is what I commented on, as viewed from the diff of 3b8ccab:

Screenshot 2025-12-12 at 22 38 49

NeilKleistGao: Right. Already reverted.

hkmc2/shared/src/test/mlscript/codegen/SanityChecks.mls
```diff
@@ -69,19 +69,14 @@ id(f)(3)(4)
...
-//│ undefined
+//│ f4 = function f(x, y) { return x + y }; f3 = f4; block$res5 = runtime.safeCall(f3(2)); undefined
 //│ = NaN
 
 :ssjs
 :re
 let f = (x, y) => x + y
 f(2)
 //│ JS:
-//│ let f6;
```
LPTK: Whoops, is it expected that `f6` is no longer declared, here?

LPTK: This might be related to the `ssjs` above.

hkmc2/shared/src/test/mlscript/codegen/ScopedBlocksAndHandlers.mls
```diff
@@ -98,3 +98,30 @@ fun start(x, f) =
     let z = 2
     fun g() = z
     g()
+
+
+
+// NOTE: `z` should be in the function top level when `:effectHandlers`
```
LPTK:
```suggestion
// NOTE: `z` should be in the function top level when `:effectHandlers` to be properly captured by the floated-out `g`.
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -366,7 +366,10 @@ object Label:
 object Scoped:
   def apply(syms: collection.Set[Local], body: Block): Block = body match
     case Scoped(syms2, body) =>
-      if syms2.isEmpty && syms.isEmpty then Scoped(Set.empty, body) else Scoped(syms ++ syms2, body)
+      if syms2.isEmpty && syms.isEmpty then Scoped(Set.empty, body)
+      else
+        assert(!syms2.exists(syms.contains), "overlapping symbols in nested Scoped")
```
LPTK: This looks uncomfortably expensive. Let's add a `validateIR: Bool` option in `Config` and only do the check when it's on.

LPTK: (Same below.)

hkmc2/shared/src/test/mlscript/codegen/SanityChecks.mls
ychenfo: Thanks for the catch! Indeed it's due to `ssjs` not outputing `preStr` (returned by `JSBuilder.worksheet`) where `f6` is properly declared now because it is in the top level `Scoped` block after `LambdaRewriter` is fixed to maintain `Scoped` blocks. I have pushed the change to also output non-empty `preStr`.

hkmc2/shared/src/main/scala/hkmc2/codegen/LambdaRewriter.scala
```diff
@@ -22,28 +22,48 @@ object LambdaRewriter:
...
         var lambdasList: List[((BlockMemberSymbol, TermSymbol), Lambda)] = Nil
+        var newLambdaSyms: List[BlockMemberSymbol] = Nil
         val lambdaRewriter = new BlockDataTransformer(SymbolSubst()):
           override def applyResult(r: Result)(k: Result => Block): Block = r match
             case lam: Lambda => 
               val sym = BlockMemberSymbol("lambda", Nil, nameIsMeaningful = false)
               val tSym = TermSymbol.fromFunBms(sym, N)
               lambdasList ::= ((sym, tSym) -> super.applyLam(lam))
+              newLambdaSyms ::= sym
               k(Value.Ref(sym, S(tSym)))
```
LPTK: Is there a reason we have to keep track of `lambdasList` and `newLambdaSyms` here instead of just inserting a `Define` here around the `k`?

ychenfo: Indeed inserting a `Define` around the `k` works the same and we don't need to collect `lambdasList`. I guess previously it was written this way because `applyResult` did not always take a `k: Result => Block`.

I think `newLambdaSyms` is still needed to be used in the `transformer` later to better respect the current `Scoped` blocks, otherwise new Scoped blocks may be created after lambda rewriter (e.g., in match branch bodies, where they are not merged).

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
ychenfo: Using [`whenValidatingIR`](https://github.com/hkust-taco/mlscript/pull/356/commits/50ba7270458972eb9b31f407ec7eaf09fcf4d50c) for now because adding `(using Config)` requires too many changes

hkmc2/shared/src/main/scala/hkmc2/codegen/LambdaRewriter.scala
LPTK: > I think `newLambdaSyms` is still needed to be used in the `transformer` later to better respect the current `Scoped` blocks, otherwise new Scoped blocks may be created after lambda rewriter (e.g., in match branch bodies, where they are not merged).

Do you have a concrete example of this in mind? I'm not sure what you mean.

ychenfo: Sure, here is an example (though I guess it's all harmless for lambda rewriter to create new nested `Scoped` blocks containing newly created lambda symbols in places where there normally won't be a nested Scoped blocks now (except for manually nested ones))

When we are not collecting `newLambdaSyms` in `rewriteOneBlk` and inserting them to the existing Scoped blocks in the main `transformer = new BlockTransformer(SymbolSubst())` below, but instead generating `Define` and `Scoped` blocks in `rewriteOneBlk`, we may get the following programs

<details>

<summary> program containing newly created nested Scoped (at line 295) after lambda rewriter </summary>

```
263: :lot
264: fun f(x) =
265:   if x then
266:     () => 4
267: //│ Lowered:
268: //│ Program:
269: //│   imports = Nil
270: //│   main = Scoped:
271: //│     syms = HashSet(member:f)
272: //│     body = Define:
273: //│       defn = FunDefn:
274: //│         owner = N
275: //│         sym = member:f
276: //│         dSym = term:f
277: //│         params = Ls of 
278: //│           ParamList:
279: //│             flags = ()
280: //│             params = Ls of 
281: //│               Param:
282: //│                 flags = ()
283: //│                 sym = x
284: //│                 sign = N
285: //│                 modulefulness = Modulefulness of N
286: //│             restParam = N
287: //│         body = Match:
288: //│           scrut = Ref:
289: //│             l = x
290: //│             disamb = N
291: //│           arms = Ls of 
292: //│             Tuple2:
293: //│               _1 = Lit of BoolLit of true
294: //│               _2 = Scoped:
295: //│                 syms = Set(member:lambda)
296: //│                 body = Define:
297: //│                   defn = FunDefn:
298: //│                     owner = N
299: //│                     sym = member:lambda
300: //│                     dSym = term:lambda
301: //│                     params = Ls of 
302: //│                       ParamList:
303: //│                         flags = ()
304: //│                         params = Nil
305: //│                         restParam = N
306: //│                     body = Return:
307: //│                       res = Lit of IntLit of 4
308: //│                       implct = false
309: //│                   rest = Return: \
310: //│                   res = Ref{disamb=term:lambda}:
311: //│                     l = member:lambda
312: //│                     disamb = S of term:lambda
313: //│                   implct = false
314: //│           dflt = S of Throw of Instantiate:
315: //│             mut = false
316: //│             cls = Select:
317: //│               qual = Ref{disamb=globalThis:globalThis}:
318: //│                 l = globalThis:globalThis
319: //│                 disamb = S of globalThis:globalThis
320: //│               name = Ident of "Error"
321: //│             args = Ls of 
322: //│               Arg:
323: //│                 spread = N
324: //│                 value = Lit of StrLit of "match error"
325: //│           rest = End of ""
326: //│       rest = End of ""
327: 
```

</details>



<details>

<summary> another shorter one (new nested Scoped at line 269) </summary>

```
260: :lot
261: let x = 3
262: () => x
263: //│ Lowered:
264: //│ Program:
265: //│   imports = Nil
266: //│   main = Scoped:
267: //│     syms = HashSet(x)
268: //│     body = Scoped:
269: //│       syms = Set(member:lambda)
270: //│       body = Assign:
271: //│         lhs = x
272: //│         rhs = Lit of IntLit of 3
273: //│         rest = Define: \
274: //│         defn = FunDefn:
275: //│           owner = N
276: //│           sym = member:lambda
277: //│           dSym = term:lambda
278: //│           params = Ls of 
279: //│             ParamList:
280: //│               flags = ()
281: //│               params = Nil
282: //│               restParam = N
283: //│           body = Return:
284: //│             res = Ref:
285: //│               l = x
286: //│               disamb = N
287: //│             implct = false
288: //│         rest = Return: \
289: //│         res = Ref{disamb=term:lambda}:
290: //│           l = member:lambda
291: //│           disamb = S of term:lambda
292: //│         implct = true
```

</details>

LPTK: Ok, but that seems perfectly fine to me. What's the problem? That's like the whole point of having scoped blocks (eemember the inlining use case – this looks pretty much similar).

Also, I wonder why the Scoped at lines 266 and 268 are not merged by the smart constructor when putting things back together upon transformation. Any ideas?

ychenfo: Indeed it should be no problem, I was doing this just to try to better maintain the `Scoped` blocks in the original program.

Scoped at line 266 and 268 is not merged because I changed the block transformer to use `new Scoped` when transforming existing Scoped blocks, it prevents destroying some manually nested Scoped blocks.

But indeed iirc we concluded that it's fine to destroy manually nested scoped blocks... I will make the relevant changes.

LPTK: Ok, please do, and simplify the lambda lifter. In my view, the current extra complexity of trying to "better maintain" Scoped blocks (for which I don't fully understand the rationale) absolutely does not pay for any corresponding observed benefits, unless I am missing something.

ychenfo: Sure, I just pushed the simplified lambda rewriter, and this simplification indeed unintentionally (in a good way) fixed a bug that the some symbols of newly created functions are not properly put in the top level scoped blocks (https://github.com/hkust-taco/mlscript/pull/356/commits/31348be97d1761004c37403e6b801bfae36747a8#diff-fe0afad08c6cbb3ce975faaf3b603a3b34411c5cbca5a226250dca4fe29eb310).

LPTK (approved):
Great work, thanks @ychenfo and @NeilKleistGao!

I made some final tweaks to the PR, and now I think it's finally ready to merge.

PR #358, opened 2025-12-01
Tail Recursion Optimization

CAG2Mark:
Implements tail recursion optimization. Also changed the lifter, handler lowering, and while loop rewriter to properly propagate functions' TermSymbols into calls.

Note that the lifter has been modified so only disambiguated references to BMS's are rewritten. This should not be a problem as all references to nested functions should be disambiguated. However, it may cause errors if passes before the lifter forget to propagate the disamb symbols.

Things that are not yet done:
- Spread arguments in calls. Currently only supported when the spread arg is in the last position and the last argument of the target function is a spread.
- Class methods are not yet optimized as they cannot yet be marked final.
- In objects and modules, calls to functions with no parameter list are represented using Selections


hkmc2/shared/src/main/scala/hkmc2/Config.scala
```diff
@@ -40,6 +41,7 @@ object Config:
     target = CompilationTarget.JS,
     rewriteWhileLoops = true,
     stageCode = false,
+    tailRecOpt = false,
```
LPTK: Any reason it's disabled by default? It should arguably be on by default.

hkmc2/shared/src/test/mlscript/HkScratch.mls
```diff
@@ -9,4 +9,5 @@
 // :todo
 
 
-
+@tailrec
```
LPTK: Don't forget to clear this file.

hkmc2/shared/src/test/mlscript/tailrec/TailRecOpt.mls
```diff
@@ -0,0 +1,259 @@
...
+//│ };
+//│ g = function g(a, b, c, d) {
+//│   return g_f(0, a, b, c, d)
+//│ };
+//│ g_f = function g_f(id, param0, param1, param2, param3) {
+//│   let scrut, scrut1, tmp, tmp1, tmp2;
+//│   loopLabel: while (true) {
+//│     if (id === 0) {
+//│       tmp = param2 + param3;
+//│       param0 = param0;
```
LPTK: Can we avoid these strange assignments? 👀

hkmc2/shared/src/test/mlscript/tailrec/TailRecOpt.mls
```diff
@@ -0,0 +1,259 @@
...
+while i < n do
+  x.push of () => i
+  set i += 1
+fun sumOf(x, idx, acc) =
+  if idx == 0 then acc
+  else 
+    sumOf(x, idx - 1, acc + x![idx]())
+sumOf(x, n - 1, 0)
+
+// check that spreads. where supported, are compiled correctly
```
LPTK:
```suggestion
// Check that spreads, where supported, are compiled correctly
```

hkmc2/shared/src/test/mlscript/tailrec/TailRecOpt.mls
```diff
@@ -0,0 +1,259 @@
...
+:sjs
```
LPTK: I think we can dispense with this one.

hkmc2/shared/src/test/mlscript/tailrec/TailRecOpt.mls
```diff
@@ -0,0 +1,259 @@
...
+//│         continue loopLabel
+//│       }
+//│     }
+//│     break;
+//│   }
+//│ };
+//│ f1(100)
+//│ = 0
+
+:todo
```
LPTK: `:todo` is a bit too loose (it will not complain if the error goes away and thus become obsolete).
```suggestion
:fixme
```

LPTK: (Same comment below.)

hkmc2/shared/src/test/mlscript/tailrec/TailRecOpt.mls
```diff
@@ -0,0 +1,259 @@
...
+    @tailcall g(x)
+//│ ╔══[ERROR] This tail call exits the current scope is not optimized.
+//│ ║  l.219: 	      @tailcall f(x)
+//│ ╙──       	                ^^^
+//│ ╔══[ERROR] This tail call exits the current scope is not optimized.
+//│ ║  l.220: 	    @tailcall g(x)
+//│ ╙──       	              ^^^^
+
+// These calls are represented as field selections and don't yet have the explicitTailCall parameter.
+:todo
```
LPTK:
```suggestion
:breakme // TODO
```

LPTK: Is the current absence of error actually known/expected, here?

hkmc2/shared/src/test/mlscript/lifter/EffectHandlers.mls
LPTK, line 1: Just curious: is this change related to tail recursion? (It's fine if not.)

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
```diff
@@ -0,0 +1,315 @@
...
+      case _ => N
+  
+  object TailCallShape:
+    def unapply(b: Block): Opt[(TermSymbol, Call)] = b match
+      case Return(c @ CallToFun(r), _) => S((r, c))
+      case Assign(a, c @ CallToFun(r), Return(b, _)) if a == b => S((r, c))
+      case _ => N
+    
+  
+  sealed abstract class CallEdge:
```
LPTK: Maybe we can make this an `enum`?

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
LPTK, line 1: Please use `===` and `=/=` rather than `==` and `!=`.

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
```diff
@@ -0,0 +1,315 @@
...
+  def buildCallGraph(fs: List[FunDefn]): List[CallEdge] =
+    fs.flatMap(f => CallFinder(f).find)
+  
+  case class SccOfCalls(funs: List[FunDefn], calls: List[CallEdge])
+  
+  def partFns(fs: List[FunDefn]): List[SccOfCalls] =
+    val defnSyms = fs.map(_.dSym)
+    val tsToDefn = fs.map(f => f.dSym -> f).toMap
+    
+    // Only care about calls to functions in the same scope
```
LPTK, line 81: This is a very significant limitation, which we should make sure to eventually lift. We do want to be able to compile tail recursion across modules.

LPTK (commented):
Awesome! I'm very excited to have this in, as it's a first step towars our eventual goal of KICE (Know In-order Call Elimination).

hkmc2/shared/src/test/mlscript/tailrec/TailRecOpt.mls
CAG2Mark: I'm not sure what you mean

hkmc2/shared/src/test/mlscript/lifter/EffectHandlers.mls
CAG2Mark: It's not related. It's for making sure the definition symbols are properly propagated in the effect handler pass, since the lifter now requires it.

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
CAG2Mark: This could be solved for class/module methods if they were lifted into normal function definitions which take an extra `this` parameter, and the methods were simply replaced with calls to those function definitions.

hkmc2/shared/src/test/mlscript/tailrec/TailRecOpt.mls
CAG2Mark: Yes, I didn't know about the `:breakme` flag.

hkmc2/shared/src/main/scala/hkmc2/Config.scala
CAG2Mark: True

CAG2Mark:
When did we establish this `KICE` name? :thinking:

hkmc2/shared/src/main/scala/hkmc2/Config.scala
LPTK: Let's give it a shot!

hkmc2/shared/src/test/mlscript/tailrec/TailRecOpt.mls
LPTK: "dispense with" = "do without"
"this one" = `:sjs`

hkmc2/shared/src/test/mlscript/lifter/EffectHandlers.mls
LPTK: Ok, but don't we already have basically similar tests that would catch that?

LPTK:
> When did we establish this `KICE` name? 🤔

Just now 😎

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
LPTK: Why would module methods need to be lifted? They seem to work fine as is. As in:
```fs
:tailrec
module A with
  fun f(x) = @tailcall f(f(x) + 1)
```

CAG2Mark: This example does not work
```fs
module A with
  fun f(x) =
    fun g(x) = @tailcall f(x)
    @tailcall g(x)
```
because `g(x)` is lifted out of the module itself.

hkmc2/shared/src/test/mlscript/lifter/EffectHandlers.mls
CAG2Mark: Those are the lifter stack safety tests that happen to use effect handlers.

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
LPTK: 1. Why is `f` lifted out of the module? Maybe that's not necessary.
2. This is an important limitation that should be better documented and tested against (it's enough to have a :fixme test with a stack overflow to track this)

CAG2Mark: `f` is not lifted out, but `g` is. It's the way the lifter currently works, as it makes the code much simpler.

CAG2Mark:
OK, turns out I didn't recompile some of the other stuff. Please wait

EDIT: Should be good now

hkmc2/shared/src/test/mlscript/codegen/TraceLog.mls
```diff
@@ -10,10 +10,19 @@ fun fib(a) = if
 //│ let fib;
 //│ fib = function fib(a) {
 //│   let scrut, tmp, tmp1, tmp2, tmp3;
-//│   scrut = a <= 1;
-//│   if (scrut === true) {
-//│     return a
-//│   } else { tmp = a - 1; tmp1 = fib(tmp); tmp2 = a - 2; tmp3 = fib(tmp2); return tmp1 + tmp3 }
+//│   loopLabel: while (true) {
```
LPTK: None of these recursive calls are in tail position, so I don't think the function should be rewritten.

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
```diff
@@ -319,10 +319,14 @@ class TailRecOpt(using State, TL, Raise):
     else (S(loopDefn), rewrittenFuns)
   
   def optFunctions(fs: List[FunDefn], owner: Opt[InnerSymbol]) =
-    partFns(fs).map(optScc(_, owner)).foldLeft[(List[FunDefn], List[FunDefn])](Nil, Nil):
+    val (newFsOpt, fsOpt) = partFns(fs).map(optScc(_, owner)).foldLeft[(List[FunDefn], List[FunDefn])](Nil, Nil):
       case ((newFns, fns), (newFnOpt, fns_)) => newFnOpt match
         case Some(value) => (value :: newFns, fns_ ::: fns)
         case None => (newFns, fns_ ::: fns)
+    // preserve the order of function defns
```
LPTK, line 328: I wonder why the diff shows that the order of functions is actually changed (eg see `Predef.mjs`).

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
```diff
@@ -0,0 +1,392 @@
...
+      c.copy(methods = mtds, companion = companion)
+  
+  def transform(b: Block) =
+    val (blk, defns) = b.floatOutDefns()
+    val (funs, clses) = 
+      defns.foldLeft[(List[FunDefn], List[ClsLikeDefn])](Nil, Nil):
+        case ((fs, cs), d) => d match
+          case f: FunDefn => (f :: fs, cs)
+          case c: ClsLikeDefn => (fs, c :: cs)
+          case _ => (fs, cs) // unreachable as floatOutDefns only floats out FunDefns and ClsLikeDefns
```
LPTK: Instead of this awkward `foldLeft`, why not use `partitionMap`?

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -265,7 +265,9 @@ class TermSymbol(val k: TermDefKind, val owner: Opt[InnerSymbol], val id: Tree.I
   override def toString: Str = s"term:${owner.map(o => s"${o}.").getOrElse("")}${id.name}${State.dbgUid(uid)}"
   
   def subst(using sub: SymbolSubst): TermSymbol = sub.mapTermSym(this)
-
+object TermSymbol:
+  def fromFunBms(b: BlockMemberSymbol, owner: Opt[InnerSymbol])(using State) =
+    TermSymbol(syntax.Fun, owner, Tree.Ident(b.nme))
```
LPTK, line 271:
```suggestion

object TermSymbol:
  def fromFunBms(b: BlockMemberSymbol, owner: Opt[InnerSymbol])(using State) =
    TermSymbol(syntax.Fun, owner, Tree.Ident(b.nme))

```

LPTK (changes requested):
Ok, I think this is almost ready to merge. But let's try to not create useless `while` loops (see comment below)!

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
CAG2Mark: The order was indeed changed in a previous commit, but I've fixed that. It's no longer in the PR diff.

hkmc2/shared/src/test/mlscript/codegen/TraceLog.mls
CAG2Mark: Oh, I did handle this but it appears there's a bug that causes things to be counted incorrectly sometimes. I'll fix it

CAG2Mark:
Done

LPTK (approved):
Ok, LGTM now, thanks!

hkmc2/shared/src/test/mlscript/tailrec/Errors.mls
```diff
@@ -80,3 +80,15 @@ module A with
 //│ ╔══[WARNING] This function does not directly self-recurse, but is marked @tailrec.
 //│ ║  l.79: 	  fun f() = 2
 //│ ╙──      	      ^
+
+:w
```
CAG2Mark: I suppose this should be a `:fixme`. The error message is a bit misleading

LPTK: Sure.
```suggestion
:fixme // TODO: support
```

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
LPTK: Ok. Did you add the test?

CAG2Mark: Yes, see https://github.com/hkust-taco/mlscript/pull/358/files#diff-2efe26f24af91d6a1257bec0dc18d23571183ef76c04a380b393f480eafd03b7R158

PR #363, opened 2025-12-14
Flow Contextuals

FlandiaYingman:
A very naive implementation of contextual parameters via flow analysis! 

There are still a lot of difficulties to research on... 😬


FlandiaYingman:
### TODOs

- [x] Basic Implicit Parameters
  
  This is nothing special than the contextual resolution in the (basic) resolution stage. I'll just have to store the "implicit context" for each resolvable term, and then query the corresponding instances if the flow-analyzed term (sel or app) really points to a function (or method) that accepts a list of implicit parameters.
  
  ```fs
  fun f(using x: Int) = x
  using Int = 42
  f // = 42
  ```

- [ ] Type Classes

  I'll add `Consumer.TypeClassInstance` as suggested in the last meeting to flow-analyze the type of the instance to be queried. This supports the usage of implicit parameters such as
  
  ```fs
  fun f(using Monoid[T]) = ...
  using Monoid[Int] = ...
  f
  // solve ?T <: TypeClassInstance(Monoid, ?ev1) for some ?ev1: Monoid[?A]
  ```

- [ ] Curried Implicit Parameter Lists
  
  I'll have to deal with the cases that a function can have multiple interleaving parameter lists, either regular or implicit. This shouldn't be too hard but requires some mechanical functions (which were already implemented in the basic resolver, I just have to adapt it to the flow resolver) to _zip_ the parameter lists and the argument lists.
  
  ```fs
  fun f(using Int)()(using Int) = ...
  using Int = 42
  f()
  ```

hkmc2/shared/src/test/mlscript/flows/ContextualsBasics.mls
```diff
@@ -0,0 +1,154 @@
...
+:sf
+id(new Foo).f
+//│ Flowed:
+//│   id⁰(new Foo⁰)‹app³›.f¹(instance$Ident(Int)⁰)‹app⁴›
+//│ where
+//│     
+//│ = 43
+
+
+:todo
```
LPTK, line 63: Oh, it looks like we do not deduplicate identical selection targets?

FlandiaYingman:
Ooohhh no my long lasting (blocking) PR!

LPTK:
Oh I guess it got closed because the base branch was merged and closed.

PR #364, opened 2025-12-14
Revert refactoring to explicit Document calls, cache DocContext regexes

Derppening:
This patchset reverts the refactoring to explicit `Document.*` calls, as the root cause of the slowdowns have been identified and an alternative way to implement WAT building using `doc`-interpolation is used instead.

This patchset also contains a minor performance improvement to `doc""`-splitting by compiling and caching all regexes used during Document splitting.

### DIFF
--- a/core/shared/main/scala/utils/package.scala
+++ b/core/shared/main/scala/utils/package.scala
@@ -153,6 +153,8 @@ package object utils {
     @inline def matches(pf: PartialFunction[A, Bool]): Bool =
       pf.lift(self).contains(true)
     
+    @inline def optionIf(cond: A => Bool): Option[A] = if (cond(self)) Some(self) else None
+    
     /** 
      * A helper to write left-associative applications, mainly used to get rid of paren hell
      * Example:
@@ -168,7 +170,6 @@ package object utils {
   implicit final class LazyGenHelper[A](self: => A) {
     
     @inline def optionIf(cond: Bool): Option[A] = if (cond) Some(self) else None
-    @inline def optionIf(cond: A => Bool): Option[A] = if (cond(self)) Some(self) else None
     
     @inline def optionUnless(cond: Bool): Option[A] = if (!cond) Some(self) else None
     @inline def optionUnless(cond: A => Bool): Option[A] = if (!cond(self)) Some(self) else None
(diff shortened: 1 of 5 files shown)


core/shared/main/scala/utils/package.scala
```diff
@@ -168,7 +170,6 @@ package object utils {
   implicit final class LazyGenHelper[A](self: => A) {
     
     @inline def optionIf(cond: Bool): Option[A] = if (cond) Some(self) else None
-    @inline def optionIf(cond: A => Bool): Option[A] = if (cond(self)) Some(self) else None
     
     @inline def optionUnless(cond: Bool): Option[A] = if (!cond) Some(self) else None
     @inline def optionUnless(cond: A => Bool): Option[A] = if (!cond(self)) Some(self) else None
```
Derppening: @LPTK `optionUnless(A => Bool)` also needs to be moved to `GenHelper`.

LPTK: Good catch!

hkmc2/shared/src/main/scala/hkmc2/utils/document/DocumentContext.scala
```diff
@@ -24,6 +26,13 @@ import Document._
  */
 
 object DocumentContext:
+  private val nestRegex = " #\\{ ".r
+  private val unnestRegex = " #\\} ".r
+  private val beginGroupRegex = "\\\\\\{".r
+  private val endGroupRegex = "\\\\\\}".r
+  private val docBreakRegex = " # ".r
+  private val forceDocBreakRegex = """\\n""".r
```
Derppening: I am not sure whether this should be in `UPPER_UNDERSCORE_CASE`, since these are all "`static final`" values in Java-speak.

LPTK: In Scala, we normally do not use screaming case, as it looks like it's straight from the dark ages of computing. Instead, we might prefer simply capitalizing constants, as in `NestRegex`.

LPTK (approved):
Thanks for the helpful diagnosis and improvement!

PR #365, opened 2025-12-16
Add compiler cache to avoid re-parsing and re-elaborating files

LPTK:
The JVM tests pass, but `concurrent.TrieMap` does not work in JS and we'll need to make a façade for it.


chengluyu (approved):
Masterpiece! A real quality-of-life update!

PR #360, opened 2025-12-07
Handler overhaul

AnsonYeung:
This PR moves away from the class-based representation of function frames in the slow path (double-compilation) to using a unified function representation and an auxiliary array for the frame. The main goal is to reduce the excessive code duplication of the previous approach.

The current approach mandates the use of the lifter. As lifter can not handle semantics of handle blocks (and it does not need to), we lower handle block to an equivalent form first, without the insertion of effect checks/resumption prelude. Now we invoke the lifter after handle block is translated and before the actual instrumentation takes place. We make use of the fact that lifter will preserve the mayRaiseEffect field.

For stack safety, I simplified the approach taken in https://github.com/hkust-taco/mlscript/pull/307, since we always include the entry state and we do not need to split into different cases when stack safety is enabled.

TODO:
- [x] Fix problem with Label/Continue/Break

Follow up:
- [ ] Spread arguments
- [ ] Debug Info (local variables/parameters)
- [ ] Use a unified array for storing the saved states in the runtime.

PR #366, opened 2025-12-16
Remove scope hack from JSBuilder

LPTK:
In https://github.com/hkust-taco/mlscript/pull/356, we changed the IR so that local symbols are explicitly allocated in `Scoped` blocks, instead of implicitly scoped. This makes the logic cleaner, resolves longstanding issues, and simplifies passes like deforestation (so they do not need to reimplement the logic to figure out what is bound where).

However, the JSBuilder still had hacks in place to generate bindings for symbols that are missing a `Scoped` defining them. This was needed because not all parts of the compiler have been fully adapted to `Scoped` yet. This PR disables the hacks and witnesses failures that seem mostly rerated to:
- [x] Code quasiquotes – @NeilKleistGao can you please address these? I think it just needs a few changes (using `registerTempSymbol` in a few places).
- [x] Tailrec optimizer – @ychenfo can you please address these in the way we discussed the other day?
- [x] Lifter – @CAG2Mark can you please address these? Now would also be the ideal time to make the refactoring you hinted at earlier, based on the more complete symbol representation we have now.
- [x] Handler lowering – but I think this should be solved by @AnsonYeung once https://github.com/hkust-taco/mlscript/pull/360 is merged
- [x] It also looks like the `@buffered` annotation generates unscoped variables; Cc @AnsonYeung


CAG2Mark:
OK. This will probably take a while...

NeilKleistGao:
Shall we push to this branch directly, or create another PR to this branch?

LPTK:
@NeilKleistGao feel free to push directly!

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -1319,6 +1334,9 @@ class Lifter()(using State, Raise, Config):
               val ctxxx = ctxx.withDefnsCur(analyzer.nestedDeep(d.sym))
               liftDefnsInCls(c, ctxxx.addBmsReqdInfo(createLiftInfoCls(c, ctxxx)))
             case _ => return super.applyBlock(b)
-          (lifted :: extra).foldLeft(applyBlock(rest))((acc, defn) => Define(defn, acc))
+          val newDefns = lifted :: extra
+          val newBms = newDefns.map(_.sym)
```
CAG2Mark: Fix for the broken lifter tests
```suggestion
          val newBms = extra.map(_.sym)
```

PR #368, opened 2025-12-19
Fix quote & region scope problems

NeilKleistGao:
It seems that I'm not authorized to push to the branch #366 directly.

In this PR:
- Collect symbols in quasiquote and region (in BBML/InvalML) for `Scoped`
- Have correct error message for cross-stage references in diff test


hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -819,6 +820,7 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
   def quoteSplit(split: Split)(k: Result => Block)(using LoweringCtx): Block = split match
     case Split.Cons(Branch(scrutinee, pattern, continuation), tail) => quote(scrutinee): r1 =>
       val l1, l2, l3, l4, l5 = new TempSymbol(N)
+      loweringCtx.collectScopedSyms(l1, l2, l3, l4, l5)
```
LPTK: Instead of creating `TempSymbol`s and collecting them later, just use `registerTempSymbol`.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -854,13 +859,21 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
         val l1, l2 = new TempSymbol(N)
+        loweringCtx.collectScopedSyms(l1, l2)
         Assign(l1, r1, setupTerm("CSRef", Value.Ref(l1) :: setupFilename :: Value.Lit(syntax.Tree.UnitLit(false)) :: Nil)(r2 =>
           Assign(l2, r2, setupTerm("Sel", Value.Ref(l2) :: Value.Lit(syntax.Tree.StrLit(name.name)) :: Nil)(k))
         ))
     case SynthSel(Ref(sym: BlockMemberSymbol), name) => // Multi-file cross-stage references
-      (t.toLoc, sym.toLoc) match
+      if config.qqEnabled then fail:
+        ErrorReport(
+            msg"Cross-stage reference to ${sym.nme}.${name.name} is only allowed in compiler test due to the lack of `import.meta` in REPL." ->
```
```suggestion
            msg"Cross-stage reference to ${sym.nme}.${name.name} is only allowed in compiled files due to the lack of `import.meta` in REPL mode." ->
```

hkmc2/shared/src/test/mlscript/codegen/Quasiquotes.mls
```diff
@@ -64,2103 +64,114 @@
...
+//│         Else(Ref(Symbol("x")))
+//│       )
 //│     )
 //│   )
 //│ )
 
-:fixme // FIXME: weird error
+
+:ge
+`1 `== `2
```
LPTK: Let's change these to `` `=== ``, then.

But it's strange that we can't use a global thing like `Predef` – that's not truly a cross-stage reference.

LPTK (approved):
Thanks!

hkmc2/shared/src/test/mlscript/codegen/Quasiquotes.mls
NeilKleistGao: It is because we need to compute which files should be included in the next stage. So, we need to use `import.meta` to get the relative path. In this example, `==` in `Predef` shadows the default one.  That's why we get the error.

hkmc2/shared/src/test/mlscript/codegen/Quasiquotes.mls
```diff
@@ -162,16 +162,15 @@ x `=> `if x `=== `0.0 then `1.0 else x
 //│ )
 
 
+// We cannot quote a symbol from another file here due to the lack of `import.meta` in REPL mode.
```
LPTK, line 165: Why would `import.meta` be needed? All we need is to make the genearted code import Predef, **_not_** the current file...

hkmc2/shared/src/test/mlscript/codegen/Quasiquotes.mls
LPTK: I still don't get it (see my other message).

hkmc2/shared/src/test/mlscript/codegen/Quasiquotes.mls
NeilKleistGao: Assume that we have a file `A.mls` containing a quoted symbol that is defined in `./B.mls` (a relative path to `A.mls`). If we generate the quoted code fragment into a file `./C.mls` (** a relative path to the pwd, which is the root dir for now **), we must know the relative path of `B.mls` to `C.mls`, which requires knowing where `A.mls` is.  In the ES Module, `__filename` is not defined, and we need to use `import.meta` to retrieve the full file path of `A.mls`.  However, it is not available in the REPL mode. That's why I did this check. Of course, it is possible to always generate an import statement for `Predef`. But for now, we are still using a relative path to import `Predef`. If we can write `import "Predef"`, and the compiler can locate the file by the given paths, then the reference to symbols in `Predef` will no longer rely on `import.meta`.

LPTK: Clearly, we should be able to know the path to`Predef` here, because it is being imported in the code! We know it at compile time, so we should be able to generate appropriate code.

PR #369, opened 2025-12-19
Lifter refactor and scope fix

CAG2Mark:
A long overdue refactor of the lifter and support for scopes. This will probably take a while.

I also plan on doing the following:
- The data structure used to store certain data about the functions uses space that is quadratic in the nesting depth anyway, so I will probably try to implement a more precise analysis on which variables are required by which functions. This analysis will take quadratic time. The lifting itself will still only rewrite the IR blocks O(1) times.
- Separate lifting into two passes: rewriting definitions to be independent of their current scope, and the actual lifting. I vaguely recall we wanted to only do the first pass at some point, so the ability to turn the "physical" lifting off may be useful.


hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
```diff
@@ -52,35 +53,27 @@ class ScopeData(using State):
...
-        Assign(`scopedWithIdSym`, Value.Lit(Tree.IntLit(id)), End(_)),
-        b: Scoped) => S((id, b))
-      case _ => N
-  
-  // Add UIDs to scopes.
-  object ScopeUidAdder extends BlockTransformer(SymbolSubst()):
-    override def applyScopedBlock(b: Block): Block = b match
-      case s: Scoped => ScopedWithId(fresh.make, s)
+  object ScopeUidAdder extends BlockTraverser:
+    var mp: IdentityHashMap[Scoped, UID] = new IdentityHashMap()
```
LPTK: The object's identity itself can serve as a key in your map. If you're going the `IdentityHashMap` route, there is no need to allocate any integer UID at all. What Anto did first was to allocate UIDs using a `val` defined directly in`Block`. But that did not work great (maybe @ychenfo can recall why), so he just switched to using object identity directly.

CAG2Mark: I'm more comfortable using my own UID because it will make debugging easier later on.

Also, I'm not sure how to get an integer representation of the object's reference anyway. I'd also rather not use an IdentityHashMap for other types, because I rely on the fact that `case class` objects with different references are still equal based on their values.

EDIT: it seems I can use `identityHashCode`, but I'm worried there could be collisions. If I'm going to use that, I need to do some more research on IdentityHashMap

CAG2Mark: I did a small experiment and the identity certainly should not be used as a key:
image

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
```diff
@@ -16,6 +16,7 @@ import scala.collection.mutable.Set as MutSet
 
 object ScopeData:
   opaque type ScopeUID = BigInt
+  val dummyUID: ScopeUID = 0
```
LPTK: What is this for?

PR #370, opened 2025-12-21
`Scoped` and tailrec

ychenfo:
This PR updates the tailrec optimization pass to maintain the `Scoped` blocks. It also tries to make minor improvements on the `:lot` pretty printing of `Scoped` blocks and on the JSBuilder about default switch branches.


hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -488,7 +488,7 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
         } # break; #} "
       val e = els match
       case S(el) =>
```
LPTK: This `case` should be indented. Let's fix it while we're at it!

LPTK (approved):
Nice work! Seems like the fix was straightforward, right?

hkmc2/shared/src/test/mlscript/codegen/NestedScoped.mls
```diff
@@ -246,88 +244,40 @@ scope.locally of scope.locally of (
...
+//│       bindings = HashMap($block$res -> block$res19, $block$res -> block$res6, member:h -> h, $runtime -> runtime, $definitionMetadata -> definitionMetadata, $prettyPrint -> prettyPrint, $Term -> Term, $block$res -> block$res7, $Block -> Block, member:f -> f, $Shape -> Shape, member:Predef -> Predef, $block$res -> block$res13, $block$res -> block$res20, $block$res -> block$res8, $block$res -> block$res9, member:f -> f1, $block$res -> block$res14, member:f_g -> f_g, $block$res -> block$res15, $block$res -> block$res10, $block$res -> block$res, $block$res -> block$res11, member:g -> g, $block$res -> block$res16, z -> z, $block$res -> block$res1, y -> y, x -> x, $block$res -> block$res17, $block$res -> block$res2, z -> z1, x -> x1, y -> y1, $block$res -> block$res3, $block$res -> block$res18, member:y -> y2, x -> x2, member:f -> f2, $block$res -> block$res12, $block$res -> block$res4, $block$res -> block$res5, member:foo -> foo, member:foo -> foo1)
 //│     curThis = S of N
 //│     bindings = HashMap()
 //│ ╔══[COMPILATION ERROR] No definition found in scope for member 'f'
```
ychenfo, line 256: Indeed this test still fails because the lifter hasn't yet put the symbols of the lifted out `f` and `g` in the correct outer `Scoped`. Only the `No definition found in scope for member 'f_g'` error is gone after this PR because `f_g` is generated during the tailrec optimization pass.

PR #371, opened 2025-12-29
Propagate Constructor Symbol

CAG2Mark:
The constructor symbol for class-like definitions was never propagated into the elaborated tree or the IR, but it could appear in resolved BMS references. For example:
```fs
:rt
class Bruh()
fun f() = Bruh()
//│ Elab: { Cls BruhParamList(‹›,List(),None) { }; fun member:f‹553›() = member:Bruh‹554›#666(); }
//│ Resolved tree:
//│ Blk:
//│   stats = Ls of 
//│     Parameterized:
//│       owner = N
//│       kind = Cls
//│       sym = class:Bruh‹555›
//│       bsym = member:Bruh‹554›
//│       tparams = Nil
//│       params = ParamList:
//│         flags = ()
//│         params = Nil
//│         restParam = N
//│       auxParams = Nil
//│       ext = N
//│       body = ObjBody of Blk:
//│         stats = Nil
//│         res = Lit of UnitLit of false
//│       companion = N
//│       annotations = Nil
//│     TermDefinition:
//│       k = Fun
//│       sym = member:f‹553›
//│       tsym = term:f‹562›
//│       params = Ls of 
//│         ParamList:
//│           flags = ()
//│           params = Nil
//│           restParam = N
//│       tparams = N
//│       sign = N
//│       body = S of App{typ=class:Bruh‹555›}:
//│         lhs = Resolved{sym=term:class:Bruh‹555›.Bruh‹557›}:
//│           t = Ref{sym=member:Bruh‹554›} of member:Bruh‹554›
//│           sym = term:class:Bruh‹555›.Bruh‹557› <-------- HERE
//│         rhs = Tup of Nil
//│       resSym = ‹result of member:f‹553››‹561›
//│       flags = ()
//│       modulefulness = Modulefulness of N
//│       annotations = Nil
//│       companion = N
```
This PR adds the symbol to the `ClassLikeDef` class in the term IR and propagates it all the way to the block IR.

Note that, previously, the only way to access this symbol was by accessing the `.tsym` field of the class's BlockMemberSymbol.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
@@ -785,6 +785,7 @@ sealed abstract class ClassLikeDef extends TypeLikeDef:
   val kind: ClsLikeKind
   val sym: DefinitionSymbol[? <: ClassLikeDef] & InnerSymbol
   val bsym: BlockMemberSymbol
+  val ctorSym: Opt[TermSymbol]
   val tparams: Ls[TyParam]
   val paramsOpt: Opt[ParamList]
   val auxParams: Ls[ParamList]
@@ -819,7 +820,8 @@ case class ModuleOrObjectDef(
   body: ObjBody,
   companion: Opt[ModuleCompanionSymbol],
   annotations: Ls[Annot],
-) extends ClassLikeDef, CompanionValue
+) extends ClassLikeDef, CompanionValue:
+  val ctorSym: Option[TermSymbol] = N
 
 case class PatternDef(
     owner: Opt[InnerSymbol],
@@ -851,11 +853,13 @@ case class PatternDef(
   val paramsOpt: Opt[ParamList] = N
   val auxParams: Ls[ParamList] = Nil
   val companion: Opt[CompanionSymbol] = N // TODO support
+  val ctorSym: Option[TermSymbol] = N
 
 
 sealed abstract class ClassDef extends ClassLikeDef:
   val kind: ClsLikeKind
   val sym: ClassSymbol
+  val ctorSym: Opt[TermSymbol]
   val tparams: Ls[TyParam]
   val paramsOpt: Opt[ParamList]
   val auxParams: Ls[ParamList]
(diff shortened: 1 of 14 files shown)


hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -897,6 +902,7 @@ object ClassDef:
       kind: ClsLikeKind,
       sym: ClassSymbol,
       bsym: BlockMemberSymbol,
+      ctorSym: Opt[TermSymbol],
```
LPTK: It seems we should make this
```suggestion
      ctorSym: S[TermSymbol],
```
right?

Parameterized classes should always have a ctor symbol.

LPTK (changes requested):
Interesting. Thanks for the fix! But I think we should tighten the types a bit.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
CAG2Mark: True

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -888,7 +888,7 @@ object ClassDef:
   ): ClassDef =
     params match
       case ps :: pss => Parameterized(owner, kind, sym.asInstanceOf// TODO: improve
-        , bsym, ctorSym
+        , bsym, S(ctorSym.get)
```
```suggestion
        , bsym, ctorSym.getOrElse(lastWords("Parameterized classes should have a ctor symbol."))
```

LPTK (approved):

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -883,7 +888,7 @@ object ClassDef:
   ): ClassDef =
     params match
       case ps :: pss => Parameterized(owner, kind, sym.asInstanceOf// TODO: improve
-        , bsym
+        , bsym, ctorSym.getOrElse(lastWords("Parameterized classes should have a ctor symbol."))
```
```suggestion
        , bsym, S(ctorSym.getOrElse(lastWords("Parameterized classes should have a ctor symbol.")))
```

PR #373, opened 2025-12-30
Preliminary Lifter Scoped Support

CAG2Mark:
Preliminary lifter support for scoped blocks. The full revamp is still ongoing.


CAG2Mark:
We really need to address those HashMap/HashSet symbol order changes...

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -14,46 +14,6 @@ import semantics.Elaborator.ctx
...
-    case Scoped(syms, body) =>
-      scopedSymForCurrentFun match
-        case None => super.applyScopedBlock(b)
-        case Some(scopedForCurrentFun) =>
-          scopedForCurrentFun.addAll(syms)
-          super.applySubBlock(body)
-    case _ => super.applySubBlock(b)
-    
-
-
```
AnsonYeung, line 56: This is used in the handler rewrite and is just unnecessary diff

AnsonYeung: Nevermind, it's moved to another file. I'm just confused by the removal of the pre handler lowering from the handler lowering itself

LPTK:
> We really need to address those HashMap/HashSet symbol order changes...

Actually, we don't really. We're only seeing changes here because there are unexpected compilation errors, which triggers the printing of internal Scope data structures. In principle, we should not be committing that, or at least not making PRs with such changes.

The printing of Scoped, on the other hand, was already handled by https://github.com/hkust-taco/mlscript/pull/366/commits/b6593edca2d0331a1fe4f27788f315ef1cb83a99.

CAG2Mark:
BTW should be ready.

Some of the tests fail because the handler fixes are not yet merged

LPTK:
(I guess it wouldn't be too bad to adapt the printing of `Scope` in a way similar to how the printing of `Scoped` was handled, as it would avoid needless changes in abnormal PRs like this one.)

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -622,7 +623,12 @@ class Lifter(handlerPaths: Opt[HandlerPaths])(using State, Raise):
     var activeClosures: Set[Local] = Set.empty
     // Map from block member symbols to initialized closures
     val closureMap: MutMap[BlockMemberSymbol, Local] = MutMap.empty
+    val extraLocals: MutSet[Local] = MutSet.empty
     
+    def rewrite(b: Block) =
+      val ret = applyBlock(b)
+      if extraLocals.isEmpty then ret
```
LPTK: I think `Scoped`'s smart constructor (`Scoped.apply`) already handles this case.

LPTK (approved):
Otherwise LGTM, thx.

PR #374, opened 2026-01-01
Added support for Tuples

FetBoba:


hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -439,6 +555,74 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+                ),
+                extraInfo = S(assign.toString)
+              )
+
+      val rstBlk = returningTerm(rst)
+      Instructions.block(
+        label = N,
+        children = Seq(assignInstr, rstBlk),
+        resultTypes = rstBlk.resultTypes.map: ty =>
+          Result(if ty is UnreachableType then RefType.anyref else ty.asValType_!)
```
LPTK: What is this `UnreachableType` special case for?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -439,6 +555,74 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+            )
+          val fieldidx = fieldSelect(selCls, selSym)
+          val objRef = ref.cast(lhsExpr, RefType(ctx.getType_!(selCls), nullable = false))
+          struct.set(fieldidx, objRef, rhsExpr)
+        case S(otherSym) =>
+          lastWords(
+            s"Expected resolved AssignField(...) expression to be a TermSymbol, but got $otherSym (${otherSym.getClass.getName})"
+          )
+        case N =>
+          nme.name.toIntOption.filter(_ >= 0).map: idx =>
```
LPTK: I don't think we want to special-case fields whose names are positive integers. Arrays are not the only ones that can have them. For now, let's just mandate the `_.[_]` syntax for index-based array accesses.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -237,25 +319,54 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
-        lastWords(s"Expected resolved Select(...) expression `$selTrmSym` to have an owner")
-      val selCls = selOwner.asBlkMember getOrElse:
-        lastWords(
-          s"Expected resolved class for Select(...) expression to be a BlockMemberSymbol, but got $selOwner (${selOwner.getClass.getName})"
+        case S(otherSym) =>
+          lastWords(
+            s"Expected resolved Select(...) expression to be a TermSymbol, but got $otherSym (${otherSym.getClass.getName})"
+          )
+        case N =>
+          id.name.toIntOption.filter(_ >= 0)
```
LPTK: I don't think we want to special-case fields whose names are positive integers. Arrays are not the only ones that can have them. For now, let's just mandate the `_.[_]` syntax for index-based array accesses.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -439,6 +555,74 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+            loc = fld.toLoc,
+            errCtx = "WatBuilder::returningTerm for AssignDynField(...)",
+            extra = assign.toString
+          )
+          val idxExpr = idxBuilder(tupleRef)
+          array.set(tupleInfo.arrayType, tupleRef, idxExpr, rhsExpr)
+        else
+          errExpr(
+            Ls(msg"WatBuilder::returningTerm for AssignDynField(...) where `arrayIdx = false` is not implemented yet" -> lhs.toLoc),
+            extraInfo = S(assign.toString)
```
LPTK: Any reason to not just use
```suggestion
            extraInfo = S(assign)
```
?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -237,25 +319,54 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+              tupleArrayGet(
+                tupleExpr = qualRes,
+                idxBuilder = _ => i32.const(idx)
+              )
+            .getOrElse:
+              errExpr(
+                Ls(
+                  msg"WatBuilder::result for field selection without a resolved symbol is not implemented (field `${id.name}`)" -> sel.toLoc
+                ),
+                extraInfo = S(sel.toString)
```
LPTK: Same question as above.

LPTK (commented):
Neat! This looks well done to me. WDYT @Derppening?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Some error expressions used  .toString. That's why assumed that this is a way to write it. But yes there is actually no need for it here.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Same as above

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Ok I now return error for access without brackets

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Same as above

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
LPTK: Yes because error messages use `Message`, which only has a conversion from strings. This is extra internal debug info, and it will be displayed to compiler developers as a useful tree (unless you stringify it as you did).

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
```diff
@@ -135,8 +135,21 @@ case class StructType(
   def toWat: Document =
     doc"(struct${fieldSeq.map(_.toWat).mkDocument(doc" ").surroundUnlessEmpty(doc" ")})"
 
+/** A type representing an array type. */
+case class ArrayType(
+    elemType: Type,
+    mutable: Bool,
+    parents: Seq[TypeIdx] = Seq.empty,
+    isSubtype: Bool = false
```
Derppening: Same here - This function appears to be unused.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -37,6 +37,11 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
 
   private val baseObjectSym: BlockMemberSymbol = BlockMemberSymbol("Object", Nil)
   private val tagFieldSym: TermSymbol = TermSymbol(syntax.MutVal, owner = N, Ident("$tag"))
+  private case class TupleArrayInfo(arrayType: TypeIdx, elemType: Type, mutable: Bool)
+  private var mutTupleArrayInfo: Opt[TupleArrayInfo] = N
+  private var tupleArrayInfo: Opt[TupleArrayInfo] = N
+  private case class ActiveLabel(sym: Local, breakLabel: Str, continueLabel: Opt[Str])
+  private var activeLabels: List[ActiveLabel] = Nil
```
Derppening: Any reason why these are not in `Ctx`?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -49,6 +54,83 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
   private def baseObjectRefType(nullable: Bool)(using Ctx): RefType =
     RefType(baseObjectTypeIdx, nullable = nullable)
 
+  private def tupleArray(mut: Bool)(using Ctx): TupleArrayInfo =
+    val cached = if mut then mutTupleArrayInfo else tupleArrayInfo
+    cached match
```
Derppening: Nit: I think this can be more succinctly expressed using `getOrElse`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Instructions.scala
```diff
@@ -336,6 +360,14 @@ object Instructions:
       resultType = S(RefType(ty, nullable = false))
     )
 
+    /** Creates a `struct.new` instruction. */
+    def new_(ty: TypeIdx, values: Seq[Expr]): FoldedInstr = FoldedInstr(
```
Derppening: This function appears to be unused.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -37,6 +37,11 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
 
   private val baseObjectSym: BlockMemberSymbol = BlockMemberSymbol("Object", Nil)
   private val tagFieldSym: TermSymbol = TermSymbol(syntax.MutVal, owner = N, Ident("$tag"))
+  private case class TupleArrayInfo(arrayType: TypeIdx, elemType: Type, mutable: Bool)
+  private var mutTupleArrayInfo: Opt[TupleArrayInfo] = N
+  private var tupleArrayInfo: Opt[TupleArrayInfo] = N
```
Derppening: I don't really understand why this is necessary. Can't we just lookup/cache the Wasm type for tuples from the module?

And, since we assume that all tuples have `anyref`s as their element type, why do we need to store `elemType` and `mutable`?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -49,6 +54,83 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+  )(using Ctx, Raise, Scope): Expr =
+    val mutInfo = tupleArray(true)
+    val immInfo = tupleArray(false)
+    val tupleIsMutable = ref.test(tupleExpr, RefType(mutInfo.arrayType, nullable = true))
+    val mutableBranch =
+      val tupleRef = ref.cast(tupleExpr, RefType(mutInfo.arrayType, nullable = false))
+      array.get(mutInfo.arrayType, tupleRef, idxBuilder(tupleRef), mutInfo.elemType)
+    val immutableBranch =
+      val tupleRef = ref.cast(tupleExpr, RefType(immInfo.arrayType, nullable = false))
+      array.get(immInfo.arrayType, tupleRef, idxBuilder(tupleRef), immInfo.elemType)
```
Derppening: I thought we established that `array.get` can be used on both mutable and immutable arrays. Is there any reason why this is necessary?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -49,6 +54,83 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+    Instructions.`if`(
+      condition = tupleIsMutable,
+      ifTrue = mutableBranch,
+      ifFalse = S(immutableBranch),
+      resultTypes = Seq(Result(mutInfo.elemType.asValType_!))
+    )
+
+  private def tupleIndexBuilder(
```
Derppening: Nit: `buildTupleIndex` - This is an operation, not a class/factory.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -49,6 +54,83 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+            val casted = ref.cast(rawIdx, RefType.i31ref)
+            i31.get(casted, signed = true)
+          case ty =>
+            val err = errExpr(
+              Ls(
+                msg"$errCtx expects an integer index but found ${ty.fold("(none)")(_.toWat.mkString())}" -> loc
+              ),
+              extraInfo = S(extra)
+            )
+            return (_: Expr) => err
```
Derppening: Nit: Just return the entire `errExpr` expression?

Derppening (commented):
Apart from some fields that appear to be better located in `Ctx` than `WatBuilder`, the rest looks good to me.

Please do add documentation where possible, including private members.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
LPTK: And does it need to be mutable? 👀

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
LPTK: I think `mutable` is needed because we do differentiate mutable and immutable tuples (see the tests).

Also, it doesn't hurt to store `elemType`. We may later generalize support for more precisely-typed arrays.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
LPTK: Also, use
```suggestion
            return (_: Expr) => errExpr(
              msg"$errCtx expects an integer index but found ${ty.fold("(none)")(_.toWat.mkString())}" -> loc
                :: Nil,
              extraInfo = S(extra)
            )
```
to avoid some of the awkward parentheses.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
Derppening: I guess I wasn't being clear here - What I meant is why is `mutTupleArrayInfo` and `tupleArrayInfo` necessary and stored as separate fields? It appears to me that what we may want is a lookup table for intrinsic types (e.g. `Unit`, `Tuple` and mutable `Tuple`) to lookup its corresponding Wasm type.

While `mutTupleArrayInfo` and `tupleArrayInfo` is probably good enough when tuples are stored as `anyref` arrays as they are now, if (and when) we generalize support for more precisely-typed arrays, these fields probably wouldn't suffice.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
FetBoba: which function? If you are talking about ArrayType, then it is used inside tupleArray on line 68 of WatBuilder.scala

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Instructions.scala
FetBoba: Ok I will delete it

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Ok

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: ok

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: I need to first cast to use array.get. I think that we can't cast immutable tuple to the mutable tuple type or vice versa.

Derppening: But can't we just use `array.get` on both mutable and immutable arrays in Wasm?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
Derppening: Sorry, I meant the `parents` and `isSubtype` fields.

FetBoba: ok

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Ok I moved it to Ctx

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Ok now I look up intrinsic types from MutMap declared in Ctx

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Yes but before using array.get we need to cast expressions

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: I thought UnreachableType can appear hear that's why I put this extra check but seems like in other parts of `returningTerm` it is not used. I will change it

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Ok I renamed it to compileTupleIndex

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
LPTK: I think his question is whether you can cast both mutable and immutable array values to the readonly array type and do the access then.

FetBoba: I don't think that it can be done. I changed this function to work as you said and I got illegal cast error. So ig the only way is to cast them and use get separately.

LPTK: Ok, fine.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -9,7 +9,7 @@ import hkmc2.utils.*
 import document.*
 import document.Document
 import semantics.*
-import text.Param as WasmParam
+import hkmc2.codegen.wasm.text.Param as WasmParam
```
Derppening: Is this change necessary?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -129,6 +129,9 @@ class TypeInfo(
 end TypeInfo
 
 object Ctx:
+  enum WasmIntrinsicType:
```
Derppening: I think this doesn't need to be nested within `Ctx` - You can hoist this outside of the `object` declaration.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -49,6 +51,79 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+      ifTrue = mutableBranch,
+      ifFalse = S(immutableBranch),
+      resultTypes = Seq(Result(elemType.asValType_!))
+    )
+
+  private def compileTupleIndex(
+      fld: Path,
+      loc: Opt[Loc],
+      errCtx: Str,
+      extra: => Str
```
Derppening: Nit: `errExtra` to clearly indicate that the `extra` is used for error reporting purposes.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -49,6 +51,79 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+              elemType = RefType.anyref,
+              mutable = mut
+            )
+          )
+        )
+    )
+
+  private def tupleArrayGet(
+      tupleExpr: Expr,
+      idxBuilder: Expr => Expr
```
Derppening, line 91: Is there a reason why this needs to be a lambda and not just a simple `Expr` that evaluates to a single non-negative integer?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -37,6 +37,8 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
 
   private val baseObjectSym: BlockMemberSymbol = BlockMemberSymbol("Object", Nil)
   private val tagFieldSym: TermSymbol = TermSymbol(syntax.MutVal, owner = N, Ident("$tag"))
+  private case class ActiveLabel(sym: Local, breakLabel: Str, continueLabel: Opt[Str])
+  private var activeLabels: List[ActiveLabel] = Nil
```
Derppening: These classes/variables appear to be unused.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -49,6 +51,79 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+    val elemType = RefType.anyref
+    val mutArrayType = tupleArrayType(true)
+    val immArrayType = tupleArrayType(false)
+    val tupleIsMutable = ref.test(tupleExpr, RefType(mutArrayType, nullable = true))
+    val mutableBranch =
+      val tupleRef = ref.cast(tupleExpr, RefType(mutArrayType, nullable = false))
+      array.get(mutArrayType, tupleRef, idxBuilder(tupleRef), elemType)
+    val immutableBranch =
+      val tupleRef = ref.cast(tupleExpr, RefType(immArrayType, nullable = false))
+      array.get(immArrayType, tupleRef, idxBuilder(tupleRef), elemType)
```
Derppening: Question: If `tupleExpr` evaluates to an expression with possible side effects (e.g. a function call), would this repeated use of `tupleExpr` here cause the expression to be emitted 3 times and invoked twice during execution? 

For example, in `makePair(1, 2).[1]`, would `makePair(1, 2)` be evaluated twice?

If so, it may be worth it to generate a temporary local to store the result of `tupleExpr`, roughly like:

```wasm
(if
  (local.tee $_tmp
    (ref.test (...))
  (then
    (array.get
      (ref.cast
        (local.get $_tmp)))
```

(`local.tee` sets the result of the child expression into the local, and returns the same value.)

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -49,6 +51,79 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+            return (_: Expr) => errExpr(
+              msg"$errCtx expects an integer index but found ${ty.fold("(none)")(_.toWat.mkString())}" -> loc
+                :: Nil,
+              extraInfo = S(extra)
+            )
+        tupleRef =>
+          Instructions.`if`(
+            condition = i32.lt_s(idxI32, i32.const(0)),
+            ifTrue = i32.add(idxI32, array.len(tupleRef)),
+            ifFalse = S(idxI32),
```
Derppening: The `idxI32` expression is emitted three times and will be executed twice during runtime, similar to [this comment](https://github.com/hkust-taco/mlscript/pull/374/changes#r2695275061).

Derppening (commented):
Please address the comments and also add some brief documentation for the private functions.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: I changed it because at some point there was an importing error. But I will look at it again

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: ok

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: ok

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: ok

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: I think it needs to be lambda otherwise we end up with duplicating the index computation in both branches because we only get correct tupleRef inside each branch. And also negative indices require `array.len(tupleRef)`

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: ok I added local.tee but it required extra handling for temp locals

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: ok

Derppening (commented):
LGTM. 

Please add some brief documentation for the newly-added private functions as mentioned in the previous review.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -49,6 +49,117 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+    val immutableBranch =
+      val tupleRef = ref.cast(tupleValue, RefType(immArrayType, nullable = false))
+      array.get(immArrayType, tupleRef, idxBuilder(tupleRef), elemType)
+    Instructions.`if`(
+      condition = tupleIsMutable,
+      ifTrue = mutableBranch,
+      ifFalse = S(immutableBranch),
+      resultTypes = Seq(Result(elemType.asValType_!))
+    )
+  /**
```
Derppening, line 112: Nit: Needs a newline between this and the previous line.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -49,6 +49,117 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+    val sym = TempSymbol(N, base)
+    val nme = scope.allocateName(sym)
+    ctx.addLocal(sym)
+    LocalIdx(SymIdx(nme))
+
+  /** 
+   * Returns locals allocated during codegen which are not in `exclude`. 
+   */
+  private def getExtraLocals(exclude: Set[Local])(using Ctx): Seq[Local] =
+    ctx.getWasmLocals._2.getOrElse(Seq.empty).filterNot(exclude.contains)
```
Derppening: I think it might be better to not have the `exclude` parameter but rather have the caller perform the `filter` operation as is necessary, since it is not very clear what the purpose of this `exclude` set is.

Derppening (commented):
This looks almost good to be merged. Please address the remaining comments.

PR #376, opened 2026-01-04
Explicit Symbol Type for Labels

CAG2Mark:
The `Label` class previously accepted any symbol type for labels, which is quite loose, and forces IR passes analyzing/modifying `Label` blocks to have loose types as well if they use this symbol as an identifier.

This PR adds a new symbol type `LabelSymbol`, and forces its usage for `Label`, `Continue` and `Break`.

@chengluyu The UCS normalization uses `TempSymbol` for label symbols, which has an optional `Term` parameter, but I noticed this term is never used. Therefore, I did not include it in `LabelSymbol`. Is this OK?
Edit: I just realized it's used in toLoc, whoops


LPTK (approved):
LGTM, thanks!

PR #377, opened 2026-01-05
Address the sharing of `argumentN$`s

ychenfo:
This PR addresses the sharing of `argumentN$` symbols, thanks @chengluyu a lot for the help!


LPTK (approved):
Great, LGTM.

Does that solve the problem of scoping conditional branches?

PR #379, opened 2026-01-08
Inverted Match

AnsonYeung:
Allow Match to be inverted, which will be used to create non null checks during handler lowering.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
@@ -494,7 +494,7 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
     case Match(scrut, hd :: tl, els, rest) =>
       val sd = result(scrut)
       def cond(cse: Case) = cse match
-        case Case.Lit(lit, inv) => doc"$sd ${if !inv then "===" else "!=="} ${lit.idStr}"
+        case Case.Lit(lit, inv) => doc"$sd ${if inv then "!==" else "==="} ${lit.idStr}"
         case Case.Cls(cls, pth) => cls match
           // case _: semantics.ModuleSymbol => doc"=== ${result(pth)}"
           // [invariant:0] If the class represented by `cls` does not exist at
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
@@ -494,7 +494,7 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
     case Match(scrut, hd :: tl, els, rest) =>
       val sd = result(scrut)
       def cond(cse: Case) = cse match
-        case Case.Lit(lit) => doc"$sd === ${lit.idStr}"
+        case Case.Lit(lit, inv) => doc"$sd ${if !inv then "===" else "!=="} ${lit.idStr}"
         case Case.Cls(cls, pth) => cls match
           // case _: semantics.ModuleSymbol => doc"=== ${result(pth)}"
           // [invariant:0] If the class represented by `cls` does not exist at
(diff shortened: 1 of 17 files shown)


hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -494,7 +494,7 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
     case Match(scrut, hd :: tl, els, rest) =>
       val sd = result(scrut)
       def cond(cse: Case) = cse match
-        case Case.Lit(lit) => doc"$sd === ${lit.idStr}"
+        case Case.Lit(lit, inv) => doc"$sd ${if !inv then "===" else "!=="} ${lit.idStr}"
```
```suggestion
        case Case.Lit(lit, inv) => doc"$sd ${if inv then "!==" else "==="} ${lit.idStr}"
```

hkmc2/shared/src/main/scala/hkmc2/codegen/llir/Interp.scala
```diff
@@ -175,7 +175,8 @@ class Interpreter(tl: TraceLogger):
           }
         case Value.Literal(lit) => 
           cases.find {
-            case (Pat.Lit(lit2), _) => lit === lit2
+            case (Pat.Lit(lit2, false), _) => lit === lit2
+            case (Pat.Lit(lit2, true), _) => !(lit === lit2)
```
LPTK: Why not
```suggestion
            case (Pat.Lit(lit2, true), _) => lit =/= lit2
```

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -718,16 +718,24 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
           cse match
-            case Case.Lit(lit) =>
+            case Case.Lit(lit, inv) =>
               val testExpr: FoldedInstr = lit match
                 case BoolLit(value) =>
                   val scrutAsI31 = ref.cast(getScrutExpr, RefType.i31ref)
                   val scrutValue = i31.get(scrutAsI31, signed = true)
-                  i32.eq(scrutValue, i32.const(if value then 1 else 0))
+                  val patValue = i32.const(if value then 1 else 0)
+                  if !inv then
```
LPTK: Swap the branches? Or maybe make the positive case `true` (and rename the variable from `inv` to `pos`).

LPTK (approved):
LGTM, but it's missing at least one test where this is actually used.

hkmc2/shared/src/main/scala/hkmc2/codegen/llir/Interp.scala
AnsonYeung: I was wondering why there is no !==

AnsonYeung:
It's probably better to change the JSBuilder and not the IR, as this change make little sense with multiple branch. I will implement the change to JSBuilder directly in the handler branch as it is small enough. So we do not need to change the IR to allow inverted match then.

LPTK:
But I don't think there's a compact/convenient way of representing that in the IR? Ideally we'd want to have such compact representation to facilitate analyses and optimizations.

Now that I think of it, though, it seems that we should allow negative match cases regardless of the pattern, not just for literal patterns.

AnsonYeung:
Right now, I express it as `Match(..., Case.Lit(...) -> End(), S(...), rst)`, I'm not sure how to add it so that it would be available for all patterns

LPTK:
We would associate a `Bool` with each `Case` in the `Match`.

Using default branches technically works but isn't compact (can't merge several into the same match). Not sure it's that important, though.

AnsonYeung:
For a negative match, I think it's typically not used in multiple branches as the only useful scenario I can think of is for early exit, where you expect the match to sucess and return if it's not matched.

LPTK:
Users can write things with negated patterns, eg

```fs
fun test(x) = if x is
  0 then "0"
  ~2 & ~3 then "not 2 and not 3"
  ~3 then "not 3"
  _ then "other"

test(0)
//│ = "0"
test(1)
//│ = "not 2 and not 3"
test(2)
//│ = "not 3"
test(3)
//│ = "other"
```

Though currently, it seems UCS lowering already does a good job producing a negation-free result:

```fs
//│ JS (unsanitized):
//│ let test;
//│ test = function test(x2) {
//│   switch (x2) {
//│     case 0:
//│       return "0";
//│       break;
//│     case 2:
//│       return "not 3";
//│       break;
//│     case 3:
//│       return "other";
//│       break;
//│     default:
//│       return "not 2 and not 3";
//│       break;
//│   }
//│ };
```

PR #357, opened 2025-12-01
Virtualize path and file system operations

chengluyu:
When I make web demos, one of the headaches is that `os.Path` doesn’t compile when the target is JavaScript. I made a hack on my web demo branch. But the implementation is a bit crappy, especially after merging the latest changes for a few times. Therefore, I want to introduce a path representation and abstract read operations that both supports a real file system or a virtual one (e.g., when the compiler is running on the web).

This PR intends to do the following things.

- [x] Provide an abstraction layer of path and file system operations.
- [x] Call the MLscript compiler from JavaScript.
  - [x] Compile single MLscript source file.
  - [x] Compile multiple MLscript source files.
- ~~Provide a minimal web demo to showcase the ability to compile multiple files.~~ This will be done in a separate PR.

PR #380, opened 2026-01-13
Remove `Term` from context initialization

NeilKleistGao:
see PR #355 .

### DIFF
--- a/hkmc2/shared/src/test/mlscript-compile/Predef.mls
+++ b/hkmc2/shared/src/test/mlscript-compile/Predef.mls
@@ -5,6 +5,7 @@ import "./RuntimeJS.mjs"
 // * TODO: proper caching of parsed/elaborated files
 import "./Runtime.mjs"
 import "./Rendering.mjs"
+import "./Term.mjs"
 // import "./Runtime.mls"
 // import "./Rendering.mls"
 
@@ -126,3 +127,6 @@ fun raiseUnhandledEffect() =
   Runtime.mkEffect(Runtime.FatalEffect, null)
 
 
+module meta with
+  fun codegen(t, file) = Term.codegen(t, file)
+  fun print(t) = Term.print(t)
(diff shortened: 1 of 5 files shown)


LPTK (approved):
LGTM, thx.

hkmc2/shared/src/test/mlscript-compile/Predef.mls
```diff
@@ -5,6 +5,7 @@ import "./RuntimeJS.mjs"
 // * TODO: proper caching of parsed/elaborated files
 import "./Runtime.mjs"
 import "./Rendering.mjs"
+import "./Term.mjs"
```
chengluyu, line 8: Why are we so obsessed with importing `Term.mls` everywhere? This is a regression against #357, which implements the detection of quote syntax and only imports `Term.mjs` when necessary.

`Term.mjs` requires `fs` and other Node.js built-in modules. Importing it would render the file less portable and unusable in browser environments.

hkmc2/shared/src/test/mlscript-compile/Predef.mls
```diff
@@ -126,3 +127,6 @@ fun raiseUnhandledEffect() =
   Runtime.mkEffect(Runtime.FatalEffect, null)
 
 
+module meta with
+  fun codegen(t, file) = Term.codegen(t, file)
+  fun print(t) = Term.print(t)
```
chengluyu, line 132: If they are purely aliases, why don't we import `Term.mjs` only in the files that need it? I checked in the codebase, only `QQImport.mls` uses `meta.codegen` and `meta.print`. 

I tried to remove the import and use of `Term` from here to `QQImport.mls`. Everything worked well. So, what is the rationale behind this change?

hkmc2/shared/src/test/mlscript-compile/Predef.mls
LPTK: I wasn't aware of that regression... I don't know why it's improted everywhere now. @NeilKleistGao is this something that's fixed in one of the many FYP branches?

hkmc2/shared/src/test/mlscript-compile/Predef.mls
LPTK: Ah, I guess we probably agreed to make these part of Predef for convenience, but it was a bad call. It's fine if the user has to import them. IIRC, before that, the code was doing some dirty hack to inject modules/functions into the current namespace.

PR #384, opened 2026-01-15
Lifter revamp

CAG2Mark:
See https://github.com/hkust-taco/mlscript/pull/369


LPTK:
Oh, I didn't realize the handler refactor PR was made against the `remove-scope-hacks` branch. We should have changed that before the merge.

By the way, your rebased commit does not compile.

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
```diff
@@ -322,8 +313,11 @@ class UsedVarAnalyzer(b: Block, scopeData: ScopeData)(using State, IgnoredScopes
           // Note that we traverse directly into scoped blocks without using handleCalledScope
           
           case l: Label if l.loop =>
-            handleCalledScope(l.label)
+            rec(l.body)(using isLinear = false) |> merge
+            applyBlock(l.rest)
           case Assign(lhs, rhs, rest) =>
+            println("assign: " + lhs + " = " + rhs)
```
LPTK: I see that you're printing debugging information and will presumably remove the printing later. Why not use the tracing debug mechanism, so we can enable debugging of the lifter (eg with `:dl`) whenever needed?

CAG2Mark: I'll consider adding it

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
```diff
@@ -18,6 +18,12 @@ import scala.jdk.CollectionConverters.*
 import java.util.IdentityHashMap
 import java.util.Collections
 import scala.collection.mutable.Buffer
+import hkmc2.ScopeData.ScopedObject.Top
+import hkmc2.ScopeData.ScopedObject.Companion
+import hkmc2.ScopeData.ScopedObject.ClassCtor
+import hkmc2.ScopeData.ScopedObject.Func
+import hkmc2.ScopeData.ScopedObject.Loop
+import hkmc2.ScopeData.ScopedObject.ScopedBlock
```
LPTK:
```suggestion
import hkmc2.ScopeData.ScopedObject.*
```

CAG2Mark: My IDE kept adding these when I was autocompleting match statements, so I've just manually removed usages of these.

CAG2Mark:
Remaining tasks:
- [x] Add back handler paths (so we are allowed to lift classes that extend handler-related classes)
- [x] Lift objects

hkmc2/shared/src/test/mlscript/handlers/Debugging.mls
```diff
@@ -157,66 +157,33 @@ module Test with
...
-//│ 	at: hkmc2.Lifter.transform$$anonfun$1(Lifter.scala:1468)
-//│ 	at: scala.collection.immutable.List.foreach(List.scala:323)
-//│ 	at: hkmc2.Lifter.transform(Lifter.scala:1467)
-//│ 	at: hkmc2.codegen.Lowering.program(Lowering.scala:1056)
-//│ 	at: hkmc2.JSBackendDiffMaker.processTerm(JSBackendDiffMaker.scala:110)
-//│ 	at: hkmc2.BbmlDiffMaker.processTerm(BbmlDiffMaker.scala:36)
-//│ 	at: hkmc2.LlirDiffMaker.processTerm(LlirDiffMaker.scala:63)
-//│ 	at: hkmc2.WasmDiffMaker.processTerm(WasmDiffMaker.scala:51)
 
 // Currently this test fails due to lifter issue. Commenting out :lift at the top of this file will make this work.
```
LPTK: Don't forget to update the comment.

hkmc2/shared/src/test/mlscript/handlers/NestedHandlers.mls
```diff
@@ -9,6 +9,7 @@ let id = 0
 class MaybeStop with
   fun f(x: Bool): ()
 
+:slot
```
LPTK: Feels like this is to much output to commit.

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
LPTK: Interesting. Which IDE do you use?

CAG2Mark: Just vscode with the nornal Scala extension

CAG2Mark:
Seems that a lot of the complexity from before came from dealing with nested modules and objects...

Things were pretty clean until I added nested modules and objects.

CAG2Mark:
OK so a new pain point: I currently lift classes by putting the extra variables in private fields. But since classes could be instantiated immutably, we are forced to put all variables they mutate into captures. Any idea on what to do? Or should we just give up and put them into captures?

EDIT: apparently private fields remain mutable after freezing, which also fixes the other problem I was worrying about (that I never mentioned)

CAG2Mark:
OK, it should be ready.

Sadly it still ended up quite complicated and long, but I hope the code is now structrued in a way that is easier to understand despite its length.

`Lifter.scala` and `UsedVarAnalyzer.scala` have been completely revamped, with only small elements from the old version being re-used, so IMO the diff is quite meaningless. I recommend just reviewing them as if they are entirely new files.

BTW, I found a bug with spreads; please see the `Spreads.mls` test.

hkmc2/shared/src/test/mlscript/staging/Syntax.mls
```diff
@@ -21,8 +21,6 @@ staged fun f() = 0
 :staging
 :w
 staged module A
-//│ ╔══[WARNING] `staged` keyword doesn't do anything currently.
-//│ ║  l.23: 	staged module A
-//│ ╙──      	       ^^^^^^^^
 //│ Pretty Lowered:
-//│   define staged class A in set block$res = undefined in end
+//│   define class A in set block$res = undefined in end
+//│ FAILURE: Unexpected lack of warnings
```
CAG2Mark: I have no clue why this changed

CAG2Mark: huh.... it seems `mlscript/staging/Syntax.mls` is excluded from the difftests. Guess I should just reset the file

LPTK: Weird! I don't see why that is. Any idea @NeilKleistGao?

CAG2Mark: All `staging` tests are excluded in the difftests scala file, but I don't know the reason. It also means we'll have to do a bisect to figure out exactly where this starts happening :upside_down_face:

CAG2Mark:
[lifter.pdf](https://github.com/user-attachments/files/24855494/lifter.pdf)

I forgot to send this. I expanded the lifter revamp doc a little bit

hkmc2/shared/src/test/mlscript/staging/Syntax.mls
LPTK: > All `staging` tests are excluded in the difftests scala file, but I don't know the reason. It also means we'll have to do a bisect to figure out exactly where this starts happening 🙃

We could just git-blame the place where the special case is made. But @NeilKleistGao probably already knows.

CAG2Mark: Oh, I meant the commit where the behaviour in the test changes.

CAG2Mark:
(For the record, Anson did not work on this or the previous lifter)

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -26,18 +27,16 @@ object Lifter:
...
-    def apply(f: BlockMemberSymbol) = mp(f)
-    private lazy val inverse = mp.flatMap:
-      case fn -> vars => vars.vars.map(v => v -> fn)
-    // gets the function to which a local belongs
-    def lookup(l: Local) = inverse.get(l)
   
+  class LazyDefn(defn: => Defn):
+    var value: Opt[Defn] = N
+    def force =
+      value = S(defn)
```
AnsonYeung: You can extend from the `mlscript.utils.Lazy` class here

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -26,18 +27,16 @@ object Lifter:
...
+  extension (l: List[LazyDefn | Defn])
+    def gatherUsed: List[Defn] = l.collect:
+      case l: LazyDefn if l.value.isDefined => l.value.get
+      case d: Defn => d
+
```
AnsonYeung: You can use `mlscript.utils.Box[Defn]` as the type

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -109,362 +81,117 @@ object Lifter:
...
     case _ => false
 
-
 /**
   * Lifts classes and functions to the top-level. Also automatically rewrites lambdas.
   * Assumes the input block does not have any `HandleBlock`s.
   */
-class Lifter()(using State, Raise, Config):
+class Lifter(topLevelBlk: Block, handlerPaths: HandlerPaths)(using State, Raise, Config):
   import Lifter.*
```
AnsonYeung, line 91: Why are we adding HandlerPaths back?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -109,362 +81,117 @@ object Lifter:
...
-    def flushModules = 
-      // called when we are lifted out while in some module, so we need to add the modules' isym paths
-      copy(curModules = Nil).addIsymPaths(curModules.map(d => d.isym -> LocalPath.Sym(d.sym)).toMap)
   
-  object LifterCtx:
-    def empty = LifterCtx()
-    def withLocals(u: UsedLocalsMap) = empty.copy(usedLocals = u)
-    
+  val handlerSyms: Set[Symbol] = Set(State.nonLocalRet, State.effectSigSymbol)
+
```
AnsonYeung: Lifter will never see effectSig under the current compilation order.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -485,18 +212,18 @@ class Lifter()(using State, Raise, Config):
...
-            case Some(path) if isIgnoredPath(path) => ()
-            case Some(Select(RefOfBms(s, _), Tree.Ident("class"))) =>
-              if clsSyms.contains(s) then extendsGraph += (s -> defn.sym)
-            case Some(RefOfBms(s, _)) =>
-              if clsSyms.contains(s) then extendsGraph += (s -> defn.sym)
-            case _ if !ignored.contains(defn.sym) =>
+            // for now, allow selecting runtime symbols
+            case Some(Select(qual = Value.Ref(l, _))) if State.runtimeSymbol is l => ()
+            case Some(RefOfBms(_, S(s: ClassSymbol))) if handlerSyms.contains(s) => ()
+            case Some(RefOfBms(s, _)) if handlerSyms.contains(s) => ()
```
AnsonYeung: 👀 Is this really needed?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
AnsonYeung: In fact using a empty set here yields no diff in tests, and you should probably just remove this variable. I think handler is not a special case here. The problem should be that a lot of the times symbols are not inserted for SynthSel right now.

hkmc2/shared/src/test/mlscript/backlog/Lifter.mls
```diff
@@ -68,6 +51,25 @@ test(2)
...
+//│ ═══[RUNTIME ERROR] Expected: '"B()"', got: '"(x) => {       this.#x = x;       return this;     }"'
+//│ = "(x) => {       this.#x = x;       return this;     }"
+
+// In fact, the correct way (using the current "constructor returns a lambda" hack) is this:
+// return (params) => {
+//   super(parentMainParams);
+//   // ctor
+//   return this(parentAuxParams);
+// }
+
```
AnsonYeung, line 72: wouldn't really work if the constructor access the fields.

AnsonYeung (commented):
I still need some time to read through rest of the code. I'll review the rest later.

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
```diff
@@ -0,0 +1,403 @@
...
+import java.util.IdentityHashMap
+import scala.collection.mutable.Map as MutMap
+import scala.collection.mutable.Set as MutSet
+
+object ScopeData:
+  opaque type ScopeUID = BigInt
+  val dummyUID: ScopeUID = 0
+  class FreshUID:
+    private val underlying = FreshInt()
+    def make: ScopeUID = underlying.make
```
AnsonYeung, line 21: why BigInt when FreshInt can only possibly generate an integer?

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
```diff
@@ -0,0 +1,403 @@
...
+
+import hkmc2.syntax.Tree
+import hkmc2.codegen.llir.FreshInt
+import java.util.IdentityHashMap
+import scala.collection.mutable.Map as MutMap
+import scala.collection.mutable.Set as MutSet
+
+object ScopeData:
+  opaque type ScopeUID = BigInt
+  val dummyUID: ScopeUID = 0
```
AnsonYeung:
```suggestion
  opaque type ScopeUID = BigInt
```

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
```diff
@@ -0,0 +1,403 @@
...
+    val nodesMap: Map[ScopedInfo, ScopeNode] = root.allChildNodes.map(n => n.obj.toInfo -> n).toMap
+  
+  type ScopeNode = ScopeNode.ScopeNode[?]
+  type TScopeNode[T] = ScopeNode.ScopeNode[T]
+  object ScopeNode:
+    case class ScopeNode[T](obj: TScopedObject[T], var parent: Opt[ScopeNode[?]], children: List[ScopeNode[?]])(using ignoredScopes: IgnoredScopes):
+      
+      lazy val allParents: List[ScopeNode[?]] = parent match
+        case Some(value) => this :: value.allParents
+        case None => this :: Nil
```
AnsonYeung: maybe ancestors could be a better name, it's a standard term used for trees.

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
```diff
@@ -0,0 +1,403 @@
...
+      lazy val allChildren: List[ScopedObject] = allChildNodes.map(_.obj)
+      
+      // does not include variables introduced by itself
+      lazy val existingVars: Set[Local] = parent match
+        case Some(value) => value.existingVars ++ value.obj.definedLocals ++ value.nestedObjSyms
+        case None => Set.empty
+      
+      lazy val isTopLevel: Bool = parent match
+        case Some(ScopeNode(obj = _: ScopedObject.Top)) => true
+        case _ => false
```
AnsonYeung:
```suggestion
      lazy val inTopLevel: Bool = parent match
        case Some(ScopeNode(obj = _: ScopedObject.Top)) => true
        case _ => false
```
`isTopLevel` is too confusing here, since function at the top level is not really the "top level" itself.

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
```diff
@@ -0,0 +1,403 @@
...
+      
+      lazy val isModOrTopLevel: Bool = parent match
+        case Some(par) => par.obj match
+          case _: ScopedObject.Companion => true
+          case _ => false
+        case None => true
```
AnsonYeung: same as above

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
```diff
@@ -0,0 +1,403 @@
...
+      
+      // Scoped blocks include the BlockMemberSymbols of their nested definitions. This removes them.
+      lazy val localsWithoutBms: Set[Local] = obj match
+        case s: ScopedObject.ScopedBlock =>
+          val rmv = children.collect:
+            case c @ ScopeNode(obj = s: ScopedObject.Referencable[?]) => s.bsym
+          obj.definedLocals -- rmv
+        case _ => obj.definedLocals
```
AnsonYeung, line 234: This sounds like an assumption that other types of blocks couldn't contain symbol of a liftable scope object?

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
```diff
@@ -0,0 +1,403 @@
...
+            case _ if ignored.contains(obj.toInfo) => false
+            case _ if isModOrTopLevel => false
+            case ScopedObject.Func(isMethod = S(true)) => false
+            case _: ScopedObject.Loop | _: ScopedObject.ClassCtor | _: ScopedObject.ScopedBlock | _: ScopedObject.Companion => false
+            case _ => true
+        impl
+        
+      lazy val liftedChildNodes: List[ScopeNode[?]] =
+        if isLifted then this :: Nil
+        else children.flatMap(_.liftedChildNodes)
```
AnsonYeung: This is unused code so it should be removed. Also why children is not added if it is lifted?

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
```diff
@@ -0,0 +1,403 @@
...
+      
+      // Finds the first parent that is a lifted object, i.e. a non-ignored definition, or the top level
```
AnsonYeung:
```suggestion
      // Finds the first ancestor that is a lifted object, i.e. a non-ignored definition, or the top level
```

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
```diff
@@ -0,0 +1,403 @@
...
+      lazy val firstLiftedParent: ScopedObject =
+        if !isLifted then
+          parent match
+          case Some(value) => value.firstLiftedParent
+          case None => obj // unreachable
```
AnsonYeung: Is this really unreachable? If so this should definitely not silently return something else here. If not then the comment is misleading.

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
```diff
@@ -0,0 +1,403 @@
...
+  def getNode(blk: Scoped): ScopeNode = getNode(getUID(blk))
+  // From the input block or definition, traverses until a function, class or new scoped block is found and appends them.
+  class ScopeFinder extends BlockTraverserShallow:
+    var objs: List[ScopedObject] = Nil
+    override def applyBlock(b: Block): Unit = b match
+      case s: Scoped =>
+        val id = fresh.make
+        objs ::= ScopedObject.ScopedBlock(id, s)
+      case l: Label if l.loop =>
+        objs ::= ScopedObject.Loop(l.label, l.body)
```
AnsonYeung: what happens to the rest block of the Label here?

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
```diff
@@ -9,386 +9,458 @@ import hkmc2.codegen.*
...
           case TryBlock(sub, finallyDo, rest) =>
             // sub and finallyDo could be executed sequentially, so we must merge
             rec(sub) |> merge
             rec(finallyDo) |> merge
-            applyBlock(rest)
+            applySubBlock(rest)
           case Return(res, false) =>
             applyResult(res)
             hasReader = Set.empty
             hasMutator = Set.empty
```
AnsonYeung, line 371: I could not figure out what these 2 lines does and commenting them out changes none of the test. Also why is this cleared anyway? Is this 2 lines still valid if the return is inside a branch?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: This doesn't work, since box has no way to get the value while not forcing the computation

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: Removing both the runtime sym check and the handler syms check causes the tests to fail, since some of the handler paths do not have a disamb symbol. But the runtime sym check subsumes the handler path check anyway

hkmc2/shared/src/test/mlscript/backlog/Lifter.mls
CAG2Mark: OK, what about
```js
return (params) => {
  let ret = super(parentMainParams)(parentAuxParams);
  // ctor
  return ret;
}
```
?

(We don't want to do this stuff anyway since it's a bit crazy)

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -736,607 +421,852 @@ class Lifter()(using State, Raise, Config):
...
+    val children = root.children
+    children.foreach: c =>
+      ctx.rewrittenScopes.put(c.obj.toInfo, createRewritten(c))
+    
+    val topLevelRewriter = new ScopeRewriter
+    
+    val (syms, top) = root.obj.contents match
+      case Scoped(syms, body) =>
+        (syms.toSet, body)
+      case b => (Set.empty, b)
```
AnsonYeung, line 1258: Does this assume root have flattened Scoped?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -109,362 +77,115 @@ object Lifter:
...
-                  msg"Modules are not yet lifted." -> N :: Nil,
-                  N, Diagnostic.Source.Compilation
-                ))
-                modules ::= c
-                ignored += c.sym
-              else if c.k is syntax.Obj then
-                objects ::= c
-            case _ => ()
-          super.applyDefn(defn)
+  private def createMetadata(s: ScopeNode): LifterMetadata =
```
AnsonYeung: Obselete comment? d is not a parameter

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -109,362 +77,115 @@ object Lifter:
...
+    
+    // hack: ClassLikeSymbol does not extend DefinitionSymbol directly, so we must
+    // use a map to convert 
     
-    // search for defns nested within a top-level module, which are unnecessary to lift
-    def inModuleDefns(d: Defn): Set[BlockMemberSymbol] =
-      val nested = ctx.nestedDefns(d.sym)
-      nested.map(_.sym).toSet ++ nested.flatMap: nested =>
-        if modules.contains(nested.sym) then inModuleDefns(nested) else Set.empty
+    val moduleObjs = nestedScopeNodes.collect:
```
AnsonYeung: Why define another variable called nestedScopeNodes which have a different name to the field and may cause confusion? I think using the field inline is better

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
```diff
@@ -0,0 +1,403 @@
...
+    
+    // Scoped nodes which could possibly be lifted to the top level.
+    sealed abstract class Liftable[T <: Defn] extends Referencable[T]:
+      val defn: T
+    
+    // The top-level scope.
+    case class Top(b: Block) extends ScopedObject[Block] // b may be a scoped block, in which case, its variables represent the top-level variables.
+    case class Class(cls: ClsLikeDefn, isObj: Bool) extends Liftable[ClsLikeDefn]:
+      val defn = cls
+    case class Companion(comp: ClsLikeBody, par: ClsLikeDefn) extends Referencable[ClsLikeBody]
```
AnsonYeung:
```suggestion
    case class Companion(body: ClsLikeBody, classComp: ClsLikeDefn) extends Referencable[ClsLikeBody]
```

why not

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -485,18 +206,16 @@ class Lifter()(using State, Raise, Config):
...
-            case Some(path) if isIgnoredPath(path) => ()
-            case Some(Select(RefOfBms(s, _), Tree.Ident("class"))) =>
-              if clsSyms.contains(s) then extendsGraph += (s -> defn.sym)
-            case Some(RefOfBms(s, _)) =>
-              if clsSyms.contains(s) then extendsGraph += (s -> defn.sym)
-            case _ if !ignored.contains(defn.sym) =>
+            // for now, allow selecting runtime symbols
+            case Some(Select(qual = Value.Ref(l, _))) if State.runtimeSymbol is l => ()
+            case Some(RefOfBms(_, S(s: ClassSymbol))) =>
+              if nestedScopes.contains(s) then extendsGraph += (s -> isym)
```
AnsonYeung: inheritanceGraph may be a better name (since this is the inheritance tree)

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -736,607 +421,852 @@ class Lifter()(using State, Raise, Config):
...
+            newPList :: duped :: Nil,
+            newPList.params.map(_.sym) ::: duped.params.map(_.sym),
+            duped.restParam.map(_.sym))
+        // we need to append an empty param list so calling this function returns a lambda
+        case Nil => 
+          (
+            newPList :: PlainParamList(Nil) :: Nil,
+            newPList.params.map(_.sym),
+            N
+          )
```
AnsonYeung: This code is unreachable by tests right now, and I suspect this code is wrong. If we have a function that have no param list reaching here, how does adding a extra param list result in the correct semantics?

AnsonYeung (commented):
I can't really look through every single line of the code. The main problem I have about the current state is that I found some code that is unreachable (never tested). If those branch is truly unreachable they should just crash instead of returning something that may not be correct. There is also some instance of dead variable and I wonder if there is some lint option to find them all and remove them.

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
CAG2Mark: I'm not sure what you mean. Only a scoped blocks have this `symbols` set that can contain symbols of nested definitions.

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
CAG2Mark: Why `classComp`

CAG2Mark:
Thanks, I'll address these tomorrow.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: Don't forget to add a regression test that does reach this case, regardless of how this is addressed.

LPTK:
> (For the record, Anson did not work on this or the previous lifter)

I know, but Anson is very familiar with what the lifter does and the IR the lifter works on. Also, I can't review the PR at the moment since I am on a family vacation.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -1010,7 +1002,7 @@ class Lifter(topLevelBlk: Block, handlerPaths: HandlerPaths)(using State, Raise,
     
     def rewriteRef(using ctx: LifterCtxNew): Call =
       if isTrivial then lastWords("tried to rewrite a ref to a trivial function")
-      aux.force
+      aux.get // forces computation
```
LPTK: We should rename `get` to `force` to make its semantics clearer (same for `get_!`).

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
CAG2Mark: It was meant to be a function that returned the "topmost" set of lifted definitions from a certain node. I forgot what it was used for

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: Yes. It will still work if it's not flattened, but it will generate less efficient code.

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
CAG2Mark: OK, that's my bad for not having a relevant test.

It changes the behavior on this test:
```fs
fun f(x) =
  fun g() =
    set x += 1
    x
  if true do return g()
  set x += 1
  x
```
Basically, it says that whatever happens inside a branch is irrelevant to the outside if we return inside that branch. Emptying these sets ensures this branch's analysis results are not merged after this recursion is complete.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: It's not possible to create a test that reaches this case. But, if a function with no parameter list somehow has a naked reference, then this will deal with that. Although I'm not even sure what the correct semantics are for that.

I will just add a warning here for now.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: The problem is that it overrides the `get` method in `Box`, which IMO should definitely not be called `force`. Perhaps we could add an extra method called `force`?

CAG2Mark:
OK, I realized there's a few tricky cases with objects and modules that I need to sort out again...

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
AnsonYeung: If I understand this correctly, this is for module companions. Then why not name the class companion of the module properly? It's not the parent of the module right?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
AnsonYeung: A warning would be fine I think as it at least say something, the problem I have with the current code is that I think it's just wrong.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
AnsonYeung: For now this works, I think later we should add disamb symbols so that lifter know these selection resolve to some other built in modules.

hkmc2/shared/src/test/mlscript/backlog/Lifter.mls
AnsonYeung: Probably? I'm just pointing this out since this is left as a comment

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
AnsonYeung: You can add both force and get to the base class

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
AnsonYeung: I confused the definedLocals method with the definedVars method. They do different things

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
AnsonYeung: hm I see, if this is not used and won't be used in the future then we should remove this. We should not leave dead code in the code base

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
AnsonYeung: I see, just add the test then.

CAG2Mark:
Note that this is not yet ready to merge. I still need to deal with some problems with objects.

Honestly, I might just rewrite all objects before lifting because that will make things 100x easier

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
AnsonYeung: In `Lowering.scala`:
```scala
    case sel @ SynthSel(prefix, nme) =>
      // * Not using `setupSelection` as these selections are not meant to be sanity-checked
      subTerm(prefix): p =>
        k(Select(p, nme)(N))
```
You can change this to
```scala
        k(Select(p, nme)(sel.sym.collect {case s: DefinitionSymbol[?] => s}))
```
Then it should propagate the symbol

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: Yes, in that case we should raise an internal compiler error, with an explanation of what went wrong. There's no point in generating wrong code when we find something thta should never happen.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: Method `get` in `Box` should also be called `force` because the semantics is fundamentally related to laziness (the box can be lazy and its computation can cycle back, hence the option).

`Eager` already has a `value` field that can be accessed when we know we're not dealing with something lazy.

CAG2Mark:
OK, objects are causing a lot of problems. There are a few options I have right now:
- Don't lift objects for now. I actually do not think having objects in the IR is a good idea, since it is trivial to create singleton classes, and we can express that they are singleton by specifying they are objects at the definition site.
- Properly support objects in the lifter, which will make it even more complicated and annoying. The main problem right now is cases like this :
  ```fs
  object M with
    object MM
  M.MM
  ```
  I have to make a new public field for `MM`. But to properly rewrite the selection `M.MM`, I need to add a lot of ad-hoc stuff just to support this case.
- Write another pass that rewrites some objects as classes. Then everything work properly, but it feels a bit unclean. I actually already have such a class written, but I haven't pushed my changes yet since there's still a few small problems.

Not sure what the best option is though...

LPTK:
> I actually do not think having objects in the IR is a good idea, since it is trivial to create singleton classes,

But the semantics is different. As discussed before, what makes object definitions useful is their initialization semantics (they allow for indirect self-reference).

```fs
fun f = print(A)
object A with
  f // A is not fully initialized yet but can still be accessed
  val x = 1
//│ > A {}

// By contrast:
fun f = print(A)
class Acls with
  f // A is not accessible at all here
  val x = 1
val A = new Acls
//│ > undefined
//│ A = Acls { x: 1 }

// We still need an initialization checker to catch things like this:
fun f = print(A.x)
object A with
  f
  val x = 1
//│ > undefined
```

That said, it is probably (?) not a crucial feature.

> and we can express that they are singleton by specifying they are objects at the definition site.

Could you elaborate what you mean by that?

> * Properly support objects in the lifter, which will make it even more complicated and annoying. The main problem right now is cases like this :
>   ```fsharp
>   object M with
>     object MM
>   M.MM
>   ```
>   I have to make a new public field for `MM`. But to properly rewrite the selection `M.MM`, I need to add a lot of ad-hoc stuff just to support this case.

Do note that this is pretty much the same situation as supporting classes in classes, which we intentionally do not support in the lifter for now (it will be possible to properly support once we resolve all selections statically).

> * Write another pass that rewrites some objects as classes. Then everything work properly

Except it doesn't, as it changes the initialization behavior.

So, either we change the initialization semantics uniformly (making top-level objects initialize the same as nested ones after lifting, ie the same as simple classes + `val` members), or we support the existing one properly in the lifter. The former seems much easier and is probably the way to go for now.

For this PR, though, I suggest omitting object support. We should add that support in a follow-up PR.

LPTK:
Also, if we go with the former approach, then there's no real need to have an `object` keyword and a corresponding IR form at all.

CAG2Mark:
> Could you elaborate what you mean by that?

OK, I previously thought the "different initialization semantics" was just that it's a singleton object and we could use a flag or something at the definition site to denote that. I didn't realize they had the semantics you mentioned.

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
CAG2Mark: I'm not following. `body` is what contains the companion, i.e. the `module` part of the definition. In the IR, the module/companion is contained within the class definition, hence the name `par`.

AnsonYeung: parent means the extend clause no? Also it's weird to use comp inside companion imo, i guess it's debatable

CAG2Mark: ok, I will just call it `cls` I guess. But imo `comp` is fine

CAG2Mark:
BTW it should be ready

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -987,11 +980,6 @@ class Lifter(topLevelBlk: Block)(using State, Raise, Config):
     
     def rewriteRef(using ctx: LifterCtxNew): Call =
       if isTrivial then lastWords("tried to rewrite a ref to a trivial function")
-      if fun.params.isEmpty then
```
LPTK: Maybe just leave `assert(fun.params.nonEmpty)`?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -1052,8 +1052,13 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
       if shouldFlattenScopes then ScopeFlattener().applyBlock(withHandlers1)
       else withHandlers1
     
+    // (new ObjectRewriter).transform(scopeFlattened)
```
LPTK: Leftover commented code? Same below.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -109,362 +77,120 @@ object Lifter:
...
-    val isMod = d match
-      case c: ClsLikeDefn => c.companion.isDefined // TODO: refine handling of companions
-      case _ => false
+    val moduleObjs: List[ScopedObject.Companion | ScopedObject.Class] = s.allChildNodes.collect:
+      case s @ ScopeNode(obj = o: ScopedObject.Companion) if !s.inModOrTopLevel => o
+      case s @ ScopeNode(obj = o: ScopedObject.Class) if !s.inModOrTopLevel && o.isObj => o
     
-    val inModTopLevel = if isMod then inModuleDefns(d) else Set.empty
-    ignored ++= inModTopLevel
+    // TODO: refine handling of companions
```
LPTK: In what way should they be refined?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -736,607 +423,842 @@ class Lifter()(using State, Raise, Config):
...
+    val cap = usedVars.reqdCaptures(s.toInfo)
 
-        val newCtx = ctx
-          .replCapturePaths(newCapturePaths)
-          .replLocalPaths(newLocalsPaths)
-          .addIsymPaths(newIsymPaths)
-          .replIgnoredBmsPaths(newBmsPaths)
+    val fresh = FreshInt()
+    
+    val sortedVars: Array[(ctorSyms: (local: Local, vs: VarSymbol), param: Param, valDefn: ValDefn)] = cap.toArray.sortBy(_.uid).map: sym =>
```
LPTK: Pls break this long line into two lines.

LPTK (approved):
As mentioned before, I can't really review the PR in depth for now. But it's  been open for long enough, so I think we should merge it. And we'll discuss the follow up work on objects/modules later in person.

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
LPTK: What does this `Companion` case class represent? Does it represent a pair of a class along with its companion? Then why not call these `clsBody` and `compDefn`?

LPTK:
> BTW it should be ready

Don't forget that we need to resolve the open conversations.

hkmc2/shared/src/main/scala/hkmc2/codegen/ScopeData.scala
CAG2Mark: It's just supposed to represent a companion. It just contains a reference to the `ClsLikeDefn` in which it is contained, for convenience

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: Actually, I remember you said we won't have nested modules in the future, so we can ignore this for now or remove the comment.

The comment originally meant that we can lift modules if they aren't dependent on local variables, but it's actually a bit more complicated. We need to reset the class's fields and run the constructor each time the module is reached. At that point, we could just lift all modules by passing the required locals to the constructor and having them as private fields in the module.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
CAG2Mark: Whoops lol

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: This is already asserted when the computation of the aux param list in `mkAuxDefn` is forced

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
CAG2Mark: I've added a TODO for now

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: @CAG2Mark will you address this?

hkmc2/shared/src/test/mlscript/backlog/Lifter.mls
LPTK: @CAG2Mark will you address this?

hkmc2/shared/src/test/mlscript/staging/Syntax.mls
LPTK: We can leave this one open, as it's not resolved yet.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: Can we resolve this?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: The scheme you describe doesn't seem right (wouldn't it violate reentrancy?). Anyway, let's leave it at that for now.

LPTK: Pls either remove or elaborate the comment.

CAG2Mark:
OK, everything is addressed

PR #385, opened 2026-01-17
Printer (`:slot`) Improvements

CAG2Mark:
Makes the following improvements to the pretty printer (`:slot`):
- Support dynamic selections
- Print class/module constructors and methods
- Fix incorrectly printed function owner
- Discriminates between different class-like objects

Hopefully this will make the printer much more helpful when dealing with classes.


LPTK (approved):
Thanks, LGTM.

PR #386, opened 2026-01-18
Fix classes inside functions

FetBoba:
Wasm backend crashes when classes defined in an outer scope are referenced inside nested functions (field access and class pattern matching). Fix: implemented lazy/on-demand creation/registration of top-level class struct types during lowering so the needed wasm type exists at each use site.


hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -201,6 +201,15 @@ class Ctx(
       namedTypes(_) = numIdx
     TypeIdx(typeInfo.id.getOrElse(numIdx))
 
+  private def resolveTypeByName(sym: BlockMemberSymbol): Opt[NumIdx] =
```
LPTK: Why do we ever need to do this? (Resolve a type by its string name.) That seems to indicate there's a design problem somehwere in the wasm backend.

Derppening:
Could you please add a description of what this specifically fixes and how you fixed the problem?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: It is a fallback method in case if there are two symbols with same name. It was used when trying to find the issue. It is not needed anymore. Sorry for not deleting it beforehand.

LPTK:
Please merge with upstream and address @Derppening's comment.

FetBoba:
> Could you please add a description of what this specifically fixes and how you fixed the problem?

Ok. I added it to description of this pr.

LPTK:
> Wasm backend crashes when classes defined in an outer scope are referenced inside nested functions (field access and class pattern matching). Fix: implemented lazy/on-demand creation/registration of top-level class struct types during lowering so the needed wasm type exists at each use site.

Why would we need this? It seems to me that you should just make one pass to register all the types of the program, and then a separate pass to generate the code of functions.

FetBoba:
> Why would we need this? It seems to me that you should just make one pass to register all the types of the program, and then a separate pass to generate the code of functions.

I thought that there might be some complications with correct order of registering types.

I will try to implement preregistering of all types. Also, by all types you mean top-level classes? Can you clarify this?

LPTK:
> Also, by all types you mean top-level classes? Can you clarify this?

Yes. Not sure what else you think I could mean.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -49,6 +49,62 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+      && (defn.k is syntax.Cls)
+      && defn.auxParams.isEmpty
+      && defn.parentPath.isEmpty
+      && defn.methods.isEmpty
+      && defn.companion.isEmpty
+      && (defn.preCtor match
+        case End(_) => true
+        case _ => false)
+
+  private def predeclareTopLevelDefnTypes(b: Block)(using Ctx): Unit = b match
```
Derppening: Nit: `declareTopLevelDefnTypes`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -49,6 +49,62 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+      if isSupportedTopLevelClass(defn) then
+        getOrCreateClassType(defn)
+      predeclareTopLevelDefnTypes(rst)
+    case Define(_, rst) =>
+      predeclareTopLevelDefnTypes(rst)
+    case Begin(_, rst) =>
+      predeclareTopLevelDefnTypes(rst)
+    case Scoped(_, body) =>
+      predeclareTopLevelDefnTypes(body)
+    case _ => ()
```
Derppening: Instead of doing a `match` on individual block-like types, would `b.subBlocks` be more comprehensive and resilient to future changes to the Block IR? @LPTK

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -49,6 +49,62 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+
+  private def getOrCreateClassType(clsLikeDefn: ClsLikeDefn)(using Ctx): TypeIdx =
+    ctx.getType(clsLikeDefn.sym).getOrElse {
```
Derppening: Nit: Use braceless syntax.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1092,6 +1122,10 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
       )
     )
     
+    // Two-pass scheme: register all supported top-level class struct types before compiling any
+    // functions, so nested function codegen never observes missing class types.
```
Derppening: Nit: Avoid double negatives - "so all class types are available during nested function codegen"

Derppening (commented):
Most of the changes look good, please address the following comments.

Please also all brief documentation for new `private`  functions.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
LPTK: I would rather make the default case BlockTail and add the missing case (Match).

FetBoba: If I make the default case BlockTail, then match becomes non-exhaustive. I will need to add cases for Assign, TryBlock, Label, HandleBlock, and Match and this function will become messy.

LPTK: That's on purpose. Default cases are unsafe as they will make you forget to handle some cases with appropriate logic. For example, why wouldn't you process the rest of Match and Assign instead of ignoring it?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -49,6 +49,64 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+    case Define(_, rst) =>
+      declareTopLevelDefnTypes(rst)
+    case Begin(_, rst) =>
+      declareTopLevelDefnTypes(rst)
+    case Scoped(_, body) =>
+      declareTopLevelDefnTypes(body)
+    case _ => ()
+
+  /** Gets or creates the Wasm struct type for a supported class definition. */
+  private def getOrCreateClassType(clsLikeDefn: ClsLikeDefn)(using Ctx): TypeIdx =
```
LPTK: Why would you need this method? We're now making one single pass to create all wasm types. They should never already be created before that, and there should never be a need to create any other after that. The first pass should just *return* the relevant type mapping to be used by further code gen passes.

hkmc2/shared/src/test/mlscript/wasm/Basics.mls
```diff
@@ -158,3 +158,64 @@ class Foo(val a)
 //│   (elem declare func $entry))
 //│ Wasm result:
 //│ = 42
+
+:wat
```
LPTK: Is there anything interesting to see here, besides the fact that it just does the right thing?
```suggestion
```

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -49,6 +49,73 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+      && defn.auxParams.isEmpty
+      && defn.parentPath.isEmpty
+      && defn.methods.isEmpty
+      && defn.companion.isEmpty
+      && (defn.preCtor match
+        case End(_) => true
+        case _ => false)
+
+  /** Declares supported top-level class types (needed for nested function codegen). */
+  private def declareTopLevelDefnTypes(b: Block)(using Ctx): Unit = b match
```
Derppening: Nit: I still don't think this function name accurately reflects what it does. It takes a `Block` and recursively creates all types that exist in the `Block`, but `main` is not the only `Block` AFAIK.

Perhaps rename it to `createDefnTypes`?

FetBoba: Sounds reasonable. Done

Derppening (approved):

PR #387, opened 2026-01-30
Add optional pass to compile first-class functions to classes

NeilKleistGao:
- Create flag `:defunc` for defunctionalization
- Collect functions and generate corresponding function classes with the call function
- Substitute all first-class functions with corresponding class instantiation
- Invoke the call function for each first-class function


LPTK:
Cool, but this is not defunctionalization, it's just compiling first class functions to classes. (That's fine, it's what we need, but the terminology is wrong and should be changed.)

LPTK:
Please add tests for the things we discussed in private and fix them appropriately.

hkmc2/shared/src/main/scala/hkmc2/codegen/FirstClassFunctionTransformer.scala
```diff
@@ -85,8 +82,8 @@ class FirstClassFunctionTransformer(using Elaborator.State, Raise) extends Block
             Assign(tmp,
               Instantiate(false, cls, fd.capturedVariables.map(v => Value.Ref(v, None).asArg)), k(Value.Ref(tmp, None))))
         case None if l.tsym.map(_.k == syntax.Fun).getOrElse(false) => // This symbol denotes a function and is defined in another module
-          val tmp = new TempSymbol(None, "tmp")
-          val cls = ref.selSN("firstCls")
+          val tmp = new TempSymbol(None)
+          val cls = Value.Ref(new BlockMemberSymbol("Lambda$" + l.nme, Nil, true), disamb)
```
LPTK: Does it really make sense to create a brand new symbol here that's immediately used in a reference? 🤔

hkmc2/shared/src/main/scala/hkmc2/codegen/FirstClassFunctionTransformer.scala
```diff
@@ -122,12 +123,20 @@ class FirstClassFunctionTransformer(using Elaborator.State, Raise) extends Block
...
             case ref @ Value.Ref(sym, _) => sym match
-              case _: VarSymbol |  _: TempSymbol =>
-                k(Call(ref.selSN("call"), args2)(c.isMlsFun, c.mayRaiseEffects, c.explicitTailCall))
-              case _ => k(Call(fun2, args2)(c.isMlsFun, c.mayRaiseEffects, c.explicitTailCall))
-            case _ => k(Call(fun2, args2)(c.isMlsFun, c.mayRaiseEffects, c.explicitTailCall))
+              case _: VarSymbol |  _: TempSymbol => k(call(ref.selSN("call")))
+              case _ => k(call(fun2))
+            case sel: Select => sel.symbol match
+              case Some(s: TermSymbol) if s.k != syntax.Fun => k(call(sel.selSN("call")))
+              case _ => k(call(fun2)) // An intra-module selection also has no symbol. e.g.,
```
LPTK: This seems like a bug. Why wouldn't there be a symbol, here?

hkmc2/shared/src/main/scala/hkmc2/codegen/FirstClassFunctionTransformer.scala
```diff
@@ -0,0 +1,187 @@
...
+              Instantiate(false, cls, fd.capturedVariables.map(v => Value.Ref(v, None).asArg)), k(Value.Ref(tmp, None))))
+        case None if l.tsym.map(_.k == syntax.Fun).getOrElse(false) => // This symbol denotes a function and is defined in another module
+          val tmp = new TempSymbol(None)
+          val cls = Value.Ref(new BlockMemberSymbol("Lambda$" + l.nme, Nil, true), disamb)
+          Scoped(Set(tmp),
+            Assign(tmp,
+              Instantiate(false, cls, Nil), k(Value.Ref(tmp, None))))
+        case _ => k(p)
+      case sel: Select => sel.symbol match
+        case Some(s: TermSymbol) if s.k == syntax.Fun =>
```
LPTK:
```suggestion
        case Some(s: TermSymbol) if s.k is syntax.Fun =>
```

hkmc2/shared/src/main/scala/hkmc2/codegen/FirstClassFunctionTransformer.scala
```diff
@@ -0,0 +1,187 @@
...
+          val blkSym = mapping.find(p => p._1.tsym match
+            case Some(t) => t == s
+            case _ => false
+          )
```
LPTK:
```suggestion
          val blkSym = mapping.find(_._1.tsym.contains(s))
```

CAG2Mark:
Hi, I just saw this PR and thought it looked interesting, so I took a look. I noticed this generates a lot of dead code. We really should add a dead code elimination pass at some point.

LPTK:
@CAG2Mark Haven't looked in detail yet. Why does it generate dead code? Would be better not to generate it in the forst place (notably, for performance reasons).

CAG2Mark:
> @CAG2Mark Haven't looked in detail yet. Why does it generate dead code? Would be better not to generate it in the forst place (notably, for performance reasons).

I ran the PR and it appears to not work for nested functions, so I assume the lifter should be run before this pass. I also noticed the PR generates an aux class for every function, but not all nested functions require an aux class, since we statically know whether they are used as first-class functions.

LPTK:
Thanks for the feedback. @NeilKleistGao how are nested functions handled? At the risk of repeating myself, anything not explicitly handled should stop the compiler (instead of miscompiling).

The useless generated code thing should be easily fixable.

hkmc2/shared/src/main/scala/hkmc2/codegen/FirstClassFunctionTransformer.scala
NeilKleistGao: I don't think so. This should correspond to the following case:
```fs
fun foo(x) = x + 1
fun bar(f) = f(0)

// in a different diff test block
bar(foo)
```
It should be an error, since we cannot find the class for function foo in the current scope. This code is replaced by an error message.

NeilKleistGao:
> Thanks for the feedback. @NeilKleistGao how are nested functions handled? At the risk of repeating myself, anything not explicitly handled should stop the compiler (instead of miscompiling).
> 
> The useless generated code thing should be easily fixable.

I tried to enable lifter and found some problems to fix. I will need to discuss some details of the lifter with @CAG2Mark later. In this PR, I rejected all nested functions and lifter-enabled cases and left TODOs. They can be submitted as another PR later.

NeilKleistGao:
> > @CAG2Mark Haven't looked in detail yet. Why does it generate dead code? Would be better not to generate it in the forst place (notably, for performance reasons).
> 
> I ran the PR and it appears to not work for nested functions, so I assume the lifter should be run before this pass. I also noticed the PR generates an aux class for every function, but not all nested functions require an aux class, since we statically know whether they are used as first-class functions.

I created aux classes for all functions, since we do not know if they will be used as first class function outside the current module. e.g.,
```fs
module Foo with
  fun foo(x) = x + 1

fun foo(f) = f(0)
foo(Foo.foo)
```
If `Foo.foo` is used 10 times in 10 different files, we will get 10 copies of the aux class. But, for sure,  it is not necessary for nested named functions, since they cannot be accessed unless we return them explicitly.

CAG2Mark:
> > Thanks for the feedback. @NeilKleistGao how are nested functions handled? At the risk of repeating myself, anything not explicitly handled should stop the compiler (instead of miscompiling).
> > The useless generated code thing should be easily fixable.
> 
> I tried to enable lifter and found some problems to fix. I will need to discuss some details of the lifter with @CAG2Mark later. In this PR, I rejected all nested functions and lifter-enabled cases and left TODOs. They can be submitted as another PR later.

Please send the problematic test cases to me when you are free.

hkmc2/shared/src/test/mlscript/codegen/FirstClassFunctionTransform.mls
```diff
@@ -209,11 +209,11 @@ bar(foo)
 // in this example, to find the class symbol,
 // we need to store it in the function foo, i.e., `foo.funcCls = Lambda$foo`
 :todo
-:ge
-:re
 bar(foo)
-//│ ═══[COMPILATION ERROR] No definition found in scope for member 'Lambda$foo'
-//│ ═══[RUNTIME ERROR] ReferenceError: Lambda$foo is not defined
+//│ ╔══[COMPILATION ERROR] Cannot transform function foo to the class for now, since it is defined in an unknown module.
```
LPTK: > the class

Which class? (Rhetorical question.)

LPTK:
> If `Foo.foo` is used 10 times in 10 different files, we will get 10 copies of the aux class.

In general, such functions can have many parameters and may thus have many ways of being partially applied. Moreover, many of them may never be partially applied. So I think eagerly creating classes for them is the wrong approach. Besides, does your approiach currently work with separate compilation (importing such a function from a file)?

hkmc2/shared/src/main/scala/hkmc2/codegen/FirstClassFunctionTransformer.scala
NeilKleistGao: Fixed.

hkmc2/shared/src/test/mlscript/codegen/FirstClassFunctionTransform.mls
NeilKleistGao: The error message is no longer reported.

NeilKleistGao:
> So I think eagerly creating classes for them is the wrong approach.

Now, if a function is not used as a first-class function, the corresponding class will not be generated.

> Besides, does your approach currently work with separate compilation (importing such a function from a file)?

Yes. I added a test case to pass `Predef.print` as an argument.

hkmc2/shared/src/main/scala/hkmc2/codegen/FirstClassFunctionTransformer.scala
```diff
@@ -0,0 +1,199 @@
...
+                k(call(fun2))
+            case _ => k(call(fun2))
+      case p: Path => updatePathWithInst(p, false): p2 =>
+        k(p2)
+      case _ => super.applyResult(r)(k)
+
+  class CheckNestedFunctions extends BlockTraverser:
+    override def applyFunDefn(fun: FunDefn) =
+      if fun.sym.nameIsMeaningful then
+        raise(ErrorReport(msg"Nested function ${fun.sym.nme} is not supported yet." -> fun.sym.toLoc :: Nil,
```
LPTK:
```suggestion
        raise(ErrorReport(msg"Nested function ${fun.sym.nme} is not supported by lambda rewriting. Lambda lifting must be performed first." -> fun.sym.toLoc :: Nil,
```

hkmc2/shared/src/main/scala/hkmc2/codegen/FirstClassFunctionTransformer.scala
LPTK: Ok. Did you add a test for it?

hkmc2/shared/src/test/mlscript/codegen/FirstClassFunctionTransform.mls
```diff
@@ -0,0 +1,852 @@
...
+
+
+:todo
+:lift
+:ge
+fun foo(x) =
+  fun bar(y) =
+    x + y
+  bar(x * 2)
+//│ ═══[COMPILATION ERROR] Cannot transform functions to classes when lifter is enabled for now.
```
LPTK: Wait how come?

hkmc2/shared/src/test/mlscript/codegen/FirstClassFunctionTransform.mls
```diff
@@ -0,0 +1,852 @@
...
+    x + y
+  bar(x * 2)
+//│ ╔══[COMPILATION ERROR] Nested function bar is not supported yet.
+//│ ║  l.808: 	  fun bar(y) =
+//│ ║         	  ^^^^^^^^^^^^
+//│ ║  l.809: 	    x + y
+//│ ╙──       	^^^^^^^^^
+
+
+:todo
```
LPTK: Everywhere, pelase stop using `:todo` and use `:fixme // TODO` instead, as I've already pointed out.

hkmc2/shared/src/test/mlscript/codegen/FirstClassFunctionTransform.mls
```diff
@@ -0,0 +1,852 @@
...
+//│     return runtime.safeCall(f2.call(this.#x))
+//│   }
+//│   toString() { return runtime.render(this); }
+//│   static [definitionMetadata] = ["class", "Foo", [null]]; 
+//│ });
+
+
+:expect 0
+:sjs
+Foo(1).foo(x => x - 1)
```
LPTK, line 228: Do you remember what we discussed about methods? I am not sure, because I don't see tests that check we aren't miscompiling things related to applied field accesses.

BTW, is the symbol of `foo` resolved here?

CAG2Mark:
It seems this does nothing for nested functions with the lifter enabled, since this doesn't compile functions with multiple param lists into classes. The tests with the lifter enabled don't break; the transformation just does nothing.

LPTK:
@CAG2Mark do you mean that it fails to do some transformation it should, or that there is nothing to do? If the former, we should come up with a case that crashes/misbehave for the tests.

CAG2Mark:
> @CAG2Mark do you mean that it fails to do some transformation it should, or that there is nothing to do? If the former, we should come up with a case that crashes/misbehave for the tests.

I don't know if rewriting functions with multiple parameter lists is an objective of this PR. Note that those contain lambdas that are created syntactically (not in the IR tree) by the JSBuilder.

LPTK:
What the JSBuilder does doesn't matter that much. What matters is that this IR pass properly handles the case where multi-parameter functions are used while not being fully applied to all their argument lists (in which case a lambda class should be created).

CAG2Mark:
In this case, it appears the call is not rewritten. But I think it quite annoying to detect whether a function is partially or fully applied, since they are rewritten into sequences of calls anyway.

In my opinion, handling that should be done in another pass.

LPTK:
I'm not following. Handling what should be done in another pass?

> But I think it quite annoying to detect whether a function is partially or fully applied, since they are rewritten into sequences of calls anyway.

Either the IR could represent calls as taking multiple argument lists (I don't think there it a strong reason why not) or we should make a pass that flattens argument lists, but the latter would be made more complicated by not having the former.

Anyway, for this PR, we can always just create a lambda if the function being accessed is given an argument list but takes more.

hkmc2/shared/src/test/mlscript/codegen/FirstClassFunctionTransform.mls
NeilKleistGao: I added more test cases:
```fs
class Foo(val x) with
  fun f(y) = y + 1


Foo(0).f(1)
//│ = 2

Foo(0).x
//│ = 0

Foo(x => x).x(1)
//│ = 1


let foo = Foo(1)
foo.x
//│ = 1
//│ foo = Foo(1)

:ge
foo.f(0)
//│ ╔══[COMPILATION ERROR] Cannot determine if f is a function object.
//│ ║  l.745: 	foo.f(0)
//│ ╙──       	^^^^^
//│ = 1

foo.Foo#f(0)
//│ = 1

let foo = Foo(x => x + 1)
//│ foo = Foo(class anonymous)


foo.Foo#f(0)
//│ = 1


:ge
foo.x(0)
//│ ╔══[COMPILATION ERROR] Cannot determine if x is a function object.
//│ ║  l.763: 	foo.x(0)
//│ ╙──       	^^^^^


foo.Foo#x(0)
//│ = 1


class Bar() with
  val x = 0
  val bar = x => x + 1


Bar().bar(1)
//│ = 2

Bar().x
//│ = 0

let bar = Bar()
bar.Bar#bar(1)
//│ = 2
//│ bar = Bar()

bar.x
//│ = 0

let bar = Bar().bar in bar(1)
//│ = 2
```

> BTW, is the symbol of foo resolved here?

In this case, yes. If the symbol can not be resolved properly, we can write `obj.ClassName#FieldName`, as shown above.

hkmc2/shared/src/test/mlscript/codegen/FirstClassFunctionTransform.mls
NeilKleistGao: As I explained above, I plan to implement rewriting with lambda lifting in another PR (since I'm not familiar with the lifter). In this PR, such code will be rejected and not be transformed incorrectly.

hkmc2/shared/src/main/scala/hkmc2/codegen/FirstClassFunctionTransformer.scala
```diff
@@ -141,15 +141,23 @@ class FirstClassFunctionTransformer(using Elaborator.State, Raise) extends Block
...
                 k(call(fun2))
+            case s: DynSelect =>
+              raise(ErrorReport(msg"Cannot determine if the dynamic selection is a function object." -> s.toLoc :: Nil,
+                  source = Diagnostic.Source.Compilation))
+                k(call(fun2))
             case _ => k(call(fun2))
       case p: Path => updatePathWithInst(p, false): p2 =>
         k(p2)
+      case _: Lambda => // TODO: Handle this correctly. Lambda blocks are only used if lifting is enabled.
+        raise(ErrorReport(msg"Lambda functions should be rewritten into function definitions first." -> r.toLoc :: Nil,
```
LPTK: It's an internal error, not a user-level compilation error.

hkmc2/shared/src/test/mlscript/codegen/FirstClassFunctionTransform.mls
LPTK: But isn't the lifter pass orthogonal? It produces a correct IR output. If your pass works correctly, it should be able to process the result correctly (including correctly rejecting cases it doesn't know how to handle).

hkmc2/shared/src/test/mlscript/codegen/FirstClassFunctionTransform.mls
```diff
@@ -0,0 +1,946 @@
...
+//│ (class Foo16 {
+//│   static {
+//│     Foo17 = this
+//│   }
+//│   constructor() {
+//│     runtime.Unit;
+//│   }
+//│   static {
+//│     let lambda11, Lambda$037, tmp43;
+//│     this.Lambda$0 = function Lambda$0() {
```
LPTK: Why are we creating public classes part of the `Foo` module? I can't see why we would want to do that. The generated classes should be local classes that are later lifted out to top-level classes (not classes in the current module) by the lifter (we can run it another time).

hkmc2/shared/src/test/mlscript/codegen/FirstClassFunctionTransform.mls
```diff
@@ -74,52 +86,14 @@ f(0, 0, true)
...
-//│   }
-//│   call(x) {
-//│     return x
-//│   }
-//│   toString() { return runtime.render(this); }
-//│   static [definitionMetadata] = ["class", "Lambda$0", []]; 
-//│ });
-//│ tmp = globalThis.Object.freeze(new Lambda$03());
-//│ tmp.call(42)
+//│ let lambda2; lambda2 = (undefined, function (x) { return x }); lambda2(42)
```
LPTK, line 100: Why is this lambda not transformed? Is there a special case?

hkmc2/shared/src/main/scala/hkmc2/codegen/FirstClassFunctionTransformer.scala
```diff
@@ -18,33 +18,33 @@ class FirstClassFunctionTransformer(using Elaborator.State, Raise) extends Block
...
-      val params = fd.params match
-        case head :: Nil => head.params.map(p => Value.Ref(p.sym).asArg)
-        case _ =>
-          raise(ErrorReport(msg"Unsupported function form." -> fd.sym.toLoc :: Nil,
-            source = Diagnostic.Source.Compilation))
-          Nil
-      
-      Return(Call(f, params)(true, false, false), false)
-
+  class ModuleTransformer(parent: Option[ModuleTransformer], outModulePath: Option[Path], newFds: ListBuffer[FunDefn], funs: HashSet[BlockMemberSymbol], scDefines: HashMap[Path, FunDefn]) extends BlockTransformer(new SymbolSubst):
```
LPTK: Why do we need all this stuff? (The hashmap, etc.)

Again, this is a trivial local transformation. It shouldn't need to keep track of anything non-local...

LPTK: If this is just to access the definition associated with a symbol, you should just define a single function to create a global map from function symbols to their defintions. In principle, the symbols should be unique.

hkmc2/shared/src/main/scala/hkmc2/codegen/FirstClassFunctionTransformer.scala
```diff
@@ -0,0 +1,198 @@
...
+            val cls = outModulePath.map(_.selSN(lamClsSym.nme)).getOrElse(Value.Ref(lamClsSym, None))
+            Return(Instantiate(false, cls, newFd.capturedVariables.map(v => Value.Ref(v, None).asArg)), false)
+          case head :: rest =>
+            val newBody = rec(rest)
+            val lamClsSym = new BlockMemberSymbol("Lambda$" + newFds.size.toString(), Nil, false)
+            val newFd = FunDefn.withFreshSymbol(fd.owner, lamClsSym, head :: Nil, newBody)(false)
+            newFds += newFd
+            val cls = outModulePath.map(_.selSN(lamClsSym.nme)).getOrElse(Value.Ref(lamClsSym, None))
+            Return(Instantiate(false, cls, newFd.capturedVariables.map(v => Value.Ref(v, None).asArg)), false)
+          case Nil => ??? // impossible
```
LPTK:
```suggestion
          case Nil => lastWords("impossible because [...]")
```
Please fill in the `[...]`.

hkmc2/shared/src/test/mlscript/codegen/FirstClassFunctionTransform.mls
CAG2Mark: It's never used as a first-class function

LPTK: I'm not sure we should make this kind of complicating reasoning in this transformation. The proper way to address this case IMO is to have a separate pass transform lambdas bound to variables and not used as first-class functions into proper function definitions, and then this pass doesn't have to worry about them at all.

LPTK: (That's something the lambda lifter could be extended to do.)

hkmc2/shared/src/main/scala/hkmc2/codegen/FirstClassFunctionTransformer.scala
NeilKleistGao: I've added comments for the fields.

hkmc2/shared/src/test/mlscript/codegen/FirstClassFunctionTransform.mls
NeilKleistGao: > It's never used as a first-class function

Yes, and the lambda rewriting has transformed it into a function definition.

hkmc2/shared/src/test/mlscript/codegen/FirstClassFunctionTransform.mls
NeilKleistGao: If I lift these classes out, I will need to duplicate the logic in the lifter to guarantee the generated paths are correct. It seems that currently, lifter cannot handle classes that extend `global.Function`:
```fs
//│ ═══[WARNING] Cannot yet lift definition `Func$0` as it extends an expression.
```

CAG2Mark: You need to add the disambiguation symbol to the `global.Function` selection, and then it will work.

hkmc2/shared/src/test/mlscript/codegen/FirstClassFunctionTransform.mls
```diff
@@ -143,13 +134,19 @@ let f = foo in bar(f)
 //│ = 1
 
 
-fun foo(x) = x + 1
-fun bar(f) = f(0)
```
LPTK: Did you really mean to remove this test?

hkmc2/shared/src/main/scala/hkmc2/codegen/FirstClassFunctionTransformer.scala
```diff
@@ -0,0 +1,160 @@
...
+            k(call(fun))
+        case _ => k(call(fun))
+    case path: Path => applyPath(path)(k)
+    case _: Lambda => lastWords("Lambda functions should be rewritten into function definitions first.")
+    case _ => super.applyResult(r)(k)
+
+  private def desugarMultipleParamList(fd: FunDefn, rest: Block) = fd.params match
+    case Nil => Define(fd, rest)
+    case _ :: Nil => Define(fd, rest)
+    case head :: tail =>
```
LPTK: What's going on, here?

All you have to do is transform
```fs
fun foo(x)(y) = x + y
```
into
```fs
fun foo(x) =
  fun tmp(y) = x + y
  tmp
```

and the rest should _already_ be handled for you! Why are you trying to keep track of free variables and doing all that complicated (and probably incorrect – cf captured mutable variables) logic?

LPTK (changes requested):

hkmc2/shared/src/test/mlscript/codegen/FirstClassFunctionTransform.mls
NeilKleistGao: These two function definitions are duplicated. The actual test cases are still in the file:
```fs
:ge
bar([foo].0)
bar([foo, x => x].1)
//│ ╔══[COMPILATION ERROR] Cannot determine if 0 is a function.
//│ ║  l.153: 	bar([foo].0)
//│ ╙──       	         ^^
//│ ═══[COMPILATION ERROR] Cannot determine if 0$__checkNotMethod is a function.
//│ ═══[COMPILATION ERROR] Cannot determine if Error is a function.
//│ ╔══[COMPILATION ERROR] Cannot determine if 1 is a function.
//│ ║  l.154: 	bar([foo, x => x].1)
//│ ╙──       	                 ^^
//│ ═══[COMPILATION ERROR] Cannot determine if 1$__checkNotMethod is a function.
//│ ═══[COMPILATION ERROR] Cannot determine if Error is a function.
//│ = 0
```

hkmc2/shared/src/main/scala/hkmc2/codegen/FirstClassFunctionTransformer.scala
NeilKleistGao: I reused the lifter's logic to compute the captured variables.

hkmc2/shared/src/main/scala/hkmc2/codegen/FirstClassFunctionTransformer.scala
NeilKleistGao: This conversation is no longer valid since the code has been removed.

hkmc2/shared/src/main/scala/hkmc2/codegen/FirstClassFunctionTransformer.scala
```diff
@@ -112,49 +96,25 @@ class FirstClassFunctionTransformer(using Elaborator.State, Elaborator.Ctx, Rais
...
+        def rec(params: List[ParamList]): Block = params match
+          case head :: Nil => 
+            val funSym = new BlockMemberSymbol("lambda$", Nil, false)
+            val funDef = FunDefn.withFreshSymbol(None, funSym, head :: Nil, fd.body)(false)
+            Scoped(Set(funSym), Define(funDef, Return(Value.Ref(funDef.sym, Some(funDef.dSym)), false)))
+          case head :: rest =>
+            val newBody = rec(rest)
+            val funSym = new BlockMemberSymbol("lambda$", Nil, false)
+            val funDef = FunDefn.withFreshSymbol(None, funSym, head :: Nil, newBody)(false)
+            Scoped(Set(funSym), Define(funDef, Return(Value.Ref(funDef.sym, Some(funDef.dSym)), false)))
```
LPTK: These cases have almost exactly the same code. Why aren't they just casing on `Nil` (which should return `fd.body`) and `::`?

hkmc2/shared/src/main/scala/hkmc2/codegen/FirstClassFunctionTransformer.scala
LPTK: What? Looking at your changes, you did not "reuse the lifter's logic to compute the captured variables". You just implemented my desugaring and left the rest of the work to the lifter, as I suggested. You never deal with captured variables explicitly, so why do you write that?

LPTK (approved):
LGTM now.

PR #335, opened 2025-09-16
Deforestation with `nofib`s in the Benchmark Sub-project

ychenfo:
TODO
- [x] Move `nofibs` to the benchmark subproject
- [x] Fix the paths

PR #389, opened 2026-01-31
Runtime rewrite for handlers

AnsonYeung:
This PR rewrites the runtime to make use of an array instead of allocating a new object for each frame. Some benchmark is necessary before this PR is ready since this is meant to improve the performance. Using a separate `Uint32Array` to store integer values in the array may yield further improvement as it will further reduce gc overhead, but at a cost of a more complicated unwind implementation.


AnsonYeung:
I have experimented with preallocating a fixed size for the array and it yields mixed result. Presumably the extra memory usage of the array might cause extra gc overhead from scanning the array.

AnsonYeung:
As a side note, will it be a good idea to have the benchmark part of #335 merged before the rest of the PR?

LPTK:
> As a side note, will it be a good idea to have the benchmark part of #335 merged before the rest of the PR?

@ychenfo could we at least merge the benchmarking part of your PR, as a separate PR, since it seems that deforestation is still not quite ready?

AnsonYeung:
| File | Throughput compared with no safety |
| ----- | ----- |
| ansi | 0.6470993333 | 
| atom | 0.4951924184 |  
| awards | 0.6192695221 |  
| banner | 0.6187800912 |  
| lcss | 0.729085165 |  
| boyer | 0.6341615584 |  
| boyer2 | 0.6295173238 |  
| calendar | 0.4578341604 |  
| cichelli | 0.5812997215 |  
| circsim | 0.5668062344 |  
| clausify | 0.5275486136 |  
| constraints | 0.5223591955 |  
| cryptarithm2 | 0.6989853689 |  
| cse | 0.7499489294 |  
| gcd | 0.6739480381 |  
| integer | 0.7082590221 |  
| knights | 0.6980119427 |  
| lambda | 0.5171208907 |  
| lastpiece | 0.7146894492 |  
| life | 0.7705185834 |  
| mandel | 0.6794940889 |  
| mandel2 | 0.514759346 |  
| mate | 0.6746412453 |  
| minimax | 0.5930816473 |  
| para | 0.647563787 |  
| power | 0.7629329129 |  
| pretty | 0.6461362767 |  
| primetest | 0.4903394771 |  
| rsa | 0.9102183599 |  
| scc | 0.706914986 |  
| secretary | 0.449899076 |  
| sorting | 0.4522511924 |  
| sphere | 0.6451101703 |  
| Average | 0.6282963069 |

hkmc2/shared/src/main/scala/hkmc2/codegen/StackSafeTransform.scala
```diff
@@ -111,8 +111,8 @@ class StackSafeTransform(depthLimit: Int, paths: HandlerPaths, stackSafetyMap: S
         methods.map(rewriteFn),
         privateFields,
         publicFields, 
-        rewriteBlk(preCtor, L(BlockMemberSymbol("TODO", Nil)), 1), // TODO: preCtor is not translated in handler lowering
-        if isTopLevel && (defn.k is syntax.Mod) then transformTopLevel(ctor) else rewriteBlk(ctor, R(isym), 1),
+        rewriteBlk(preCtor, L(BlockMemberSymbol("TODO", Nil))), // TODO: preCtor is not translated in handler lowering
```
LPTK: I just noticed this weird `"TODO"`. What's going on? Are we generating incorrect code in some cases?

AnsonYeung: This is completely useless bc rewriteBlk won't do anything if the symbol is not translated during handler lowering

LPTK:
@AnsonYeung the benchmarks are merged. Do we need to conseuqently change anything here before merging?

AnsonYeung:
No change needed

LPTK:
Before we merge this, could you please just add the result of a couple of experiments, just so we know we're going in the right direction:
* Performance of nofib benchmarks after ensuring that no effect is ever thrown, which is apparently not the case above.
* Performance of some example code that throws lots deep effects before and after this PR (we expect to see a big improvement). An easy way to do this would be to artificially lower the stack limit for the nofib benchmarks to force the stack effect to be thrown repeatedly.

AnsonYeung:
With outliers removed:
no-outliers
With outlier:
with-outlier

Without the outlier, the average throughput with unlimited stack with stack safety instrumentation is 0.6329537819 and average throughput with 1k stack limit is 0.5655368698. The numbers might be a bit different because I included the whole nofib suite this time (except the cryptarithm1).

For cryptarithm1, the baseline is 16.5s, instrumented is 50.1s, and stack safety with 1k limit is 5.8s. The first 2 both suffer from extremely slow gc stack walking during minor gc.

LPTK:
Oof 16.5s to 50.1s after instrumentation is a bit extreme. I don't really know why there would be such a big difference, given that the programs are supposed to execute still quite similarly when effects are not raised... Any idea? Seems hard to believe that it's just because the stack of the instrumented version is just more expensive to walk (?).

By the way, looks like the tests fail.

AnsonYeung:
Looking into the flamegraph it seems that aside from the gc overhead there is also a big (lazy) deoptimization overhead. And a major gc runs too for some reason
stacksafe-unlimited

AnsonYeung:
~~ok apparently 1 million stack depth is not enough, i run it with a higher depth and now the only overhead is gc at 37.5s (82% IterateRoot, i.e. 30s), the baseline have 70.15% iterate root i.e. 11.55s out of 16s~~ If I run the function the second time, the time become 88s total. So it only helps for the first run for some reason.
flamegraph

my guess is that the stack is twice as large somehow and the gc need to travel twice as many elements on the stack?

CAG2Mark:
BTW when shall we merge this?

LPTK:
When we get the promised benchmarks from @AnsonYeung that show the new system is indeed faster (and not, say, a billion times slower) than the old one.

AnsonYeung:
perf changes at stack depth 100, not all test can run at this stack depth (presumably the Rendering issue bc I can't instrument it).
chart
Seems like it improve on some test but regress on others

LPTK:
Thanks for the results! Is the Y axis the relative execution time?

Wow, that's a bit surprising. I would have expected more of an improvement.

> not all test can run at this stack depth (presumably the Rendering issue bc I can't instrument it)

What do you mean?

AnsonYeung:
Yes it's the relative execution time, some test run slower and the average ratio is basically 1.

There is a circular import in Rendering and Runtime if stack safety is turned on.

LPTK:
> Yes it's the relative execution time, some test run slower and the average ratio is basically 1.

Omg 💀

How do you allocate and extend the big arrays? Do you just push into them? Do you preallocate any space?

> There is a circular import in Rendering and Runtime if stack safety is turned on.

But wasn't this the case before? Is this a new bug? This seems serious, shouldn't it worry us?

AnsonYeung:
> How do you allocate and extend the big arrays? Do you just push into them? Do you preallocate any space?

Yes just push into them. I tried to preallocate 512 elements by new Array(512) before with extra bookkeeping due to not being able to push and there is no improvement and I didn't commit that to this branch.

> But wasn't this the case before? Is this a new bug? This seems serious, shouldn't it worry us?

This is not a new bug of this pr, it had only worked when the rendering file is very small. I think I can introduce a new HandlerLowering mode for translating Rendering.mls that just skip top level checks

AnsonYeung:
In fact iirc there is a slowdown if the array is preallocated (maybe the gc need to scan the whole array)

So I considered in a previous comment somewhere that we use Uint32Array to store statically known integers like param counts/local counts but it will complicate the unwind function, which means it might be slower due to the extra stuff done during unwind

AnsonYeung:
OK it turns out to be a very annoying bug which is easy to fix. The new runtime calling convention failing silently is kind of annoying.

hkmc2/shared/src/main/scala/hkmc2/Config.scala
```diff
@@ -68,7 +68,10 @@ object Config:
     // Whether we check `Instantiate` nodes for effects. Currently, effects cannot be raised in constructors.
     checkInstantiateEffect: Bool = false,
     // A debug option that allows codegen to continue even if an unlifted definition is encountered.
-    softLifterError: Bool = false
+    softLifterError: Bool = false,
+    // Skips instrumenting module constructors, this can be used when the file is statically known to not
+    // raise any effect and cannot use runtime during module construction.
```
LPTK: What do you mean by "cannot use runtime"? Do you mean that it is forbidden from using the Runtime.mls module due to cyclic dependencies? Would be better to be more explicit.

hkmc2/shared/src/main/scala/hkmc2/Config.scala
```diff
@@ -68,13 +68,16 @@ object Config:
...
+    skipModCtor: Bool = false,
```
LPTK: Shouldn't this be called
```suggestion
    doNotInstrumentTopLevelModCtor: Bool = false,
```
?

LPTK:
> OK it turns out to be a very annoying bug which is easy to fix.

What is "it"? Do you mean the inability to run some of tthe benchmarks?

How many benchmarks were omitted, and does the overall result change, now that they can be included?

>  The new runtime calling convention failing silently is kind of annoying.

I suppose you mean the calling convention of instrumented functions. Was that different before?

AnsonYeung:
> What is "it"? Do you mean the inability to run some of tthe benchmarks?

Yes, turns out Runtime.isArrayLike is inserted by JSBuilder directly and it raise stack safety effect in places where it shouldn't. This is fixed.

> How many benchmarks were omitted, and does the overall result change, now that they can be included?

5 benchmarks were omitted

nofib-runtime-compare


Nothing really stands out.

> I suppose you mean the calling convention of instrumented functions. Was that different before?

It's not that much different but the fact that a special object is returned means that the toString() or valueOf() can be overriden to throw error, which will catch most illegal usage. But now we return null and it's simply treated as the unit in mlscript, which means it's easier for it to silently fail.

hkmc2/shared/src/main/scala/hkmc2/Config.scala
LPTK: You marked this as resolved, but I don't see a resolution.

LPTK:
> Nothing really stands out.

Now, the numbers look even less convincing (eg, cse used to be faster but no longer is). Are we just observing noise?

hkmc2/shared/src/main/scala/hkmc2/Config.scala
AnsonYeung: I did change the comment. This is used for resolving the cyclic dependency

LPTK: But there's a reason I pointed to the expression "use runtime", which is grammatically incorrect and too vague, and explicitly rephrased it for you as "the Runtime.mls module".

LPTK:
Superseded by #398.

To quote from @AnsonYeung:

> I found the improved one only raise the effect in single digit but those that regressed quite a lot raised thousands
I think we should just revert the runtime changes and only merge in the fixes
>
> secretary raised 50k times and there is a serious regression
>
> mandel is at 4k, most test I look through is 0-2k

PR #390, opened 2026-02-01
Add Singletons and Unit

b-temirov:
This PR adds singleton support in the wasm backend (lazy global instance + hidden getter) and implements Unit as singleton.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
@@ -10,6 +10,7 @@ import mlscript.utils.*, shorthands.*
 import utils.*
 
 import hkmc2.Message.MessageContext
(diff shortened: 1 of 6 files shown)


LPTK:
> lazy global instance

Any reason it's lazy? It's currently not lazy in JS. We should strive to use the same semantics for both.

b-temirov:
> Any reason it's lazy?

Lazy initialization was suggested by @Derppening as one approach in our Jan 7 meeting. My earlier eager approach inserted singleton constructor into `entry` as a preamble, but that may cause problems when linking multiple wasm modules (future work).

LPTK:
Lazy semantics is probably harder to get right and incurring more overhead. I'd rather we settle on whatever a direct port of the JS semantics implies, which I think simply loads and initializes modules in the order they are imported.

b-temirov:
This PR is superseded by #391, therefore closing it.

PR #391, opened 2026-02-11
Add Singletons (eager init) + Unit

b-temirov:
This PR adds singleton support in the wasm backend using eager initialization (named singleton globals + one module start init function), and implements `Unit` by synthesizing it as one such singleton. This PR supersedes #390.


hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -90,6 +90,36 @@ class FuncInfo(
...
+ *   Symbolic identifier for the global, or `N` if the global is anonymous.
+ * @param valType
+ *   The value type of the global.
+ * @param mutable
+ *   Whether the global is mutable.
+ * @param init
+ *   The initializer expression for the global.
+ */
+class GlobalInfo(
+    val id: Opt[SymIdx],
```
LPTK: Why is this optional? It looks like it's always `Some`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1100,6 +1210,10 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
       Raise,
       Scope
   ): (Document, Str) =
+    singletonByBms.clear()
+    singletonByIsym.clear()
+    singletonInitActions.clear()
```
LPTK: Why these weird `clear` statements?

hkmc2/shared/src/test/mlscript/wasm/Singletons.mls
```diff
@@ -0,0 +1,101 @@
+
+:global
+:wasm
+
+:wat
+object Foo with
```
LPTK, line 7: Does this work with
```suggestion
module Foo with
```
?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -37,6 +37,13 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
 
   private val baseObjectSym: BlockMemberSymbol = BlockMemberSymbol("Object", Nil)
   private val tagFieldSym: TermSymbol = TermSymbol(syntax.MutVal, owner = N, Ident("$tag"))
+  private case class SingletonInfo(
+      globalName: Str,
+      globalTy: RefType
+  )
+  private val singletonByBms: MutMap[BlockMemberSymbol, SingletonInfo] = MutMap.empty
+  private val singletonByIsym: MutMap[ModuleOrObjectSymbol, SingletonInfo] = MutMap.empty
+  private val singletonInitActions: ArrayBuf[Expr] = ArrayBuf.empty
```
Derppening: All of these belong in `Ctx`.

hkmc2/shared/src/test/mlscript/wasm/Singletons.mls
```diff
@@ -0,0 +1,101 @@
...
+//│         (nop))
+//│       (return
+//│         (local.get $this))))
+//│   (export "Bar" (func $Bar))
+//│   (elem declare func $Bar)
+//│   (func $entry (type 5) (result (ref null any))
+//│     (local $Bar (ref null any))
+//│     (local $Foo (ref null any))
+//│     (local $Foo1 (ref null any))
+//│     (local $Bar1 (ref null any))
```
Derppening, line 72: Why are there these object symbols in the local scope of the entry function?

hkmc2/shared/src/test/mlscript/wasm/Singletons.mls
```diff
@@ -0,0 +1,101 @@
...
+//│       (block
+//│         (struct.set $Foo 1
+//│           (ref.cast (ref $Foo)
+//│             (local.get $this))
+//│           (ref.i31
+//│             (i32.const 1)))
+//│         (nop))
+//│       (return
+//│         (local.get $this))))
+//│   (export "Foo" (func $Foo))
```
Derppening: I think singleton constructors shouldn't need to be exported - they are automatically invoked when the Wasm module is initialized and there is no reason why they should be called outside of their defining module.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -292,18 +328,23 @@ class Ctx(
   def containsLocal(sym: Local): Bool = locals.head.contains(sym)
 
   /** Adds a new variable into the global variable scope. */
-  def addGlobal(sym: Symbol): GlobalIdx =
-    val numIdx = NumIdx(locals.last.size)
+  def addGlobal(sym: Symbol, globalInfo: GlobalInfo): GlobalIdx =
+    val numIdx = NumIdx(globals.size)
+    globals += globalInfo
     locals.last(sym) = numIdx
```
Derppening: I am not sure it is a good idea to duplicate the set of global variables in both `globals` and `locals.last(sym)`. Is it feasible to refactor all globals to be read from the `globals` `ArrayBuf`?

Derppening (commented):
As usual, please add documentation to all `private` functions as well.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Instructions.scala
```diff
@@ -270,6 +270,14 @@ object Instructions:
   end array
 
   object ref:
+    /** Creates a `ref.null` instruction. */
+    def null_(heapType: HeapType): FoldedInstr = FoldedInstr(
```
Derppening: Nit: Use `` `null` `` as the function name for consistency.

hkmc2/shared/src/test/mlscript/wasm/Singletons.mls
b-temirov: I tried
```
module Foo with
  val x = 1
Foo.x
```
but the output is a compilation error: `WatBackend::returningTerm for ClsLikeDefn(...) where companion.isDefined not implemented yet`

LPTK: Ok, it's fine not to handle this for now.

hkmc2/shared/src/test/mlscript/wasm/Singletons.mls
b-temirov: I think it's caused by `blockPreamble(t.definedVars)` predeclaring all defined symbols as locals:
```scala
val locals = blockPreamble(t.definedVars) // TODO: remove use of `definedVars` now that we properly put everything in proper Scoped blocks (see the change already done in JSBuilder)
```
so it shouldn't be introduced by the singleton logic.

LPTK: FYI @Derppening @b-temirov we have moved to a different, simpler way of handling scopes and locally-defined variables. We should no longer rely on `definedVariables`. Check out how JSBuilder now does it.

b-temirov: > ```scala
> // TODO: remove use of `definedVars` now that we properly put everything in proper Scoped blocks (see the change already done in JSBuilder)
> ```
I could raise a separate PR that targets this TODO comment. Should I do so?

b-temirov:
> As usual, please add documentation to all `private` functions as well.

The documentation is now added in the new commit.

hkmc2/shared/src/test/mlscript/wasm/Singletons.mls
LPTK: Ok, sure.

hkmc2/shared/src/test/mlscript/wasm/Singletons.mls
```diff
@@ -0,0 +1,98 @@
...
+  val x = 1
+object Bar with
+  val y = 2
+Bar.y
+//│ Wat:
+//│ (module
+//│   (type $Object (sub (struct (field $$tag (mut i32)))))
+//│   (type $Foo (sub $Object (struct (field $$tag (mut i32)) (field $x (mut (ref null any))))))
+//│   (type $Bar (sub $Object (struct (field $$tag (mut i32)) (field $y (mut (ref null any))))))
+//│   (type (func (result (ref null any))))
```
LPTK, line 19: Are these types used? Why aren't they named? And why the duplication?

LPTK (approved):
Otherwise LGTM. Please respond to/address my last comment and open the promised other PR.

hkmc2/shared/src/test/mlscript/wasm/Singletons.mls
b-temirov: These are constructor functions for singleton objects `Foo`, `Bar`. They're unnamed because `id = N` for `FuncInfo` for these constructors. If `id` exists, it emits both function label and export. @Derppening  suggested to remove export earlier.
> I think singleton constructors shouldn't need to be exported - they are automatically invoked when the Wasm module is initialized and there is no reason why they should be called outside of their defining module.

LPTK: Is there no way of adding a label (for documentation) without exporting the type?

Derppening: It is possible. A label for the `type` section can exist regardless of whether the type is external or not.

LPTK: Yeah, so we should use labels here @b-temirov

hkmc2/shared/src/test/mlscript/wasm/Singletons.mls
```diff
@@ -0,0 +1,98 @@
...
+//│   (export "entry" (func $entry))
+//│   (elem declare func $entry)
+//│   (func  (type 6)
+//│     (block
+//│       (global.set $Foo$inst
+//│         (ref.cast (ref null $Foo)
+//│           (call 0)))
+//│       (global.set $Bar$inst
+//│         (ref.cast (ref null $Bar)
+//│           (call 1))))))
```
Derppening: Nit: I suggest placing this before the `entry` function.

Derppening (commented):
There are still several private functions that are missing documentation. Please fix those as well.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
```diff
@@ -138,7 +138,7 @@ case class StructType(
 /** A type representing an array type. */
 case class ArrayType(
     elemType: Type,
-    mutable: Bool,
+    mutable: Bool
```
LPTK: 👎

b-temirov:
> There are still several private functions that are missing documentation. Please fix those as well.

I found that `GlobalInfo.idDoc` was missing documentation, so I added it to new commit. Plus, added docs for `Ctx` for all singleton-related functions.

PR #392, opened 2026-02-12
Add Control Flow

b-temirov:
This PR adds support for control flow constructs (`Label`, `Break`, `Continue`) and fixes pattern matching.


hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -68,17 +61,26 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
-    case _ => N
+  /** Returns singleton metadata when `sym` resolves to a registered singleton object. */
+  private def singletonInfoFor(sym: Local)(using Ctx): Opt[SingletonInfo] =
+    ctx.getSingletonInfo(sym)
 
+  /** Loads the singleton object reference from its backing mutable global. */
   private def singletonGlobalGet(info: SingletonInfo): Expr =
     global.get(GlobalIdx(SymIdx(info.globalName)), info.globalTy)
 
+  /** True when the lowered main block references `Unit` and needs synthesized singleton definition. */
```
LPTK: Can't you just codegen the main body of the program and set a flag to true if you encounter Unit? Then it's correct by construction and you don't need an awkward separate traversal.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -712,7 +721,19 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+  def returningTerm(t: Block)(using Ctx, Raise, Scope): Expr =
+    def asStatement(expr: Expr): Expr =
+      expr.resultType match
+        case S(UnreachableType) => expr
+        case S(_) => FoldedInstr(
+            mnemonic = "drop",
+            instrargs = Seq.empty,
+            stackargs = Seq(expr),
+            resultTypes = Seq.empty
+          )
```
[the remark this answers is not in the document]
LPTK: Do we not have any test that uses `return` in a non-trivial way?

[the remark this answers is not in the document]
b-temirov: Added new test cases to ControlFlow.mls. One of them is `nonTailMatchReturn` - return inside a non-tail match arm, so it covers updated `asStatement` lowering path.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -69,52 +69,39 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+    if ctx.containsSingleton(unitDefn.sym) then return
+
+    if ctx.getType(unitDefn.sym).isEmpty then
+      createDefnTypes(Define(unitDefn, End("")))
+
+    ctx.pushLocal()
+    try
+      returningTerm(Define(unitDefn, End("")))
+      ()
+    finally ctx.popLocal()
```
LPTK: `try`/`finally` is useful if we expect to throw exceptions that can be meaningfully caught while continuing codegen. I don't think that's the case. And other uses of `pushLocal`/`popLocal` don't use it.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -182,6 +182,10 @@ object Ctx:
       globalName: Str,
       globalTy: RefType
   )
+  case class LabelTarget(
```
Derppening, line 169: Nit: Add newline before this class definition.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -260,10 +264,22 @@ class Ctx(
...
+  private var labelTargets: List[(LabelSymbol, Ctx.LabelTarget)] = Nil
   private val singletonByBms: MutMap[BlockMemberSymbol, Ctx.SingletonInfo] = MutMap.empty
   private val singletonByIsym: MutMap[ModuleOrObjectSymbol, Ctx.SingletonInfo] = MutMap.empty
   private val singletonInitActions: ArrayBuf[Expr] = ArrayBuf.empty
 
+  /** Pushes a label target for the dynamic extent of `body` and pops it afterwards. */
+  def withLabel[T](label: LabelSymbol, target: Ctx.LabelTarget)(body: => T): T =
+    labelTargets = (label, target) :: labelTargets
+    try body
+    finally labelTargets = labelTargets.tail
```
Derppening: Is this necessary? I don't think we throw exceptions during codegen as part of normal code generation (except `lastWords`, which signals an immediate termination of the compilation process).

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -69,44 +69,37 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+      privateFields = Nil,
+      publicFields = Nil,
+      preCtor = End(""),
+      ctor = End(""),
+      companion = N,
+      bufferable = N
+    )
+
+  /** Registration path for synthetic Unit runtime state. */
+  private def RegisterUnitSingleton()(using Ctx, Raise, Scope): Unit =
```
Derppening, line 104: Nit: Lower camel case please - `registerUnitSingleton`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -764,7 +762,23 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+    def asStatement(expr: Expr): Expr =
+      if isControlTransfer(expr) then expr
+      else
+        expr.resultType match
+          case S(_) => FoldedInstr(
+              mnemonic = "drop",
+              instrargs = Seq.empty,
+              stackargs = Seq(expr),
+              resultTypes = Seq.empty
+            )
```
Derppening: `FoldedInstr` should never be instantiated manually - You should create a `drop` instruction by using `Instructions.drop` (and add that if it isn't already present).

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1108,11 +1122,99 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+            case S(_) => local.set(target, expr)
+            case N => expr
+
+      def lowerMatchBody(expr: Expr): Expr =
+        matchResLocal match
+          case S(localIdx) => assignTailResult(localIdx, expr)
+          case N => asStatement(expr)
+
+      val matchResInitExpr = matchResLocal.map: localIdx =>
+        local.set(localIdx, result(Value.Ref(State.unitSymbol)))
```
Derppening: It appears that this change causes `Unit` to be unnecessarily emitted for every single module, even if it is not used. Is it possible to change this to initialize to `ref.null any`, and only emit `Unit` when it is necessary?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -69,44 +69,37 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+  private def RegisterUnitSingleton()(using Ctx, Raise, Scope): Unit =
+    val unitDefn = syntheticUnitDefn
+    if ctx.containsSingleton(unitDefn.sym) then return
+
+    if ctx.getType(unitDefn.sym).isEmpty then
+      createDefnTypes(Define(unitDefn, End("")))
+
+    ctx.pushLocal()
+    returningTerm(Define(unitDefn, End("")))
+    ctx.popLocal()
```
Derppening, line 113: I don't really understand why we need to push/pop the local scope here. Could you explain why?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -69,44 +69,37 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+      methods = Nil,
+      privateFields = Nil,
+      publicFields = Nil,
+      preCtor = End(""),
+      ctor = End(""),
+      companion = N,
+      bufferable = N
+    )
+
+  /** Registration path for synthetic Unit runtime state. */
```
Derppening: Nit:


```suggestion
  /** Registers the synthetic `Unit` singleton. */
```

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
Derppening: I think we need to traverse `arms` as well to create types declared within match arms.

Derppening (commented):
Please address the comments and resolve the merge conflicts.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
Derppening: Nit (and unrelated to this PR): Shouldn't we use `scrutExpr` here? Right now the `scrutExpr` is evaluated twice.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
b-temirov: I use this because compiling `Define(unitDefn, ...)` through `returningTerm` allocates `this` local for the object ctor. It ends up in the current function’s local frame and is emitted as extra locals of the caller. So the push/pop is to prevent local leakage.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
b-temirov: Since this change is under control flow scope, I applied the fix.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
LPTK: What is the currrent function frame where `Define` occurs? In principle, all `Define`s at this point are top-level.

LPTK: Basically the way this pushing/popping of a local variable buffer stack doesn't make much sense, as @Derppening knows. Eventually we'll want to simplify this away.

hkmc2/shared/src/test/mlscript/wasm/ControlFlow.mls
```diff
@@ -0,0 +1,291 @@
...
+    while true do
+      if i is
+        3 then return i
+        else set i += 1
+loopMatchReturn()
+//│ Wasm result:
+//│ = 3
+
+
+:wat
```
LPTK, line 198: Is this useful? Don't we already have a similar literal pattern matching `:wat`?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
Derppening: Why did you resolve this? This is clearly not yet addressed.

LPTK:
I'll merge this for now and let @Derppening address the remaining discussions in a follow-up PR.

PR #393, opened 2026-02-13
Lifter fix

CAG2Mark:


hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -373,7 +373,8 @@ class Lifter(topLevelBlk: Block)(using State, Raise, Config):
                 k(Value.Ref(newSym, N))
             
             // Other naked references to BlockMemberSymbols.
-            case S(r) =>
+            // Do not rewrite selections.
```
LPTK: Ok, but can you please elaborate a bit in the comment what's going on, here? Why do we not rewrite selections?

LPTK (approved):

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: I've expanded the comment in the code.

PR #394, opened 2026-02-14
Add nofib tests to the benchmark sub-project

ychenfo:


hkmc2/shared/src/test/mlscript/nofib/mate.mls
LPTK, line 1: Is there any reason to keep the same code as diff-tests? Why duplicate the code of every nofib benchmark? Sounds like a maintenance nightmare.

hkmc2Benchmarks/src/test/bench/nofibs-output.mls
LPTK, line 1: Do we really want all these test in a single file? Why not separate them, so that we can test them individually, if needed?

LPTK (approved):
Q: is there any reason for grouping the benchmarks in these three files, or is this arbitrary?

ychenfo:
> Q: is there any reason for grouping the benchmarks in these three files, or is this arbitrary?

The grouping is arbitrary. Previously when I run the benchmarks if all the tests were in one file it would take too long, so I break all the tests into 3 parts.

> Is there any reason to keep the same code as diff-tests? Why duplicate the code of every nofib benchmark? Sounds like a maintenance nightmare.
> Do we really want all these test in a single file? Why not separate them, so that we can test them individually, if needed?

Oh I see... I guess in the main diff tests (`hkmc2/shared/src/test/mlscript/nofib/*.mls`) I can import the corresponding compiled nofib test in the benchmark project in `hkmc2Benchmarks/src/test/bench/mlscript-compile/nofib` and then run them.

AnsonYeung (approved):
LGTM

PR #395, opened 2026-02-16
Fix `.sbtopts`

CAG2Mark:
The newest version of sbt seems to not parse `.sbtopts` correctly. It seems to parse some of the comments as commands, and results in the following error:
```
Error: Could not find or load main class #
Caused by: java.lang.ClassNotFoundException: #
mkdir: cannot create directory ‘’: No such file or directory
Error: Could not find or load main class #
Caused by: java.lang.ClassNotFoundException: #
Error: Could not find or load main class #
Caused by: java.lang.ClassNotFoundException: #
```
I've changed the file to work around this error for now.

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -8,7 +8,12 @@
 
 # The options below were inspired by a message I got on some runs with more than usual memory retention:
 #   [warn] In the last 10 seconds, 5.067 (50.7%) were spent in GC. [Heap: 0.01GB free of 1.00GB, max 1.00GB] Consider increasing the JVM heap using `-Xmx` or try a different collector, e.g. `-XX:+UseG1GC`, for better performance.
--J-Xms2G # reduces warm-up time the first time the tests are run
--J-Xmx4G # reduces time spent doing GC
+
+# Reduces warm-up time the first time the tests are run
+-J-Xms2G
+
+# Reduces time spent doing GC
+-J-Xmx4G
+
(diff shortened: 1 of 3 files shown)


CAG2Mark:
OK... it looks there was some regression and the entire CI is now borked :upside_down_face:

LPTK:
Thanks.

I have _no idea_ why this changed when switching to Scala `3.8.1` with JDK 17:

image
 
Seems like potentially flaky concurrency problem, but I don't see how that would happen.

Well, I don't want to deal with this right now, so reverted back to Scala `3.7.4`.

LPTK (approved):

LPTK:
> image
> Seems like potentially flaky concurrency problem, but I don't see how that would happen.

For the record, the problem is almost certainly that the handling of symbol state is quite messy at the moment. The compiler context will reuse the `Elaborator.State` of the module being compiled in order to on-demand elaborate any imported module (if not already elaborated). But `State` is not meant to be shared in this way and is not thread-safe. We currently shove in it a `dbg` method that, in the diff-tests, accesses mutable state – this is what governs the printing of the extra symbol info.

Instead of doing this, we should share all the common builtin symbols (like `runtimeSymbol`) in some synchronized pre-initialized cache that's shared by everyone (so they see builtin symbols as having the same identity) but every compilation unit should get its own instance of `State`.

PR #398, opened 2026-02-25
Handler fixes

AnsonYeung:
Supersedes #389 , as the runtime changes does not yield performance improvement.

This PR introduces various fixes related to effect handlers:
- Add option for compiling without instrumentation of top level module constructor
- Add stack depth when runtime function is called
- lifter fixes from @CAG2Mark 
- Runtime.isArrayLike no longer trigger stack safety effect
- Fix short_or/short_and not handling effects properly
- Fixes benchmark paths
- Fixes some situation where the benchmark result is not written, since the ReplHost could be killed before node finished writing the file if `fs.writeFile` is used instead of `fs.writeFileSync`.

---

- [X] Decide how to deal with host-language exceptions.
    -> We now use a finally blocks (`set`/`in`) to gracefully handle user-code exceptions.


hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -1063,7 +1063,7 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
       
     val stackSafe = config.stackSafety match
       case N => withHandlers2
-      case S(sts) => StackSafeTransform(sts.stackLimit, handlerPaths, stackSafetyInfo).transformTopLevel(withHandlers2)
+      case S(sts) => StackSafeTransform(sts.stackLimit, config.effectHandlers.exists(_.doNotInstrumentTopLevelModCtor), handlerPaths, stackSafetyInfo).transformTopLevel(withHandlers2)
```
CAG2Mark: Why not add `using Config` to `StackSafeTransform`?

hkmc2/shared/src/main/scala/hkmc2/codegen/StackSafeTransform.scala
```diff
@@ -111,8 +111,8 @@ class StackSafeTransform(depthLimit: Int, paths: HandlerPaths, stackSafetyMap: S
         methods.map(rewriteFn),
         privateFields,
         publicFields, 
-        rewriteBlk(preCtor, L(BlockMemberSymbol("TODO", Nil)), 1), // TODO: preCtor is not translated in handler lowering
-        if isTopLevel && (defn.k is syntax.Mod) then transformTopLevel(ctor) else rewriteBlk(ctor, R(isym), 1),
+        preCtor,
+        if isTopLevel && (defn.k is syntax.Mod) then transformTopLevel(ctor) else ctor,
```
CAG2Mark: I think this code was written before `ClsLikeBody` was added. But now that modules are represented using that, the `defn.k is syntax.Mod` branch is no longer reachable, so it should be removed.

hkmc2/shared/src/main/scala/hkmc2/Config.scala
```diff
@@ -68,13 +68,16 @@ object Config:
     // Whether we check `Instantiate` nodes for effects. Currently, effects cannot be raised in constructors.
     checkInstantiateEffect: Bool = false,
     // A debug option that allows codegen to continue even if an unlifted definition is encountered.
-    softLifterError: Bool = false
+    softLifterError: Bool = false,
+    // Skips instrumenting module constructors, this can be used when the file is statically known to not
+    // raise any effect and cannot use the Runtime.mls module during module construction due to cyclic dependency.
```
CAG2Mark, line 73: Could you pls explain what this cyclic dependency is?

hkmc2/shared/src/test/mlscript-compile/Runtime.mls
CAG2Mark: Is this TODO now resolved?

hkmc2/shared/src/test/mlscript/handlers/StackSafety.mls
```diff
@@ -6,7 +6,7 @@
 
 // preserve tail calls
 // MUST see "return hi(tmp)" in the output
```
CAG2Mark: Not related to this PR, but this comment should probably be rewritten now that tailrec funcs are rewritten. (Or turn off the optimization for this file)

hkmc2/shared/src/test/mlscript/handlers/StackSafety.mls
```diff
@@ -6,7 +6,7 @@
...
-:stackSafe 5
+:stackSafe 50
```
CAG2Mark: Is this because the stack limit being too low now results in an infinite loop? If so, this would probably be a good time to add a guard preventing this from being set too low.

CAG2Mark (approved):
LGTM. Just some small comments.

hkmc2/shared/src/test/mlscript-compile/Runtime.mls
```diff
@@ -389,8 +389,12 @@ fun handleEffect(cur) =
     handlerFrame.nextHandler = null
 
   // handle the effect
-  set curEffect = null
+  let oldDepth = stackDepth
+  set
+    curEffect = null
+    stackDepth += 2
   let tmp = cur.handlerFun(resume(cur.contTrace))
+  set stackDepth = oldDepth
```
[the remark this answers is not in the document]
LPTK: Actually, how should we handle exceptions, here?

If we want to do what Copilot suggests, in MLscript we can use:

```suggestion
  set curEffect = null
  let tmp = 
    set stackDepth += 2 in cur.handlerFun(resume(cur.contTrace))
```

(And then we don't need `oldDepth`.)

hkmc2/shared/src/test/mlscript-compile/Runtime.mjs
```diff
@@ -1115,24 +1119,26 @@ let Runtime1;
...
     curDepth = Runtime.stackDepth;
+    newDepth = Runtime.stackDepth + 3;
     lbl: while (true) {
       let scrut, scrut1, scrut2, tmp1, tmp2, tmp3, tmp4, tmp5;
       if (cont instanceof Runtime.FunctionContFrame.class) {
         Runtime.curEffect = null;
+        Runtime.stackDepth = newDepth;
         tmp1 = runtime.safeCall(cont.resume(value));
         value = tmp1;
+        Runtime.stackDepth = curDepth;
```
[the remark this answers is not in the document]
LPTK: Same remark as above.

hkmc2/shared/src/test/mlscript-compile/Runtime.mjs
```diff
@@ -1185,25 +1191,36 @@ let Runtime1;
...
         Runtime.stackDepth = 1;
-        tmp = runtime.Unit;
+        tmp2 = runtime.safeCall(saved(runtime.Unit));
+        result = tmp2;
+        scrut2 = Runtime.curEffect !== null;
+        if (scrut2 === true) {
+          throw globalThis.Object.freeze(new globalThis.Error("Effect crossed through stack safe boundary"))
+        } else {
+          tmp3 = runtime.Unit;
+        }
```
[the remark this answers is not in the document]
LPTK: @copilot this is generated code...

hkmc2Benchmarks/src/test/scala/hkmc2/BenchTestRunner.scala
```diff
@@ -22,5 +22,5 @@ class BenchTestRunner
   override protected def createDiffMaker
       (file: Path, preludePath: Path, predefPath: Path, relativeName: String)
       : DiffMaker =
-    new BenchDiffMaker((os.pwd/"hkmc2Benchmarks").toString, file, preludePath, predefPath, relativeName)(using state.cctx)
+    new BenchDiffMaker(state.workingDir.toString, file, preludePath, predefPath, relativeName)(using state.cctx)
```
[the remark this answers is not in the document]
LPTK: I think it has a point!

hkmc2/shared/src/test/mlscript-compile/Runtime.mls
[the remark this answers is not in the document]
LPTK: Omg same pattern again.

hkmc2/shared/src/test/mlscript/handlers/StackSafety.mls
LPTK: Good point, let's have both versions (with tailrec rewriting on and off).

hkmc2/shared/src/test/mlscript/handlers/StackSafety.mls
LPTK: I'd like to know as well 🤔

LPTK:
So, we have to decide how to handle exceptions coming from user code. Is it really safe to ignore them, as we do now? If we do what Copilot suggests, it might have an impact on performance.

Note that a valid option might also be to decree that host-lang exceptions should never be raised in effect-handled code (and an effect mirrorring the exception should be used instead).

hkmc2/shared/src/test/mlscript-compile/Runtime.mls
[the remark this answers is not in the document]
AnsonYeung: We can't deal with try finally with effect handlers enabled. The problem is the following: the resumption parameter may or may not be called, and can even be "leaked" to outer context. If you do not call the resumption, the restore code will never run. Consider something like
```
handle h ...

set x = 2 in h.raise()
```

If h.raise does not resume, x will never be restored. This is regardless of whether error is used or not.

hkmc2/shared/src/main/scala/hkmc2/Config.scala
AnsonYeung: rn the Rendering does not depend on Runtime during construction, so we can import Runtime, and it'll import Rendering correctly. But after stack safety Rendering will depend on Runtime during construction (calling `runtime.runStackSafe` and using `runtime.curEffect`). This causes problem.

hkmc2/shared/src/test/mlscript-compile/Runtime.mls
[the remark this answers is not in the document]
LPTK: This is in `Runtime.mls`, not in a user program.

Besides, let's add a check that try/finally is not supported in the presence of effect handlers (it should probably fail compilation or warn). Later, we can design a proper fix (there are some, eg based on the type system).

hkmc2/shared/src/test/mlscript/handlers/StackSafety.mls
AnsonYeung: It's already disabled (by you xd). We can add another test to check tail call optimization is applied afterwards but we shouldn't mess with this test file as it's meant to use stack intentionally.

LPTK: Oh, right!

@CAG2Mark I guess you misread the test output? The expected code is in there (almost).

```suggestion
// MUST see "return hi(tmp1)" in the output
```

CAG2Mark:
> So, we have to decide how to handle exceptions coming from user code. Is it really safe to ignore them, as we do now? If we do what Copilot suggests, it might have an impact on performance.
> 
> Note that a valid option might also be to decree that host-lang exceptions should never be raised in effect-handled code (and an effect mirrorring the exception should be used instead).

It _might_ be possible by wrapping all states determined to be inside try-catch statements with the corresponding try-catch statements in the state machine, and catch/finally could jump to the states corresponding to the blocks inside the original catc/finally blocks. But I don't know, as it seems a bit complicated.

hkmc2/shared/src/test/mlscript/handlers/StackSafety.mls
CAG2Mark: Whoops lol, I copy-pasted it into HkScratch for some testing and saw there was a while loop

hkmc2/shared/src/main/scala/hkmc2/Config.scala
CAG2Mark: I see. Can you pls add this explanation to the code? It's not that obvious to an outsider

hkmc2/shared/src/test/mlscript/handlers/StackSafety.mls
CAG2Mark: I tried the code with a stack depth of 5 and the run never terminates.
(Actually I guess that is where I saw the tailrec function was rewritten into while loop and got mistaken, lol)

AnsonYeung:
> > So, we have to decide how to handle exceptions coming from user code. Is it really safe to ignore them, as we do now? If we do what Copilot suggests, it might have an impact on performance.
> > Note that a valid option might also be to decree that host-lang exceptions should never be raised in effect-handled code (and an effect mirrorring the exception should be used instead).
> 
> It _might_ be possible by wrapping all states determined to be inside try-catch statements with the corresponding try-catch statements in the state machine, and catch/finally could jump to the states corresponding to the blocks inside the original catc/finally blocks. But I don't know, as it seems a bit complicated.

`catch` was never an issue as it's just a usual effect handler. `finally` on the other hand is problematic, see above.

hkmc2/shared/src/main/scala/hkmc2/Config.scala
AnsonYeung: well this is just one possible instance to justify the option and is not that important but i can add that in the comment

hkmc2/shared/src/main/scala/hkmc2/Config.scala
```diff
@@ -68,13 +68,18 @@ object Config:
...
+    // One specific scenario is Rendering.mls which is depended by Runtime.mls, and hence using stack safety will
+    // reference Runtime.mls during construction of Rendering module which cause cyclic dependency.
```
```suggestion
    // One specific scenario is Rendering.mls, which Runtime.mls depends on, and hence using stack safety will
    // reference Runtime.mls during construction of the Rendering module, causing a cyclic dependency error.
```

hkmc2/shared/src/test/mlscript-compile/Runtime.mls
```diff
@@ -467,10 +457,9 @@ fun checkDepth() =
     ()
 
 fun runStackSafe(limit, f) =
-  set
-    stackLimit = limit
-    stackDepth = 1
-    stackHandler = StackDelayHandler
+  set stackLimit = limit in...
```
LPTK, line 462: Neat use of `...` here!

LPTK:
Side note: dealing with `finally` in user code shouldn't be a problem once we:
* Add a way to drop a continuation, which should clean up its resources and run its `finally` blocks.
* Track in user code that continuations are used linearly (ie always end up being either called or destroyed).

I believe Jonathan/@b-studios has a paper preprint about similar stuff.

hkmc2DiffTests/src/test/scala/hkmc2/MLsDiffMaker.scala
```diff
@@ -97,6 +97,10 @@ abstract class MLsDiffMaker extends DiffMaker:
                 failures += 1
                 output("/!\\ Stack limit must be positive, but the stack limit here is set to " + value)
                 S(StackSafety.default)
+              else if value < 10 then
+                failures += 1
+                output("/!\\ Stack limit is too low, the minimum supported is 10.")
```
LPTK: BTW where does this arbitrary-looking number come from?

(Would be good to add a comment about it here.)

AnsonYeung: Should I use the real minimum here

hkmc2/shared/src/test/mlscript/handlers/StackSafety.mls
AnsonYeung: I changed the limit to reflect the new minimum

hkmc2/shared/src/test/mlscript-compile/Runtime.mls
AnsonYeung: It's quite nice. It would be even better if multiple backtracking assignment is supported.

hkmc2DiffTests/src/test/scala/hkmc2/MLsDiffMaker.scala
AnsonYeung: Changed this to 6, this includes 1 for initial depth, 3 for resume increment, and 1 for function call. It needs to be strictly greater than 5 to not raise effect immediately upon resumption.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -347,7 +347,8 @@ class HandlerLowering(paths: HandlerPaths, opt: EffectHandlers)(using TL, Raise,
       case _: Return => blk
 
       // ignored cases
-      case TryBlock(sub, finallyDo, rest) => ??? // ignore
+      case TryBlock(sub, finallyDo, rest) =>
+        lastWords("try-finally is not supported in the presence of effect handlers")
```
LPTK: This case is reachable from normal user code/config, so it shouldn't be an "Internal Error". It should probably just be a compilation error.

hkmc2DiffTests/src/test/scala/hkmc2/MLsDiffMaker.scala
```diff
@@ -97,6 +97,11 @@ abstract class MLsDiffMaker extends DiffMaker:
                 failures += 1
                 output("/!\\ Stack limit must be positive, but the stack limit here is set to " + value)
                 S(StackSafety.default)
+              // Minimum: 1 for initial depth, 3 for resuming in the trampoline, 1 for function entry. The limit need to be strictly greater.
```
LPTK:
```suggestion
              // Minimum: 1 for initial depth, 3 for resuming in the trampoline, 1 for function entry.
              // The limit needs to be strictly greater than 1 + 3 + 1 = 5.
```

hkmc2DiffTests/src/test/scala/hkmc2/MLsDiffMaker.scala
LPTK: Sure.

CAG2Mark:
> > > So, we have to decide how to handle exceptions coming from user code. Is it really safe to ignore them, as we do now? If we do what Copilot suggests, it might have an impact on performance.
> > > Note that a valid option might also be to decree that host-lang exceptions should never be raised in effect-handled code (and an effect mirrorring the exception should be used instead).
> > 
> > 
> > It _might_ be possible by wrapping all states determined to be inside try-catch statements with the corresponding try-catch statements in the state machine, and catch/finally could jump to the states corresponding to the blocks inside the original catc/finally blocks. But I don't know, as it seems a bit complicated.
> 
> `catch` was never an issue as it's just a usual effect handler. `finally` on the other hand is problematic, see above.

Actually there's an OOPSLA paper that might be related to this: https://dl.acm.org/doi/abs/10.1145/3763155
(and I happen to have chosen it for a course reading assignment :P)

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
AnsonYeung: This is so that the expression type checks, since raise won't crash the compiler. There is a compilation error during elaboration.

hkmc2/shared/src/test/mlscript-compile/Runtime.mls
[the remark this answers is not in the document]
LPTK: This is resolved, right?

hkmc2/shared/src/test/mlscript-compile/Runtime.mls
LPTK: Right, I was thinking the same.

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
LPTK: Try
```suggestion
        raise(...)
        term(Term.Error)
```

LPTK:
> [x] Decide how to deal with host-language exceptions.
> -> We now use a finally blocks (`set`/`in`) to gracefully handle user-code exceptions.

Ok, so this is setlled for now. I hope it doesn't have too much of a performance impact.

hkmc2/shared/src/test/mlscript-compile/Runtime.mls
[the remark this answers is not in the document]
AnsonYeung: Yes, aside from the actual support for try/finally

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -343,7 +354,13 @@ class HandlerLowering(paths: HandlerPaths, opt: EffectHandlers)(using TL, Raise,
...
 
       // ignored cases
-      case TryBlock(sub, finallyDo, rest) => ??? // ignore
+      case TryBlock(sub, finallyDo, rest) =>
+        raise(ErrorReport(
+          msg"Try finally are not supported with effect handlers enabled." ->
+          N :: Nil,
+          source = Diagnostic.Source.Compilation))
+        containsError = true
+        rtThrowMsg("Try finally are not supported with effect handlers enabled.")
```
```suggestion
        rtThrowMsg("`try`-`finally` blocks are not currently supported with effect handlers enabled.")
```

LPTK: And instead of using this custom `rtThrowMsg` method, prefer using `Lowering`'s `fail` method (you can pass the enclosing instead of `Lowering` as a parameter of `HandlerLowering` or alternatively (perhaps better) extract `fail` to a `Lowering` companion object).

https://github.com/hkust-taco/mlscript/blob/1857fda7a955b59761df45ff76c153eeb76b5e06/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala#L132-L137

LPTK (approved):
Otherwise LGTM!

LPTK:
It would be nice, though, to quantify the performance overhead of using `finally` blocks (thgouth `set`/`in`) vs the old less safe approach.

CAG2Mark:
> It would be nice, though, to quantify the performance overhead of using `finally` blocks (thgouth `set`/`in`) vs the old less safe approach.

To my knowledge, try-catch blocks are not that expensive.

AnsonYeung:
> It would be nice, though, to quantify the performance overhead of using `finally` blocks (thgouth `set`/`in`) vs the old less safe approach.

Stack safe (depth 100):

chart-set-in

LPTK:
Ok, let's say it's good enough for now.

PR #399, opened 2026-02-26
Overhaul semantics and compilation of UCS `while`

### LINKED ISSUE #382: Bug in UCS while loop that merge else branch
Original code:
```fs
while f() is
  A() do ()
  B(xxx) do g(xxx)
```

When f return A(), the while loop terminates as the A() branch was merged with the else branch.

Full example to demonstrate that changing else branch affect the loop behaviour:
```fs
let x = 1
while x ===
  1 do
    set x = 2
  2 do
    set x = 3
    print("valid")
else
  set x = 2
print("not inf loop")
//│ > not inf loop
//│ x = 2

let x = 1
while x ===
  1 do
    set x = 2
  2 do
    set x = 3
    print("valid")
else
  set x = 3
print("not inf loop")
//│ > valid
//│ > not inf loop
//│ x = 3
```

LPTK:
Fixes #382.

Supersedes #397.

* Now, we no longer have the funky and buggy logic surrounding "top-level" default branches in UCS `while` loops, and both default branches and `else` now has a uniform semantics.

* There is no longer a difference between using `then` and `do` in the consequents of a `while`; we should probably forbid the former.

* `while` now always return a `()` value, regardless of their branches.

* [x] To do later: warn about pure expressions in statement position in `while` branches.


hkmc2/shared/src/test/mlscript/codegen/While.mls
```diff
@@ -31,9 +28,9 @@
 let x = true
 //│ x = true
 
-:re
+// TODO: forbid uses of keywords `then` as well as `else` in `while`
```
chengluyu, line 33: Will we do this?

hkmc2/shared/src/test/mlscript/codegen/While.mls
chengluyu, line 1: Do we have any tests of nested `while`/`if` in `while`? For example, `while`
- contains nested `if` or `case`
- whose `else` contains another `while`

chengluyu (approved):
LGTM with questions.

hkmc2/shared/src/test/mlscript/codegen/While.mls
LPTK: Not sure, but why would that be interesting? Do you foresee any weird interactions? If so, feel free to propsoe one and I'll add it.

hkmc2/shared/src/test/mlscript/codegen/While.mls
LPTK: It's not urgent, so maybe later.

PR #400, opened 2026-02-27
Port JS scoped-blocks and remove definedVars

b-temirov:
This PR ports JSBuilder scoped-block local allocation semantics to wasm backend. It removes `definedVars` local collection and switches to scope-driven allocation.


hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -943,7 +960,10 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
 
       `return`(S(resWat))
 
-    case Scoped(_, body) => returningTerm(body)
+    case Scoped(syms, body) =>
+      scope.nest givenIn:
+        blockPreamble(syms)
+        returningTerm(body)
```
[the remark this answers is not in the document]
LPTK: @b-temirov Ah, it seems WASM doesn't actually have local scopes! So you need to flatten the function body's scopes so that all the variables are defined at the top. Luckily, we already have that implemented: use `ScopeFlattener().applyBlock(...)`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -717,6 +717,23 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
       errExpr(
         Ls(msg"This code requires effect handler instrumentation but was compiled without it." -> N)
       )
+    case Assign(l, r, rst) if l is State.noSymbol =>
```
LPTK, line 887: Good catch of the recent change!

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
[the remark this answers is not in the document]
LPTK: Or just apply Copilot's suggestion; looks like it should actually just work.

LPTK (changes requested):
Following the change proposed by Copilot, I realize that there's a bunch of junk in the WatBuilder that was mindlessly copied from JSBuilder, such as `def body` now being unused. As a more serious example, it doesn't make any sense to use things like `scope.nestRebindThis`, as this deals with details of how JavaScript deals with the `this` keyword (which doesn't even exist in Wasm) in nested definitions (which aren't even possible in Wasm).

Please make a pass to remove and simplify this unnecessary logic.

b-temirov:
1. Removed dead code:
  - Deleted unused `body` helper.
  - Removed unused `name` parameter from `setupFunction`.

2. Removed JS-specific `this` rebinding flow:
  - Replaced `scope.nestRebindThis(...)` with `scope.nest` in `Define` handling.
  - Removed `thisProxy` / `thisProxyDefined` branch and error path.
  - Replaced `scope.findThis_!` usages with explicit scope/local lookup.

3. Added explicit constructor self binding:
  - Introduced `bindCtorThis(...)` to bind constructor `isym` directly to local name `"this"` in wasm scope.

LPTK:
> 3\. Added explicit constructor self binding:

There is no test to accompany this change. There should be at least one WASM test that uses `this`.

b-temirov:
Added two new test cases to `Basics.mls`.

hkmc2/shared/src/test/mlscript/wasm/Basics.mls
```diff
@@ -164,3 +164,68 @@ fun getX(f: Foo) = f.x
...
+//│           (nop)))
+//│       (return
+//│         (local.get $this))))
+//│   (export "Foo" (func $Foo))
+//│   (elem declare func $Foo)
+//│   (func $entry (type 3) (result (ref null any))
+//│     (local $Foo (ref null any))
+//│     (local $tmp (ref null any))
+//│     (local $x (ref null any))
+//│     (local $Foo1 (ref null any))
```
LPTK: What is this unused `$Foo1`  variable and where does it come from? Is it generated by mistake?

LPTK (approved):
Thanks, LGTM now. Except for this one small question.

hkmc2/shared/src/test/mlscript/wasm/Basics.mls
b-temirov: It came from compiling class constructors in the enclosing locals frame, so ctor params and temps were being collected into `entry` locals. Added a helper that compiles constructors under their own frame.

PR #402, opened 2026-02-28
Add JS-backed string literals

FetBoba:
This PR adds string support to the WASM backend by lowering MLX string literals to JS strings through a host import.


hkmc2DiffTests/src/test/scala/hkmc2/WasmDiffMaker.scala
```diff
@@ -132,18 +134,36 @@ abstract class WasmDiffMaker extends LlirDiffMaker:
...
       end mkQuery
 
-      val importObj =
-        doc"""{ #{  # "system": { #{  # "mem": new WebAssembly.Memory({initial: 100}) #}  # } #}  # }"""
       val jsStr =
-        doc"""await wasm.binaryenPrintFuncRes( #  #{ `$modWat # `, # $importObj, # exports => exports.${mainFnNme}(), #}  # );"""
+        doc"""
+          await (() => {
+            # const watSrc = $modWatJsLit;
+            # const memMatch = watSrc.match(/\(import\s+"system"\s+"mem"\s+\(memory\s+(\d+)\)\)/);
```
LPTK: What is this crazy regexp for? Please don't do that.

First of all, could you please explain why you think we need more than one memory here? And what's the strange business with the computation of `minPages`?

FetBoba: Yes, it was very stupid to do it in this way. Sorry for that. I complicated things too much.
We only need one memory (system.mem), but its initial pages must satisfy the module’s declared minimum. minPages is now computed in codegen as ceil(max(totalStringBytes,1)/65536) and passed directly.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -287,12 +303,63 @@ class Ctx(
...
+        if newMin =/= existing.minPages then
+          memoryImports(idx) = existing.copy(minPages = newMin)
+      case N =>
+        val idx = memoryImports.size
+        memoryImports += MemoryImport(module, name, minPages)
+        cachedMemoryImport(key) = idx
+
+  /** Returns the minimum page requirement of memory import (`module`, `name`) if present. */
+  def getMemoryImportMinPages(module: Str, name: Str): Opt[Int] =
+    memoryImports.find(m => m.module == module && m.name == name).map(_.minPages)
```
LPTK:
```suggestion
    memoryImports.find(m => m.module === module && m.name === name).map(_.minPages)
```

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -425,6 +494,6 @@ class Ctx(
 
   def toWat: Document =
     val startDef = startFunc.toSeq.map(funcIdx => doc"(start ${funcIdx.toWat})")
-    doc"(module #{  # ${(types.toSeq.map(_.toWat) ++ globals.toSeq.map(_.toWat) ++ startDef ++ funcs.toSeq.map(_.toWat)).mkDocument(doc" # ")}) #} "
+    doc"(module #{  # ${(types.toSeq.map(_.toWat) ++ memoryImports.toSeq.map(_.toWat) ++ functionImports.toSeq.map(_.toWat) ++ dataSegments.toSeq.map(_.toWat) ++ globals.toSeq.map(_.toWat) ++ startDef ++ funcs.toSeq.map(_.toWat)).mkDocument(doc" # ")}) #} "
```
LPTK: Try to make lines not exceed 120 columns.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -200,6 +205,61 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+        nextStringDataOffset = offset + byteLen
+        StringLitInfo(offset = offset, byteLen = byteLen, watBytes = watBytes)
+    )
+
+  /**
+   * Ensures imports required for string materialization exist and returns the constructor function.
+   */
+  private def ensureStringImports(using Ctx): FuncIdx =
+    val minBytes = nextStringDataOffset max 1
+    val minPages = (minBytes + 65535) / 65536
```
LPTK: Please avoid using magic numbers like `65536`. You should bind it to a named Scala constant with an appropriate name. Here, something like `pageSize`, I guess?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1211,7 +1282,10 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
   def program(p: Program, exprt: Opt[BlockMemberSymbol], wd: io.Path)(using
       Raise,
       Scope
-  ): (Document, Str) =
+  ): (Document, Str, Int) =
+    stringLits.clear()
```
LPTK: What's this `clear` for?

It's the second time you add this sort of weird statements. Is this generated by AI?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1211,7 +1282,10 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+    nextStringDataOffset = 0
```
LPTK: Same question.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1231,6 +1305,18 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
 
     val ctx = Ctx.empty
     given Ctx = ctx
+
+    // Check whether there is strigns in block because function imports
```
```suggestion
    // Check whether there are strings in the block [because function imports]
```
But "because function imports" doesn't make grammatical sense; please rephrase.

hkmc2DiffTests/src/test/scala/hkmc2/WasmDiffMaker.scala
```diff
@@ -132,18 +134,34 @@ abstract class WasmDiffMaker extends LlirDiffMaker:
...
-          case n if n >= 0 => out.substring(0, n)
-          case _ => ""
+        // Omit the REPL tail only when it is explicitly "undefined" or unit.
+        val lines = out.linesIterator.toList
+        val result =
+          lines.lastOption match
+            case S(last) if last.trim === "undefined" || last.trim === "()" =>
+              lines.dropRight(1).mkString("\n")
+            case _ =>
+              lines.mkString("\n")
```
LPTK: Can you explain why you made this change?

This code was originally copied from `JSBackendDiffMaker.scala`, which still has the old version. If your version is better, explain why, and share the code between the two so they don't diverge again.

LPTK (approved):
Otherwise LGTM, but would like to see reviews from others.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -10,12 +10,13 @@ import document.*
 import document.Document
 import js.CodeBuilder
 import semantics.*, Elaborator.State
-import syntax.Tree.{BoolLit, IntLit, Ident}
+import syntax.Tree.{BoolLit, IntLit, StrLit, Ident}
 import text.Param as WasmParam
 import Message.MessageContext
 import Scope.scope
+import hkmc2.codegen.BlockTraverser
```
Derppening: Nit: This import is redundant.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -287,12 +303,63 @@ class Ctx(
...
       namedFuncs(_) = numIdx
     FuncIdx(funcInfo.id.getOrElse(numIdx))
 
+  /**
+   * Adds a function import into this context.
+   *
+   * Returns the function index in the global function index space.
+   */
+  def addFunctionImport(sym: Opt[Symbol], funcImport: FuncImport): FuncIdx =
+    val numIdx = NumIdx(functionImports.size)
```
Derppening: It seems like this would break if this function is called after `addFunc` is invoked at least once. 

Could we just derive the `NumIdx` of this function import using `functionImports.size + funcs.size` similar to `addFunc`? This would also address this comment: https://github.com/hkust-taco/mlscript/pull/402/changes#r2877874447

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -200,6 +205,61 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+    )
+
+  /**
+   * Ensures imports required for string materialization exist and returns the constructor function.
+   */
+  private def ensureStringImports(using Ctx): FuncIdx =
+    val minBytes = nextStringDataOffset max 1
+    val minPages = (minBytes + 65535) / 65536
+    val systemImportMod = "system"
+    val stringFromUtf16ImportNme = "mlx_str_from_utf16"
```
Derppening: I think these can also be made named constants outside of the function scope. Perhaps store them in an `object ExternIntrinsics`?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -200,6 +205,61 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+        val offset = (nextStringDataOffset + 1) & ~1
+        val byteLen = value.length * 2
+        nextStringDataOffset = offset + byteLen
+        StringLitInfo(offset = offset, byteLen = byteLen, watBytes = watBytes)
+    )
+
+  /**
+   * Ensures imports required for string materialization exist and returns the constructor function.
+   */
+  private def ensureStringImports(using Ctx): FuncIdx =
```
Derppening: Nit: I suggest `getOrLoadStrCtorFunction` - `ensure`-prefix may be confused with assertion-like functions.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -287,12 +303,63 @@ class Ctx(
...
+
+  /**
+   * Returns the cached function import for (`module`, `name`), creating it with `createImport`
+   * if needed.
+   */
+  def getOrCreateFunctionImport(
+      module: Str,
+      name: Str,
+      createImport: => FuncImport
+  ): FuncIdx =
```
Derppening: Nit: Rewrite into this form to allow for braceless syntax

```suggestion
  def getOrCreateFunctionImport(
    module: Str,
    name: Str,
  )(createImport: => FuncImport): FuncIdx =
```

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -200,6 +205,61 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+            )
+          )
+        )
+        FuncImport(
+          module = systemImportMod,
+          name = stringFromUtf16ImportNme,
+          id = S(SymIdx(stringFromUtf16ImportNme)),
+          typeIdx = importTy
+        )
+    )
```
Derppening: Re https://github.com/hkust-taco/mlscript/pull/402/changes#r2877854309:

```suggestion
    ctx.getOrCreateFunctionImport(
      module = systemImportMod,
      name = stringFromUtf16ImportNme,
    ):
      val importTy = ctx.addType(
        sym = N,
        TypeInfo(
          id = N,
          FunctionType(
            params = Seq(
              WasmParam(N, RefType.anyref),
              WasmParam(N, RefType.anyref)
            ),
            results = Seq(Result(RefType.anyref))
          )
        )
      )
      FuncImport(
        module = systemImportMod,
        name = stringFromUtf16ImportNme,
        id = S(SymIdx(stringFromUtf16ImportNme)),
        typeIdx = importTy
      )
```

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1231,6 +1305,18 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+    // must be declared before function definitions to keep indices stable.
+    var hasStringLiteral = false
+    val traverser = new BlockTraverser:
+      override def applyValue(v: Value): Unit =
+        v match
+          case Value.Lit(StrLit(_)) => hasStringLiteral = true
+          case _ => ()
+    traverser.applyBlock(p.main)
+    if hasStringLiteral then
+      ensureStringImports
```
Derppening: Rather than doing this, I think we should just lazily include the string extern functions as it is used.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
```diff
@@ -191,6 +191,26 @@ case class LocalIdx(idx: Index) extends CtxIdx(idx)
...
+    id: Opt[SymIdx],
+    typeIdx: TypeIdx,
+) extends ToWat:
+  def toWat: Document =
+    doc"""(import "${module}" "${name}" (func ${id.fold(doc"")(_.toWat)} (type ${typeIdx.toWat})))"""
+
+/** A data segment entry. */
+case class DataSegment(offset: Int, bytes: Str) extends ToWat:
+  def toWat: Document =
+    doc"""(data (i32.const ${offset}) "${bytes}")"""
```
Derppening: Please do not hardcode the offset like this. The wasm spec explicitly allows an instruction in place of `(i32.const $offset)`, so accept an instruction there instead.

If you want to keep the `offset: Int` overload for convenience, you can have a secondary constructor that specifies a compile-time known offset.

hkmc2/shared/src/test/mlscript/wasm/Strings.mls
```diff
@@ -0,0 +1,90 @@
...
+:wasm
+
+:wat
+"Hello World!"
+//│ Wat:
+//│ (module
+//│   (type (func (param (ref null any)) (param (ref null any)) (result (ref null any))))
+//│   (type $Object (sub (struct (field $$tag (mut i32)))))
+//│   (type (func (result (ref null any))))
+//│   (import "system" "mem" (memory 1))
```
Derppening, line 24: Can we use `memory 0` by default?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1310,7 +1400,8 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
 
     ctx.addFunc(S(entrySym), entryFnInfo)
 
-    (ctx.toWat, entryNme)
+    val systemMemMinPages = ctx.getMemoryImportMinPages("system", "mem").getOrElse(1)
```
Derppening: Nit: I think this can be `0` if no strings are present.

hkmc2/shared/src/test/mlscript/wasm/Strings.mls
```diff
@@ -0,0 +1,90 @@
+
+:global
+:wasm
+
+:wat
+"Hello World!"
```
Derppening: I think it is also worth adding test cases for empty strings and duplication strings. In particular, the duplicate string test case should show the WAT output to verify that the string deduplication is indeed performed.

hkmc2DiffTests/src/test/scala/hkmc2/WasmDiffMaker.scala
```diff
@@ -66,9 +67,10 @@ abstract class WasmDiffMaker extends LlirDiffMaker:
       val low = ltl.givenIn:
         codegen.Lowering()
       val le = low.program(trm)
-      val (modWat, mainFnNme) = ltl.givenIn:
+      val (modWat, mainFnNme, systemMemMinPages) = ltl.givenIn:
         baseScp.nest.givenIn:
           WatBuilder().program(le, N, wd)
+      val modWatJsLit = Tree.StrLit(modWat.mkString()).idStr
```
Derppening: I don't think `Tree.StrLit` is supposed to be used this way. Could you explain why this change is needed?

hkmc2DiffTests/src/test/scala/hkmc2/WasmDiffMaker.scala
```diff
@@ -132,18 +134,34 @@ abstract class WasmDiffMaker extends LlirDiffMaker:
...
+            # const watSrc = $modWatJsLit;
+            # const mem = new WebAssembly.Memory({ initial: $systemMemMinPages });
+            # const decodeUtf16 = new TextDecoder("utf-16le");
+            # const imports = {
+                "system": {
+                  "mem": mem,
+                  "mlx_str_from_utf16": (ptr, byteLen) =>
+                    decodeUtf16.decode(new Uint8Array(mem.buffer, ptr, byteLen))
+                }
+              };
```
Derppening: Nit: Please keep the `importObj` variable and inline the highlighted contents into that `doc`-interpolated string.

hkmc2/shared/src/test/mlscript/wasm/Strings.mls
LPTK: Let's _only_ show the WAT of the duplicated case, to avodi pollution.

hkmc2DiffTests/src/test/scala/hkmc2/WasmDiffMaker.scala
LPTK: If you do want to reuse the escaping behavior, it would be better to refactor it into a utility function. Also see the ocmment on `idStr`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: Ok done

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Ok done

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Ok removed

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: Ok done

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Ok done

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: Ok changed

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Ok renamed

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: I removed comment and string scan code as was suggested by @Derppening.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: Ok I changed it. But I also introduced funcInfosByIndex in Ctx otherwise last test case in String.mls fails with  `Missing function definition for symbol `member:id`` error (which means that indices are not mapped correctly in `getFuncInfo`)

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Ok I removed this

hkmc2/shared/src/test/mlscript/wasm/Strings.mls
FetBoba: Yes

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Ok changed

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
FetBoba: Ok I changed it. I think overload is redundant because it will have only one call site in WatBuilder.scala

hkmc2/shared/src/test/mlscript/wasm/Strings.mls
FetBoba: Ok added these tests and removed other wat flags

hkmc2DiffTests/src/test/scala/hkmc2/WasmDiffMaker.scala
FetBoba: Ok done

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Ok removed

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Ok done

hkmc2DiffTests/src/test/scala/hkmc2/WasmDiffMaker.scala
FetBoba: I changed to old version. I thought that there might be a case where output doesn't have a trailing sentinel line. But turns out that output is always wrapped in `try { ... }`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Yes I was using AI to assist me but didn't check this part thoroughly after that (which I usually do). I removed it.

hkmc2DiffTests/src/test/scala/hkmc2/WasmDiffMaker.scala
FetBoba: I changed it to just use `JSBuilder.makeStringLiteral`. Is it ok?

LPTK: I guess it's ok for now.

LPTK:
@FetBoba Please fix the conflicts and then I think we can merge.

PR #403, opened 2026-03-03
Fix while loop rewriting

### LINKED ISSUE #401: `while` rewriting into a tail-recursive function does not always return `()`
With https://github.com/hkust-taco/mlscript/pull/399, `while` loops no longer return a value, so their lowering should explicitly return `()`. But I just noticed that the alternative `while` lowering of https://github.com/hkust-taco/mlscript/pull/351 still does not do that for default branches:

```fs
:sjs
fun f = while false do ()
print(f)
//│ JS (unsanitized):
//│ let f, tmp;
//│ f = function f() {
//│   lbl1: while (true) {
//│     let scrut2;
//│     scrut2 = false;
//│     if (scrut2 === true) {
//│       continue lbl1
//│     } else {}
//│     break;
//│   }
//│   return runtime.Unit
//│ };
//│ tmp = f();
//│ Predef.print(tmp)
//│ > ()

:sjs
:rewriteWhile
fun f = while false do ()
print(f)
//│ JS (unsanitized):
//│ let f1, tmp1;
//│ f1 = function f() {
//│   let while1, tmp2, tmp3, tmp4;
//│   tmp4 = undefined;
//│   while1 = (undefined, function () {
//│     let scrut2;
//│     scrut2 = false;
//│     if (scrut2 === true) {
//│       return while1()
//│     } else {}
//│     return runtime.LoopEnd
//│   });
//│   tmp2 = while1();
//│   tmp3 = tmp2 !== runtime.LoopEnd;
//│   if (tmp3 === true) { return tmp2 } else { return tmp4 }
//│ };
//│ tmp1 = f1();
//│ Predef.print(tmp1)
//│ FAILURE LOCATION: mkQuery (JSBackendDiffMaker.scala:161)
//│ ═══[RUNTIME ERROR] Error: MLscript call unexpectedly returned `undefined`, the forbidden value.
```

AnsonYeung:
Fixes #401

### DIFF
--- a/hkmc2/shared/src/test/mlscript-compile/Runtime.mls
+++ b/hkmc2/shared/src/test/mlscript-compile/Runtime.mls
@@ -10,6 +10,9 @@ module Runtime with ...
 object Unit with
   fun toString() = "()"
 
+// This object is used as a marker for the end of a while loop in the compiled codegen
+// It is possible for the loop body to contain a return statement,
+// in which case we want to distinguish between returning from the loop body and loop termination.
 object LoopEnd
 val short_and = RuntimeJS.short_and
 val short_or = RuntimeJS.short_or
--- a/hkmc2/shared/src/test/mlscript-compile/Runtime.mls
+++ b/hkmc2/shared/src/test/mlscript-compile/Runtime.mls
@@ -10,7 +10,7 @@ module Runtime with ...
 object Unit with
   fun toString() = "()"
 
-// This object is used as a marker for the end of a while loop in the compiled codegen
+// This object is used as a marker for the end of a while loop in the compiled codegen.
 // It is possible for the loop body to contain a return statement,
 // in which case we want to distinguish between returning from the loop body and loop termination.
 object LoopEnd
--- a/hkmc2/shared/src/test/mlscript/codegen/ScopedBlocks.mls
+++ b/hkmc2/shared/src/test/mlscript/codegen/ScopedBlocks.mls
@@ -181,7 +181,8 @@ fun f(x) =
 //│   });
 //│   tmp2 = while1();
 //│   tmp3 = tmp2 !== runtime.LoopEnd;
-//│   if (tmp3 === true) { return tmp2 } else { return runtime.Unit }
+//│   if (tmp3 === true) { return tmp2 }
+//│   return runtime.Unit
 //│ };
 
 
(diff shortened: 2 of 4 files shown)


hkmc2/shared/src/test/mlscript/codegen/ScopedBlocks.mls
```diff
@@ -182,7 +181,7 @@ fun f(x) =
 //│   });
 //│   tmp2 = while1();
 //│   tmp3 = tmp2 !== runtime.LoopEnd;
-//│   if (tmp3 === true) { return tmp2 } else { return tmp4 }
+//│   if (tmp3 === true) { return tmp2 } else { return runtime.Unit }
```
LPTK: Can you remind me why we need LoopEnd? Now would be a good time to document this (such doc is missing). You could the doc to LoopEnd's definition in `Runtime.mls`.

LPTK: Why do we need to return `tmp2`, here? Being a `while` loop, it shouldn't be anything useful.

I would have thought the code would be more like

```suggestion
//│   if (tmp3 === true) { ??? }
//│   return runtime.Unit
```

LPTK (approved):
Otherwise LGTM, thanks!

hkmc2/shared/src/test/mlscript/codegen/ScopedBlocks.mls
AnsonYeung: Done. For easy reference: this is used to distinguish between a loop termination and a return statement within the while loop.

AnsonYeung: See above. The tmp2 is for storing return values. The suggested change is applied.

hkmc2/shared/src/test/mlscript-compile/Runtime.mls
```diff
@@ -10,6 +10,9 @@ module Runtime with ...
 object Unit with
   fun toString() = "()"
 
+// This object is used as a marker for the end of a while loop in the compiled codegen
```
LPTK:
```suggestion
// This object is used as a marker for the end of a while loop in the compiled codegen.
```

PR #406, opened 2026-03-08
Deforestation mk4

ychenfo:
Supersedes and closes https://github.com/hkust-taco/mlscript/pull/335


hkmc2/shared/src/test/mlscript/deforest/module.mls
```diff
@@ -0,0 +1,285 @@
...
+//│ deforest > 	Ref(tmp:arg$XX$0$⁰,None)
+//│ deforest > 	Select(Ref(tmp:arg$XX$0$⁰,None),Ident(b))
+//│ deforest > <<< fusing <<<
+//│ = 123
+
+
+:deforest
+:sjs
+:expect Cons(7, Cons(5, Cons(3, Cons(1, Nil))))
+data object Nil
```
LPTK: Please clean up: this test is duplicated, and also maybe you shouldn't declare the data structure in this block?

hkmc2/shared/src/test/mlscript/deforest/todos.mls
```diff
@@ -0,0 +1,310 @@
...
+//│ deforest > >>> fusing >>>
+//│ deforest > <<< fusing <<<
+
+object None
+data class Some(x)
+//│ deforest > >>> fusing >>>
+//│ deforest > <<< fusing <<<
+
+
+// TODO: many dead code contain duplicated `t + 5 * 4 - 3 + 2 - 1`
```
LPTK: Could you comment on why it's currently duplicated and what will need to be done to avoid this?

ychenfo: Done in https://github.com/hkust-taco/mlscript/pull/406/changes/4a523458b74b5643324e6ea4d4674b2eeb4a8256#diff-d68d7ee2586f8cb350e1c5288bd6dcc802c32cc3b765ff8e5cbd90b1f461ca9dR31

hkmc2/shared/src/main/scala/hkmc2/codegen/LambdaRewriter.scala
```diff
@@ -13,14 +13,11 @@ object LambdaRewriter:
...
           val sym = BlockMemberSymbol("lambda", Nil, nameIsMeaningful = false)
-          val tSym = TermSymbol.fromFunBms(sym, N)
-          sym.tsym = S(tSym)
-          val lamDefn =
-            val Lambda(params, body) = super.applyLam(lam)
-            FunDefn(N, sym, tSym, params :: Nil, body)(false)
-          Scoped(Set(sym), Define(lamDefn, k(Value.Ref(sym, S(tSym)))))
+          val Lambda(params, body) = super.applyLam(lam)
+          val lamDefn = FunDefn.withFreshSymbol(N, sym, params :: Nil, body)(false)
+          Scoped(Set(sym), Define(lamDefn, k(lamDefn.asPath)))
```
LPTK: Better use `Set.single`.

core/shared/main/scala/utils/algorithms.scala
```diff
@@ -50,8 +50,12 @@ object algorithms {
     )
 
     // pre-process: assign each node an id
-    val edgesSet = edges.toSet
-    val nodesUniq = (edgesSet.flatMap { case (a, b) => Set(a, b) } ++ nodes.toSet).toList
+    val nodesUniq = {
```
LPTK, line 53: No need for braces here.

ychenfo: I think this file is part of the `core` subproject, which uses Scala 2.

LPTK: Ah ok, never mind.

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Analyze.scala
```diff
@@ -0,0 +1,820 @@
...
+          || privateFields.nonEmpty
+          || publicFields.nonEmpty
+          || !preCtor.matches:
+            case End("") => true
+          || !ctor.matches:
+            case Return(Select(Value.Ref(runtimeSym, None), Tree.Ident("Unit")), true) =>
+              runtimeSym is elabState.runtimeSymbol
+        then ()
+        else
+          mod.foreach(applyClsLikeBody)
```
LPTK: Strangely convoluted way of writing
```suggestion
        if ctorSym.isEmpty
          && paramsOpt.isEmpty
          && auxParams.nonDefined
          && parentPath.isEmpty
          && methods.isEmpty
          && privateFields.isEmpty
          && publicFields.isEmpty
          && preCtor.matches:
            case End("") => true
          && ctor.matches:
            case Return(Select(Value.Ref(runtimeSym, None), Tree.Ident("Unit")), true) =>
              runtimeSym is elabState.runtimeSymbol
        then mod.foreach(applyClsLikeBody)
```

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Analyze.scala
```diff
@@ -0,0 +1,820 @@
...
+  def allConstraints = globalCollector.constraints
+  val funToSccGroups = MutMap.empty[TermSymbol, Ls[TermSymbol]]
+  def funToSccRep(tSym: TermSymbol): Option[TermSymbol] = funToSccGroups.get(tSym).map(_.head)
+
+  // for fusing strictly internal parts of functions
+  val synthesizedInstIdToFunSym = LinkedHashMap.empty[InstantiationId, TermSymbol]
+  
+  // ===================================================
+  
+  locally {
```
LPTK:
```suggestion
  locally:
```

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Analyze.scala
```diff
@@ -0,0 +1,820 @@
...
+  case class FinalDestSel(dtors: Set[CtorDtorId], field: SelField) extends FinalDest
+  private val ctorDests = LinkedHashMap.empty[ConcreteProducer, Set[ConcreteConsumer | NoCons.type]].withDefaultValue(Set.empty)
+  private val dtorSrcs = LinkedHashMap.empty[ConcreteConsumer, Set[ConcreteProducer | NoProd.type]].withDefaultValue(Set.empty)
+  
+  val finalCtorDests = LinkedHashMap.empty[CtorDtorId, FinalDest]
+  val finalDtorSrcs = LinkedHashMap.empty[CtorDtorId, Set[CtorDtorId]]
+  val fusingCtorInfo = MutMap.empty[CtorDtorId, ConcreteProducer]
+  val fusingDtorInfo = MutMap.empty[CtorDtorId, ConcreteConsumer]
+  // propagate
+  locally {
```
LPTK, line 696:
```suggestion
  
  locally {
```

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Analyze.scala
```diff
@@ -0,0 +1,820 @@
...
+        case Value.Ref(l, disamb) =>
+          disamb.fold(generatedProdVars(l))(generatedProdVars.apply).asProdStrat
+        case Value.Lit(lit) => NoProd
+        case other => lastWords(other.toString())
+      case _ => die
+  }
+end DeforestConstraintsCollector
+
+
+class DeforestConstrainSolver(val collector: DeforestConstraintsCollector):
```
LPTK, line 669: Doesn't this belong in its own file?

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Rewrite.scala
```diff
@@ -0,0 +1,615 @@
...
+          case None => super.applyResult(ctor)(k)
+          case Some(FinalDestSel(_, field)) =>
+            val ctorInfo = solver.fusingCtorInfo(ctor.uid.toCtorDtorId)
+            val idx = ctorInfo.args.unzip._1.indexOf(field)
+            val fieldSyms = mkCtorFieldSyms(ctor.uid.toCtorDtorId)
+            args.zip(fieldSyms).foldRight(k(Value.Ref(fieldSyms(idx)))):
+              case (Arg(N, a) -> s, rest) =>
+                applyPath(a): fusedField =>
+                  Scoped(Set.single(s), Assign(s, fusedField, rest))
+              case _ => die
```
LPTK: I think this should be
```suggestion
              case (Arg(spd, a) -> s, rest) => TODO(spd)
```

LPTK: Please only use `die` and `lastWords` for things that _can't_ happen, not for things that are left to implement.

hkmc2/shared/src/main/scala/utils/TraceLogger.scala
```diff
@@ -27,7 +27,7 @@ abstract class TraceLogger(using val debugPrinter: DebugPrinter):
   inline def enter() = indent += 1
   inline def exit() = indent -= 1
   
-  protected def emitDbg(str: Str): Unit = scala.Predef.println(str)
+  protected[hkmc2] def emitDbg(str: Str): Unit = scala.Predef.println(str)
```
LPTK, line 30: I don't think this change is needed?

LPTK (approved):
Ok, I don't have time to review everything in detail at this time, but I believe this is more or less ready to merge!

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Analyze.scala
ychenfo: I think maybe it's reasonable to keep them here for now, they are indeed both part of the analysis phase and the classes are not particularly large (~260 loc).

hkmc2/shared/src/main/scala/utils/TraceLogger.scala
ychenfo: In `Lowering.scala` I make a new traceLogger to output fusion opportunities, and its `emitDbg` need to use the `emitDbg` from lowering traceLogger (not inherited), so I think I need `protected[hkmc2]` to widen access...

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Analyze.scala
LPTK: Ok, fair enough.

hkmc2/shared/src/main/scala/utils/TraceLogger.scala
LPTK: Why not pass the same TraceLogger as for lowering?

ychenfo: Oh, my intention is that deforest's tracelogger can be enabled independently of lowering tracing. If we reused the same tracelogger, we would have to enable lowering debug just to see deforest summary. The `emitDbg` calling the lowering tracelogger's `emitDbg` is just so the output still goes to the same place, prefixed by "deforest > ".

PR #412, opened 2026-03-12
Improve Symbol information in Block IR

ChingLongTin:
This adds more types of Symbols to Block IR, and addresses both comments in #355.
https://github.com/hkust-taco/mlscript/pull/355#discussion_r2888514921: we now require `:ftc` to be enabled for functions with multiple parameter lists
https://github.com/hkust-taco/mlscript/pull/355#discussion_r2888671974: we collect all class definitions during compilation into  `defnMap` so that we can retrieve the definition for desugared classes generated from `:ftc`

In the future, we will need change `StagedSymbols.mls` to show that we have staged the Symbols correctly.
Instead of relying on the printout from `showBlock`, we probably will need to print the global map of created symbols, since `showBlock` will be used for pretty-printing code for the next stage and will not show a difference between the different types of symbols.


LPTK:
Thanks. Please mark the PR as ready when it's ready for a review.

hkmc2/shared/src/test/mlscript/block-staging/Functions.mls
```diff
@@ -121,9 +128,9 @@ staged module A with
   fun g() = {1 : 2}
 //│ ═══[COMPILATION ERROR] Other Blocks not supprted in staged module: AssignField(Ref(term:A/x,None),Ident(a),Lit(IntLit(0)),Return(Select(Ref(tmp:runtime,None),Ident(Unit)),false))
```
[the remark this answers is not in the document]
ChingLongTin: golden output...?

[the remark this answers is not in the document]
LPTK: This style of regression testing is sometimes referred to as "golden tests".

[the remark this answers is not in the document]
ChingLongTin: huh, interesting.

LPTK:
Still waiting for @NeilKleistGao's review.

hkmc2/shared/src/main/scala/hkmc2/codegen/Instrumentation.scala
```diff
@@ -356,3 +371,16 @@ class Instrumentation(using State, Raise) extends BlockTransformer(new SymbolSub
       val newModule = c.copy(sym = sym, companion = S(newCompanion))
       Define(newModule, rest)
     case b => b
+
+  def mkDefnMap(b: Block): Unit =
+    val transformer = new BlockTransformer(new SymbolSubst()):
```
NeilKleistGao: If you only collect information and leave the block IR untouched, you can just use `BlockTraverser` instead of `BlockTransformer`

NeilKleistGao (changes requested):

hkmc2/shared/src/main/scala/hkmc2/codegen/Instrumentation.scala
LPTK: Copilot gave you the same suggestion. How come you did not follow it @ChingLongTin?

ChingLongTin: oops, I applied it now.

hkmc2/shared/src/main/scala/hkmc2/codegen/Instrumentation.scala
```diff
@@ -356,3 +371,16 @@ class Instrumentation(using State, Raise) extends BlockTransformer(new SymbolSub
...
       Define(newModule, rest)
     case b => b
+
+  def mkDefnMap(b: Block): Unit =
+    val transformer = new BlockTraverser:
+      override def applyDefn(defn: Defn) = defn match
+      case c: ClsLikeDefn =>
+        defnMap.addOne(c.isym, c)
+        super.applyDefn(defn)
+      case _ => super.applyDefn(defn)
```
LPTK: The bodies of definitions should be indented
```suggestion
        case c: ClsLikeDefn =>
          defnMap.addOne(c.isym, c)
          super.applyDefn(defn)
        case _ => super.applyDefn(defn)
```

LPTK (approved):
Otherwise LGTM.

PR #413, opened 2026-03-13
Fix CI

AnsonYeung:
This fixes the issue of the non deterministic test failure.


hkmc2/shared/src/test/mlscript/handlers/StackSafety.mls
```diff
@@ -269,19 +275,10 @@ fun handler_stack(n) =
 handler_stack(100)
 //│ ═══[RUNTIME ERROR] RangeError: Maximum call stack size exceeded
 
-
-// * For some reason, this works locally with values 100 and 100,
```
LPTK, line 273: It would be nice to still leave a comment explaining what was happening here before this change (the cause of the flakiness).

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -721,5 +721,9 @@ class HandlerLowering(paths: HandlerPaths, opt: EffectHandlers)(using TL, Raise,
     val preTransformed = new PreHandlerLowering().applyBlock(b)
     val ctx = HandlerCtx.TopLevel
     val transformed = translateBlock(preTransformed, ctx, Set.empty)
-    (transformed, stackSafetyMap)
+    val blk = blockBuilder
+      .assignFieldN(paths.runtimePath, Tree.Ident("curEffect"), unit)
+      .assignFieldN(paths.runtimePath, Tree.Ident("resumePc"), intLit(-1))
```
LPTK: Can we rather just call a function like `runtime.resetEffects()` instead, to reduce the bloat a bit? (You should assign it to `noSymbol`.)

LPTK (changes requested):

hkmc2/shared/src/test/mlscript/handlers/StackSafety.mls
```diff
@@ -270,18 +273,12 @@ handler_stack(100)
...
 
 
-// * For some reason, this works locally with values 100 and 100,
-// * but it makes the CI sometimes fail unexpectedly (see https://github.com/hkust-taco/mlscript/pull/410).
-// * Since the test is flaky, I'll disable it for now
-:exit
-====================================================================================================
-
+// See https://github.com/hkust-taco/mlscript/pull/410
+// This test was failing non-deterministically due to handler state not being properly reset.
```
LPTK: Come on. That's a very poor effort at documentation. That PR does nor povide any explanation.

```suggestion
// This test was failing non-deterministically due to handler state not being properly reset.
// The flakiness was due to the stack overflow in the previous block not always interrupting the code in the same place,
// sometimes leaving the effect handlers runtime in an inconsistent state.
```

LPTK (approved):
Otherwise LGTM.

PR #416, opened 2026-03-14
Fix tailrec bug

### LINKED ISSUE #415: Miscompilation of tail-recursive optimized functions with lambdas
Compare

```fs
let lam = () => 0
//│ lam = fun lam

fun foo(x) =
  assert x < 5 else x
  if x == 2 do set lam = () => x
  id(foo(x + 1))

foo(0)
//│ = 5

lam()
//│ = 2
```
with
```fs
let lam = () => 0
//│ lam = fun lam

fun foo(x) =
  assert x < 5 else x
  if x == 2 do set lam = () => x
  foo(x + 1)

foo(0)
//│ = 5

lam()
//│ = 5
```

We can probably reuse an analysis defined in the lifter.

CAG2Mark:
Fixes https://github.com/hkust-taco/mlscript/issues/415.

Also refactored some parts of the lifter and removed some dead code.


hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
```diff
@@ -416,6 +447,26 @@ class TailRecOpt(using State, TL, Raise):
...
+     * fun f(x) =
+     *   set lam = () => x
+     *   f(x + 1)
+     * 
+     * we need to do some analysis on what nested functions use what variables. We
+     * re-use the analysis from the lifter to do this.
+     */
+    
+    given Opt[(ScopeData, AccessMap)] = 
+      if config.liftDefns.isDefined then N
```
LPTK: I think you are making the assumption that if the `liftDefns` option is enabled, there will be no lambdas to deal with here. This uncorroborated assumption is basically asking for silently introducing miscompilation bugs. What tells you some other pass will not introduce lambdas in intermediate output (in fact, that's exactly what deforestation does)? At the very least, you should check there are no lambdas and gracefully handle the case otherwise.

You must design compiler passes so they work on any valid IR, not just so they seem to work when the stars align... 😉

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
```diff
@@ -269,10 +272,35 @@ class TailRecOpt(using State, TL, Raise):
...
+      
+      val copiedParamSyms = copiedParams.map:
+          case x => x -> VarSymbol(x.id)
+        .toMap
+      
       val symRewriter = new BlockTransformer(SymbolSubst()):
-        def applyVarSym(l: VarSymbol): VarSymbol = paramsIdxes.get(l) match
-          case Some(idx) => paramSymsArr(idx)
-          case _ => l
+        def applyVarSym(l: VarSymbol): VarSymbol = copiedParamSyms.getOrElse(
```
LPTK: By the way, any reason we're not using the `SymbolSubst` parameter, here?

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
```diff
@@ -331,9 +359,12 @@ class TailRecOpt(using State, TL, Raise):
           case None => super.applyBlock(b)
         case _ => super.applyBlock(b)
       
-      def rewrite(b: Block) =
-        applyBlock(symRewriter.applyBlock(b))
-    
+      def rewrite(b: Block): Block =
+        val blk = applyBlock(symRewriter.applyBlock(b))
+        val withCopied = copiedParamSyms.toList.sortBy(_._1.uid).foldRight(blk):
```
LPTK: I suppose this might be slightly less inefficient
```suggestion
        val withCopied = copiedParamSyms.toArray.sortBy(_._1.uid).foldRight(blk):
```

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
```diff
@@ -269,10 +272,35 @@ class TailRecOpt(using State, TL, Raise):
...
       
+      val copiedParams: Set[VarSymbol] = accessInfo match
+        case Some((scopeData, accessMap)) =>
+          val node = scopeData.getNode(f)
+          val (_, childDefNodes) = 
+            node.partitionTree2:
+              case _: ScopedObject.Class => true
+              case _: ScopedObject.Func => true
+              case _ => false 
+          childDefNodes.map(_.obj)
```
LPTK: Avoid allocating tons of intermediate lists
```suggestion
          childDefNodes.iterator.map(_.obj)
```

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
```diff
@@ -269,10 +272,35 @@ class TailRecOpt(using State, TL, Raise):
...
+            .collect:
+              case r: ScopedObject.Referencable[?] => r.sym
+            .toSet
+            .flatMap(s => accessMap(s).accessed)
+            .collect:
+              case x: VarSymbol => x 
+            .intersect(params.toSet)
```
LPTK: Can we just
```suggestion
            .flatMap(s => accessMap(s).accessed)
            .collect:
              case x: VarSymbol => x 
            .toSet
            .intersect(params.toSet)
```
?

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
```diff
@@ -269,10 +272,35 @@ class TailRecOpt(using State, TL, Raise):
       val paramsSet = f.params.toSet
       val paramsIdxes = params.zipWithIndex.toMap
       
+      val copiedParams: Set[VarSymbol] = accessInfo match
```
LPTK: Without being familiar with the lifter code, I had trouble understand this complex-looking logic. Could you (very briefly) document what this does and why it works?

LPTK (changes requested):

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
CAG2Mark: There could be lots of duplicated symbols in the different access maps.

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
CAG2Mark: True

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
LPTK: Does it matter? This is just an iterator, which will go through them efficiently using internal mutation and without materializing intermediate results. Seems much better than building so many intermediate set data structures (which _also_, in fine, need to go through all the occurrences anyway).

CAG2Mark: Sure, it probably doesn't matter that much

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
CAG2Mark: Please see the latest commit

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
```diff
@@ -269,18 +272,41 @@ class TailRecOpt(using State, TL, Raise):
...
+          case _: ScopedObject.Func => true
+          case _ => false
+        childDefNodes.iterator.map(_.obj)
+          .collect:
+            case r: ScopedObject.Referencable[?] => r.sym // Obtains the definition symbol of the nested class/function.
+          .flatMap(s => accessMap(s).accessed) // All local variables that each nested class/function could access.
+          .collect:
+            case x: VarSymbol => x 
+          .toSet
+          .intersect(params.toSet)
```
LPTK: Oh, and in fact it should be even better to just do
```suggestion
          .filter(params.toSet)
          .toSet
```
for the same reason of avoiding needless intermediate data structures.

LPTK: Notice that the set before intersection could be huge, so this could be a signficiant optimization.

LPTK (approved):
Otherwise LGTM, thanks!

PR #417, opened 2026-03-14
wasm: Refactor fully to use symbolic indices

Derppening:
Refactor all usages of numeric indices to symbolic indices, so that all top-level module constructs are named.

No functionality changes, except all functions are exposed as element segments. This allows any function to be called using `call_ref` for consistency.


hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -30,14 +31,17 @@ import scala.collection.mutable.{ArrayBuffer as ArrayBuf, Map as MutMap}
...
   */
 class FuncInfo(
-    val id: Opt[SymIdx],
+    val id: SymIdx,
     val typeIdx: TypeIdx,
     params: Seq[Local -> Str],
     nResults: Int,
     locals: Seq[Local -> Str],
     val body: Expr,
+    val exports: Seq[Str],
```
LPTK: Any reason this is a Seq rather than an Option? If so, pls document

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -60,13 +64,33 @@ class FuncInfo(
...
+    SymIdx(sym.optionIf(_.nameIsMeaningful).fold(summon[Scope].allocateName(sym))(_.nme)),
     typeIdx,
     params,
     nResults,
     locals,
     body,
+    sym.optionIf(_.nameIsMeaningful).map(_.nme).toSeq,
+  )
+
+  @deprecated("Consider providing a symbolic identifier by using `Scope.allocateName` with a `TempSymbol`.")
```
LPTK, line 77: If this constructor can do exactly this for us, should it really be deprecated?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -241,22 +273,25 @@ end Ctx
...
-    namedGlobals: MutMap[Symbol, NumIdx],
-    var locals: Ls[MutMap[Local, NumIdx]],
+    namedGlobals: MutMap[Symbol, Int],
+    var locals: Ls[MutMap[Local, Int]],
     private var startFunc: Opt[FuncIdx],
 ) extends ToWat:
 
   import Ctx.prettyString
 
+  /** Monotonically increasing counter for generating object tags. */
```
LPTK: Do you know many counters that are not "monotonically increasing"?
```suggestion
  /** Counter for generating object tags. */
```

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -281,37 +316,54 @@ class Ctx(
     labelTargets.collectFirst:
       case (sym, target) if sym eq label => target
 
+  /** Returns a new number to be used as an object tag. */
+  def getFreshObjectTag(): Int =
+    val tag = objectTagNum
+    objectTagNum += 1
+    tag ensuring objectTagNum > tag
```
LPTK: That's a bit silly, no?
```suggestion
    tag
```

LPTK (approved):
Neat, thanks!

But is there any plan to move away from the deprecated stuff? Did you leave it there to help the FYP students? Also, does it really need to be deprecated?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
Derppening: The only reason is that the WAT syntax allows it. I have changed this into an `Opt` in 0d69b6e.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
Derppening: Good catch, I have un-deprecated this in d0950a37.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
Derppening: Updated in 87097c8.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
Derppening: Removed the redundant check in 4ef7968.

Derppening:
> Neat, thanks!
> 
> But is there any plan to move away from the deprecated stuff? 

The implementation has fully moved away from the deprecated stuff. The only locations that still retains numeric indices is `getType` and `getFunc` for (temporary) backwards compatibility.

> Did you leave it there to help the FYP students?

Yes. I have tried to make all changes API-compatible to ensure that their changes can be applied on top of this with minimal changes.

> Also, does it really need to be deprecated?

I am not sure. It just seems like a good idea to force people to move away from using numeric indices before it is fully removed or made `private` within the `Ctx` implementation.

PR #420, opened 2026-03-16
Switch specialization

CAG2Mark:
Specialize the following, where `ai` are ints:

```fs
if scrut is a1 do
  body1
  set scrut = a2
if scrut is a2 do
  body2
  set scrut = a3
if scrut is an do
  bodyn
```

into a switch statement:

```js
switch (scrut) {
  case a1:
    body1
    scrut = a2;
  case a2:
    body2
    scrut = a3;
  // ...
  case an:
    bodyn
}
```
Note that `scrut` is guaranteed to not change between `set scrut = ai` and `if scrut is ai`, because the JS event loop waits until the entire call stack is cleared before running any other code. Hence, this transformation is safe.

This can be used to reduce the overhead of the state machine transformation for effect handlers.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
@@ -215,8 +215,10 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
     * switch (scrut) {
     *   case a1:
     *     body1
+    *     scrut = a2;
     *   case a2:
     *     body2
+    *     scrut = a3;
     *   ...
     *   case an:
     *     bodyn
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
@@ -17,6 +17,7 @@ import Scope.scope
 import hkmc2.syntax.Tree.UnitLit
 import hkmc2.semantics.Elaborator.ctx
 import hkmc2.syntax.Tree.{IntLit, StrLit}
+import scala.annotation.tailrec
 
 
 // TODO factor some logic for other codegen backends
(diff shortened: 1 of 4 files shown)


AnsonYeung:
> Note that scrut is guaranteed to not change

Pretty sure this technically false but I think this is a relatively safe assumption to make.

CAG2Mark:
> > Note that scrut is guaranteed to not change
> 
> Pretty sure this technically false but I think this is a relatively safe assumption to make.

For JS specifically, when wouldn't this be the case?

LPTK:
First, whether something works due to JS semantics or not is not a sufficient argument to make. We target other backends as well, such as WASM and C++. What matters is that we establish clear and practical semantic expectations that all IR passes need to follow.

Now, one assumption we _can_ make is that we don't have to worry about race conditions for local variables. And arguably, regardless of the target, we should design the language so that race condition considerations are not really needed (simialr to Rust's "fearless" concurrency). And it's true that in JS and WASM this property should hold. So, I agree that the optimization seems to be safe.

@AnsonYeung 

> technically false but I think this is a relatively safe assumption to make

For the record, "seems like it would mostly work" is absolutely not a good enough approach to designing compiler optimizations, outside of one-off student projects and the like. Our transformation passes should be rigorously justified to not affect the semantics of the IR. Ideally we'd want to always have an idea about how we would prove them correct at least on paper.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -248,35 +248,33 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
...
-        Value.Ref(`scrutSym`, _),                    // the scrutinee is ref to `scrutSym`
-        (Case.Lit(Tree.IntLit(`curVal`)), b) :: Nil, // there is only one case matching the previously set int literal
-        S(End(_)), rest)                             // default case exists and does nothing
-        => lastBlkAssign(b) match
+        scrut_ @ Value.Ref(scrutSym_, _),                   // the scrutinee is ref
+        (Case.Lit(Tree.IntLit(curVal_)), b) :: Nil,         // there is only one case matching an int literal
+        S(End(_)), rest                                     // default case exists and does nothing
+      )
+        if scrutSym.map(_ === scrutSym_).getOrElse(true)    // the scrutinee is the same as the one before
+        && curVal.map(_ === curVal_).getOrElse(true)        // the matched int literal is one previously set
```
LPTK: Please use capitalization to mark the beginning of comments. I was super confused trying to parse a single sentence because there is no break between these two comments.

CAG2Mark:
> First, whether something works due to JS semantics or not is not a sufficient argument to make. We target other backends as well, such as WASM and C++. What matters is that we establish clear and practical semantic expectations that all IR passes need to follow.
> 
> Now, one assumption we _can_ make is that we don't have to worry about race conditions for local variables. And arguably, regardless of the target, we should design the language so that race condition considerations are not really needed (simialr to Rust's "fearless" concurrency). And it's true that in JS and WASM this property should hold. So, I agree that the optimization seems to be safe.
> 
> @AnsonYeung
> 
> > technically false but I think this is a relatively safe assumption to make
> 
> For the record, "seems like it would mostly work" is absolutely not a good enough approach to designing compiler optimizations, outside of one-off student projects and the like. Our transformation passes should be rigorously justified to not affect the semantics of the IR. Ideally we'd want to always have an idea about how we would prove them correct at least on paper.

Yes, this is a specific optimization for the JS builder and my comment is that we do not need to worry about race conditions for the `scrut` variable.

LPTK:
> First, whether something works due to JS semantics or not is not a sufficient argument to make.

(I am aware that this is only a change to JSBuilder, but I do think we should try to reuse the same pattern to compile to WASM.)

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -194,6 +195,87 @@ class JSBuilder(using TL, State, Ctx) extends CodeBuilder:
...
+      case Match(rest = rest) => lastBlkAssign(rest)
+      case Scoped(body = rest) => lastBlkAssign(rest)
+      case Label(rest = rest) => lastBlkAssign(rest)
+      case Begin(rest = rest) => lastBlkAssign(rest)
+      case TryBlock(rest = rest) => lastBlkAssign(rest)
+      case Assign(rest = rest) => lastBlkAssign(rest)
+      case AssignField(rest = rest) => lastBlkAssign(rest)
+      case AssignDynField(rest = rest) => lastBlkAssign(rest)
+      case Define(rest = rest) => lastBlkAssign(rest)
+      case HandleBlock(rest = rest) => lastBlkAssign(rest)
```
LPTK: Can we just define a `NonBlockTail` trait with a `val res: Block` so this can be rewritten to
```suggestion
      case nbt: NonBlockTail => lastBlkAssign(nbt.rest)
```
?

LPTK (approved):
Otherwise LGTM, thanks!

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
CAG2Mark: Sure. BTW, I was a bit confused as to why `Scoped` is a `BlockTail`. Any idea why?

LPTK: No idea; looks like it's @ychenfo who made it this way: https://github.com/hkust-taco/mlscript/commit/2d197c4cd41981d7a6e18673af8bdae81911fc60#diff-72bbb753659dd44a5e47f3cd0b3954e4ee95fda127ca491a2c267461779c60fbR339

The `body` of the `Scoped` block does act like a block's `rest`, so `Scoped` should probably be a `NonBlockTail`.

ychenfo: Oh I originally made it this way mainly just because `Scoped` blocks don't have a `rest` field, so we cannot know their immediate continuations by looking at scoped blocks themselves.

But indeed I am not sure if this is a good choice since scoped blocks are not like other `BlockTails` and they contain nested blocks.

LPTK: I think that conceptually the `body` of `Scoped` really is like a `rest`, meaning  that's what the code goes on to do after traversing the block.

AnsonYeung:
> > > Note that scrut is guaranteed to not change
> > 
> > 
> > Pretty sure this technically false but I think this is a relatively safe assumption to make.
> 
> For JS specifically, when wouldn't this be the case?

```js
> Object.defineProperty(globalThis, "x", {value: 3});
<ref *1> Object [global] {
  global: [Circular *1],
  clearImmediate: [Function: clearImmediate],
  setImmediate: [Function: setImmediate] {
    [Symbol(nodejs.util.promisify.custom)]: [Getter]
  },
  clearInterval: [Function: clearInterval],
  clearTimeout: [Function: clearTimeout],
  setInterval: [Function: setInterval],
  setTimeout: [Function: setTimeout] {
    [Symbol(nodejs.util.promisify.custom)]: [Getter]
  },
  queueMicrotask: [Function: queueMicrotask],
  structuredClone: [Function: structuredClone],
  atob: [Getter/Setter],
  btoa: [Getter/Setter],
  performance: [Getter/Setter],
  fetch: [Function: fetch],
  navigator: [Getter],
  crypto: [Getter]
}
> x
3
> x = 4
4
> x
3
>
```

We just need to be careful to not do stuff like this.

LPTK:
Sure, but this is not a local variable, even though the JS surface language syntax makes it look like one. In the IR, it would be represented with a `Select` node, and we would never generate this sort of code for a local variable.

(Similar story for JS's crazy `arguments` semantics, which can't be observed from MLscript:
```js
> f = function(x){ console.log(x); arguments[0] = "oops"; console.log(x) }
[Function: f]
> f("Hello")
Hello
oops
```
)

PR #421, opened 2026-03-17
Support "match only" mode in pattern compilation

chengluyu:


hkmc2/shared/src/main/scala/hkmc2/semantics/ups/Compiler.scala
```diff
@@ -21,6 +21,16 @@ import scala.annotation.tailrec
   */
 class Compiler(using Context)(using tl: TL)(using Ctx, State, Raise) extends TermSynthesizer:
   import Compiler.*, tl.*
+
+  private def bool(value: Bool): Term = Term.Lit(BoolLit(value))
+
+  private def isMatchOnly(using mode: ResultMode): Bool = mode === ResultMode.MatchOnly
```
LPTK:
```suggestion
  private def isMatchOnly(using mode: ResultMode): Bool = mode is ResultMode.MatchOnly
```

hkmc2/shared/src/test/mlscript-compile/ups/EvaluationContext.mls
```diff
@@ -0,0 +1,83 @@
...
+  (App(Value as lhs, Ctx(Hole) as rhs) => App(lhs, rhs)) |
+  (App(Ctx(Hole) as lhs, rhs)          => App(lhs, rhs))
+
+pattern Redex = App(Abs(Var(x), m), Abs as n) => subst(m, x, n)
+
+pattern Step = (Ctx(Redex) as res) => res
+
+fun isStep_naive(t) = t is Step
+fun isStep_optimized(t) =
+  if t is (@compile Step) as s then s else ()
```
LPTK, line 82: Technically we should be smart enough to still compile it as match-only. Ie we should be able to see that this `as` just binds the scrutinee since `Step` doesn't perform any transformation.

hkmc2/shared/src/main/scala/hkmc2/semantics/ups/Compiler.scala
```diff
@@ -173,13 +183,9 @@ class Compiler(using Context)(using tl: TL)(using Ctx, State, Raise) extends Ter
...
         val split = makeSplit(
-          // There's no transform in the topmost pattern. So, we just make and
-          // return a `MatchSuccess` instance.
           makeConsequent = (outputSymbol, bindings) => Split.Else:
-            makeMatchSuccess(outputSymbol.use, bindings.use),
-          // Here the string "topmost" is just to indicate the failure is
-          // passed from the topmost split for the purpose of debugging.
-          alternative = Split.Else(makeMatchFailure(str("topmost"))))
+            successfulMatchResult(outputSymbol, bindings),
+          alternative = Split.Else(emptyMatchResult("topmost")))
```
LPTK, line 217: Was the previous version of this code and
```scala
          // There's no transform in the topmost pattern. So, we just make and
          // return a `MatchSuccess` instance.
```
incorrect?

hkmc2/shared/src/main/scala/hkmc2/semantics/ups/Compiler.scala
```diff
@@ -409,25 +442,72 @@ object Compiler:
         fillImplicitArgs(sel.prefix)
         sel.resolve
       case _: Term => term
+    type ClassLikeDefnSymbol = ClassSymbol | ModuleOrObjectSymbol | PatternSymbol
+    def isMatchingSymbol(candidate: ClassLikeDefnSymbol): Bool =
```
LPTK: What's the logic behind these strange name-based comparisons? It doesn't seem this could be correct. What if distinct definitions happen to share the same name?

hkmc2/shared/src/main/scala/hkmc2/semantics/ups/Compiler.scala
```diff
@@ -409,25 +442,72 @@ object Compiler:
...
-      .flatten
+      val direct = elem.symbol.iterator.flatMap(classLikeCandidates).collectFirst:
+        case candidate if isMatchingSymbol(candidate) =>
+          elem.ref(new Ident(candidate.nme)).withLoc(loc).resolved(candidate)
+      direct.orElse:
+        elem.symbol.iterator.collectFirst:
+          case owner: (ClassSymbol | ModuleOrObjectSymbol) =>
+            ownerMemberReference(owner)
+        .flatten
+    def ownerChainReference(symbol: ClassLikeDefnSymbol): Opt[Term] =
```
LPTK, line 600: Can you explain why we suddenly need this?

hkmc2/shared/src/main/scala/hkmc2/semantics/ups/SplitCompiler.scala
```diff
@@ -123,6 +124,23 @@ class SplitCompiler(using tl: TL)(using State, Ctx, Raise) extends TermSynthesiz
...
+      _.symbol.defn.exists(_.extractionParams.nonEmpty)
+
+  private def hasExplicitExtractionMatches(scrutinee: Scrut, pattern: SP): Bool = pattern match
+    case Constructor(target, arguments) => target.resolvedSym.flatMap(_.asPat).exists: patternSymbol =>
+      val defn = patternSymbol.defn.getOrElse:
+        lastWords(s"Pattern `${patternSymbol.nme}` has not been elaborated.")
+      // This check is only used to pick full vs match-only compilation.
+      // Reuse the argument pairing logic, but suppress diagnostics here so we
+      // don't report the same user error twice.
+      given Raise = Function.const(())
```
LPTK: Bad idea. Please just add an option to "the argument pairing logic" that makes it omit the errors in question.

LPTK (changes requested):

hkmc2/shared/src/main/scala/hkmc2/semantics/ups/Compiler.scala
chengluyu: No, it was not incorrect. We used to return `MatchSuccess` at this point, but now, when `MatchOnly` is set, we return `true` here.

hkmc2/shared/src/main/scala/hkmc2/semantics/ups/Compiler.scala
chengluyu: A few days ago, while benchmarking the compiled patterns, I discovered a bug: when using `@compiled` in the `mlscript-compile` folder, it was unable to correctly synthesize references pointing to `class`/`module`/`object`s. They produced a lot of `Error` terms. So I need to strengthen the reverse lookup in `reference` method here.

I admit that this part is not well-written because I didn't know exactly what was in `Ctx`. So I ended up developing the logic here by testing as I wrote. I will clean this up properly. At the very least, I will stop using the name-based lookup mentioned in the above comment.

hkmc2/shared/src/main/scala/hkmc2/semantics/ups/Compiler.scala
chengluyu: 🥲 I just discovered that `isMatchingSymbol` can be safely deleted. I likely added this code for easier comparison during debugging, but I've found that the system still works exactly the same without it.

hkmc2/shared/src/test/mlscript-compile/ups/EvaluationContext.mls
chengluyu: I think `as s` here binds the output of `Step`. Am I understanding this correctly?

LPTK: Oh right, I missed the context that `Step` here does indeed perform transformation.

Can you add a test with `P as x` where `P` does not perform any transformation and check that we're still compiling it in match-only mode?

chengluyu: I tried it out. While we indeed didn't perform the transformation, we still carry the scrutinees using `MatchSuccess` through out the entire matching process. I feel like this part can be optimized. Let me give it a try.

chengluyu: Hmmm, we need to add another analysis to determine whether the compiled pattern always returns the original scrutinee without any bindings to be used in transformations.

LPTK: Ok, this should probably be done in a follow-up PR, then.

chengluyu: I think I've done it. Please have a look.

hkmc2/shared/src/main/scala/hkmc2/semantics/ups/SplitCompiler.scala
```diff
@@ -897,35 +958,57 @@ class SplitCompiler(using tl: TL)(using State, Ctx, Raise) extends TermSynthesiz
...
+    // We must stay in `Full` mode when
+    // - the pattern explicitly matches extraction arguments, or when any
+    //   reachable instantiated pattern definition has extraction slots;
+    // - otherwise, if the use site needs an output, we only need `Full`
+    //   mode when a successful match does have transformations.
+    // 
+    // When the instantiated pattern is transform-free and thus preserves the
+    // original scrutinee as its output, `MatchOnly` is used and the caller can
+    // obtain the scrutinee directly.
+    val canReuseScrutineeOutput = preservesScrutineeOutput(synonym, context)
```
LPTK: Shouldn't this be
```suggestion
    val canReuseScrutineeAsOutput = preservesScrutineeAsOutput(synonym, context)
```
?

hkmc2/shared/src/test/mlscript/ups/examples/Extraction.mls
LPTK: Is this comment outdated?

hkmc2/shared/src/test/mlscript/ups/MatchResult.mls
```diff
@@ -22,24 +22,550 @@ data class MatchSuccess(x)
...
+//│       set patternOutput = matchRecord.output⁰;
+//│       set bindings = matchRecord.bindings⁰;
+//│       return true
+//│     else
+//│       return false
+//│   end
+//│ };
+//│ end
+
+:sir
```
LPTK: I think this is too much output to commit.

hkmc2/shared/src/test/mlscript/ups/MatchResult.mls
LPTK, line 1: Please clean up this file: only keep one or two _small, representative_ uses of `:sir`.

hkmc2/shared/src/test/mlscript/ups/TransformFree.mls
LPTK, line 1: Please clean up this file: only keep one or two _small, representative_ uses of `:sir`.

LPTK (approved):
Ok, let's get this merged, thanks. However there is a pre-existing problem with the difference between compiled and interpreted patterns (see https://github.com/hkust-taco/mlscript/issues/422).

hkmc2/shared/src/test/mlscript-compile/ups/EvaluationContext.mls
```diff
@@ -0,0 +1,83 @@
...
+    name === name' then term
+    let target' = if target.freeVars.has(name')
+      then subst(target, name', target.freshVar(name'))
+      else target
+    else Abs(Var(name'), subst(body, name, target'))
+  App(lhs, rhs) then App(subst(lhs, name, target), subst(rhs, name, target))
+
+pattern Value = Abs
+
+pattern Ctx(pattern Hole) =
```
LPTK, line 71: Don't forget to update this to the new approach (which doesn't require `=>`) and to use prefix-`|`.

hkmc2/shared/src/test/mlscript/ups/transformation/BindingLess.mls
```diff
@@ -0,0 +1,82 @@
...
+toUppercaseList of "a" :: "X" :: "c" :: Nil
+//│ = Cons("A", Cons("X", Cons("C", Nil)))
+
+pattern ToUppercase = ToUppercase ~ ("" | ToUppercase)
+
+fun toUppercase(input) = if input is ToUppercase as output then output
+
+:re
+toUppercase of "aXc"
+//│ ═══[RUNTIME ERROR] RangeError: Maximum call stack size exceeded
```
LPTK: 🤔

chengluyu: This is what I said didn't work during the meeting. I'm now fixing it.

hkmc2/shared/src/test/mlscript/ups/transformation/BindingLess.mls
```diff
@@ -0,0 +1,82 @@
...
+
+:expect false
+ps is @compile IncFst
+//│ = false
+
+:expect true
+pi is @compile IncFst
+//│ = true
+
+pattern Char = (Str as s where s.length is 1) => s
```
LPTK: `Char` should probably be a builtin pattern, right? Otherwise, because of the `where` clause, I guess it won't behave as expected within compiled string patterns with concatenation.

hkmc2/shared/src/test/mlscript/ups/transformation/BindingLess.mls
chengluyu: I've implemented a preliminary fix here. The issue will be fully resolved in the string compilation PR.

hkmc2/shared/src/test/mlscript/ups/transformation/BindingLess.mls
chengluyu: I placed it into `Char.mls` for now. When I implement the string compilation later, I will make it a built-in pattern.

LPTK: I'm leaving  this "unresolved" for reference. We'll still merge the PR.

hkmc2/shared/src/test/mlscript/syntax/WeirdBrackets.mls
```diff
@@ -71,7 +71,9 @@ fun interesting(x) =
 //│ >    x is Box(Bool) then 2
 //│ Expanded split with flattened patterns:
 //│ >  ‹if|while›
-//│ >    x is Box(arg$Box$0$) and tmp:arg$Box$0$ is Int then 1
+//│ >    x is Box(arg$Box$0$) and
+//│ >      tmp:arg$Box$0$ is Int then 1
+//│ >      x is Box(arg$Box$0$) and tmp:arg$Box$0$ is Bool then 2
```
LPTK: I wonder why we have more splits now, and some duplication, whereas there was none before.

hkmc2/shared/src/test/mlscript/ups/transformation/BindingLess.mls
```diff
@@ -1,7 +1,9 @@
 :js
 
+import "../../../mlscript-compile/Char.mls"
 import "../../../mlscript-compile/Stack.mls"
 
+open Char { AnyCharacter }
```
LPTK: Why not `AnyChar`?

hkmc2/shared/src/test/mlscript/syntax/WeirdBrackets.mls
chengluyu: I found it. It is because of the changes in f88a6f891.

chengluyu: Yeah, I just forgot to commit the changes here.

hkmc2/shared/src/main/scala/hkmc2/codegen/Printer.scala
```diff
@@ -37,6 +37,7 @@ class Printer(using Raise, ShowCfg, SymbolPrinter):
         case Case.Cls(cls, path) => doc"${print(cls)}"
         case Case.Tup(len, inf) => doc"Array($len${if inf then "+" else ""})"
         case Case.Field(name, safe) => doc"${if safe then "" else "Object "}{ ${name.name} }"
+        case _ => TODO(c)
```
LPTK: I think you added this after mis-merging the main branch. Please make a pass on your whole diff to make sure there aren't other things like this, and take a look at the compilation warnings.

LPTK (changes requested):

hkmc2/shared/src/test/mlscript-compile/ups/EvenOddTree.mls
```diff
@@ -0,0 +1,30 @@
...
+
+object A
+object B
+
+data class Node[T](left: Node[T], value: T, right: Node[T])
+
+// The patterns generate trees that have odd or even numbers of node `A`.
+
+pattern OddTree =
+  A |
```
LPTK: Please fix thsi ugly syntax everywherem not JUST in the place I previously commented...

hkmc2/shared/src/test/mlscript/ups/TransformFree.mls
```diff
@@ -0,0 +1,26 @@
...
+// site binds the pattern's output, it still creates `MatchSuccess` and
+// `MatchFailure`.
+
+// :sir
+fun toNumPair(t) = if t is (@compile NumPair) as s then s else null
+
+
+
+// :sir
+fun toSumPair(t) = if t is (@compile SumPair) as s then s else null
```
LPTK:
```suggestion
// :sir
fun toSumPair(t) = if t is (@compile SumPair) as s then s else null

```

LPTK (approved):
Do you have any tests that shows the IR of a match-only pattern? So we can make sure it doesn't regress later.

Otherwise LGTM, thanks!

hkmc2/shared/src/test/mlscript/ups/MatchResult.mls
```diff
@@ -43,8 +43,37 @@ Cross.unapply("0")
 // :sir
 fun foo(x) = x is Cross
 
-// :sir
+:sir
+// Keep the IR output to prevent regressions.
```
LPTK, line 47: I think the one below is enough.

hkmc2/shared/src/test/mlscript/ups/examples/Computation.mls
```diff
@@ -7,12 +7,12 @@ data class Celsius(value: Num)
 data class Fahrenheit(value: Num)
 
 pattern ToFahrenheit =
-  ((Celsius(c)) => Fahrenheit((c * 9 / 5) + 32)) |
-  ((Fahrenheit as f) => f)
+  | ((Celsius(c)) => Fahrenheit((c * 9 / 5) + 32))
```
LPTK: You don't need parentheses in all these examples.

hkmc2/shared/src/test/mlscript/ups/TransformFree.mls
LPTK: Why did you close this? You haven't applied the change.

PR #423, opened 2026-03-21
Inliner, Break Optimization and Fixing UCS Tail Calls

AnsonYeung:
This PR implements inlining. As naive inlining generates unnecessary labels and break, this PR adds break optimization to optimize away useless breaks. The UCS part of #435 is addressed.

Method inlining is not enabled with this PR.


hkmc2/shared/src/test/mlscript/tailrec/Errors.mls
```diff
@@ -11,19 +11,17 @@ fun f(x) =
 :e
 fun f(x) = 2
 fun g(x) = @tailcall f(x)
-//│ ╔══[COMPILATION ERROR] This call is not optimized as it does not directly recurse through its parent function.
-//│ ║  l.13: 	fun g(x) = @tailcall f(x)
-//│ ╙──      	                     ^^^^
+//│ FAILURE: Unexpected lack of compilation or type error
```
LPTK: Pls disable inlining for this test to retain the failure.

hkmc2/shared/src/test/mlscript/tailrec/TailRecOpt.mls
```diff
@@ -155,6 +155,11 @@ fun f(x) =
   fun g() =
     @tailcall f(x)
   @tailcall g()
+//│ FAILURE: Unexpected compilation error
```
LPTK: Pls disable inlining for this test to retain the failure.

hkmc2/shared/src/main/scala/hkmc2/Config.scala
```diff
@@ -27,6 +27,7 @@ case class Config(
   rewriteWhileLoops: Bool,
   tailRecOpt: Bool,
   deforest: Opt[Deforest],
+  inliner: Opt[Inliner],
```
```suggestion
  inlining: Opt[Inliner],
```

hkmc2/shared/src/main/scala/hkmc2/codegen/inline/Inliner.scala
```diff
@@ -0,0 +1,285 @@
...
+object Inliner:
+  object TermSymbolPath:
+    def unapply(p: Path) = p match
+      case Value.Ref(l, S(ts: TermSymbol)) => S(ts)
+      case s: Select => s.symbol match
+        case S(ts: TermSymbol) => S(ts)
+        case _ => N
+      case _ => N
+  
+  def matchArgs(args: List[Arg], params: ParamList): Option[List[(Local, Result)]] =
```
LPTK: Could you please illustrate each handled case with a concrete example?

hkmc2/shared/src/main/scala/hkmc2/codegen/inline/Inliner.scala
```diff
@@ -0,0 +1,285 @@
...
+
+import mlscript.utils.*, shorthands.*
+import utils.*
+
+import hkmc2.codegen.*
+import hkmc2.semantics.*
+import semantics.Elaborator.State
+
+object Inliner:
+  object TermSymbolPath:
```
LPTK: Pls document why we need this strange extractor.

hkmc2/shared/src/main/scala/hkmc2/codegen/inline/Inliner.scala
```diff
@@ -0,0 +1,285 @@
...
+      if doRename then
+        needsSub += sym
+    
+    def doSymbolSubst(orig: Symbol, newSym: => Symbol): Symbol =
+      if needsSub(orig) then
+        subMap.getOrElseUpdate(orig, newSym)
+      else
+        orig
+
+    object Subst extends SymbolSubst:
```
LPTK: This cumbersome logic should probably be extracted into a new `SubstMap extends Subst` that maintains its mapping and the messy casting internally.

hkmc2/shared/src/test/mlscript/codegen/InnerNameHygiene.mls
LPTK, line 1: Let's disable inlining in this file to keep seeing the `proxy$`s that are being tested.

LPTK (commented):
The logic looks a bit more complex than I expected. I am curious: what is the main source of complexity?

AnsonYeung:
> The logic looks a bit more complex than I expected. I am curious: what is the main source of complexity?

The main complexity imo is that function definition and usage is not in order, which requires at least 2 traversals. The function could be mutually recursive for example. Even in this case, one function could be inlined into another if one of the function is only used once or is small enough.

Other complexities are symbol renaming, and spread arguments. Multiple chained calls to multiple param list is not supported for inlining currently.

A global function symbol to definition map would simplify this a bit, it would mean no more Define node for functions however.

LPTK:
> The main complexity imo is that function definition and usage is not in order, which requires at least 2 traversals. The function could be mutually recursive for example. Even in this case, one function could be inlined into another if one of the function is only used once or is small enough.

Oh, this makes me think about that old GHC paper where they had what looked like a reasonable solution: [Secrets of the Glasgow Haskell Compiler inliner](https://www.microsoft.com/en-us/research/wp-content/uploads/2002/07/inline.pdf). It would be great if you could see whether you could implement something like their solution (sec. 4.2). Would also be a nice paper to present in a group meeting/seminar.

> Other complexities are symbol renaming, and spread arguments. Multiple chained calls to multiple param list is not supported for inlining currently.

Would it help to make `Call` take multiple argument lists? I've been thinking that doing so would make sense, to reflect multiple parameter lists.

> A global function symbol to definition map would simplify this a bit, it would mean no more Define node for functions however.

Can't you just define a reusable pass that creates that mapping? We could use it in other places as well, so that would be useful. We can't really do away with `Define` nodes because functions are scoped, at least nested ones. And carrying around/updating a map all over the place isn't necessarily a win, as far as I can see.

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Rewrite.scala
```diff
@@ -425,112 +425,31 @@ class DeforestRewriter(val solver: DeforestConstrainSolver)(using Raise):
...
-            case None => 
-              mapping.get(l) match
-                case None => k(Value.Ref(l, x))
-                case Some(newBms: BlockMemberSymbol) => k(Value.Ref(newBms, newBms.tsym))
-                case Some(newSym) => k(Value.Ref(newSym, N))
             case Some(bms) =>
               k(Value.Ref(bms, l.asMod))
+            case None => super.applyValue(v)(k)
         case _ => super.applyValue(v)(k)
-    end RefreshSymbol
```
LPTK, line 533: Let's not remove existing `end`s.

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Rewrite.scala
```diff
@@ -425,112 +425,31 @@ class DeforestRewriter(val solver: DeforestConstrainSolver)(using Raise):
...
-                newBms
-              case varSym: VarSymbol => new VarSymbol(varSym.id)
-              case _ => lastWords(s"unexpected symbol kind: $s")
-            mapping(s) = newS
-            newSyms.add(newS)
-          val res = Scoped(newSyms, applyBlock(body))
-          for s <- syms do mapping.remove(s)
-          res
-        case _ => super.applyScopedBlock(b)
+          case Scoped(syms, body) =>
```
LPTK: `match` body doesn't need indentation.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -181,9 +240,258 @@ class BlockSimplifier(symbolsToPreserve: Set[Local])(using DebugPrinter, State):
...
+          map = map + (f.dSym -> InlinerFunInfo(f, isMethod, 0, false, false))
+        
+        override def applyDefn(defn: Defn): Unit = defn match
+          case f: FunDefn =>
+            addFunctionAndApplyBody(f, false)
+          case c: ClsLikeDefn =>
+            c.parentPath.foreach(applyPath)
+            c.methods.foreach: f =>
+              addFunctionAndApplyBody(f, true)
+            // Note: no tracking, since instantiate will not be inlined and won't cause cycles.
```
LPTK: Please make sure to write grammatically-understandable documentation
```suggestion
            // Note: no tracking, since `Instantiate` will not be inlined and won't cause cycles.
```

hkmc2/shared/src/test/mlscript-compile/Predef.mjs
```diff
@@ -156,31 +156,23 @@ let Predef1;
...
-                break split_root$
-              } else {
-                break split_1$
-              }
-            } else {
-              break split_1$
+      scrut = a === b;
+      if (scrut === true) {
+        tmp = true;
+        break split_root$
```
LPTK: Shouldn't this `break` also be removed? It breaks straight to the enclosing label `split_root$`.

hkmc2/shared/src/test/mlscript/deforest/todos.mls
```diff
@@ -30,6 +30,7 @@ data class Some(x)
 // in "rest function"s and after the early return (and becomes dead code).
 // A dead code elimination pass that detects code after early return would solve this problem.
 :sir
+:sjs
```
LPTK: We probably don't want that in the final changes, right?

LPTK:
When I try to run the tests locally, I get in `CamlLightTest.mls`:
```
```
Does inlining remove tail-recursion optimization opportunities?

hkmc2/shared/src/test/mlscript-compile/Predef.mjs
AnsonYeung: It's hard to see on the diff but this break skips code. The bracket on line 151 close on line 242

hkmc2/shared/src/test/mlscript/backlog/TailRecFailure.mls
```diff
@@ -65,7 +65,6 @@ fun go(x) = if x is
 //│ end
 //│ —————————————————| Output |—————————————————————————————————————————————————————————————————————————
 
-// FIXME: not tail-rec optimized due to consequent-sharing
```
LPTK, line 68: Is the problem fully fixed? Or can some UCs expressions still trigger it (then, we should reproduce it here).

hkmc2/shared/src/test/mlscript-compile/Predef.mjs
LPTK: Oh ok, got it.

hkmc2/shared/src/main/scala/hkmc2/codegen/inline/Inliner.scala
AnsonYeung: See the test file `Inliner.mls`. These cases are handling the spread argument in different position.

hkmc2/shared/src/test/mlscript/backlog/TailRecFailure.mls
AnsonYeung: It should be fixed. The code generating the break statement will directly generate the return if `k` is `Ret` now

hkmc2/shared/src/main/scala/hkmc2/codegen/inline/Inliner.scala
AnsonYeung: Since the Call node takes in a path, this allow matching for a specific function or method. However method can be virtual and also the `this` parameter needs to be adjusted for inlining and so method is not actually inlined right now. I added this as the doc:
```
// Reference to a function body can occur as a.f or f, this handles both cases.
```

hkmc2/shared/src/main/scala/hkmc2/codegen/inline/Inliner.scala
AnsonYeung: This part is superseded by SymbolRefresher

hkmc2/shared/src/main/scala/hkmc2/codegen/inline/Inliner.scala
LPTK: This will also mathc virtual calls, right? But presumably virtual functions won't be in the mapping we're maintaining so that's fine?

AnsonYeung: Right now method is not considered at all for inlining so it's fine. However I think later if a class have 2 methods f and g where f calls g, then we need to know that g is not virtual / final in order to inline. Inlining methods should be somewhat important since most functions in a program will be methods.

LPTK: Yes, we can consider method inlining later. Following the previous version of MLscript, I intend to make methods not virtual by default, requiring a `virtual` modifier to make them so.

hkmc2/shared/src/test/mlscript/backlog/TailRecFailure.mls
LPTK: Great. Please add this change/fix to the PR title and description (which are currently not very informative).

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -37,7 +37,7 @@ object Thrw extends TailOp:
 
 class LoweringCtx(
   initMap: Map[Local, Value], // No longer in meaningful use and could be removed if we don't find a use for it
-  val mayRet: Bool, // TODO[Anson]]: document what this is for...
+  val mayRet: Bool, // For rewriting while loop into tail recursive function
```
LPTK: This is insufficient. Pls explain what's the semantics of this field and what it's doing in `LoweringCtx`.

hkmc2/shared/src/test/mlscript-compile/apps/parsing/Lexer.mls
```diff
@@ -1,4 +1,4 @@
-
+#config(liftDefns: Some(LiftDefns))
```
LPTK, line 1: Neat, so it works, now!

LPTK, line 1:
```suggestion
#config(liftDefns: Some(LiftDefns))

```

AnsonYeung:
I'm not sure whether SymbolRefresher really handles everything. The logic is quite weird since a symbol can be introduced in many weird ways.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
AnsonYeung: changed

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -37,7 +37,7 @@ object Thrw extends TailOp:
 
 class LoweringCtx(
   initMap: Map[Local, Value], // No longer in meaningful use and could be removed if we don't find a use for it
-  val mayRet: Bool, // For rewriting while loop into tail recursive function
+  val mayRet: Bool, // For rewriting while loop into tail recursive function, represent whether a explicit return is legal in the current block
```
LPTK: an*

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
CAG2Mark: This should fix the issue with the switch specialization
```suggestion
        S(End(_)) | N, rest                                 // Default case exists and does nothing.
```

CAG2Mark: Oh, for some reason the line in the suggestion is empty. Pls apply the suggestion manually.

LPTK: Thanks, changing this line worked.

hkmc2/shared/src/test/mlscript/codegen/Inliner.mls
```diff
@@ -0,0 +1,242 @@
+:js
+
+:sjs
```
LPTK: Why not use `:soir` consistently, instead of `:sjs`? Here, and everywhere else in the PR.

hkmc2/shared/src/test/mlscript/codegen/Inliner.mls
```diff
@@ -0,0 +1,242 @@
...
+//│ let f2;
+//│ f2 = function f() {
+//│   let done, done1;
+//│   done1 = function done(x1) { return x1 };
+//│   done = done1;
+//│   return done
+//│ };
+//│ —————————————————| Output |—————————————————————————————————————————————————————————————————————————
+
+:soir
```
LPTK, line 40:
```suggestion
// * TODO: support spread arguments like this one
:soir
```

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -155,13 +167,60 @@ class BlockSimplifier(symbolsToPreserve: Set[Local])(using DebugPrinter, State):
...
           else
             registerChange
-            Begin(applyBlock(bod), applyBlock(rst))
+            Begin(freshLabelCtx(applyBlock(bod)), applyBlock(rst))
+      
+      // * Remove useless break
+      case Break(label) if tailLabels.contains(label) =>
+        tl.log(s"Break ${label} is eliminated: current tail label list is ${tailLabels}")
+        registerChange
+        End()
```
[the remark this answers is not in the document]
LPTK: `tl` is a methods defined in `hkmc2/shared/src/main/scala/utils/TraceLogger.scala`.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -181,9 +240,258 @@ class BlockSimplifier(symbolsToPreserve: Set[Local])(using DebugPrinter, State):
...
+      
+      def matchArgs(args: List[Arg], params: ParamList): Option[List[(VarSymbol, Result)]] =
+        if args.exists(_.spread.isDefined) then
+          // we require a precise match when any arg is a spread arg
+          if params.restParam.isEmpty then return N
+          if args.exists(_.spread.exists(!_.isEager)) then return N
+          val pairs = args.zip(params.params.iterator.map((_, false)) ++ params.restParam.map((_, true)))
+          if pairs.exists((arg, param) => arg.spread.isDefined =/= param._2) then return N
+          S(pairs.map((arg, param) => (param._1.sym, arg.value)))
+        else
```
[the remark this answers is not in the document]
LPTK: @AnsonYeung What happens when there is a length mismatch and `zip` drops some elements from either side? Is the call semantics of the non-inlined version preserved? Please add a comment about that as well as regression tests for both cases.

hkmc2/shared/src/main/scala/hkmc2/codegen/SymbolRefresher.scala
```diff
@@ -0,0 +1,119 @@
...
+      val newSyms = MutSet.empty[Symbol]
+      val oldSyms = MutSet.empty[Symbol]
+      for s <- syms.toList.sortBy(_.uid) do
+        assert(!mapping.isDefinedAt(s), s"already defined: $s")
+        val newS = s match
+          case tmpSym: TempSymbol => new TempSymbol(N, tmpSym.nme)
+          case bms: BlockMemberSymbol =>
+            val newBms = new BlockMemberSymbol(bms.nme, Nil, bms.nameIsMeaningful)
+            newBms.tsym = bms.tsym.map: t =>
+              val newOwner = t.owner.map(o => existingMapping.getOrElse(o, o).asInstanceOf)
```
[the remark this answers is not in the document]
LPTK: The code proposed by Copilot does seem better?

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -77,6 +82,13 @@ class BlockSimplifier(symbolsToPreserve: Set[Local])(using DebugPrinter, State):
       
       applyProgram(prog)
     
+    def freshLabelCtx[T](thunk: => T)(using Line) =
```
LPTK: Please document what this is for. (Make an _actual_ effort so that someone unfamiliar with this code can understand.)

LPTK:
```suggestion
    inline def nestLabelCtx[T](inline thunk: => T) =
```

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -155,13 +167,60 @@ class BlockSimplifier(symbolsToPreserve: Set[Local])(using DebugPrinter, State):
...
-          else Begin(applyBlock(bod), unr)
+          then Label(lbl, loop, freshLabelCtx(applyBlock(bod)), unr)
+          else Begin(freshLabelCtx(applyBlock(bod)), unr)
         else
-          if usedLabels.contains(lbl) then super.applyBlock(b)
+          if usedLabels.contains(lbl) then
+            def computeBod =
+              tailLabels += lbl
+              val result = applyBlock(bod)
+              tailLabels -= lbl
```
LPTK: Why not use `freshLabelCtx`/`nestLabelCtx`?

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -155,13 +167,60 @@ class BlockSimplifier(symbolsToPreserve: Set[Local])(using DebugPrinter, State):
...
+              val (cse, blk) = tup
+              val blk2 = freshLabelCtx(applySubBlock(blk))
+              applyCase(cse): cse2 =>
+                if (cse2 is cse) && (blk is blk2) then k(tup) else k(cse2 -> blk2)
+          ): arms2 =>
+              val dflt2 = freshLabelCtx(dflt.mapConserve(applySubBlock))
+              val rst2 = applySubBlock(rst)
+              if (scrut2 is scrut) &&
+                  (arms2 is arms) &&
+                  (dflt2 is dflt) && (rst2 is rst)
```
LPTK:
```suggestion
              if (scrut2 is scrut)
              && (arms2 is arms)
              && (dflt2 is dflt)
              && (rst2 is rst)
```

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -155,13 +167,60 @@ class BlockSimplifier(symbolsToPreserve: Set[Local])(using DebugPrinter, State):
...
+      
+      // * Create fresh label contexts for non tails
+      case Begin(sub, _: End) => super.applyBlock(b)
+      case Begin(sub, rst) =>
+        val sub2 = freshLabelCtx(applySubBlock(sub))
+        val rst2 = applySubBlock(rst)
+        if (sub2 is sub) && (rst2 is rst) then b else Begin(sub2, rst2)
+      
+      case Match(scrut, arms, dflt, _: End) => super.applyBlock(b)
+      case Match(scrut, arms, dflt, rst) =>
```
LPTK: Instead of duplicating this error-prone code, can we just modify `BlockTransformer` so that it uses an `applyMatchArmBlock` in these positions, that we can easily override? Try to keep it DRY.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -181,9 +240,261 @@ class BlockSimplifier(symbolsToPreserve: Set[Local])(using DebugPrinter, State):
...
+      case class InlinerFunInfo(
+        defn: FunDefn,
+        isMethod: Bool,
+        private[InlinerAnalyzer] var useCount: Int,
+        private[InlinerAnalyzer] var hasNakedRef: Bool,
+        private[InlinerAnalyzer] var isLoopBreaker: Bool,
+      ):
+        def isPrivate = !symbolsToPreserve.contains(defn.sym)
+        
+        def canBeInlineEliminated =
```
LPTK, line 278: Is this doc correct?
```suggestion
        /** Whether the definition can be removed 1from the program after it has been inlined. */
        def canBeInlineEliminated =
```
or please update with a more accurate doc.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -181,9 +240,261 @@ class BlockSimplifier(symbolsToPreserve: Set[Local])(using DebugPrinter, State):
...
+          isPrivate && !isMethod && useCount <= 1 && !hasNakedRef && !isLoopBreaker
```
LPTK, line 279: Can't we eliminate functions used several times?

LPTK (approved):
Nice work! The code looks pretty clean, although still under-documented.

Please address the comments, and otherwise LGTM.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
AnsonYeung: Maybe it's a bit misnamed, but yes. This is for one used function which will always benefit from inlining as they won't contribute to code duplication

LPTK: How about `def noBrainerInlinable`? 😄

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -77,6 +82,25 @@ class BlockSimplifier(symbolsToPreserve: Set[Local])(using DebugPrinter, State):
...
     
+    // Evaluate `thunk` with a new tail label set. This is used for evaluating any sub blocks that is not in the tail position.
+    // For example, the match arms within a `Match` node are not in the tail position unless the rest block is `End`.
+    // When evaluating the match arms, the tail labels should not be considered to be at tail.
+    // The tail label set is restored after `thunk` completes.
+    inline def freshLabelCtx[T](inline thunk: => T): T =
+      val oldTailLabels = tailLabels
+      tailLabels = MutSet.empty
+      val result = thunk
+      assert(tailLabels.isEmpty)
```
LPTK: Ok, that's a bit silly, no?
```suggestion
```

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockTransformer.scala
```diff
@@ -40,16 +46,17 @@ class BlockTransformer(subst: SymbolSubst):
       applyResult(exc): exc2 =>
         if exc2 is exc then b else Throw(exc2)
     case Match(scrut, arms, dflt, rst) =>
+      val applySub = if rst.isEmpty then applySubBlock else applySubBlockNonTail
```
LPTK: Less likely to be inefficient at runtime:
```suggestion
      def applySub(b: Block) = if rst.isEmpty then applySubBlock(b) else applySubBlockNonTail(b)
```
(This part of the code is quite performance-sensitive.)

LPTK: Same remark below.

LPTK:
Now, please remember to make a pass on _all_ the open comments and either resolve them or respond to them if something is unclear or unresolved. (https://github.com/hkust-taco/mlscript/blob/hkmc2/CONTRIBUTING.md)

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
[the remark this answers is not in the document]
AnsonYeung: This is indeed a bug, it's fixed and a regression test is added.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
AnsonYeung: I extracted this to new method, this is not the same as fresh label ctx

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
AnsonYeung: I think the current name is ok. The elimination of the function used several times is done in the separate dce pass. If the function have any naked reference then the function cannot be removed even after inlining. This should have a clear meaning during inlining and I added a comment in the code here.

AnsonYeung:
@LPTK I just went through the comments.

LPTK:
The CI is not terminating. I can reproduce problems locally:

```
ScalaTest can't report this exception because another preceded it, so printing its stack trace:
java.lang.StackOverflowError
```

AnsonYeung:
It was a typo

PR #424, opened 2026-03-21
Implement rudimentary Block simplifier that performs basic DCE

LPTK:
Note: the change to `Predef.mjs` is not due to the new block simplifier but to an unrelated fix in `Term#size`, which was incorrectly implemented.


hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -352,16 +358,27 @@ object Scoped:
...
 object TryBlock:
-  def apply(sub: Block, finallyDo: Block, rest: Block): Block = rest match
-    case Scoped(syms, body) => Scoped(syms, TryBlock(sub, finallyDo, body))
-    case _ => new TryBlock(sub, finallyDo, rest)
+  def apply(body: Block, finallyDo: Block, rest: Block): Block =
+    body match
+    case _: Unreachable => body
+    case _ =>
+      rest match
+      case Scoped(syms, body) => Scoped(syms, TryBlock(body, finallyDo, body))
```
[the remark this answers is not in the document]
LPTK: Wow, well done, Copilot!!!

hkmc2/shared/src/test/mlscript/basics/Drop.mls
```diff
@@ -4,21 +4,22 @@
...
-//│ tmp3 = globalThis.Object.freeze({ b: b });
 //│ runtime.Unit
+//│ ———————————————| Lowered IR |———————————————————————————————————————————————————————————————————————
+//│ let a⁰, b⁰, tmp, tmp1;
+//│ set a⁰ = 0;
+//│ set tmp = { "a": a⁰ };
+//│ set b⁰ = 1;
+//│ set tmp1 = { "b": b⁰ };
+//│ runtime⁰.Unit⁰
 //│ —————————————————| Output |—————————————————————————————————————————————————————————————————————————
```
AnsonYeung, line 23: The code looks quite weird, is this parsing expected?

hkmc2/shared/src/main/scala/hkmc2/codegen/CachedAnalysis.scala
```diff
@@ -0,0 +1,27 @@
...
+import hkmc2.utils.*
+
+
+trait CachedAnalysis[A, R]:
+  
+  def analyzeUncached(a: A): R
+  
+  import java.util.IdentityHashMap
+  
+  val cache: IdentityHashMap[A, R] = new IdentityHashMap
```
AnsonYeung, line 16: Should we also use weak keys here? The map may prevent old Block from gc

hkmc2/shared/src/test/mlscript/handlers/NonLocalReturns.mls
```diff
@@ -115,8 +115,8 @@ f()
 //│ ╔══[INTERNAL ERROR] Unexpected nested class: lambdas may not function correctly.
 //│ ║  l.108: 	  object A with
 //│ ╙──       	         ^
-//│ ═══[RUNTIME ERROR] Error: Effect $_non$_local$_return$_effect$_18 is raised in a constructor
-//│ ═══[RUNTIME ERROR] Expected: '100', got: 'undefined'
```
AnsonYeung, line 119: Seems like object constructor are eliminated by mistake

AnsonYeung (commented):
Seems like there is a bug. Otherwise LGTM

hkmc2/shared/src/test/mlscript/basics/Drop.mls
LPTK: Yeah, that's actually how records elaborate at the moment. The fields are bound in follow-on code/fields.
```fs
let r = { a: 1, print(a), b: a + 1}
//│ > 1
//│ r = {a: 1, b: 2}

// Or, equivalently
let r =
  a: 1
  print(a)
  b: a + 1
//│ > 1
//│ r = {a: 1, b: 2}
```

hkmc2/shared/src/main/scala/hkmc2/codegen/CachedAnalysis.scala
LPTK: I'm not sure. Weak reference have some overhead, and `CachedAnalysis` instances aren't supposed to be super long-lived – they should only be created to make an optimization pass and then promptly dropped.

hkmc2/shared/src/test/mlscript/handlers/NonLocalReturns.mls
LPTK: Oh, looks like the whole object is removed, which shouldn't happen. Will fix tonight.

hkmc2/shared/src/main/scala/hkmc2/codegen/CachedAnalysis.scala
AnsonYeung: i see

PR #425, opened 2026-03-24
Pretty printing for Block IR

ChingLongTin:
implements turning Block IR back into a mlscript string


hkmc2/shared/src/main/scala/hkmc2/codegen/ReflectionInstrumenter.scala
LPTK, line 1: It looks like most of these changes are not about pretty printing. Are they?

ChingLongTin: the diff now reflect that most changes are about the pretty printing, after the merge.

hkmc2/shared/src/test/mlscript/block-staging/Functions.mls
```diff
@@ -56,35 +91,47 @@ staged module ClassInstrumentation with
...
   fun call() = 
     1 + 1
     Nonstaged.f()
     Staged.f()
-//│ > FunDefn(Symbol("ctor$"), ([]), End, true)
-//│ > FunDefn(Symbol("call"), ([]), Assign(Symbol("$_no$_symbol$_"), Call(Select(Ref(ModuleSymbol("Nonstaged", class Nonstaged)), Symbol("f")), []), Return(Call(Select(Ref(ModuleSymbol("Staged", class Staged)), Symbol("f_gen")), []), false)), true)
+//│ > fun ctor_() =
+//│ >   ()
+//│ > fun call() =
+//│ >   __no__symbol__ = Nonstaged.f()
```
LPTK: This can just be printed as
```suggestion
//│ >   Nonstaged.f()
```

hkmc2/shared/src/test/mlscript/block-staging/PrintCode.mls
```diff
@@ -2,7 +2,134 @@
...
+//│ >   fun f(x) = 1
+//│ > class B
+
+printCode(Scoped([Symbol("x"), Symbol("y")], Assign(Symbol("x"), ValueLit(4.2), Return(ValueRef(Symbol("x")), true))))
+//│ > let x
+//│ > let y
+//│ > x = 4.2
+//│ > x
+
+printCode(Define(ValDefn(Some(aSym), Symbol("x"), ValueLit(1)), End()))
```
LPTK, line 61: Note that `Define` nodes correspond to new definitions. `Define(ValDefn)` correspond to a `val` definition, not to an assignment. When you properly print a constructor within the scope of its enclosing object/class/module, it will all make sense.

ChingLongTin: I have changed it back to using `val` instead of `set`. Later I will skip printing the `ValDefn` in a constructor, if the parameter list of the enclosing class already defined the symbol, i.e.

```
staged class C(val x) with
  val y = 1

fun ctor(x) = Define(ValDefn(Some(sym), Symbol("x"), ValueRef(Symbol("x"))), ...)

class C(val x) with
  val y = 1 // skip printing implicit ValDefn to x
```

LPTK:
The CI fails (looks like test output is not up to date).

hkmc2/shared/src/test/mlscript-compile/Block.mls
```diff
@@ -28,7 +49,7 @@ class Arg(val value: Path)
 class Case with
   constructor
     Lit(val lit: Literal)
-    Cls(val cls: Symbol, val path: Path)
+    Cls(val cls: ClassSymbol, val path: Path)
```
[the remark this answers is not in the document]
LPTK: You could make a `ConcreteClassSymbol` class that extends `ClassSymbol` and make `VirtualClassSymbol` also extend `ClassSymbol`. (However, beware that pattern matching is [currently a bit broken](https://github.com/hkust-taco/mlscript/issues/151#issuecomment-3501159558) if you mix clauses related to subclasses.)

hkmc2/shared/src/main/scala/hkmc2/codegen/ReflectionInstrumenter.scala
```diff
@@ -277,6 +278,14 @@ class ReflectionInstrumenter(using State, Raise, Ctx) extends BlockTransformer(n
...
                 blockCtor("ClsLikeDefn", Ls(c, methods, none)): cls =>
                   blockCtor("Define", Ls(cls, p))(k(_, ctx))
+    case Define(v: ValDefn, rest) =>
+      // TODO: only allow ValDefn inside ctors
+      transformBlock(rest): p =>
+        transformOption(v.tsym.owner, transformSymbol(_, N, "test")): owner =>
+          transformSymbol(v.sym): sym =>
+            transformPath(v.rhs): rhs =>
+              blockCtor("ValDefn", Ls(owner, sym, rhs)): v =>
+                blockCtor("Define", Ls(v, p))(k(_, ctx))
```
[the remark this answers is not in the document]
ChingLongTin: I will change the ValDefn constructor, so this is unchanged.

hkmc2/shared/src/test/mlscript-compile/Block.mls
```diff
@@ -19,8 +19,9 @@ type ParamList = Array[Symbol]
 
 class Symbol(val name: Str)
 // this is so that we're able to retrieve information about the class from the symbol
-class VirtualClassSymbol(val name: Str) extends Symbol(name)
-class ClassSymbol(val name: Str, val value, val paramsOpt: Option[ParamList], val auxParams: Array[ParamList]) extends Symbol(name)
+class ClassSymbol(val name: Str) extends Symbol(name)
+class VirtualClassSymbol(val name: Str) extends ClassSymbol(name)
+class ConcreteClassSymbol(val name: Str, val value, val paramsOpt: Opt[ParamList], val auxParams: Array[ParamList]) extends ClassSymbol(name)
```
LPTK: Aren't these missing an `extends` clause?

LPTK:
Please fix the conflicts and merge.

hkmc2/shared/src/test/mlscript-compile/Block.mls
ChingLongTin: what do you mean? they both have an `extends` at the end, similar to how ClassSymbol extends Symbol in `class ClassSymbol(val name: Str) extends Symbol(name)`

LPTK: Never mind, I think I read them too fast...

LPTK:
The CI fails.

hkmc2/shared/src/test/mlscript-compile/Block.mls
```diff
@@ -46,22 +68,20 @@ class Path extends Result with
...
 
 class Block with
   constructor
-    Match(val scrut: Path, val arms: Array[[Case, Block]], val dflt: Opt[Block], val rest: Block)
+    Match(val scrut: Path, val arms: Array[Arm], val dflt: Opt[Block], val rest: Block)
     Return(val res: Result, val implct: Bool)
     Assign(val lhs: Symbol, val rhs: Result, val rest: Block)
     Define(val defn: Defn, val rest: Block)
-    // TODO: [fyp] handle Scoped, Label, Break nodes
+    // TODO: [fyp] handle Scoped nodes
```
NeilKleistGao: Remove the TODO if it is done.

hkmc2/shared/src/test/mlscript/block-staging/Functions.mls
```diff
@@ -38,14 +39,45 @@ staged module Expressions with
...
-//│ > FunDefn(Symbol("match1"), ([]), Scoped([Symbol("scrut")], Assign(Symbol("scrut"), Lit(9), Match(Ref(Symbol("scrut")), [Cls(VirtualClassSymbol("Bool"), Select(Ref(Symbol("runtime")), Symbol("unreachable"))) -> Return(Lit(1), false), Lit(8) -> Return(Lit(2), false), Cls(VirtualClassSymbol("Int"), Select(Ref(Symbol("runtime")), Symbol("unreachable"))) -> Return(Lit(3), false)], Return(Lit(0), false), End))), true)
-//│ > FunDefn(Symbol("match2"), ([]), Scoped([Symbol("element1$"), Symbol("element0$"), Symbol("a")], Match(Ref(Symbol("x1")), [Tup(0) -> Return(Lit(1), false), Tup(2) -> Assign(Symbol("element0$"), Call(Select(Select(Ref(Symbol("runtime")), Symbol("Tuple")), Symbol("get")), [Ref(Symbol("x1")), Lit(0)]), Assign(Symbol("element1$"), Call(Select(Select(Ref(Symbol("runtime")), Symbol("Tuple")), Symbol("get")), [Ref(Symbol("x1")), Lit(1)]), Match(Ref(Symbol("element0$")), [Lit(1) -> Match(Ref(Symbol("element1$")), [Lit(2) -> Return(Lit(2), false)], Assign(Symbol("a"), Ref(Symbol("element0$")), Return(Lit(3), false)), End)], Assign(Symbol("a"), Ref(Symbol("element0$")), Return(Lit(3), false)), End)))], Return(Lit(0), false), End)), true)
+//│ > fun ctor_() = ()
```
NeilKleistGao, line 42: Shall we have the module printing in this PR and handle the constructor generation correctly?

hkmc2/shared/src/test/mlscript/block-staging/Functions.mls
```diff
@@ -97,42 +138,106 @@ staged module OtherBlocks with
...
 // nested module
 module A with
   staged module B with
     fun f() = 1
-//│ > FunDefn(Symbol("ctor$"), ([]), End, true)
-//│ > FunDefn(Symbol("f"), ([]), Return(Lit(1), false), true)
+//│ > fun ctor_() = ()
+//│ > fun f() = 1
+
+:todo force enable :ftc to desugar Label and Break
```
NeilKleistGao: The `:ftc` flag does not desugar Label or Break

hkmc2/shared/src/test/mlscript/block-staging/Functions.mls
ChingLongTin: probably not, the printing uses the class `FunCache`, which is introduced in `add-cache`.

hkmc2/shared/src/test/mlscript-compile/Block.mls
```diff
@@ -70,96 +89,109 @@ fun concat(b1: Block, b2: Block) = if b1 is
...
-    String then "\"" + l.toString() + "\""
+    Str then "\"" + l.toString() + "\""
     else l.toString()
 
-fun showSymbol(s: Symbol) =
-  s.toString()
+fun showSymbol(s: Symbol) = if s.name is
+  "runtime" then "Runtime"
+  // FIXME: this is a hack for local symbols, and will fail if global symbols contain them, like the generated classes of :ftc.
+  else s.name.replaceAll("$", "_")
```
NeilKleistGao: It is fine to substitute invalid characters (we also did it in `Scope.scala`)

hkmc2/shared/src/test/mlscript-compile/Block.mls
```diff
@@ -70,96 +89,109 @@ fun concat(b1: Block, b2: Block) = if b1 is
...
     undefined then "undefined"
     null then "null"
-    String then "\"" + l.toString() + "\""
+    Str then "\"" + l.toString() + "\""
     else l.toString()
 
-fun showSymbol(s: Symbol) =
-  s.toString()
+fun showSymbol(s: Symbol) = if s.name is
+  "runtime" then "Runtime"
```
NeilKleistGao: In what case, we still need to access `runtime`, besides `runtime.Tuple.get`? Can we translate them all into normal MLscript syntax (e.g., translate `runtime.Tuple.get` to select syntax)?

NeilKleistGao (approved):
The rest LGTM.

hkmc2/shared/src/test/mlscript-compile/Block.mls
ChingLongTin: since spread operators are not supported, `runtime.Tuple.slice` will not appear, so `runtime.Tuple.get` is the only time we use `runtime`. I have turned it into a `Select` node during instrumentation.

hkmc2/shared/src/test/mlscript/block-staging/Functions.mls
```diff
@@ -61,7 +61,7 @@ staged module Expressions with
 //│ >     Int then 3
 //│ >     else 0
 //│ > fun match2() =
-//│ >   let {element1_, element0_, a}
+//│ >   let {element0_, a, element1_}
```
LPTK: This change seems to indicate nondeterminism. Is that the case? Where does this order come from, and why did it change?

ChingLongTin: This comes from printing the `Set` of symbols from a `Scoped` block. I suppose it is non-deterministic, since `set.toList`. Perhaps we can sort them by uid first to make it deterministic?

LPTK: Yes, you should never iterate on a set and use the resulting order as part of compiler output. Sorting by `uid` is the way to go, indeed.

PR #426, opened 2026-03-24
Add cache to staged modules

ChingLongTin:
Populates the function cache and generatorMap for each staged module


hkmc2/shared/src/test/mlscript-compile/Block.mls
```diff
@@ -64,7 +56,7 @@ class Case with
     Cls(val cls: Symbol, val path: Path)
     Tup(val len: Int)
 
-class Result extends CachedHash with
```
LPTK: Any reason for removing this extension clause?

ChingLongTin: it isn't used for anything within the PR. Originally I needed it to construct a Map of `Path`s to the corresponding (generator) function in the current stage, but this is done with `generatorMap` now, so we don't need to hash `Result`s.

hkmc2/shared/src/main/scala/hkmc2/codegen/ReflectionInstrumenter.scala
```diff
@@ -16,39 +16,55 @@ import syntax.{Literal, Tree}
...
-        k =>
-          headCont: head =>
-            tailCont: tail =>
-              k(head :: tail)
-      )(f)
+  // recover `defn` for when `sym.defn` is `None`, when the definition was generated by other compiler passes
+  val defnMap = HashMap[Symbol, ClsLikeDefn | ClsLikeBody]()
+  val symbolMapSym: Symbol = TempSymbol(N, "symbolMap")
+  // only create symbolMap if we create entries to avoid changing IR for unrelated tests
+  var symbolMapUsed: Bool = false
```
NeilKleistGao: It could be good if we use a mutable `Opt[Symbol]` here. When it is used, you can set it to `S(TempSymbol(...))`.

hkmc2/shared/src/test/mlscript/block-staging/Classes.mls
```diff
@@ -0,0 +1,175 @@
+:js
+:staging
+
+import "../../mlscript-compile/Block.mls"
+
+// TODO: copy C# annotation to Block IR
```
NeilKleistGao, line 6: Not sure if it is necessary.

hkmc2/shared/src/test/mlscript/block-staging/Classes.mls
```diff
@@ -0,0 +1,175 @@
...
+//│ > Map(1) {
+//│ >   [class I] { Symbol(mlscript.definitionMetadata): [ 'class', 'I' ] } => ConcreteClassSymbol {
+//│ >     name: 'I',
+//│ >     value: [class I] { Symbol(mlscript.definitionMetadata): [Array] },
+//│ >     paramsOpt: None {},
+//│ >     auxParams: []
+//│ >   }
+//│ > }
+
+// :sjs
```
NeilKleistGao: Please remove the commented flags.

hkmc2/shared/src/test/mlscript/block-staging/Classes.mls
```diff
@@ -0,0 +1,175 @@
...
+class C
+staged module M with
+  fun f() = C
+//│ > fun ctor_() = ()
```
NeilKleistGao: Shall we have the module printing in this PR and handle the constructor generation correctly (discussed in #425)?

NeilKleistGao (changes requested):

hkmc2/shared/src/test/mlscript/block-staging/Classes.mls
LPTK: What is a C# annotation? 🤔

ChingLongTin: this is referring to the extra selection information, where `x.C#call()` indicates that the function is from a class `C`. Currently we just ignore this information, so the instrumented code becomes `x.call()` instead.

LPTK:
Please resolve the conflicts.

ChingLongTin:
hmm, `ClassInFun.mls` fails from printing `let y` in a different order...

LPTK:
Unfortunately, I still don't know why the order of private fields does not seem stable :/

EDIT: It's being investigated and fixed at https://github.com/hkust-taco/mlscript/pull/459.

hkmc2/shared/src/test/mlscript/block-staging/Classes.mls
ChingLongTin: Although this information might be useful for shape propagation to filter out impossible shapes, it is not implemented, and so I'll leave this TODO here for now.

hkmc2/shared/src/test/mlscript-compile/Block.mls
```diff
@@ -84,121 +91,128 @@ class Block with
...
+      Match(scrut, arms, dflt, rest) then
+        "if " + showPath(scrut) + " is"
+        + indent("\n" + arms.map(showArm(_)).join("\n"))
+        + if dflt is Some(db) then indent("\nelse" + indent((if db is Return then " " else "\n") + showBlock(db))) else ""
+        + showRestBlock(rest)
+      Scoped(symbols, rest) then
+        // but this one must use the function argument instead of a lambda...
+        "let {" + symbols.map(showSymbol).join(", ") + "}" + showRestBlock(rest)
+      End() then "()"
+      _ then "<unknown block>" + b
```
NeilKleistGao:
```suggestion
      _ then "<unknown block>"
```

NeilKleistGao (commented):
Otherwise, LGTM.

LPTK:
> Otherwise, LGTM.

Why not make an _approving_ review?

LPTK:
Please fix the conflicts, then we can merge.

ChingLongTin:
@NeilKleistGao How should we handle classes with private fields here? In `block-staging/Nested.mls` we have lifted staged classes (for lamdas) that have private fields, when previously we restricted staging to classes with all public fields only.

ChingLongTin:
The new test file `Nested.mls` uses `:ftc` to turn functions that capture variables into staged classes. however, this case cannot be instrumented as they will contain private variables, which are disallowed for now. For this PR, I put a `:fixme` on the given tests.

I will address those tests in a later PR. Unless there is that approach that works for classes with private fields, I will look into turning the private fields generated classes of `:ftc` to public fields.

LPTK:
Note that some changes to private fields are upcoming and may make this easier to deal with:
* https://github.com/hkust-taco/mlscript/pull/492
* https://github.com/hkust-taco/mlscript/pull/494

LPTK:
Pls revert most of the changes in dcc22f7.

PR #427, opened 2026-03-25
Add flags to disable `Object.freeze` and module checks

NeilKleistGao:


hkmc2/shared/src/test/mlscript/codegen/NoFreeze.mls
```diff
@@ -0,0 +1,96 @@
+:js
+
+:sjs
```
LPTK: Please just use `ssjs` here and remove the `Bar` cases lower in the file, they are mostly redundant.

LPTK (approved):
Otherwise LGTM.

PR #428, opened 2026-03-28
Remove empty constructor noise from IR/codegen

LPTK:
@ychenfo Could you please check why this change leads to unexpected deforestation regressions?

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
@@ -969,7 +969,7 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
         sym
     val ctor =
       inScopedBlock:
-        term_nonTail(Blk(clsBody.nonMethods, clsBody.blk.res))(ImplctRet)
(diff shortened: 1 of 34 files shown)


ychenfo:
I checked and found that the regression comes from the lone-module check on my side: it was matching a very specific module ctor shape, while this PR ends up producing the equivalent but shorter `End("")` form instead.

This diff below could be a rather flimsy fix to the regression. But I realize that this lone-module check is brittle and probably not very meaningful in the first place, and the changes I'm working on to factor out common analysis are aligned with removing this check entirely.

```diff
@@ -366,8 +366,7 @@ class DeforestPreAnalyzer(
             case End("") => true
           && ctor
             .matches:
-              case Return(Select(Value.Ref(runtimeSym, None), Tree.Ident("Unit")), true) =>
-                runtimeSym is elabState.runtimeSymbol
+              case End("") => true
         then mod.foreach(applyCompanionModule)
       else
         ctxTracker.markAsNonHandleable()
```

LPTK:
@ychenfo Ok, thanks! It works.

As you saym this logic should be replaced. We of course want to deforest class and module constructor bodies as well.

PR #429, opened 2026-03-30
Implement instance methods

b-temirov:
This PR adds class-method support for non-inheriting classes by lowering instance methods to standalone direct-call wasm functions with an explicit `this` parameter, and a prescan that predeclares constructors and methods so self recursion, mutual recursion, later-method calls, and constructor calls to same or later classes work by construction.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
@@ -194,6 +194,23 @@ object Ctx:
       globalTy: RefType,
   )
   
+  enum MethodShape:
+    case Getter
+    case Callable
+
+  case class MethodInfo(
+      ownerIsym: InnerSymbol,
+      dSym: TermSymbol,
+      funcId: SymIdx,
+      shape: MethodShape,
+  )
+
+  case class ConstructorInfo(
+      classBms: BlockMemberSymbol,
+      funcId: SymIdx,
+      exportName: Opt[Str],
+  )
+  
   case class LabelTarget(
       breakLabel: Str,
       continueLabel: Opt[Str],
@@ -232,6 +249,8 @@ object Ctx:
     namedFuncs = MutMap.empty,
     tags = ArrayBuf.empty,
     namedGlobals = MutMap.empty,
+    methodInfoBySymbol = MutMap.empty,
+    constructorInfoByClass = MutMap.empty,
     locals = MutMap() :: Nil,
     startFunc = N,
   )
@@ -279,6 +298,8 @@ class Ctx(
     namedFuncs: MutMap[Symbol, Int],
     tags: ArrayBuf[TagInfo],
     namedGlobals: MutMap[Symbol, Int],
+    methodInfoBySymbol: MutMap[TermSymbol, Ctx.MethodInfo],
+    constructorInfoByClass: MutMap[BlockMemberSymbol, Ctx.ConstructorInfo],
     var locals: Ls[MutMap[Local, Int]],
     private var startFunc: Opt[FuncIdx],
 ) extends ToWat:
@@ -503,6 +524,26 @@ class Ctx(
 
   /** Checks whether the global variable scope contains the variable `sym`. */
   def containsGlobal(sym: Symbol): Bool = namedGlobals.contains(sym)
+
+  /** Returns method metadata for resolved method symbol `sym`, if present. */
+  def getMethodInfo(sym: TermSymbol): Opt[Ctx.MethodInfo] =
+    methodInfoBySymbol.get(sym)
+
+  /** Registers method metadata under its resolved method symbol. */
+  def registerMethodInfo(info: Ctx.MethodInfo): Unit =
+    methodInfoBySymbol(info.dSym) = info
+
+  /** Checks whether constructor metadata has been registered for class symbol `sym`. */
+  def containsConstructorInfo(sym: BlockMemberSymbol): Bool =
+    constructorInfoByClass.contains(sym)
+
+  /** Returns constructor metadata for class symbol `sym`, if present. */
+  def getConstructorInfo(sym: BlockMemberSymbol): Opt[Ctx.ConstructorInfo] =
+    constructorInfoByClass.get(sym)
+
+  /** Registers constructor metadata under its class symbol. */
+  def registerConstructorInfo(info: Ctx.ConstructorInfo): Unit =
+    constructorInfoByClass(info.classBms) = info
 
   /** Checks whether singleton metadata has been registered for class symbol `sym`. */
   def containsSingleton(sym: BlockMemberSymbol): Bool = singletonByBms.contains(sym)
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
@@ -98,7 +98,7 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
       ctor = End(""),
       companion = N,
       bufferable = N,
-    )
+    )(N)
 
   /** Registers the synthetic `Unit` singleton. */
   private def RegisterUnitSingleton()(using Ctx, Raise, Scope): Unit =
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
@@ -803,9 +803,13 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
           singletonInfoFor(l) match
             case S(info) => singletonGlobalGet(info)
             case N =>
-              ctx.getFunc(l) match
-                case S(funcIdx) => ref.func(funcIdx, RefType(ctx.getFuncInfo_!(l).typeIdx, nullable = false))
-                case N => getVar(l, r.toLoc)
+              if disamb.exists(_.isInstanceOf[ClassSymbol]) then
+                errExpr:
+                  Ls(msg"Plain class references are not supported in Wasm; instantiate the class instead." -> r.toLoc)
+              else
+                ctx.getFunc(l) match
+                  case S(funcIdx) => ref.func(funcIdx, RefType(ctx.getFuncInfo_!(l).typeIdx, nullable = false))
+                  case N => getVar(l, r.toLoc)
 
     case Call(Value.Ref(l: BuiltinSymbol, _), lhs :: rhs :: Nil) if !l.functionLike =>
       if l.binary then
@@ -980,8 +984,7 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
 
     case Instantiate(_, cls, as) =>
       cls match
-        // TODO: Implement proper lowering for Errors with string and unit payloads.
-        // Currently exceptions are encoded as i31 payloads; unsupported payloads are lossy.
+        // TODO: Implement proper lowering for Errors with unit payloads.
         case Select(Value.Ref(sym, _), id)
             if (sym eq State.globalThisSymbol) && id.name == "Error" =>
           return as.headOption match
@@ -989,6 +992,7 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
                 case Value.Lit(BoolLit(value)) => ref.i31(i32.const(if value then 1 else 0))
                 case Value.Lit(IntLit(value)) =>
                   withValidIntLit(value, arg.value.toLoc)(intVal => ref.i31(i32.const(intVal)))
+                case Value.Lit(StrLit(_)) => result(arg.value)
                 case unsupported =>
                   warnExpr(
                     msg"WatBuilder::result for Instantiate(...) of `globalThis.Error(...)` with payload `${
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
@@ -36,7 +36,7 @@ object WatBuilder:
 
 class WatBuilder(using TraceLogger, State) extends CodeBuilder:
   import Ctx.ctx
-  import Ctx.{SingletonInfo, binaryOps, unaryOps, wasmIntrinsicArities, wasmIntrinsicNameSet}
+  import Ctx.{SingletonInfo, binaryOps, unaryOps, wasmIntrinsicArities, wasmIntrinsicNameSet, ConstructorInfo, MethodInfo, MethodShape}
   import Instructions.{block as blockInstr, *}
   import WatBuilder.ExternIntrinsics
 
@@ -66,7 +66,7 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
       && ((defn.k is syntax.Cls) || (defn.k is syntax.Obj))
       && defn.auxParams.isEmpty
       && defn.parentPath.isEmpty
-      && defn.methods.isEmpty
+      && (defn.methods.isEmpty || (defn.k is syntax.Cls))
       && defn.companion.isEmpty
       && (defn.preCtor match
         case End(_) => true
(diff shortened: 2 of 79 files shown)


hkmc2/shared/src/test/mlscript/wasm/ScopedLocals.mls
```diff
@@ -204,11 +204,12 @@ class Foo(val a, val b)
 //│       (nop)
 //│       (block (result (ref null any))
 //│         (local.set $tmp
-//│           (call $Foo_ctor
-//│             (ref.i31
-//│               (i32.const 42))
-//│             (ref.i31
-//│               (i32.const 1))))
+//│           (ref.cast (ref $Foo)
```
LPTK: Why is this new cast needed?

hkmc2/shared/src/test/mlscript/wasm/Methods.mls
```diff
@@ -0,0 +1,343 @@
...
+//│ ╔══[COMPILATION ERROR] WatBuilder::result for method value extraction is not implemented yet
+//│ ║  l.322: 	Box().f
+//│ ╙──       	     ^^
+
+
+:ge
+class GetterOnly() with
+  fun f = 1
+GetterOnly().f()
+//│ ╔══[COMPILATION ERROR] Getter-style methods are not callable; use member access instead of `()`
```
LPTK: What's the point of raising an error like this? What if the returned value was a lambda? (We'll eventually support them.)

hkmc2/shared/src/test/mlscript/wasm/Methods.mls
LPTK, line 1: Interesting. Can you please add tests that call more than one method, and on something else than an object instantiation directly? Eg, chained metod calls on variables. You'll need to use the `x.ClassName#methodName` syntax for now, so it resolves statically.

hkmc2/shared/src/test/mlscript/wasm/ScopedLocals.mls
b-temirov: Previously we did
```
val objType = ctx.getFuncInfo_!(ctorFuncIdx).body.resultType_!
call(funcidx = ctorFuncIdx, as.map(argument), Seq(Result(objType.asValType_!)))
```
in `Instantiate(...)` and used the body result type for the `call`. But now we are using predeclared `ConstructorInfo`, which doesn't include body result type, so the call using `RefType.anyref` is less precise, hence the cast
```
private def directConstructorCall(
      ctorInfo: ConstructorInfo,
      args: Seq[Expr],
  )(using Ctx): Expr =
    ref.cast(
      call(
        funcidx = FuncIdx(ctorInfo.funcId),
        operands = args,
      ),
      RefType(ctx.getType_!(ctorInfo.classBms), nullable = false),
    )
```

hkmc2/shared/src/test/mlscript/wasm/Methods.mls
b-temirov: Changed to `Calling the result of getter-style method access is not implemented yet; use member access without ()`

LPTK: But why do you want to add a special case for this specific scenario? What happens if you just remove this special-cased error handling code? Just let the compiler naturally fail.

hkmc2/shared/src/test/mlscript/wasm/Methods.mls
```diff
@@ -316,21 +316,66 @@ Box(41).mk()
...
+  fun inc() = 
+    set x += 1
+    this
+  fun double() =
+    set x *= 2
+    this
+  fun get() = x
+let c = Counter(1)
+c.Counter#inc().Counter#double().Counter#get()
+//│ FAILURE: Unexpected exception
```
LPTK: To make this test not fail, use `:fixme`.

hkmc2/shared/src/test/mlscript/wasm/Methods.mls
```diff
@@ -347,42 +345,22 @@ class Counter(val x) with
...
-
-
-:ge
-class NullaryOnly() with
-  fun f() = 1
-NullaryOnly().f
-//│ ╔══[COMPILATION ERROR] WatBuilder::result for method value extraction is not implemented yet
-//│ ║  l.385: 	NullaryOnly().f
-//│ ╙──       	             ^^
+//│ /!!!\ Uncaught error: java.lang.Exception: Internal Error: Missing field `term:Box/f` in struct `member:Box` with type `(type $Box (sub $Object (struct (field $$tag (mut i32)))))`
```
LPTK: This shouldn't crash the compiler. It should just say that `term:Box/f` is neither a callable method nor a field.

hkmc2/shared/src/test/mlscript/wasm/Methods.mls
```diff
@@ -347,42 +345,22 @@ class Counter(val x) with
...
-//│ ╙──       	     ^^
-
-
-:ge
-class GetterOnly() with
-  fun f = 1
-GetterOnly().f()
-//│ ╔══[COMPILATION ERROR] Calling the result of getter-style method access is not implemented yet; use member access without `()`
-//│ ║  l.376: 	GetterOnly().f()
-//│ ╙──       	^^^^^^^^^^^^^^^^
```
LPTK: Why remove this test?

hkmc2/shared/src/test/mlscript/wasm/ScopedLocals.mls
LPTK: > But now we are using predeclared `ConstructorInfo`, which doesn't include body result type

Why not?

hkmc2/shared/src/test/mlscript/wasm/Methods.mls
b-temirov: I misinterpreted your earlier comment and thought we need to let compiler crash, and it wasn't so useful to include.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -842,7 +850,15 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
                 methodInfo.shape match
                   case MethodShape.Callable =>
                     S(directMethodCall(methodInfo, result(Value.This(methodInfo.ownerIsym)), args.map(argument)))
-                  case MethodShape.Getter => N
+                  case MethodShape.Getter =>
+                    S(
+                      errExpr(
+                        Ls(
+                          msg"`${methodInfo.dSym.toString}` is neither a field access nor a callable method" -> c.toLoc,
```
LPTK: Instead of duplicating this error message four times, please use a helper function. Don't repeat yourself.

hkmc2/shared/src/test/mlscript/wasm/ScopedLocals.mls
b-temirov: We register `ConstructorInfo` in the prescan before body lowering so methods can reference constructors defined later in code. At that stage there is no emitted constructor body to get the result. But anyways removing cast seems to be fine.

LPTK: 🤔

LPTK:
Is this ready for review/merge?

b-temirov:
> Is this ready for review/merge?

Not quite. Inheritance coming soon on top. Or should I raise inheritance separately?

LPTK:
Yes, better keep the PRs small and merge them regularly.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -192,6 +192,103 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+    case AssignDynField(_, _, _, _, rst) =>
+      registerDefnCallableMetadata(rst)
+    case HandleBlock(_, _, _, _, _, _, _, rst) =>
+      registerDefnCallableMetadata(rst)
+    case Label(_, _, body, rst) =>
+      registerDefnCallableMetadata(body)
+      registerDefnCallableMetadata(rst)
+    case Scoped(_, body) =>
+      registerDefnCallableMetadata(body)
+    case _: BlockTail => ()
```
LPTK, line 223: Wouldn't this work?
```suggestion
  private def registerDefnCallableMetadata(b: Block)(using Ctx, Raise, Scope): Unit = b match
    case Define(defn: ClsLikeDefn, rst) =>
      registerClassCallableMetadata(defn)
      registerDefnCallableMetadata(rst)
    case _ => b.subBlocks.foreach(subBlocks)
```

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -528,6 +675,120 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+    sel.symbol.flatMap(_.asTrm).flatMap(ctx.getMethodInfo)
+
+  /** Returns whether `path` is an explicit owner reference such as a class or module path. */
+  private def isDirectOwnerRef(path: Path): Bool = path match
+    case Value.Ref(l, disamb) =>
+      val resolved = l match
+        case _: BlockMemberSymbol => disamb.getOrElse(l)
+        case sym => sym
+      resolved match
+        case _: ClassSymbol | _: ModuleOrObjectSymbol => true
```
LPTK, line 720: Wouldn't this trigger in
```fs
:lot
object Test
Test
//│ —————————————| Lowered IR Tree |————————————————————————————————————————————————————————————————————
//│ Program:
//│   main = Scoped(syms = {member:Test⁰}): 
//│     body = Define:
//│       defn = ClsLikeDefn:
//│         isym = object:Test¹
//│         sym = member:Test⁰
//│         k = Obj
//│       rest = Return: \
//│       res = Ref{disamb=object:Test¹}:
//│         l = member:Test⁰
//│         disamb = S of object:Test¹
//│       implct = true
```
?

Again, self-references never have a disamb and always contain the inner symbol directly.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -528,6 +675,120 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+      case _: BlockMemberSymbol => disamb.flatMap(_.asTrm)
+      case sym: Symbol => sym.asTrm
+    resolvedSym.flatMap(ctx.getMethodInfo)
+
+  /** Resolves method metadata for a selection when it denotes a registered class method. */
+  private def methodInfoForSelection(sel: Select)(using Ctx): Opt[MethodInfo] =
+    sel.symbol.flatMap(_.asTrm).flatMap(ctx.getMethodInfo)
+
+  /** Returns whether `path` is an explicit owner reference such as a class or module path. */
+  private def isDirectOwnerRef(path: Path): Bool = path match
```
LPTK: Is this supposed to be named
```suggestion
  private def isSelfReference(path: Path): Bool = path match
```
?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -594,80 +861,112 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
-                    }`\n... which has type `${
-                      ty.fold("(none)")(_.toWat.toString)
-                    }`",
-                ),
-              )
-          val baseTypeInfo = ctx.getTypeInfo_!(baseTypeIdx)
+          val ctorCall = constructorInfoForPath(fun).map: ctorInfo =>
+            directConstructorCall(ctorInfo, args.map(argument))
+
+          val methodCall = fun match
```
LPTK, line 867: This will always be computed, even when `ctorCall` is defined! You should move this expensive code to the `orElse` argument or make this `val` a `def`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -919,33 +1216,67 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
-                msg"WatBuilder::returningTerm for AssignField(...) without a resolved symbol is not implemented (field `${
-                    nme.name
-                  }`). Use `_.[_]` for index-based accesses." ->
-                  nme.toLoc,
-              ),
-              extraInfo = S(assign),
-            )
+            lhs match
+              case ownerRef: Value.Ref if isDirectOwnerRef(ownerRef) =>
+                val resolvedOwner = ownerRef.l match
```
LPTK, line 1261: For the same reason stated above, I don't think this makes a lot of sense.

hkmc2/shared/src/test/mlscript/wasm/Methods.mls
```diff
@@ -0,0 +1,383 @@
...
+//│           (local.get $tmp)
+//│           (ref.i31
+//│             (i32.const 40))))))
+//│   (export "entry" (func $entry))
+//│   (elem declare func $entry))
+//│ Wasm result:
+//│ = 41
+
+
+:wat
```
LPTK, line 128: We already have the `:wat` above, which I think is sufficient.
```suggestion
```

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -919,33 +1216,67 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+                      selOwner.getClass.getName
+                    })",
+                )
+              directFieldAssign(selCls, selSym, lhsExpr, rhsExpr)
+          case S(selSym: BlockMemberSymbol) =>
+            selSym.asTrm match
+              case S(fieldSym) =>
+                if ctx.getMethodInfo(fieldSym).nonEmpty then
+                  errExpr(
+                    Ls(msg"WatBuilder::returningTerm for AssignField(...) to class methods is not implemented yet" -> nme.toLoc),
```
LPTK, line 1239: I don't think we ever want to allow assigning to an object method. This error should be a standard compilation error, not a "not implemented yet".

LPTK (changes requested):
The logic seems quite complicated, for what this achieves. But let's try to merge this soon anyway, as soon as you have fixed what I point out.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -194,6 +194,23 @@ object Ctx:
       globalTy: RefType,
   )
   
+  enum MethodShape:
+    case Getter
+    case Callable
```
Derppening, line 199: I didn't know there is a meaningful difference between bracketless vs bracketed invocations. AFAICT it seems like bracketless invocations are statically resolved and bracketed invocations are dynamically resolved.

@LPTK is this intended behavior?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -194,6 +194,23 @@ object Ctx:
...
+      dSym: TermSymbol,
+      funcId: SymIdx,
+      shape: MethodShape,
+  )
+
+  case class ConstructorInfo(
+      classBms: BlockMemberSymbol,
+      funcId: SymIdx,
+      exportName: Opt[Str],
+  )
```
Derppening, line 212: Nit: Please add missing documentation.

Derppening (commented):
Just a few preliminary comments for now. I will need to look into this in more detail.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
LPTK: The resolution is the same. It's just the difference between taking an argument list, as in `x.f()`, and not taking one, as in `x.f` (the latter is declared as `fun f = ...`).

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
LPTK: Please also update `createDefnTypes` similarly.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
LPTK: TODO: we don't really need this enum; just store the number of parameter lists that each method accepts.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
b-temirov: As discussed in meeting, moved to using `BlockTraverserShallow` in `collectTopLevelClassDefns` and removed `createDefnTypes ` entirely.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
b-temirov: No longer relevant.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
b-temirov: No longer relevant.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
b-temirov: No longer relevant.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
b-temirov: No longer relevant.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
b-temirov: No longer relevant.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
b-temirov: No longer relevant.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
b-temirov: No longer relevant.

b-temirov:
This PR has been superseded by #456.

PR #430, opened 2026-03-30
Proper Error for Plain Class Refs

FetBoba:
Added check for plain class references and proper error message for them.


hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -556,9 +556,25 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
       singletonInfoFor(l) match
         case S(info) => singletonGlobalGet(info)
         case N =>
-          ctx.getFunc(l) match
-            case S(funcIdx) => ref.func(funcIdx, RefType(ctx.getFuncInfo_!(l).typeIdx, nullable = false))
-            case N => getVar(l, r.toLoc)
+          val isThisLikeRef = l.isInstanceOf[InnerSymbol] && disamb.isEmpty
```
```suggestion
          val isSelfReference = l.isInstanceOf[InnerSymbol] && disamb.isEmpty
```

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -556,9 +556,25 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+          val isPlainClassRef =
+            !isThisLikeRef && (l match
+              case _: ClassSymbol => true
```
LPTK: All "naked" references to inner symbols are self-reference, so I think this should never trigger. Or do you have an example that does?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -556,9 +556,25 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+              case _ =>
+                disamb.exists:
+                  case _: ClassSymbol => true
+                  case ds => ds.defn.exists:
+                      case _: ClassDef => true
```
LPTK: Can this really happen? When the thing has a class definition but not a class symbo?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -556,9 +556,25 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+                      case _ => false
+            )
+          if isPlainClassRef then
+            errExpr(
+              Ls(msg"Plain class references are not supported in Wasm; instantiate the class instead." -> r.toLoc),
+            )
```
```suggestion
            errExpr:
              Ls(msg"Plain class references are not supported in Wasm; instantiate the class instead." -> r.toLoc),
```

FetBoba: Done

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: ok I see. Ill remove it

LPTK (approved):

PR #431, opened 2026-03-30
Default match case and string exceptions

FetBoba:
Added support for exceptions with string payloads and proper error for missing default match case.


hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1436,7 +1436,7 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
           val defaultExpr =
             val rawDefaultExpr = dflt match
               case S(defaultBody) => returningTerm(defaultBody)
-              case N => unreachable
+              case N => `throw`(exnTagIdx, Seq(result(Value.Lit(StrLit("match error")))))
```
LPTK: The error-throwing statement, indcluding its string payload, is already in the IR:
```fs
:re
:sir
if false then 0
//│ let scrut;
//│ set scrut = false;
//│ match scrut
//│   true =>
//│     0
//│   else
//│     throw new globalThis⁰.Error⁰("match error")
//│ end
//│ ═══[RUNTIME ERROR] Error: match error
```
What happens to it, here? Why do you recreate a throw statement with a hardcoded string payload here?

FetBoba: Sorry, I didn't notice it. Since there was a task to implement default match case in kanban I thought that I need to add throw there. I'll remove it.

LPTK (approved):
Well... that was an easy fix.

PR #432, opened 2026-03-30
Wasm imports for REPL cross-block support

FetBoba:
This PR adds REPL cross-block imports.

Approach:
Keep a session export table on the Scala side that records top-level Wasm entities exported by previous blocks and how they should be imported. When compiling a new block, the backend finds free references, resolves the ones from the existing session, and emits the corresponding Wasm imports before generating the rest of the module.

Notes:
- top-level functions, globals, classes, and singleton-related runtime state can now be imported across REPL blocks
- session state is updated only after successful execution, so failed blocks do not corrupt later import/export state
- only top-level REPL bindings are persisted; internal scoped locals are not
- previous bindings are emitted as Wasm imports before normal definitions


hkmc2/shared/src/test/mlscript/wasm/ReplImports.mls
```diff
@@ -0,0 +1,66 @@
...
+inc(Foo.x)
+//│ Wasm result:
+//│ = 2
+
+if C(7) is C(v) then v else 0
+//│ Wasm result:
+//│ = 7
+
+:wat
+set counter = counter + step
```
LPTK:
```suggestion
set counter += step
```

hkmc2/shared/src/test/mlscript/wasm/ReplImports.mls
```diff
@@ -0,0 +1,66 @@
...
+counter
+//│ Wat:
+//│ (module
+//│   (type $Object (sub (struct (field $$tag (mut i32)))))
+//│   (type $plus_impl (func (param $lhs (ref null any)) (param $rhs (ref null any)) (result (ref null any))))
+//│   (type $entry1 (func (result (ref null any))))
+//│   (import "repl" "counter" (global $counter (mut (ref null any))))
+//│   (import "repl" "step" (global $step (mut (ref null any))))
+//│   (func $plus_impl1 (type $plus_impl) (param $lhs (ref null any)) (param $rhs (ref null any)) (result (ref null any))
```
LPTK: Is this `plus_impl` function generated in every single wasm block that uses addition?

Since imports are supported now, why not just import it?

hkmc2DiffTests/src/test/scala/hkmc2/WasmDiffMaker.scala
```diff
@@ -44,11 +49,16 @@ abstract class WasmDiffMaker extends LlirDiffMaker:
   lazy val prettifyBinaryenWat = (content: Str) =>
     content.substring(2, content.length() - 2).replace("\\\\n", "\n").replace("\\\\\"", "\"")
 
+  /** Resets Wasm REPL session state for a fresh diff run. */
```
LPTK: This makes no sense, as the DiffMaker is re-created anew every time diff-tests are run. Pretty sure this is more LLM slop?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -29,19 +89,27 @@ import scala.collection.mutable.{ArrayBuffer as ArrayBuf, Map as MutMap}
...
     val id: SymIdx,
     val typeIdx: TypeIdx,
     params: Seq[Local -> Str],
-    nResults: Int,
     locals: Seq[Local -> Str],
-    val body: Expr,
-    val `export`: Opt[Str],
+    val bodyOpt: Opt[Expr],
+    val resultTypes: Seq[Result],
+    val importModule: Opt[Str] = N,
```
LPTK: Never use default arguments in core logic.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1045,13 +1203,19 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
                       val funcInfo =
                         FuncInfo(
                           sym,
-                          typeIdx = funcTy,
```
LPTK: Why did you remove these named arguments? Or did the LLM remove them automatically? They improve clarity.

**Please review your entire diff yourself and remove all the changes that were made that you cannot justify.**

LPTK (changes requested):

hkmc2/shared/src/test/mlscript/wasm/ReplImports.mls
FetBoba: Done

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1494,7 +1683,8 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
       p: Program,
       exprt: Opt[BlockMemberSymbol],
       wd: io.Path,
-  )(using Raise, Scope): (Document, Str, Int) =
+      sessionImports: Seq[WasmSessionBinding] = Nil
+  )(using Raise, Scope): (Document, Str, Int, Seq[WasmSessionBinding]) =
```
LPTK: If you're going to return a tuple, at least be more informative by adding names

```suggestion
  )(using Raise, Scope): (wat:Document, entryName: Str, systemMemMinPages: Int, sessionExports: Seq[WasmSessionBinding]) =
```

But TBH I liked the `CompiledWasmModule` class better. What was wrong with it?

FetBoba: Nothing was wrong with it but since this change was not adding anything I thought you would prefer removing it. And also I thought that it's little bit more convenient to use tuple in WasmDiffMaker. If you want I can add it back

LPTK: In what way is it more convenient to use tuple in WasmDiffMaker?

If not for this, putting it back would be good.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: Ok I removed them

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -33,19 +86,27 @@ import scala.reflect.ClassTag
...
     val typeUse: TypeUse,
     params: Seq[Local -> Str],
-    nResults: Int,
     locals: Seq[Local -> Str],
-    val body: Expr,
-    val `export`: Opt[Str],
+    val bodyOpt: Opt[Expr],
+    val resultTypes: Seq[Result],
+    val importModule: Opt[Str] = N,
+    val importName: Opt[Str] = N,
```
Derppening: Imported functions are its own type - See `Import[ExternType.Func]`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -90,51 +153,110 @@ class FuncInfo(
...
+ * @param exportName
+ *   Optional export name.
+ */
+class GlobalInfo(
+    val id: SymIdx,
+    val valType: ValType,
+    val mutable: Bool,
+    val init: Opt[Expr],
+    val importModule: Opt[Str] = N,
+    val importName: Opt[Str] = N,
```
Derppening: Imported globals should follow the design for other imported functions/memories. See `Ctx` for more detail.

hkmc2DiffTests/src/test/scala/hkmc2/WasmDiffMaker.scala
```diff
@@ -57,15 +62,23 @@ abstract class WasmDiffMaker extends LlirDiffMaker:
...
+          .distinctBy(_.bindingKey)
+      val compiled = ltl.givenIn:
         baseScp.nest.givenIn:
-          WatBuilder().program(le, N, wd)
+          WatBuilder().program(le, N, wd, sessionImports)
+      val modWat = compiled.wat
+      val mainFnNme = compiled.entryName
+      val systemMemMinPages = compiled.systemMemMinPages
+      val sessionExports = compiled.sessionExports
       val modWatJsLit = JSBuilder.makeStringLiteral(modWat.mkString(output.ColWidth))
```
LPTK: Instead of these, you could just use
```scala
val CompiledWasmModule(modWat, .../* etc. */) = ltl.givenIn:
```

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -29,23 +89,23 @@ import scala.reflect.ClassTag
...
   */
 class FuncInfo(
     val id: SymIdx,
     val typeUse: TypeUse,
     params: Seq[Local -> Str],
-    nResults: Int,
     locals: Seq[Local -> Str],
-    val body: Expr,
-    val `export`: Opt[Str],
+    val bodyOpt: Opt[Expr],
```
Derppening: This also needs to be reverted to `Expr`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -115,26 +177,41 @@ class FuncInfo(
...
+ * @param init
+ *   The initializer expression for the global.
+ * @param exportName
+ *   Optional export name.
+ */
+class GlobalInfo(
+    val id: SymIdx,
+    val valType: ValType,
+    val mutable: Bool,
+    val init: Opt[Expr],
```
Derppening: Same here.

hkmc2DiffTests/src/test/scala/hkmc2/WasmDiffMaker.scala
FetBoba: Ok done

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: ok changed

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: ok changed

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: ok I changed code to use it instead

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: Ok I added imported globals according to design

hkmc2/shared/src/test/mlscript/wasm/ReplImports.mls
LPTK: @FetBoba this is fixed, right?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -195,6 +207,123 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+      case HandleBlock(_, _, _, _, _, _, _, rst) => S(rst)
+      case _ => N
+
+    def recur(block: Block, scopedSyms: Set[Local]): Set[Symbol] = block match
+      case Scoped(syms, body) =>
+        recur(body, scopedSyms ++ syms)
+      case Begin(sub, rst) =>
+        recur(sub, scopedSyms) ++ recur(rst, scopedSyms)
+      case Define(ValDefn(_, sym, _), rst) =>
+        recur(rst, scopedSyms) + sym
```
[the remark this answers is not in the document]
LPTK: @FetBoba Actually, the proper approach is to only export those symbols that are in the `symbolsToPreserve` set, in `JSBackendDiffMaker.scala`.

hkmc2DiffTests/src/test/scala/hkmc2/WasmDiffMaker.scala
```diff
@@ -27,6 +28,11 @@ abstract class WasmDiffMaker extends LlirDiffMaker:
 
   private val baseScp: utils.Scope =
     utils.Scope.empty(utils.Scope.Cfg.default)
+  private val wasmReplImportsNme = s"${wasmSuppNme}ReplImports"
+  private val wasmReplImportsRef = s"globalThis.$wasmReplImportsNme"
+  private val sessionImportsBySymbol = mutable.Map.empty[Local, Vector[WasmSessionBinding]]
```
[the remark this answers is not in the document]
LPTK: Copilot's remark is fair and should be addressed, but for the love of God, please don't follow its suggestion to define an `AbstractMap` subclass. Just directly use existing data structures.

hkmc2/shared/src/test/mlscript/wasm/ReplImports.mls
FetBoba: Yes, it's fixed

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -112,11 +112,13 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
       createDefnTypes(Define(unitDefn, End("")))
 
     ctx.pushLocal()
+    given ArrayBuf[WasmSessionBinding] = ArrayBuf.empty
     returningTerm(Define(unitDefn, End("")))
     ctx.popLocal()
```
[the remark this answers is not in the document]
FetBoba: Ok I changed it. But there is another issue with Units. Unit$inst is redefined in each block where () is used.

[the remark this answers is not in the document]
LPTK: Is this specific to Unit? Why isn't Unit treated like a normal singleton definition?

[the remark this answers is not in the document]
FetBoba: Yes, because Unit has own lowering path (we generate it in RegisterUnitSingleton)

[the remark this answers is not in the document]
FetBoba: When we collect free vars we don't collect unit that's why import of Unit$inst is not generated.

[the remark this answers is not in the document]
LPTK: Does it really need to? It would be nice to share as much logic as possible (and to minimize special-cased logic). In any case, I guess you need to address the issue you mention.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -18,6 +18,66 @@ import scala.collection.immutable.ListMap
 import scala.collection.mutable.{ArrayBuffer as ArrayBuf, Map as MutMap}
 import scala.reflect.ClassTag
 
+/** Metadata for a REPL binding that can be imported by later Wasm modules. */
+sealed trait WasmSessionBinding:
```
Derppening: Nit: I don't think these classes require the `Wasm` prefix, since they are already contained within the `wasm` module. We can always import them as `WasmSessionBinding` (etc.) if we ever need to disambiguate from other similarly named classes.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -18,6 +18,66 @@ import scala.collection.immutable.ListMap
...
+  /** Returns the deduplication key for this binding. */
+  def bindingKey: Str
+  /** Returns the symbols that should resolve to this binding. */
+  def bindingSyms: Seq[Local]
+  /** Returns the export name if this binding is re-exported. */
+  def exportNameOpt: Opt[Str] = N
+
+object WasmSessionBinding:
+  val replModuleName: Str = "repl"
```
Derppening: Nit: `ReplModuleName` - Scala's convention uses upper camel case for `val`s in `object`s (equivalent to `static final`s in Java).

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -18,6 +18,66 @@ import scala.collection.immutable.ListMap
...
+
+final case class WasmSessionFunc(
```
Derppening: Nit: I think `case class`es are by default final? @LPTK

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -29,23 +89,23 @@ import scala.reflect.ClassTag
...
   */
 class FuncInfo(
     val id: SymIdx,
     val typeUse: TypeUse,
     params: Seq[Local -> Str],
-    nResults: Int,
     locals: Seq[Local -> Str],
     val body: Expr,
-    val `export`: Opt[Str],
+    val resultTypes: Seq[Result],
```
Derppening: Nit: `resultTypes` should be at the same relative order as `nResults`, no?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -115,26 +175,41 @@ class FuncInfo(
...
+class GlobalInfo(
+    val id: SymIdx,
+    val valType: ValType,
+    val mutable: Bool,
+    val init: Expr,
+    val exportName: Opt[Str]
+) extends ToWat:
+
+  /** Returns the symbolic identifier document used in global declarations. */
+  private def idDoc: Document = id.toWat
```
Derppening: Why is this necessary? `${id.toWat}` is simple and concise enough.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -115,26 +175,41 @@ class FuncInfo(
...
 
   def toWat: Document =
     val typeDoc =
       if mutable then doc"(mut ${valType.toWat})"
       else valType.toWat
-    doc"(global ${id.toWat} $typeDoc ${init.toWat})"
+    doc"(global${idDoc.surroundUnlessEmpty(doc" ")} ${typeDoc} ${init.toWat})${
+      exportName.fold(doc""): name =>
+        doc""" # (export "${name}" (global ${idDoc}))"""
```
Derppening: We now inline `export` clauses into the `global` definition itself, i.e. `(global $name (export "...") ...)` - See `FuncInfo.toWat`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
```diff
@@ -233,6 +233,14 @@ object ExternType:
   case class Func(override val id: SymIdx, typeUse: TypeUse) extends ExternType(id):
     def toWat: Document = doc"""(func ${id.toWat} ${typeUse.toWat})"""
 
+  /** A global entry that is externally addressable. */
+  case class Global(override val id: SymIdx, valType: ValType, mutable: Bool) extends ExternType(id):
```
Derppening: I prefer classes in `Wasm.scala` to more closely resemble grammar elements in the WAT format.

Please split the `valType`/`mutable` fields into its own `GlobalType` class (similar to `MemType`) - See [here](https://webassembly.github.io/spec/core/text/types.html#global-types).

hkmc2DiffTests/src/test/scala/hkmc2/WasmDiffMaker.scala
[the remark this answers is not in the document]
Derppening: Also, why are you using `Vector` here and not an `ArrayBuf`? New elements will be inserted [here](https://github.com/hkust-taco/mlscript/pull/432/changes#diff-ac87f7f2a3b1745d5d60fd3ad5d8c51a5b0d848b3e002e7c146aa942839b385fR203) and using `:+` will create a new `Vector`.

Derppening:
Also, please run `scalafmt` (I can see missing trailing commas), and how many times do I have to remind you to please write documentation...

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
LPTK: They are not! It's fine to mark this one as `final`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -660,15 +798,14 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
             lastWords(
-              s"Expected resolved class for Select(...) expression to be a BlockMemberSymbol, but got $selOwner (${
-                  selOwner.getClass.getName
-                })",
+              s"Expected resolved class for Select(...) expression to be a BlockMemberSymbol, but got ${ownerInfo.fold("(none)")(
+                  _.toString
+                )}",
             )
           val fieldidx = fieldSelect(selCls, selSym)
           struct.get(
```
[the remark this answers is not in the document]
FetBoba: Ok changed

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: Ok changed

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
FetBoba: Ok changed

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: Ok changed

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: Ok changed

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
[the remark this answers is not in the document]
FetBoba: Ok done

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: Ok changed

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: Yes. Done

hkmc2DiffTests/src/test/scala/hkmc2/WasmDiffMaker.scala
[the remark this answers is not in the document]
FetBoba: I changed it to LinkedHashMap

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -91,8 +91,8 @@ final case class CompiledWasmModule(
   *   [[Seq]] of parameter local variables and their names.
   * @param locals
   *   [[Seq]] of local variables (excluding parameters) and their names.
-  * @param bodyOpt
-  *   The expression of the function body.
+ * @param body
+ *   The expression of the function body.
```
Derppening: 👎

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -115,26 +174,34 @@ class FuncInfo(
...
+ *   Symbolic identifier for the global.
+ * @param valType
+ *   The value type of the global.
+ * @param mutable
+ *   Whether the global is mutable.
+ * @param init
+ *   The initializer expression for the global.
+ * @param exportName
+ *   Optional export name.
+ */
```
Derppening: 👎

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -491,21 +586,81 @@ class Ctx extends ToWat:
...
+          globals.zipWithIndex.collectFirst:
+            case ((_, gi), i) if gi === globalEntry => GlobalIdx(NumIdx(i))
+      case sym: Symbol =>
+        namedGlobals.get(sym).map: globalEntry =>
+          GlobalIdx(globalExternType(globalEntry).id)
+
+  /** Same as [[getGlobal]] but throws an exception when the `globalref` is not found. */
+  def getGlobal_!(globalref: GlobalIdx | Symbol, resolveSymIdx: Bool = false): GlobalIdx =
+    getGlobal(globalref, resolveSymIdx).getOrElse:
+      lastWords(s"Missing global definition for ${globalref.prettyString}")
```
Derppening: Please look carefully at `getFunc` or `getType` - We are moving *away* from numeric indices, which is why functions with `resolveSymIdx` is *deprecated*!

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -116,7 +116,8 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
     ctx.popLocal()
 
   /** Registers eager singleton runtime state by creating its global and start-init action. */
-  private def registerSingletonInit(clsLikeDefn: ClsLikeDefn, typeref: TypeIdx)(using Ctx, Raise, Scope): Unit =
+  private def registerSingletonInit(clsLikeDefn: ClsLikeDefn, typeref: TypeIdx)(using Ctx, Raise, Scope,
+      ArrayBuf[SessionBinding]): Unit =
```
Derppening: I prefer having all implicit parameters on the same line:

```suggestion
  private def registerSingletonInit(
      clsLikeDefn: ClsLikeDefn,
      typeref: TypeIdx
  )(using Ctx, Raise, Scope, ArrayBuf[SessionBinding]): Unit =
```

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1537,6 +1734,10 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
         ),
       )
 
+    val sessionExports = ArrayBuf.empty[SessionBinding]
+    given Set[Local] = preservedSessionSymbols
+    given ArrayBuf[SessionBinding] = sessionExports
```
Derppening: Since these two variables are related as they both track the symbols and WAT identifiers that are exported by a module, it's better to add a new class and pass that class around implicitly. Looking at the diff it is very difficult to figure out what `Set[Local]` and `ArrayBuf[SessionBinding]` is supposed to be.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -195,6 +198,120 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+      exportName = exportName,
+      globalType = GlobalType(RefType.anyref, mutable = true)
+    )
+
+  /** Registers imported REPL bindings into the current module before codegen starts. */
+  private def registerSessionImports(sessionImports: Seq[SessionBinding])(using
+      Ctx,
+      Raise,
+      Scope
+  ): Unit =
```
Derppening: Same here - See above.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -919,14 +1060,12 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
           case S("global") =>
-            errExpr(
-              Ls(msg"WatBuilder::returningTerm for Assign(...) to global variable not implemented yet" -> l.toLoc),
-              extraInfo = S(s"Block IR: ${t.showAsTree}"),
-            )
-          case S("local") => local.set(idx, rExpr)
+            global.set(lExpr.instrargs(0).asInstanceOf[GlobalIdx], rExpr)
+          case S("local") =>
+            val idx = lExpr.instrargs(0).asInstanceOf[LocalIdx]
+            local.set(idx, rExpr)
```
Derppening: Nit: Make it consistent - Either inline the `idx` into the `local.set` call, or extract `idx` for the `global.set` call.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1577,11 +1786,10 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
       id = SymIdx(entryNme),
       typeUse = TypeUse(entryFnTy),
       params = Seq.empty,
-      nResults = 1,
-      // TODO(Derppening): Should we place top-level scope variables in the global section?
       locals = (entryFnLocals ++ entryExtraLocals).map(l => l -> scope.allocateOrGetName(l)),
       body = entryFnExpr,
-      `export` = S(entryNme),
+      resultTypes = Seq(Result(RefType.anyref)),
```
Derppening: Nit: Reorder this to below `params`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1616,10 +1824,10 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
           id = SymIdx(scope.allocateName(TempSymbol(N, "start"))),
           typeUse = TypeUse(initTy),
           params = Seq.empty,
-          nResults = 0,
           locals = Seq.empty,
           body = initBody,
-          `export` = N,
+          resultTypes = Seq.empty,
```
Derppening, line 2133: Same here.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -18,6 +18,65 @@ import scala.collection.immutable.ListMap
...
+  def bindingKey: Str = s"singleton:$moduleName:$exportName"
+  def bindingSyms: Seq[Local] = blockSym +: objectSym.toSeq
+  override def exportNameOpt: Opt[Str] = S(exportName)
+
+final case class CompiledWasmModule(
+    wat: Document,
+    entryName: Str,
+    systemMemMinPages: Int,
+    sessionExports: Seq[SessionBinding]
+)
```
Derppening, line 137: You still haven't added the docs for these classes, and haven't run `scalafmt` (there are still missing trailing commas)!

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -18,6 +18,65 @@ import scala.collection.immutable.ListMap
 import scala.collection.mutable.{ArrayBuffer as ArrayBuf, Map as MutMap}
 import scala.reflect.ClassTag
 
+/** Metadata for a REPL binding that can be imported by later Wasm modules. */
+sealed trait SessionBinding:
```
Derppening, line 22: If `sym`, `moduleName`, `exportName` are all common among all derived classes of `SessionBinding`, why aren't they abstract fields?

In fact, why does `SessionBinding` need to be a `sealed trait` and not an `abstract sealed class`? All derived classes of `SessionBinding` are just `case class`es, and I don't see the need for them to inherit other abstract classes.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Ok done

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Ok changed

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Ok done

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Ok done

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Ok done

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
FetBoba: Ok done

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: `SessionClass` doesn't have `moduleName` and `exportName`. `SessionSingleton` has different structure of `sym`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: Ok I changed them. Should I delete old version of these functions or mark them deprecated?

LPTK: Does anyone use them anywhere? If not, why would you deprecate them isntead of removing them?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
Derppening: Ugh that's annoying. Let's leave it as is then.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
Derppening: If these functions (`getGlobal`, `getGlobal_!`) are newly introduced in this PR, then by all means remove them. I might be mistaken that they were present before.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
FetBoba: Done

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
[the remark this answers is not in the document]
LPTK: By the way, I was responding to:

> Yes, because Unit has own lowering path (we generate it in RegisterUnitSingleton)

Can you please respond and address this, or comment how you addressed it?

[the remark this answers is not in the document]
FetBoba: I addressed it. `Unit` still has a special lowering path in Wasm, but I fixed REPL reuse by making its synthetic block-member symbol stable across blocks and recording its session info once available.

hkmc2DiffTests/src/test/scala/hkmc2/WasmDiffMaker.scala
```diff
@@ -86,8 +86,13 @@ abstract class WasmDiffMaker extends LlirDiffMaker:
           errored = true
           outerRaise(d)
         case d => outerRaise(d)
+      val referencedSessionSymbols = mutable.LinkedHashSet.empty[Local]
+      new BlockTraverser:
+        override def applySymbol(sym: Symbol): Unit =
+          referencedSessionSymbols += sym
+      .applyBlock(pgrm.main)
```
LPTK: Isn't this going to make a hash set of _all_ the symbols in the _entire_ program? That seems very inefficient and weird. I much preferred the `freeVars` approach. What prompted this change?

FetBoba: Yes it wasn't full fix. Now I fixed it.

FetBoba: `freeVars` approach didn't collect `Unit`.

hkmc2DiffTests/src/test/scala/hkmc2/WasmDiffMaker.scala
```diff
@@ -45,29 +52,59 @@ abstract class WasmDiffMaker extends LlirDiffMaker:
...
         case d @ ErrorReport(source = Source.Compilation) =>
           errored = true
-          reportedMessages += d.mainMsg
           outerRaise(d)
         case d => outerRaise(d)
-      val (modWat, mainFnNme, systemMemMinPages) = ltl.givenIn:
+      val sessionImportSymbols = mutable.LinkedHashSet.from(pgrm.main.freeVars)
+      new BlockTraverser:
+        override def applyPath(p: Path): Unit = p match
+          case sel @ Select(_, _) =>
```
LPTK:
```suggestion
          case sel: Select =>
```

hkmc2DiffTests/src/test/scala/hkmc2/WasmDiffMaker.scala
LPTK: Ok, good.

LPTK:
Please fix the conflicts, so we can merge.

LPTK:
(Remember diff-test conflicts can often be resolved by just running the diff-tests.)

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -18,34 +18,179 @@ import scala.collection.immutable.ListMap
...
+  override def exportNameOpt: Opt[Str] = S(exportName)
+
+/** Metadata for a class type made visible to later Wasm REPL modules.
+  *
+  * @param sym
+  *   The block member symbol of the class.
+  * @param typeInfo
+  *   The Wasm type information that must be recreated in importing modules.
+  * @param runtimeTags
+  *   The class's runtime tag together with descendant class tags.
```
LPTK:
```suggestion
  *   The class' runtime tag together with descendant class tags.
```

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -366,10 +529,12 @@ class Ctx extends ToWat:
     getTypeInfo(typeref).getOrElse:
       lastWords(s"Missing type definition for ${typeref.prettyString}")
 
+  /** Records the class's runtime tag together with descendant class tags for `sym`. */
```
LPTK:
```suggestion
  /** Records the class' runtime tag together with descendant class tags for `sym`. */
```

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -366,10 +529,12 @@ class Ctx extends ToWat:
...
   def registerRuntimeClassTags(sym: BlockMemberSymbol, tags: LinkedHashSet[Int]): Unit =
     runtimeClassTags(sym) = tags
 
-  def getRuntimeClassTags(sym: BlockMemberSymbol): Opt[LinkedHashSet[Int]] =
+  /** Returns the class's runtime tag together with descendant class tags for `sym`. */
```
LPTK:
```suggestion
  /** Returns the class' runtime tag together with descendant class tags for `sym`. */
```

LPTK (approved):
Otherwise LGTM.

Derppening (approved):
No further comments - I'll perform some cleanup as I see fit after the merge.

PR #445, opened 2026-04-05
add Class to Prelude.mls

ChingLongTin:
provides a type for the runtime class value used in ClassSymbol and ModuleSymbol


LPTK (changes requested):
LGTM, thx.

LPTK (approved):
(Oops, wrong option.)

PR #451, opened 2026-04-07
Wasm Class Inheritance

b-temirov:
This PR adds basic class inheritance support to wasm backend by introducing a top-level prescan, inherited struct-type predeclaration, ctor/init splitting for class initialization, and subtype class matching.

I thought supporting forward constructor calls to classes defined later is useful to have before implementing methods. For that to work, I added early predeclaration and topological ordering. But topological ordering would remain unexercised without real parent-child relationships, so I ended up implementing inheritance first.

This PR supersedes #436.


hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -146,54 +146,202 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
-      createDefnTypes(rst)
-    case Scoped(_, body) =>
-      createDefnTypes(body)
-    case _: BlockTail => ()
+      defn.optionIf(isSupportedTopLevelClass).toList ::: collectTopLevelClassDefns(rst)
+    case Begin(sub, rst) =>
+      collectTopLevelClassDefns(sub) ::: collectTopLevelClassDefns(rst)
+    case b: NonBlockTail =>
+      collectTopLevelClassDefns(b.rest)
+    case _: BlockTail => Nil
```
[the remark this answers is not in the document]
LPTK: @b-temirov why not use `subBlocks`, as IIRC I've already shown you on onother PR.

LPTK:
Please address Copilot's review of the old PR: https://github.com/hkust-taco/mlscript/pull/436#pullrequestreview-4044306440

If a comment is no longer relevant, please say so in a comment reply and resolve it, in the original PR.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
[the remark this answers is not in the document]
LPTK: @b-temirov From the meeting, I suggested using a `BlockTraverserShallow` for this.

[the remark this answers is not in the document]
b-temirov: > `collectTopLevelClassDefns` only follows the `rest` chain (plus `Begin.sub`), but it no longer traverses nested blocks like `Match.arms/dflt`...

I noticed that no test exercises nested definitions, and nested classes are rejected by wasm backend. Therefore I removed it.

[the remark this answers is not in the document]
LPTK: You removed what?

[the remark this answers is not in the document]
b-temirov: I removed the traversal into match arms, default case and label body:
```
case Match(_, arms, dflt, rst) =>
      arms.foreach((_, body) => createDefnTypes(body))
      dflt.foreach(createDefnTypes)
      createDefnTypes(rst)
...
case Label(_, _, body, rst) =>
      createDefnTypes(body)
...
```

b-temirov:
I addressed Copilot's comments on #436. One comment was relevant and the issue it mentioned was already fixed in this PR.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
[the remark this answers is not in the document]
LPTK: Ok, but then please document this and add a regression test making sure that classes defined in, eg, match arms, are indeed rejected with a proper error.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -289,6 +289,7 @@ class Ctx extends ToWat:
   private val singletonByBms = MutMap.empty[BlockMemberSymbol, Ctx.SingletonInfo]
   private val singletonByIsym = MutMap.empty[ModuleOrObjectSymbol, Ctx.SingletonInfo]
   private val singletonInitActions = ArrayBuf.empty[Expr]
+  private val classTags = MutMap.empty[BlockMemberSymbol, Seq[Int]]
```
Derppening: I can see this conflicting with [this from the REPL changes](https://github.com/hkust-taco/mlscript/pull/432/changes#diff-8eed170ee9e51aeff4f8a4007b4ba2ecfcc1be769582c62cbc5f7192c3e80282R368). Are the two designs compatible (apart from the value type being `Int` vs `Seq[Int]`)?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -365,6 +366,12 @@ class Ctx extends ToWat:
     getTypeInfo(typeref).getOrElse:
       lastWords(s"Missing type definition for ${typeref.prettyString}")
 
+  def setClassTags(sym: BlockMemberSymbol, tags: Seq[Int]): Unit =
+    classTags(sym) = tags
+
+  def getClassTags(sym: BlockMemberSymbol): Opt[Seq[Int]] =
+    classTags.get(sym)
+
```
Derppening: Similar functions are defined in #432: `registerRuntimeClassTag` and `getRuntimeClassTag`. I suggest renaming these functions to better match the changes in the other PR.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -146,54 +146,208 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+      val sym = zeroIndegree.dequeue()
+      ordered += defnsBySym(sym)
+      childrenBySym(sym).foreach: childSym =>
+        indegrees(childSym) -= 1
+        if indegrees(childSym) == 0 then
+          zeroIndegree.enqueue(childSym)
+
+    if ordered.size != defns.size then
+      raise(ErrorReport(
+        msg"Wasm inheritance ordering detected an inheritance cycle." ->
```
Derppening: Nit: I think just stating that inheritance cycles are not supported is good enough.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -146,54 +146,208 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+        source = Diagnostic.Source.Compilation,
+      ))
+
+    ordered.toList
+
+  /** Declares one supported top-level class type for early wasm registration. */
+  private def predeclareClassType(defn: ClsLikeDefn)(using Ctx, Raise, Scope): Unit =
+    val parentTypeIdx =
+      if defn.parentPath.isEmpty then baseObjectTypeIdx
+      else ctx.getType_!(resolveParentSym(defn).getOrElse(lastWords("unreachable")))
```
Derppening: `lastWords("unreachable")` is not very informative if this is really exercised. Consider giving a more informative error message.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -890,17 +1126,17 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
   private def getI32FromAnyref(name: Str): Expr =
     i31.get(ref.cast(getLocalAnyref(name), RefType.i31ref), true)
 
-  def returningTerm(t: Block)(using Ctx, Raise, Scope): Expr =
-    def isControlTransfer(expr: Expr): Bool =
-      expr.resultType.contains(UnreachableType) || expr.mnemonic == "return"
+  private def isControlTransfer(expr: Expr): Bool =
+    expr.resultType.contains(UnreachableType) || expr.mnemonic == "return"
```
Derppening: This might be better implemented as an extension function of `Expr`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
Derppening: Also:

- I think this stores the mapping of a symbol to the class tag of itself and its children - Why its children and not its parents?
- Why `Seq[Int]` instead of `Set[Int]` for faster lookup and deduplication?

Derppening (commented):
Please also run `scalafmt` and add missing documentation for private functions.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
b-temirov: 1. After discussing with @FetBoba, we decided to choose `LinkedHashSet`, which preserves insertion order while giving set semantics. @FetBoba needs to access first element to get a class' own tag, while the rest of the inserted elements are descendent tags. 

2. For each class the mapping stores the set of tags that should satisfy a pattern for that class: the class’s own tag and the tags of all its subclasses. When we lower `case Case.Cls(cls, _)`, we read the scrutinee object’s tag and test whether that tag belongs to the accepted tag set for `cls`. This makes a parent class match instances of that class and instances of descendant classes.

b-temirov:
Added documentation to private functions and applied small fixes.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -289,6 +289,7 @@ class Ctx extends ToWat:
   private val singletonByBms = MutMap.empty[BlockMemberSymbol, Ctx.SingletonInfo]
   private val singletonByIsym = MutMap.empty[ModuleOrObjectSymbol, Ctx.SingletonInfo]
   private val singletonInitActions = ArrayBuf.empty[Expr]
+  private val runtimeClassTags = MutMap.empty[BlockMemberSymbol, LinkedHashSet[Int]]
```
Derppening, line 292: > For each class the mapping stores the set of tags that should satisfy a pattern for that class: the class’s own tag and the tags of all its subclasses. When we lower case Case.Cls(cls, _), we read the scrutinee object’s tag and test whether that tag belongs to the accepted tag set for cls. This makes a parent class match instances of that class and instances of descendant classes.

This doesn't make a lot of sense.

1. When compiling a single MLwasm file, we only know the set of descendants that exist *within the current module*. What if classes in other files also extend the same base class?
2. Wouldn't this cause a state explosion if we are pattern matching on a commonly-derived class, e.g. `Object`?

A better way would probably be creating a `typeinfo` struct for all classes, which stores the object tag and its direct parent (or all its parents). We can then perform a runtime traversal to check if the scrutinee has the pattern-matched class as one of its ancestors.

b-temirov: I agree that current implementation is not ideal and needs to be changed. IIRC, @LPTK mentioned in our last meeting that we should use vtable to pattern match classes; and while current implementation can be merged for the time being, a later separate PR needs to target pattern matching. 

Since inheritance is needed to implement vtables, I believe it's more convenient to merge inheritance first, merge methods, and raise a PR targeting virtual dispatch and pattern matching.

LPTK: I did not mean that you should use virtual methods to implement pattenr matching, which you seem think. (How would that work?)

When I mentioned the vtable, it was just as the standard name of the data structure that's attached to each object instance – also called run time type infortmation, or RTTI. Each object should have a pointer to an RTTI struct that contains the list of parents, which pattern matching can query.

Storing a list of the descendents of each class indeed doesn't make much sense. But I'm not opposed to merging this PR now and addressing this in a follow-up.

Derppening: Please create an item in the backlog explaining what is missing and add a `TODO` comment linking to the issue.

b-temirov: > When I mentioned the vtable, it was just as the standard name of the data structure that's attached to each object instance – also called run time type infortmation, or RTTI.

I mean the same idea, meaning class metadata/descriptor struct, which also contains virtual method slots.

I created an item in the backlog and added a TODO comment.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1394,6 +1631,9 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
 
                 val expectedTag = typeinfo.objectTag.getOrElse:
                   lastWords(s"Expected class $clsBlkMemberSym to have an object tag")
+                
+                // TODO: replace with RTTI ancestry checks once each object carries runtime type information.
```
Derppening: Is this the comment referring to the issue? If so, please add the link (URL) to this comment for easy reference.

Derppening (commented):
Other than the nit, LGTM. Thanks a lot!

Derppening (approved):

PR #452, opened 2026-04-08
Factor out flow analysis; dead parameter elimination

ychenfo:
<del>
TODO:
- [x] merge upstream, resolve conflict
  - [x] can reuse symbol refresher
- [x] fix bugs related to fun defns without param list
  - [ ] make sure no more bugs...
- [ ] ~~update block simplifier for assignment to dead params~~
- [x] I used LLM for some tedious but intricate changes, I need to further review them
</del>

------

- Factored deforestation’s analysis out into a shared analysis with mono and poly mode
- Added a dead parameter elimination pass
- Added public/private modifiers to functions
- Extended symbol refresher to freshen class/module definitions


hkmc2/shared/src/test/mlscript/dead-param-elim/basic.mls
```diff
@@ -0,0 +1,282 @@
...
+//│ define f_3$f⁰ as fun f_3$f¹() {
+//│   return 0
+//│ };
+//│ define f_1$f¹⁴ as fun f_1$f¹⁵() {
+//│   return 0
+//│ };
+//│ define f¹⁸ as fun f¹⁹(a) { return 0 };
+//│ do f_1$f¹⁵();
+//│ do f_3$f¹();
+//│ f_5$f¹()
```
LPTK: Wait, are we really copying functions at each call site?

I think that when we discussed it, we said polymorphic splitting should not be used for this optimization. Or at least, we should try to reuse existing identical copies instead of generating new ones.

What are your thoughts on this?

hkmc2/shared/src/test/mlscript/HkScratch.mls
```diff
@@ -8,5 +8,3 @@
 // :d
 // :todo
 
-
-
```
LPTK, line 12: Don't forget to revert these changes.

LPTK (commented):
The PR looks great!

It's really nice that we can remove dead parameters even in higher-order code such as
```fs
fun apply(f, x) = f(x)
apply((a) => 0, 42)
```

And also, that the analysis is now general and reusable.

However, I have some concerns with code duplication.

hkmc2/shared/src/test/mlscript/dead-param-elim/basic.mls
ychenfo: Update: currently there are no more such duplications due to mono mode by default and private modifiers added to functions

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
```diff
@@ -398,7 +398,7 @@ class TailRecOpt(using State, TL, Raise):
     val loopDefn = FunDefn(
       owner, bms, dSym,
       PlainParamList(params) :: Nil,
-      loop)(false, N)
+      loop)(false, N, Visibility.Public)
```
LPTK:
```suggestion
      loop)(false, N, Visibility.Public) // Q: maybe should be Private?
```

hkmc2/shared/src/main/scala/hkmc2/Config.scala
```diff
@@ -315,6 +316,11 @@ object ConfigParser:
       parseInt(value) match
         case S(v) => _.copy(patMatConsequentSharingThreshold = S(v))
         case N => identity
+    case "inline" =>
+      parseInt(value) match
+        case S(x) if x < 0 => _.copy(inlining = N)
+        case S(v) => _.copy(inlining = S(Inliner(v)))
+        case N => identity
```
LPTK: Pls parse `inlining` (not `inline`) correctly – it's an Option.

hkmc2/shared/src/test/mlscript/dead-param-elim/dead-ref.mls
```diff
@@ -1,61 +1,42 @@
 :js
 
-
+:noInline
```
LPTK, line 2: Why not just put this at the top of the file? We are not interested in inlining anything here; just disable it.

LPTK (approved):
Otherwise LGTM, thanks!

PR #457, opened 2026-04-13
Share UCS splits using new constructs

chengluyu:
During UCS normalization, when the same _alternative_ split appears on both the positively (`+`) and negatively (`-`) specialized splits of a match, the old implementation duplicated the code. This PR introduces a way to let-bind splits to share the common splits instead.

### New `Split` constructs

Two new `Split` variants were introduced:
- **`LetSplit(sym, tail)`**: declares a named split. The symbol's `body` holds the shared split.
- **`UseSplit(sym)`**: references a previously declared named split.

A new symbol class **`SplitSymbol`** holds the shared body and an optional `LabelSymbol` for lowering.

#### Method `normalize`

`normalizeImpl` returns `(Split, Set[SplitSymbol])`: the normalized split plus any `SplitSymbol`s whose `UseSplit` references survive but whose `LetSplit` hasn't been created yet. This lets `LetSplit` be placed at the lowest common ancestor of all `UseSplit` references.

A `LetSplit` candidate is created only when:
- The alternative is **not trivial** (`End` or `UseSplit`), and
- The alternative **doesn't reference the same scrutinee** (which would make positive/negative specialization produce different results on each side).

After normalization, the existing `patMatConsequentSharingThreshold` policy decides whether the candidate is kept as `LetSplit` or inlined. The decision scales with the surviving `UseSplit` count, so small alternatives can still be shared when they are reused enough.

#### Method `specialize`

`specialize` returns `Opt[Split]` (`N` = unchanged, `S(...)` = changed), so `UseSplit` references are preserved when specialization doesn't change the split.

### Changes to lowering

`lowerSplit` translates `LetSplit` to a `Label` block and `UseSplit` to `Break`. When the continuation is `Ret`/`Thrw`, the label is used directly to preserve tail-call position. Otherwise, an exit label with a temp variable is used.

### What's removed?

- `createLabelsForDuplicatedBranches` and the `Labels` class (old label-based deduplication) are gone!

### What remains configurable?

- `patMatConsequentSharingThreshold` and the `:patMatConsequentSharingThreshold` test command are still supported. They now control whether a normalized join-point candidate survives or is inlined.


hkmc2/shared/src/test/mlscript/codegen/ClassMatching.mls
```diff
@@ -48,44 +48,34 @@ if s is
...
+//│ end
 //│ ——————————————| Optimized IR |——————————————————————————————————————————————————————————————————————
-//│ let arg$Some$0$, tmp;
+//│ let arg$Some$0$;
 //│ match s⁰
 //│   Some⁰ =>
 //│     set arg$Some$0$ = s⁰.value⁰;
 //│     match arg$Some$0$
 //│       1 =>
-//│         begin
```
LPTK, line 79: FIXME why was there a `begin` left?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -36,7 +35,9 @@ class Normalization(lowering: Lowering)(using tl: TL)(using Raise, Ctx, State) e
       else (these match
         case Split.Cons(head, tail) => Split.Cons(head, tail ++ those)
         case Split.Let(name, term, tail) => Split.Let(name, term, tail ++ those)
-        case Split.Else(_) /* impossible */ | Split.End => those)
+        case Split.Else(_) /* impossible */ | Split.End => those
```
LPTK, line 38:
```suggestion
        case Split.Else(_) => softAssert(false); those
        case Split.End => those
```

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -611,11 +630,12 @@ object Normalization:
 
   /** Immutable context tracking pending join point symbols whose LetSplit
     * placement is deferred to the lowest common ancestor of their UseSplit references. */
-  case class JoinPointCtx(pending: Set[SplitSymbol]):
-    def +(sym: SplitSymbol): JoinPointCtx = JoinPointCtx(pending + sym)
+  case class JoinPointCtx(pending: Set[SplitSymbol], sharingThreshold: Opt[Int]):
```
LPTK, line 633: Is there a reason to put that field in here, when the Config is already available everywhere?

hkmc2/shared/src/test/mlscript/block-staging/Functions.mls
```diff
@@ -16,7 +16,6 @@ staged module Expressions with
 //│ >   x = 42
 //│ >   x
 
-:fixme // Label/Break blocks from LetSplit lowering not yet supported in ReflectionInstrumenter
```
LPTK, line 19: Please reintroduce this fixme and the conditions to make it break.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -77,40 +78,123 @@ class Normalization(lowering: Lowering)(using tl: TL)(using Raise, Ctx, State) e
         case S(_) | N => lhs
       case _ => lhs
 
-  inline def apply(split: Split): Split = normalize(split)(using VarSet())
+  inline def apply(split: Split): Split = normalize(split)(using VarSet(), JoinPointCtx.empty)._1
```
LPTK, line 81: Note: bug. Should be fixed by moving out the needless field.

hkmc2/shared/src/test/mlscript/opt/AbortivePrefix.mls
```diff
@@ -71,30 +67,31 @@ fun f =
 //│ ———————————————| Lowered IR |———————————————————————————————————————————————————————————————————————
 //│ let f²;
 //│ define f² as fun f³() {
-//│   let scrut, scrut1;
-//│   block split_root$:
-//│     block split_1$:
-//│       set scrut = true;
+//│   let scrut, scrut1, tmp, tmp1;
+//│   set scrut = true;
+//│   block σ$x:
```
LPTK: This label seems unused. Is this on purpose?

hkmc2/shared/src/test/mlscript/ucs/normalization/DeduplicationWhile.mls
```diff
@@ -17,17 +17,19 @@ while true and true and false do print(1) else print(2)
...
 //│     let scrut, scrut1, scrut2;
 //│     scrut = true;
-//│     if (scrut === true) {
-//│       scrut2 = true;
-//│       if (scrut2 === true) {
-//│         scrut1 = false;
-//│         if (scrut1 === true) {
-//│           Predef.print(1);
-//│           continue lbl
+//│     σ$x: {
```
LPTK: Why do we have a label here? It seems there is no sharing.

hkmc2/shared/src/test/mlscript/ucs/normalization/ExcessiveDeduplication.mls
LPTK, line 107: TODO: Update comment.

LPTK (changes requested):
Nice changes, thanks!

Please avoid duplicating specialized splits when they're used several times with the same specialization. (Later, we might also try to preserve more sharing between specializations.)

Also, please add tests for these cases (sharing of specialized splits and lack thereof).

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
chengluyu: Fixed in 52d4be8.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
chengluyu: Resolved in a1e437f2aba08ac284fc670d5af12f06a9f2dcea.

hkmc2/shared/src/test/mlscript/ucs/normalization/ExcessiveDeduplication.mls
chengluyu: Updated in 05347d479fcc0488468e5927452c5655416845b2.

hkmc2/shared/src/test/mlscript/opt/AbortivePrefix.mls
chengluyu: Fixed in 70b37dbf6fba7358969696d39780d9ad53075100.

hkmc2/shared/src/test/mlscript/block-staging/Functions.mls
chengluyu: Fixed in 151f3238b90a3e131143c534dfb282fd687fcc08.

hkmc2/shared/src/test/mlscript/codegen/ClassMatching.mls
chengluyu: The stale `begin` is no longer present in the current golden output; this block now lowers directly to the `match`/`end` shape shown in `ClassMatching.mls`.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
chengluyu: This is covered by the same cleanup as the adjacent Config-field comment: `Normalization` now uses the contextual `Config` directly instead of carrying the extra field.

hkmc2/shared/src/test/mlscript/ucs/normalization/DeduplicationWhile.mls
chengluyu: Fixed in 3244d8be36c0584524b5663210e1beeee0732d67: `LetSplit` lowering now recognizes `while` continuations as terminal, so it can use the direct join-label form instead of wrapping the branch in the extra `σ$x` exit label. The updated `DeduplicationWhile.mls` snapshot now emits direct `continue lbl` paths instead of that label.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -77,40 +79,166 @@ class Normalization(lowering: Lowering)(using tl: TL)(using Raise, Ctx, State) e
...
+  private def normalize(split: Split)(using vs: VarSet, jpctx: JoinPointCtx, spctx: SpecializedSplitCtx): (Split, Set[SplitSymbol]) = trace(
     pre = s"normalize <<< ${split.prettyPrint}",
-    post = (res: Split) => "normalize >>> " + res.prettyPrint,
+    post = (res: (Split, Set[SplitSymbol])) => "normalize >>> " + res._1.prettyPrint,
   ):
     normalizeImpl(split)
+
+  private def shouldShareJoinPoint(body: Split, refCount: Int): Bool =
+    refCount > 1 && (config.patMatConsequentSharingThreshold match
+      case S(threshold) => body.size * 2 > threshold
```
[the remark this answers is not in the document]
chengluyu: Fixed in c7a10d846: the sharing threshold now scales by `refCount` (`body.size * refCount > threshold`), preserving the old two-reference behavior while allowing heavily reused alternatives to pass the threshold. Added a `LetSplitJoinPoints.mls` case with three surviving references and threshold 2.

Authorship note: this reply was written by GPT-5.5 via Codex.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -77,40 +79,166 @@ class Normalization(lowering: Lowering)(using tl: TL)(using Raise, Ctx, State) e
...
+      * terms are abortive, and `UseSplit` compiles to a `Break`. */
+    private def alwaysTerminates: Bool = split match
+      case Split.End => true
+      case Split.Else(t) => t match
+        case _: Term.Ret | _: Term.Throw => true
+        case _ => false
+      case Split.Cons(Branch(_, _, cons), tail) =>
+        cons.alwaysTerminates && tail.alwaysTerminates
+      case Split.Let(_, _, tail) => tail.alwaysTerminates
+      case Split.LetSplit(_, tail) => tail.alwaysTerminates
```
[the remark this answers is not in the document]
chengluyu: Fixed in f002b6f03: `alwaysTerminates` now accounts for both the join-point body and the tail of `LetSplit`, while `UseSplit` delegates to the referenced body instead of assuming it terminates.

Authorship note: this reply was written by GPT-5.5 via Codex.

hkmc2/shared/src/main/scala/hkmc2/semantics/Split.scala
```diff
@@ -47,31 +54,52 @@ enum Split extends AutoLocated with ProductWithTail:
       case Cons(head, tail) => Cons(head, tail.duplicate)
       case Let(name, term, tail) => Let(name, term, tail.duplicate)
       case Else(default) => Else(default)
-      case End => End).setDuplicated
+      case End => End
+      case LetSplit(sym, tail) => LetSplit(sym, tail.duplicate)
+      case UseSplit(sym) => UseSplit(sym)).setDuplicated
```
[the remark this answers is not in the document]
chengluyu: Fixed in 53f4c809c: `Split.mkClone` and `Split.duplicate` now carry a freshening map for `SplitSymbol`s. `LetSplit` creates a fresh symbol/body copy, and branch continuations plus `UseSplit` references are rewritten through the same map, so lowering no longer mutates/shared-labels the original symbol across duplicated split trees.

Authorship note: this reply was written by GPT-5.5 via Codex.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -974,6 +974,14 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
...
       Assign(l, r, setupTerm("Else", Value.Ref(l) :: Nil)(k))
     case Split.End => setupTerm("End", Nil)(k)
+    case Split.LetSplit(sym, tail) =>
+      quoteSplit(sym.body): r1 =>
+        val l1 = loweringCtx.registerTempSymbol(N)
+        blockBuilder.assign(l1, r1)
+          .chain(b => quoteSplit(tail)(r2 => Assign(loweringCtx.registerTempSymbol(N), r2, b)))
+          .rest(setupTerm("LetSplit", Value.Ref(l1) :: Nil)(k))
+    case Split.UseSplit(sym) =>
+      setupTerm("UseSplit", Nil)(k)
```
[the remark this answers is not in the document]
chengluyu: Fixed in 9ad9b65b8: `quoteSplit` now quotes a `LetSplit` as `(sym, body, tail)`, binds the split symbol through the same symbol setup path used elsewhere, and emits `UseSplit(sym)` instead of dropping the reference. I also extended the quoted `Term.Split` model with `LetSplit` and `UseSplit` constructors so the emitted terms exist at runtime.

Authorship note: this reply was written by GPT-5.5 via Codex.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -77,40 +79,166 @@ class Normalization(lowering: Lowering)(using tl: TL)(using Raise, Ctx, State) e
...
     pre = s"normalize <<< ${split.prettyPrint}",
-    post = (res: Split) => "normalize >>> " + res.prettyPrint,
+    post = (res: (Split, Set[SplitSymbol])) => "normalize >>> " + res._1.prettyPrint,
   ):
     normalizeImpl(split)
+
+  private def shouldShareJoinPoint(body: Split, refCount: Int): Bool =
+    refCount > 1 && (config.patMatConsequentSharingThreshold match
+      case S(threshold) => body.size * 2 > threshold
+      case N => false)
```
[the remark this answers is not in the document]
chengluyu: Updated the PR description: it now says `patMatConsequentSharingThreshold` and `:patMatConsequentSharingThreshold` are retained, and that the threshold controls whether normalized join-point candidates survive or are inlined.

Authorship note: this reply was written by GPT-5.5 via Codex.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -77,40 +79,166 @@ class Normalization(lowering: Lowering)(using tl: TL)(using Raise, Ctx, State) e
...
-  def normalizeImpl(split: Split)(using vs: VarSet): Split = split match
+  extension (split: Split)
+    /** Check if any branch in the split tests the given scrutinee. */
+    private def referencesScrutinee(scrutinee: Term.Ref): Bool = split match
+      case Split.Cons(Branch(thatScrutinee, _, continuation), tail) =>
+        (scrutinee === thatScrutinee) || continuation.referencesScrutinee(scrutinee) || tail.referencesScrutinee(scrutinee)
+      case Split.Let(_, _, tail) => tail.referencesScrutinee(scrutinee)
+      case Split.Else(_) | Split.End => false
+      case Split.LetSplit(_, tail) => tail.referencesScrutinee(scrutinee)
+      case Split.UseSplit(_) => false
```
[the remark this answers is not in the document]
chengluyu: Fixed in f002b6f03: `referencesScrutinee` and `countUseSplit` now traverse `LetSplit` bodies as well as tails, and `UseSplit` follows the referenced body where appropriate.

Authorship note: this reply was written by GPT-5.5 via Codex.

chengluyu (commented):
LGTM. 

Next, I will implement the deduplication for the consequent and alternative splits of disjunctive patterns. There is some logic over there that might be shared.

hkmc2/shared/src/test/mlscript/codegen/CurriedClasses.mls
```diff
@@ -77,5 +77,3 @@ h2.y
 h.y
 //│ ═══[RUNTIME ERROR] Expected: '2', got: '3'
 //│ = 3
-
-
```
LPTK, line 81: Why did you make these changes?

hkmc2/shared/src/test/mlscript/codegen/CurriedFunctions.mls
```diff
@@ -94,5 +94,3 @@ foo(loudly(1))(loudly(2))(loudly(3))(loudly(4))
 //│ > world
 //│ > hi 4
 //│ > 1 2 3 4
-
-
```
LPTK, line 98: Ditto

hkmc2/shared/src/main/scala/hkmc2/semantics/Split.scala
```diff
@@ -3,15 +3,30 @@ package semantics
...
+
+  /** Duplicate this branch during split rewriting. Unlike `mkClone`, this keeps
+    * the existing scrutinee and pattern nodes, but it still has to thread the
+    * split-symbol renaming map through the continuation. */
+  private[semantics] def duplicate(freshSplitSymbols: MutMap[SplitSymbol, SplitSymbol]): Branch =
+    Branch(scrutinee, pattern, continuation.duplicate(freshSplitSymbols))
+
   override def children: Vector[Located] = Vector.triple(scrutinee, pattern, continuation)
-  
+
```
LPTK: Why remove this indentation?

hkmc2/shared/src/main/scala/hkmc2/semantics/Split.scala
```diff
@@ -124,7 +124,7 @@ enum Split extends AutoLocated with ProductWithTail:
     case Split.End => true
     case Split.LetSplit(_, tail) => tail.isEmpty
     case Split.UseSplit(sym) => sym.body.isEmpty
-  
+
```
LPTK: Ditto

hkmc2/shared/src/test/mlscript/ucs/normalization/DeduplicationWhile.mls
LPTK: There is no scuh commit. What is this referring to?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -77,40 +79,167 @@ class Normalization(lowering: Lowering)(using tl: TL)(using Raise, Ctx, State) e
...
+        cons.countUseSplit(sym) + tail.countUseSplit(sym)
+      case Split.Let(_, _, tail) => tail.countUseSplit(sym)
+      case Split.Else(_) | Split.End => 0
+      case Split.LetSplit(s, tail) => s.body.countUseSplit(sym) + tail.countUseSplit(sym)
+      case Split.UseSplit(s) => if s eq sym then 1 else 0
+
+    /** Whether every leaf of `split` unconditionally transfers control away
+      * (no fall-through). `End` compiles to `throw`, explicit `return`/`throw`
+      * terms are abortive, and `UseSplit` transfers to its referenced body. */
+    private def alwaysTerminates: Bool = split match
```
LPTK, line 148: This name is inadequate. Termination obviously cannot be determined syntactically.

hkmc2/shared/src/test/mlscript/ucs/normalization/DeduplicationWhile.mls
chengluyu: It refers to `3244d8be3`. I did a rebase.CleanShot 2026-05-07 at 14 56 56

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -297,13 +434,43 @@ class Normalization(lowering: Lowering)(using tl: TL)(using Raise, Ctx, State) e
...
+    case Split.LetSplit(sym, tail) =>
+      // Lower the join point: the body goes into the Label's `rest`, and
+      // UseSplit generates Break(joinLabel) to reach it.
+      val joinLabel = new LabelSymbol(N, sym.nme)
+      sym.label = S(joinLabel)
+      // When every leaf of both sides transfers control away (explicit
+      // `return`/`throw`, `UseSplit` → break, or `End` → throw), no value
+      // can flow into `cont`, so the exit-label wrapper is unnecessary even
+      // when `cont` is not a TailOp.
+      if (cont eq Ret) || (cont eq Thrw) || (form is IfLikeForm.While) ||
```
LPTK, line 452: Why is this using brittle `eq` tests instead of `isInstanceOf[TailOp]`?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -297,13 +434,43 @@ class Normalization(lowering: Lowering)(using tl: TL)(using Raise, Ctx, State) e
...
+          (sym.body.alwaysTerminates && tail.alwaysTerminates) then
+        val bodyBlock = lowerSplit(sym.body, cont)
+        Label(joinLabel, false, lowerSplit(tail, cont), bodyBlock)
+      else
+        // Other continuations (including ImplctRet, which generates `expr;`
+        // without `return`) can fall through the Label body into the rest.
+        // Wrap with an exit label and temp variable so every path stores its
+        // result, breaks to exitLabel, then the original cont runs once.
```
LPTK, line 460: Why are all the code and comments compressed to 80 columns? 100 is more appropriate.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -1461,4 +1473,3 @@ object MergeMatchArmTransformer extends BlockTransformer(SymbolSubst.Id):
               dfltRewritten.fold(restRewritten)(Begin(_, restRewritten)) |> some, rest)
       case _ => m
     case b => b
-
```
LPTK, line 1464: 👎

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -77,40 +79,167 @@ class Normalization(lowering: Lowering)(using tl: TL)(using Raise, Ctx, State) e
...
+    private def countUseSplit(sym: SplitSymbol): Int = split match
+      case Split.Cons(Branch(_, _, cons), tail) =>
+        cons.countUseSplit(sym) + tail.countUseSplit(sym)
+      case Split.Let(_, _, tail) => tail.countUseSplit(sym)
+      case Split.Else(_) | Split.End => 0
+      case Split.LetSplit(s, tail) => s.body.countUseSplit(sym) + tail.countUseSplit(sym)
+      case Split.UseSplit(s) => if s eq sym then 1 else 0
+
+    /** Whether every leaf of `split` unconditionally transfers control away
+      * (no fall-through). `End` compiles to `throw`, explicit `return`/`throw`
```
LPTK, line 146: Why does this comment say that "`End` compiles to `throw`"? It's clearly not true. Does it want to say something more specific in this special context?

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -77,40 +79,167 @@ class Normalization(lowering: Lowering)(using tl: TL)(using Raise, Ctx, State) e
...
-      val whenFalse = normalizeImpl(specialize(alternative, -, scrutinee, pattern).clearFallback)
-      Branch(scrutinee, pattern, whenTrue) ~: whenFalse
+      if alternative.isTrivial || alternative.referencesScrutinee(scrutinee) then
+        // The alternative is trivial (End or UseSplit — no code worth sharing),
+        // or it references the same scrutinee (so positive and negative
+        // specialization may transform it differently on each side, making
+        // sharing unsound). Duplicate and specialize separately.
+        // TODO: detect when both specializations agree and share even here.
+        val positiveSplit = consequent ++ alternative.duplicate
+        val (whenTrue, trueRefs) = normalize(specialize(positiveSplit, +, scrutinee, pattern).getOrElse(positiveSplit))
```
LPTK, line 181: `specialize` is now specified to return `N` when no chnage was made. But this info is simply discarded here. Wasn't it supposed to be used to potentially share the split here? (I guess it would be possible when both positive and negative specialization return `N`.)

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -135,39 +264,40 @@ class Normalization(lowering: Lowering)(using tl: TL)(using Raise, Ctx, State) e
...
+              S(aliasBindings(pattern, thatPattern)(rec(continuation).getOrElse(continuation) ++ rec(tail).getOrElse(tail)))
             else if thatPattern <:< pattern then
               log(s"Case 1.1.2: $pattern <:< $thatPattern")
-              pattern.markAsRefined; split.copy(tail = rec(tail))
+              pattern.markAsRefined
+              rec(tail).map(newTail => split.copy(tail = newTail))
             else if split.isFallback then
               log(s"Case 1.1.3: $pattern is unrelated with $thatPattern")
-              rec(tail)
+              S(rec(tail).getOrElse(tail))
```
LPTK, line 291: Is this really the correct logic? Why do we discard the lack of change and replace it wirh `S`, here? This is an open question; I just want to understand.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
chengluyu: "Termination" is not appropriate here. It means either `return` or `throw`.

hkmc2/shared/src/test/mlscript/ucs/normalization/LetSplitJoinPoints.mls
```diff
@@ -0,0 +1,458 @@
...
+//│ set tmp2 = B¹(2);
+//│ f3¹(tmp2)
+//│ —————————————————| Output |—————————————————————————————————————————————————————————————————————————
+//│ = "B"
+
+
+// === Join point cases ===
+
+
+// * Join point: non-trivial Else alternative on different scrutinee
```
LPTK, line 84: This testcase comment says there is a join point, but I don't see one. What's going on?

hkmc2/shared/src/test/mlscript/ucs/normalization/LetSplitJoinPoints.mls
```diff
@@ -0,0 +1,458 @@
...
+//│       end
+//│   end
+//│ };
+//│ do f4¹(0, 9);
+//│ do f4¹(9, 1);
+//│ f4¹(9, 9)
+//│ —————————————————| Output |—————————————————————————————————————————————————————————————————————————
+//│ = "neither"
+
+// * Join point: multi-branch alternative, different scrutinee
```
LPTK, line 114: Same.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
chengluyu: I do a newline check in the end.

hkmc2/shared/src/test/mlscript/ucs/normalization/LetSplitJoinPoints.mls
```diff
@@ -0,0 +1,458 @@
...
+//│ do f5¹(0, 9);
+//│ do f5¹(9, 1);
+//│ f5¹(9, 2)
+//│ —————————————————| Output |—————————————————————————————————————————————————————————————————————————
+//│ = "y=2"
+
+// * Join point: deeply nested class patterns share the default branch
+data class S(value)
+
+:sir
```
LPTK, line 149: Same.

hkmc2/shared/src/test/mlscript/ucs/normalization/LetSplitJoinPoints.mls
```diff
@@ -0,0 +1,458 @@
...
+//│ set tmp6 = S¹(42);
+//│ set tmp7 = S¹(tmp6);
+//│ do f6¹(tmp7);
+//│ set tmp8 = S¹(42);
+//│ do f6¹(tmp8);
+//│ f6¹(42)
+//│ —————————————————| Output |—————————————————————————————————————————————————————————————————————————
+//│ = "default"
+
+// * Join point: default branch with a computation should not be duplicated
```
LPTK, line 204: It has nothing to do with the presence of a computation; what matters is the term size.

hkmc2/shared/src/test/mlscript/ucs/normalization/LetSplitJoinPoints.mls
LPTK, line 1: Looks like this test file should use bigger consequents to share or a different sharing threshold, as they don't demonstrate what they purport to.

hkmc2/shared/src/test/mlscript/ucs/normalization/SpecializedSplitSharing.mls
```diff
@@ -0,0 +1,157 @@
...
+//│             end
+//│           Dog⁰ =>
+//│             match z
+//│               1 =>
+//│                 return ["dog-one"]
+//│               else
+//│                 return [101, 102, 103, 104, 105, 106, 107, 108, 109, 110, 111, 112]
+//│             end
+//│           else
+//│             return [101, 102, 103, 104, 105, 106, 107, 108, 109, 110, 111, 112]
```
LPTK, line 104: Is this really supposed to be duplicated?

(This might be related with the future improvement what you told me before; if so, pls document it.)

hkmc2/shared/src/test/mlscript/codegen/ClassMatching.mls
LPTK: I'm not asking you what's currently the case. I'm asking you, LUYU @chengluyu (not ChatGPT) why we used to have a `begin` there. I just want to understand that, as it's surprising to me.

chengluyu: I'm not sure, but it looks like it's related to the previous consequent sharing logic.

At that time, how it handled pure expression `if`s was to _always_ first place the results into a `tmp` var, and then pass that `tmp` into subsequent tails. 

But now we will check whether the consequents are actually used. In this case, since nothing uses `0` or `1`, it simply won't create a block and place them into `tmp` var

LPTK: Hmm ok.

LPTK: So do we still observe `begin`s in such positions in some other cases?

LPTK: Asking because these should probably have been flattened.

chengluyu: > Asking because these should probably have been flattened.

I don't know why they were not flattened.

> So do we still observe `begin`s in such positions in some other cases?

The `begin` here looks weird indeed. But it's fixed by this commit anyway. Do we really need to dig into this?

LPTK: Probably not.

LPTK:
Subsumed by https://github.com/hkust-taco/mlscript/pull/486.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
chengluyu: Case 1.1.3 was dead code because I found isFallback is never set. I removed it in `dee50a3dd`. I also found Case 1.1.4 was returning S when assuming removed no fields, and I fixed it in `0a8e240c7`. I also refined the comment of the method in `005b7510a`.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
chengluyu: Fixed in `aa2f49e03`.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
chengluyu: Fixed in `afde9f038`.

hkmc2/shared/src/test/mlscript/ucs/normalization/LetSplitJoinPoints.mls
chengluyu: Deleted in `fb52c5815` because consequent sharing is already covered by `Deduplication.mls`, `DeduplicationWhile.mls`, and `ExcessiveDeduplication.mls`.

hkmc2/shared/src/test/mlscript/ucs/normalization/LetSplitJoinPoints.mls
chengluyu: See https://github.com/hkust-taco/mlscript/pull/457#discussion_r3294234064

hkmc2/shared/src/test/mlscript/ucs/normalization/LetSplitJoinPoints.mls
chengluyu: See https://github.com/hkust-taco/mlscript/pull/457#discussion_r3294234064

hkmc2/shared/src/test/mlscript/ucs/normalization/LetSplitJoinPoints.mls
chengluyu: See https://github.com/hkust-taco/mlscript/pull/457#discussion_r3294234064

hkmc2/shared/src/test/mlscript/ucs/normalization/LetSplitJoinPoints.mls
chengluyu: See https://github.com/hkust-taco/mlscript/pull/457#discussion_r3294234064

PR #458, opened 2026-04-13
Effect handlers: Straight-line optimization

CAG2Mark:
Our implementation of effect handlers relies on creating a state machine from function bodies to suspend/resume them. This was previously done by creating a control flow graph partitioned along the points where the function could suspend and resume, and then jumping between the states using a loop one big switch statement.

This PR:
- Optimizes the switch statement to use fallthrough. In particular, we partition the control flow graph into "straight line" segments, and then the consecutive states in these segments will fall through to the next without jumping back to the start of the loop. However, it will still jump to the start of the loop at the end of each straight line segment.
- Adds more cases to the switch specialization to handle more tricky fallthrough cases, namely those where the scrutinee is not assigned to directly at the end of a case.


hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -553,24 +612,84 @@ class HandlerLowering(paths: HandlerPaths, opt: EffectHandlers)(using TL, Raise,
...
 
     val mainLoop =
       if parts.states.size <= 1 then
-        postTransform.applyBlock(parts.states.head._2.blk)
+        segmentTailTransform.applyBlock(parts.states.head._2.blk)
       else
+        val matches = straightLines.map(straightLineToArms).foldLeft[Block](End()):
+          case (acc, f) => f(acc)
+        Label(mainLoopLbl, true, matches, End())
+        /*
```
LPTK: TODO: remove comment?

hkmc2/shared/src/main/scala/hkmc2/codegen/SpecializedSwitch.scala
```diff
@@ -67,13 +68,95 @@ private enum MatchType:
...
-// N: None of the cases
-@tailrec
-private def isTailAssign(b: Block, scrutSym: Local): Opt[Literal] = b match
-  case a @ Assign(`scrutSym`, Value.Lit(l), End(_)) => S(l)
-  case b: NonBlockTail => isTailAssign(b.rest, scrutSym)
-  case _: BlockTail => N
+private object PostCondRes:
+  val empty = PostCondRes(false, false, Map.empty)
+
+private case class PostCondRes(stop: Bool, isAbortive: Bool, varsMap: Map[Local, Literal]):
```
LPTK: Pls use a more informative name.

LPTK (commented):
Can we sort the states in the switch by uid?

LPTK:
Is this ready for review now?

CAG2Mark:
Oh, I forgot to sort the switch by UID

CAG2Mark:
We should probably benchmark it before merging though

LPTK:
> We should probably benchmark it before merging though

I think it's fine. We can always revert it later, if needed (quite unlikely).

LPTK (approved):
LGTM, nice changes!

CAG2Mark:
> > We should probably benchmark it before merging though
> 
> I think it's fine. We can always revert it later, if needed (quite unlikely).

In any case, I just got some initial benchmark numbers from @AnsonYeung and it seems it is faster across the board. We can see a 5-10% increase generally.

LPTK:
> We can see a 5-10% increase generally.

Awesome. That's great to hear!

PR #463, opened 2026-04-16
wasm: Remove redundant `MemberSymbol` resolution by ID

Derppening:
This is no longer needed as well, and all tests pass without this ID-based lookup of struct fields.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
@@ -919,19 +919,9 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
     val symToField = structInfo.compType match
       case ty: StructType => ty.fieldsBySym
       case _ => lastWords(s"Cannot select field from non-struct type: ${structInfo.compType.toWat}")
-    val fieldIdx = symToField.get(sym)
-      .orElse:
-        sym match
-          case memSym: MemberSymbol if fieldOwner(memSym).contains(thisSym) =>
-            symToField.find((fieldSym, _) => fieldSym.nme == sym.nme).map((_, v) => v)
-          case _ => N
-      .map(_.id)
-    FieldIdx(SymIdx(
-      fieldIdx getOrElse:
-        lastWords(
-          s"Missing field `${sym.toString}` in struct `${thisSym.toString}` with type `${structInfo.toWat.mkString()}`",
-        ),
-    ))
+    val fieldIdx = symToField.get(sym).map(_.id).getOrElse:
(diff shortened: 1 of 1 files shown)


hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1033,9 +1025,9 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
                       s"Block IR: `${
                           fun.toString
                         }`\nCompiled WAT: `${
-                          base.toWat.toString
+                          base.toWat.mkString()
```
LPTK, line 1028: `toString` shouldn't have `()`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -918,20 +918,12 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
-            symToField.find((fieldSym, _) => fieldSym.nme == sym.nme).map((_, v) => v)
-          case _ => N
-      .map(_.id)
-    FieldIdx(SymIdx(
-      fieldIdx getOrElse:
-        lastWords(
-          s"Missing field `${sym.toString}` in struct `${thisSym.toString}` with type `${structInfo.toWat.mkString()}`",
-        ),
-    ))
+      case _ => lastWords(s"Cannot select field from non-struct type: ${structInfo.compType.toWat.mkString()}")
```
LPTK, line 921: Can this be triggers by user code? If so, it shoould be a proper compilation error.

LPTK (approved):
Otherwise LGTM, thx.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
Derppening: Removing the `()` from `mkString` causes the error message to contain the lambda signature instead:

```
//│ Compiled WAT: `hkmc2.codegen.wasm.text.WatBuilder$$Lambda/0x00007faa50340000@258ee7de`
//│ ... which has type `hkmc2.codegen.wasm.text.WatBuilder$$Lambda/0x00007faa503403d0@4930539b`
```

I guess it's because `mkString` takes a default argument?

LPTK: Whoops, I misread this as `toString()`. Never mind, carry on.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
Derppening: I don't think this can be triggered by user code.

PR #468, opened 2026-04-18
Handler IR changes

AnsonYeung:
Remove `HandleBlock` from the IR, as it create extra complexity that is not inherent to effect handlers.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
@@ -327,7 +327,6 @@ class HandlerLowering(paths: HandlerPaths, opt: EffectHandlers)(using TL, Raise,
           source = Diagnostic.Source.Compilation))
       case Throw(_) => blk
       case Scoped(_, body) => go(body) // PreHandlerLowering
-      case _: Suspend | _: HandleSuspension => lastWords("unexpected handleBlock") // already translated at this point
 
     val initId = allocId()
     // Note: initial part will only be resumed if stack safety is on.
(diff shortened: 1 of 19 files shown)


hkmc2/shared/src/test/mlscript/decls/Prelude.mls
```diff
@@ -245,6 +245,8 @@ declare module js with
     bitor
     shl
     try_catch
+    suspend
+    handle_suspension
```
LPTK: Why put these in `js`? Effect handlers should not be backend-specific (even though support for other backends still needs to be added). Let's create a new virtual module called `runtime` and put these in.

LPTK (approved):
Otherwise, LGTM. Nice simplifications!

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -556,7 +555,14 @@ class HandlerLowering(paths: HandlerPaths, opt: EffectHandlers)(using TL, Raise,
         FunDefn(fun.owner, fun.sym, fun.dSym, fun.params, bod2)(fun.forceTailRec, fun.configOverride, fun.visibility)
       (debugInfoSym, debugInfo, fun2)
 
-    val subblockTransform = new BlockTransformer(SymbolSubst.Id):
+    // transform inner function/class and effect handler intrinsics to the JS runtime functions.
```
```suggestion
    // transform inner function/class and effect handler intrinsics to the runtime functions.
```

AnsonYeung: Some backend may have a more efficient implementation available (i.e. native stack manipulation). So effect handler can be backend-specific.

PR #473, opened 2026-04-21
Fix missing configs for `JSBuilder` in compile test

NeilKleistGao:


hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -26,7 +26,7 @@ abstract class CodeBuilder:
   type Context
   
 
-class JSBuilder(using TL, State, Ctx, Config) extends CodeBuilder:
+class JSBuilder(using Config, TL, State, Ctx) extends CodeBuilder:
```
LPTK, line 29: Why this change?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -92,8 +92,8 @@ class Lowering()(using Config, TL, Raise, State, Ctx):
           r.expanded
       case t => t
   
-  var lowerHandlers: Bool = config.effectHandlers.isDefined
-  var lift: Bool = config.liftDefns.isDefined
+  val lowerHandlers: Bool = config.effectHandlers.isDefined
+  val lift: Bool = config.liftDefns.isDefined
```
LPTK, line 96: Are these still used?

LPTK (commented):
Does it work with compiled files? Could you please add a test to check that? (By committing the corresponding `.mjs`.)

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
NeilKleistGao: Yes. e.g.,
```scala
case h @ Handle(lhs, rhs, as, cls, defs, bod) =>
      if !lowerHandlers then
        return fail:
          ErrorReport(
            msg"Effect handlers are not enabled" ->
            h.toLoc :: Nil,
            source = Diagnostic.Source.Compilation) 
```

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
NeilKleistGao: When we create `JSBuilder`, we only need to pass the config explicitly:
```scala
val jsb = ltl.givenIn:
  JSBuilder(using effectiveConfig)
```

NeilKleistGao:
> Does it work with compiled files?

Yes. The corresponding `.mjs` file is committed.

hkmc2/shared/src/test/mlscript-compile/.gitignore
```diff
@@ -3,3 +3,4 @@
 !/Runtime.mjs
 !/RuntimeJS.mjs
 !/apps/parsing/vendors/railroad/railroad.mjs
+!/NoFreeze.mjs
```
LPTK:
```suggestion
!/NoFreeze.mjs

```

LPTK (approved):
LGTM, thanks!

PR #474, opened 2026-04-22
wasm: Refactor more `Str` to `SymIdx`, miscellaneous cleanup

Derppening:
- Refactor `Str` to `SymIdx` where it makes sense(TM)
- Remove deprecated functions in `Ctx`
- Refactor usages of `Instructions.*` (due to conflict with Scala builtins or `Elaborator.*`) to `*Instr`
- Refactor `.map(...).getOrElse(...)` to `fold`
- Use postfix syntax for `getOrElse` where applicable


LPTK (approved):
LGTM, thx.

PR #476, opened 2026-04-23
Track annotations in block IR and stage definitions lifted from staged modules

NeilKleistGao:
Related: #355 
- <del>Add `isStaged` field for `Defn`</del>Track `annotations`  for `Defn`
- Show `staged` keyword for staged definitions when printing the block IR
- If a definition is in a staged module and lifted, mark the definition staged as well.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
@@ -234,7 +234,7 @@ sealed abstract class Block extends Product:
           val newBody = d.body.flattened
           if newBody is d.body
           then d
-          else d.copy(body = newBody)(forceTailRec = d.forceTailRec, configOverride = d.configOverride, annotations = d.annotations)
+          else d.copy(body = newBody)(configOverride = d.configOverride, annotations = d.annotations)
         case v: ValDefn => v
         case c: ClsLikeDefn =>
           val newPreCtor = c.preCtor.flattened
@@ -242,7 +242,7 @@ sealed abstract class Block extends Product:
           def flattenMethods(ms: List[FunDefn]) = ms.mapConserve:
             case f@FunDefn(owner, sym, dSym, params, body) =>
               val newBody = body.flattened
-              if newBody is body then f else f.copy(body = newBody)(forceTailRec = f.forceTailRec, configOverride = f.configOverride, annotations = f.annotations)
+              if newBody is body then f else f.copy(body = newBody)(configOverride = f.configOverride, annotations = f.annotations)
           val newMethods = flattenMethods(c.methods)
           val newCompanion = c.companion.mapConserve: c =>
             val newCtor = c.ctor.flattened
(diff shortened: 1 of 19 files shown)


LPTK (commented):
Shouldn't we rather preserve all annotations and make things like `isStaged` and `visibility` methods/lazy vals that retrieve the relevant information?

NeilKleistGao:
> Shouldn't we rather preserve all annotations and make things like `isStaged` and `visibility` methods/lazy vals that retrieve the relevant information?

Yes, I think we can do it.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -806,6 +806,11 @@ class Lifter(topLevelBlk: Block)(using State, Raise, Config):
       val captureArgs = capturesOrdered.map(c => ctx.capturesMap(c).asArg)
       val localArgs = passedSymsOrdered.map(l => ctx.symbolsMap(l).asArg)
       defnsArgs ::: captureArgs ::: localArgs
+
```
```suggestion
    
```

LPTK (approved):
LGTM, nice simplification!

LPTK:
@CAG2Mark Could you pls take a quick look at the part affecting the lifter?

hkmc2/shared/src/main/scala/hkmc2/codegen/ReflectionInstrumenter.scala
```diff
@@ -368,15 +368,15 @@ class ReflectionInstrumenter(using State, Raise, Ctx) extends BlockTransformer(n
             val (stagedMethods, debugPrintCode) = c.methods
               .map(applyFunDefnInner)
               .unzip
-            val newModule = c.copy(methods = c.methods ++ stagedMethods)(c.configOverride)
+            val newModule = c.copy(methods = c.methods ++ stagedMethods)(c.configOverride, Nil)
             Define(newModule, rest)
```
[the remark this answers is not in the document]
LPTK: Oof, good catch.

CAG2Mark:
> @CAG2Mark Could you pls take a quick look at the part affecting the lifter?

Hi, been a bit busy with something today. I'll look tomorrow.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -562,24 +566,29 @@ final case class FunDefn(
...
-    val visibility: Visibility,
+    val annotations: Ls[Annot],
 ) extends Defn:
   val innerSym = N
   val asPath = Value.Ref(sym, S(dSym))
+  def forceTailRec: Bool = annotations.contains(Annot.TailRec)
+  def visibility: Visibility = annotations.collectFirst:
+    case Annot.Modifier(Keyword.`private`) => Visibility.Private
+    case Annot.Modifier(Keyword.`public`) => Visibility.Public
+  .getOrElse(Visibility.Public)
```
CAG2Mark: I think these would be better as `lazy val`s.

CAG2Mark:
The lifter changes seem fine to me.

I don't know what staged definitions are, but the PR seems to do what it says it does with respect to the lifter.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
LPTK: Probably doesn't hurt, and it would hurt more if these are `defs` and they happen to be called many times in some contexts.

NeilKleistGao: I think it is reasonable to use `lazy val`.

NeilKleistGao:
I think this PR is ready to be merged.

LPTK:
Thanks!

PR #477, opened 2026-04-23
Implement Wasm RTTI

b-temirov:
This PR adds RTTI to the Wasm backend by introducing per-class `typeinfo` objects, virtual-method slot dispatch, and RTTI-based class pattern matching. It also migrates runtime tags from object headers into `typeinfo`.


hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -261,6 +377,45 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
     ordered.toList
   end sortTopLevelClasses
 
+  /** Returns the elaborated source methods for this class. */
+  private def semanticMethodDefs(defn: ClsLikeDefn)(using Raise): List[TermDefinition] =
+    val clsDef = defn.isym.defn.get.asInstanceOf[hkmc2.semantics.ClassLikeDef]
+    clsDef.body.methods.filter(_.body.nonEmpty)
+
```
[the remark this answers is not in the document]
b-temirov: For this PR, abstract classes are out of scope.

hkmc2/shared/src/test/mlscript/wasm/Basics.mls
```diff
@@ -134,16 +139,28 @@ class Foo(val a)
...
+//│       (global.set $Foo_typeinfo
+//│         (struct.new_default $Foo_typeinfo))
+//│       (struct.set $Foo_typeinfo $$tag
+//│         (ref.cast (ref $Foo_typeinfo)
+//│           (global.get $Foo_typeinfo))
+//│         (i32.const 1))
+//│       (struct.set $Foo_typeinfo $$parent
+//│         (ref.cast (ref $Foo_typeinfo)
+//│           (global.get $Foo_typeinfo))
+//│         (ref.null $TypeInfoBase))))
```
LPTK: There's probably no need for this function; just initialize the global variables appropriately using `new`, not `new_default`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -46,25 +46,142 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
 
-  private def baseObjectStruct(using Ctx): StructType =
-    ctx.getTypeInfo_!(baseObjectSym).compType match
-      case struct: StructType => struct
-      case other => lastWords(s"Base Object type must be a struct, found ${other.toWat.mkString()}")
+  /** Returns the Wasm type index of the synthetic base RTTI struct. */
+  private def typeInfoBaseTypeIdx(using Ctx): TypeIdx =
+    ctx.getType_!(typeInfoBaseSym)
+
+  /** Resolves the field index for a symbolic field inside a previously registered struct type. */
```
LPTK: What's a _symbolic_ field?
```suggestion
  /** Resolves the field index for a field inside a previously registered struct type. */
```

hkmc2/shared/src/test/mlscript/wasm/VirtualMethods.mls
```diff
@@ -0,0 +1,201 @@
...
+class A() with
+  virtual fun f(x) = x + 1
+class B() extends A() with
+  fun f(x) = x * 2
+fun callF(a: A) = a.f(5)
+callF(B())
+//│ Wat:
+//│ (module
+//│   (type $TypeInfoBase (sub (struct (field $$tag (mut i32)) (field $$parent (mut (ref null any))))))
+//│   (type $Object (sub (struct (field $$typeinfo (mut (ref null any))))))
```
LPTK: Use more precise type than `ref null any`, avoiding the cast later on.

hkmc2/shared/src/test/mlscript/wasm/VirtualMethods.mls
```diff
@@ -0,0 +1,201 @@
...
+//│   (elem $entry declare func $entry)
+//│   (start $start1))
+//│ Wasm result:
+//│ = 10
+
+
+class A(val a) with
+  virtual fun f(x) = x + a
+class B(val b) extends A(b) with
+  fun f(x) = x * b
```
LPTK:
```suggestion
  virtual fun f(x) = x * b
```

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -2102,11 +2440,15 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
         val ordered = sortTopLevelClasses(collectTopLevelClassDefns(p.main))
+        ordered.foreach(predeclareClassVirtualTable)
+        ordered.foreach(predeclareClassTypeInfoType)
+        ordered.foreach(predeclareClassTypeInfoGlobal)
         ordered.foreach(predeclareClassType)
         predeclareClassTags(ordered)
         ordered.foreach(predeclareClassInit)
         ordered.foreach(predeclareClassConstructor)
         ordered.foreach(predeclareClassMethods)
+        ordered.foreach(registerClassTypeInfoInitAction)
```
Derppening: At this point, I wonder if it makes sense to consolidate all of these functions and have a single function that registers everything necessary for a class.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1892,43 +2228,31 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
-                    val clsTypeIdx = ctx.getType_!(clsBlkMemberSym)
-                    val typeinfo = ctx.getTypeInfo_!(clsTypeIdx)
-
-                    val expectedTag = typeinfo.objectTag getOrElse:
-                      lastWords(s"Expected class $clsBlkMemberSym to have an object tag")
-
-                    // TODO (https://github.com/orgs/hkust-taco/projects/14/views/1?pane=issue&itemId=174476970):
-                    // replace with RTTI ancestry checks once each object carries runtime type information.
-                    val matchTags = ctx.getAllRuntimeTags(clsBlkMemberSym).getOrElse(LinkedHashSet(expectedTag))
-
```
Derppening, line 1872: As discussed in the meeting, can we remove `ctx.getAllRuntimeTags` (and its related code)?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -348,6 +529,19 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
       exportName: Opt[Str],
   )(using Ctx, Raise, Scope): Unit =
     val funcTy = declareClassFuncType(defn, suffix, params)
+    predeclareClassFuncWithType(defn, suffix, params, sym, id, exportName, funcTy)
+  end predeclareClassFunc
```
Derppening: Nit: This `end` marker is redundant.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -505,6 +699,96 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+      globalSym,
+      GlobalInfo(
+        id = SymIdx(scope.allocateName(globalSym)),
+        globalType = GlobalType(globalTy, mutable = true),
+        init = ref.`null`(typeInfoTypeIdx),
+        exportName = N,
+      ),
+    )
+    typeInfoGlobals(defn.sym) = globalIdx
+  end predeclareClassTypeInfoGlobal
```
Derppening: Nit: Same here - `end` marker is redundant.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -505,6 +699,96 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+        objectTag = N,
+      ),
+    )
+    typeInfoTypeIdxs(defn.sym) = typeInfoType
+  end predeclareClassTypeInfoType
+
+  /** Predeclares the shared runtime `typeinfo` global for one supported top-level class. */
+  private def predeclareClassTypeInfoGlobal(defn: ClsLikeDefn)(using Ctx, Raise, Scope): Unit =
+    val typeInfoTypeIdx = typeInfoTypeIdxs(defn.sym)
+    val globalSym = BlockMemberSymbol(s"${defn.sym.nme}_typeinfo", Nil, nameIsMeaningful = false)
```
Derppening: Nit: I think it's better to use the Wasm identifier name instead of the class symbol name for the prefix to the `_typeinfo` global. It would make it easier to match the `typeinfo` instance to its corresponding `type`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -505,6 +699,96 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
   end registerSessionImports
 
+  /** Predeclares the per-class `typeinfo` struct type for one supported top-level class. */
+  private def predeclareClassTypeInfoType(defn: ClsLikeDefn)(using Ctx, Raise, Scope): Unit =
+    val parentTypeInfoIdx =
+      if defn.parentPath.isEmpty then ctx.getType_!(typeInfoBaseSym)
+      else typeInfoTypeIdxs(resolveParentSym(defn).get)
+
+    val inheritedFields = ctx.getTypeInfo_!(parentTypeInfoIdx).compType match
+      case struct: StructType => struct.fields
```
Derppening, line 695: Nit: Wouldn't this cause a `non-exhaustive match` warning?

Derppening (commented):
Please address these comments first and re-request a review, thanks!

b-temirov:
Addressed the comments.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -2072,12 +2377,26 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+        id = SymIdx("TypeInfoBase"),
+        StructType(Seq(
+          tagFieldSym -> Field(I32Type, mutable = true, id = "$tag"),
+          parentFieldSym -> Field(RefType.anyref, mutable = true, id = "$parent"),
+        )),
+        objectTag = N,
+      ),
+    )
+
+    // Create base Object struct with typeinfo field that all other structs will inherit
```
Derppening: Nit: This comment can be moved to before `TypeInfoBase`, and the comment updated to reflect that the two `addType` calls creates the two Wasm intrinsic types.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -349,6 +352,7 @@ end TagInfo
 
 enum WasmIntrinsicType:
   case TupleArray(mutable: Bool)
+  case VirtualMethod(arity: Int)
```
Derppening, line 327: Missing documentation.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -477,6 +481,11 @@ object Ctx:
       globalTy: RefType,
   )
 
+  case class VirtualTable(
+      virtualMethods: List[BlockMemberSymbol],
+      virtualMethodSlots: Map[BlockMemberSymbol, Int],
+  )
```
Derppening, line 466: Same here - Missing documentation.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -46,25 +46,142 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+      ref.cast(objRef, baseObjectRefType(nullable = false)),
+      RefType(typeInfoBaseTypeIdx, nullable = true),
+    )
+
+  /** Follows one direct-parent RTTI reference from a shared class `typeinfo` object. */
+  private def readTypeInfoParent(typeInfoRef: Expr)(using Ctx): Expr =
+    struct.get(
+      structFieldIdx(typeInfoBaseSym, parentFieldSym),
+      ref.cast(typeInfoRef, RefType(typeInfoBaseTypeIdx, nullable = false)),
+      RefType.anyref,
```
Derppening: If I am not mistaken, this instruction obtains the (possibly null) reference to the parent class's `typeinfo` instance correct? If so, the result type can be `(ref null $TypeInfoBase)`?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -46,25 +46,142 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+    typeInfoGlobals.get(sym).map: globalIdx =>
+      val globalTy = ctx.getGlobalType_!(globalIdx).globalType.valType
+      global.get(globalIdx, globalTy)
+
+  /** Reads the RTTI pointer stored in an object's common header. */
+  private def readObjectTypeInfo(objRef: Expr)(using Ctx): Expr =
+    struct.get(
+      structFieldIdx(baseObjectSym, typeInfoFieldSym),
+      ref.cast(objRef, baseObjectRefType(nullable = false)),
+      RefType(typeInfoBaseTypeIdx, nullable = true),
```
Derppening: Is there a case where the first field of `$$Object` is a `ref.null $TypeInfoBase`? If not, it seems like this should be `nullable = false`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -46,25 +46,142 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+  /** Returns `1` when `scrutTypeInfo` is equal to or descends from `targetTypeInfo`, else `0`. */
+  private def isSubtypeByTypeInfo(
+      scrutTypeInfo: Expr,
+      targetTypeInfo: Expr,
+  )(using Ctx, FunctionCtx, Raise, Scope): Expr =
+    val currentTmp = mkTempLocal("currentTypeInfo")
+    val targetTmp = mkTempLocal("targetTypeInfo")
+    val resultTmp = mkTempLocal("typeInfoMatch")
+    val endLabel = scope.allocateName(TempSymbol(N, "typeInfoEnd"))
+    val loopLabel = scope.allocateName(TempSymbol(N, "typeInfoLoop"))
```
Derppening: Please use `FunctionCtx.withLabel`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -502,9 +685,103 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+        val globalSym = BlockMemberSymbol(ctx.getTypeInfo_!(typeInfoTypeIdx).id.id, Nil, nameIsMeaningful = false)
+        val globalIdx = ctx.addGlobal(
+          globalSym,
+          GlobalInfo(
+            id = SymIdx(scope.allocateName(globalSym)),
+            globalType = GlobalType(RefType(typeInfoTypeIdx, nullable = false), mutable = false),
+            init = struct.`new`(typeInfoTypeIdx, initFields),
+            exportName = N,
+          ),
+        )
```
Derppening: This looks strange to me - Shouldn't these `typeinfo` structs be *imported* rather than reconstructed?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -46,25 +46,142 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+  private def readRuntimeTag(objRef: Expr)(using Ctx): Expr =
+    val typeInfoRef = ref.cast(
+      readObjectTypeInfo(objRef),
+      RefType(typeInfoBaseTypeIdx, nullable = false),
+    )
+    struct.get(
+      structFieldIdx(typeInfoBaseSym, tagFieldSym),
+      typeInfoRef,
+      I32Type,
+    )
```
Derppening: This function appears to be unused.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -2072,12 +2377,26 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
       CompiledWasmModule(ctx.toWat, entryName, systemMemMinPages, sessionExportCtx.collectedBindings.toSeq)
 
-    // Create base Object struct with tag field that all other structs will inherit
+    ctx.addType(
+      sym = S(typeInfoBaseSym),
+      TypeInfo(
+        id = SymIdx("TypeInfoBase"),
+        StructType(Seq(
+          tagFieldSym -> Field(I32Type, mutable = true, id = "$tag"),
+          parentFieldSym -> Field(RefType.anyref, mutable = true, id = "$parent"),
```
Derppening: Now that all fields in `$TypeInfoBase` are constructed using `struct.new`, the fields no longer need to be `mutable`.

Also, since `$parent` always points to a (nullable) `$TypeInfoBase`, could we just make the type `RefType(SymIdx("TypeInfoBase"), nullable = false)`?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1892,43 +2219,21 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
-                      lastWords(s"Expected class $clsBlkMemberSym to have an object tag")
-
-                    // TODO (https://github.com/orgs/hkust-taco/projects/14/views/1?pane=issue&itemId=174476970):
-                    // replace with RTTI ancestry checks once each object carries runtime type information.
-                    val matchTags = ctx.getAllRuntimeTags(clsBlkMemberSym).getOrElse(LinkedHashSet(expectedTag))
-
                     val scrutExpr = getScrutExpr
-                    val isStructCompatible = ref.test(scrutExpr, baseObjectRefType(nullable = true))
+                    val isStructCompatible = ref.test(scrutExpr, baseObjectRefType(nullable = false))
+                    val scrutRtti = readObjectTypeInfo(scrutExpr)
```
Derppening: Nit: The `scrutExpr` is emitted twice. You should use a temp local to store the result of `scrutExpr` and reuse that.

Derppening (commented):
The rest of the PR looks good to me, but please address these comments.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
LPTK: Isn't this what we discussed in the group meeting?

Somehow, the corresponding https://github.com/hkust-taco/mlscript/pull/477#discussion_r3135219968 is marked as resolved 🤔

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
b-temirov: I changed, but still left `nullable = true` since `TypeInfoBase`'s  `$parent` is `null`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
b-temirov: I thought we discussed changing `Object.$$typeinfo` field type from `ref null any` to `ref null $TypeInfoBase`. I applied this change and resolved the comment in the earlier commit.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
b-temirov: I tried importing `typeinfo` structs, but this error came up:
```
//│ Wasm result:
//│ FAILURE LOCATION: mkQuery (WasmDiffMaker.scala:165)
//│ ═══[RUNTIME ERROR] LinkError: WebAssembly.instantiate(): Import #1 "repl" "Foo_typeinfo": JS object does not match expected wasm type
```
I felt like there was no immediate fix I could apply in the wasm backend? Correct me if I'm wrong. So, importing metadata and constructing locally was the workaround I used.

hkmc2/shared/src/test/mlscript/wasm/ClassInheritance.mls
```diff
@@ -226,15 +227,18 @@ B() is A
 //│               (local.set $matchRes
 //│                 (ref.null any))
 //│               (if
-//│                 (ref.test (ref $Object)
-//│                   (local.get $scrut))
+//│                 (block (result i32)
+//│                   (local.set $scrut1
+//│                     (local.get $scrut))
```
LPTK: Do you know why there's now an intermediate/duplicated `scrut1` variable?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
Derppening: What do the WATs of the exporting and importing modules look like?

b-temirov: For the exporting module:
```
object Foo with
	val x = 1
class C(val x)
fun inc(x) = x + 1
val step = 2
let counter = 1
let unit = ()
```
its WAT is:
```
//│ Wat:
//│ (module
//│      (type $TypeInfoBase (sub (struct (field $$tag i32) (field $$parent (ref null $TypeInfoBase)))))
//│      (type $Object (sub (struct (field $$typeinfo (mut (ref null $TypeInfoBase))))))
//│      (type $Foo_typeinfo (sub $TypeInfoBase (struct (field $$tag i32) (field $$parent (ref null $TypeInfoBase)))))
//│      (type $Foo (sub $Object (struct (field $$typeinfo (mut (ref null $TypeInfoBase))) (field $x (mut (ref null any))))))
...
//│      (global $unit (export "unit") (mut (ref null any)) (ref.null any))
//│      (global $Foo_typeinfo (export "Foo_typeinfo") (ref $Foo_typeinfo) (struct.new $Foo_typeinfo
//│          (i32.const 1)
//│          (ref.null $TypeInfoBase)))
...
```

For the importing module:
```
inc(Foo.x)
```
its output:
```
//│ Wat:
//│ (module
//│      (type $TypeInfoBase (sub (struct (field $$tag i32) (field $$parent (ref null $TypeInfoBase)))))
//│      (type $Object (sub (struct (field $$typeinfo (mut (ref null $TypeInfoBase))))))
//│      (type $Foo (sub $Object (struct (field $$typeinfo (mut (ref null $TypeInfoBase))) (field $x (mut (ref null any))))))
//│      (type $tmp (func (param $x (ref null any)) (result (ref null any))))
//│      (type $Foo_typeinfo (sub $TypeInfoBase (struct (field $$tag i32) (field $$parent (ref null $TypeInfoBase)))))
//│      (type $entry1 (func (result (ref null any))))
//│      (import "repl" "inc" (func $inc (type $tmp)))
//│      (import "repl" "Foo_typeinfo" (global $Foo_typeinfo (ref $Foo_typeinfo)))
//│      (import "repl" "Foo$inst" (global $Foo (mut (ref null $Foo))))
//│      (func $entry (export "entry") (type $entry1) (result (ref null any))
//│          (call $inc
//│                (struct.get $Foo $x
//│                     (ref.cast (ref $Foo)
//│                          (global.get $Foo)))))
//│      (elem $entry declare func $entry))
//│ Wasm result:
//│ FAILURE LOCATION: mkQuery (WasmDiffMaker.scala:165)
//│ ═══[RUNTIME ERROR] LinkError: WebAssembly.instantiate(): Import #1 "repl" "Foo_typeinfo": JS object does not match expected wasm type
```

Derppening: Claude seems to tell me that Wasm treats the `$Foo_typeinfo` type in the two modules as distinct types, which is why this doesn't link. There is a [type imports proposal](https://github.com/WebAssembly/proposal-type-imports/blob/main/proposals/type-imports/Overview.md) but it is currently in phase 1 of the feature stabilization process.

Could you try making the type of `$Foo_typeinfo` global `(ref any)` and insert downcasts as necessary, and see if that works?

LPTK: IIRC, in principle, as long as the two types have the exact same shape, they should be treated as equal?

LPTK: I note that TypeInfoBase appears to be recursive. Does that require special care? I remember seeing a "recursive" keyword in the spec

Derppening: > IIRC, in principle, as long as the two types have the exact same shape, they should be treated as equal?

What Claude was suggesting is that this is true within the same module, but not across modules.

> I note that TypeInfoBase appears to be recursive. Does that require special care? I remember seeing a "recursive" keyword in the spec

I think `rec` is only used if two types are recursive with respect to each other (e.g. `A` has a field of type `B` and vice versa).

b-temirov: Tried importing `C_typeinfo` as `ref null any`. Later a downcast to local `TypeInfoBase` is still failing.
Using the same exporting module as above, for the importing module:
```
:wat
if C(7) is C(v) then v else 0
...
(import "repl" "C_typeinfo" (global $C_typeinfo (ref null any)))
...
(local.set $currentTypeInfo
  (struct.get $Object $$typeinfo
    (ref.cast (ref $Object)
      (local.get $scrut))))
(local.set $targetTypeInfo
  (global.get $C_typeinfo))
...
(ref.eq
  (ref.cast (ref null eq)
    (local.get $currentTypeInfo))
  (ref.cast (ref null eq)
    (local.get $targetTypeInfo)))
...
(local.set $currentTypeInfo
  (struct.get $TypeInfoBase $$parent
    (ref.cast (ref $TypeInfoBase)
      (local.get $currentTypeInfo))))
//│ Wasm result:
//│ FAILURE LOCATION: mkQuery (WasmDiffMaker.scala:165)
//│ ═══[RUNTIME ERROR] RuntimeError: illegal cast
//│ at wasm://wasm/c38afb5e:wasm-function[1]:0xa2
//│ at REPL38:1:3896
//│ at lambda1 (file:///Users/bakhtiyartemirov/Desktop/FYP/mlscript/hkmc2/shared/src/test/mlscript-compile/wasm/Wasm.mjs:97:33)
//│ at Runtime.try_catch (file:///Users/bakhtiyartemirov/Desktop/FYP/mlscript/hkmc2/shared/src/test/mlscript-compile/RuntimeJS.mjs:15:18)
//│ at lambda (file:///Users/bakhtiyartemirov/Desktop/FYP/mlscript/hkmc2/shared/src/test/mlscript-compile/wasm/Wasm.mjs:102:39)
//│ at async REPL38:1:22
//│ at async node:repl:647:29
```

Derppening: The other way I can think of is to encapsulate the `typeinfo` instance into a global function, and store a `funcref` in the object header instead of the actual `typeinfo` value.

Derppening: @b-temirov I just tested this manually and it seems to work:

```
:silent
fun compileWatToWasm(wat) = 
    let mod = binaryen.parseText of wat
    mod.setFeatures(binaryen.Features.All)
    Predef.js_assert of mod.validate()
    val modBuf = mod.emitBinary()
    mod.dispose()
    modBuf
fun instantiateWasm(wasm, importObj) = 
    assert WebAssembly.validate of wasm
    let mod = new! WebAssembly.Module of wasm
    let inst = new! WebAssembly.Instance of mod, importObj
    inst
let wasm1 = compileWatToWasm of 
    """
    (module
        (type $Object (sub
            (struct
                (field $$tag i32))))
        (type $A 
            (sub $Object
                (struct
                    (field $$tag i32)
                    (field $i (mut i32)))))
        (global (export "a") (ref $A)
            (struct.new $A
                (i32.const 1)
                (i32.const 2))))
    """
let inst1 = instantiateWasm of wasm1, {}
let mod2 = compileWatToWasm of 
    """
    (module
        (type $Object (sub
            (struct
                (field $$tag i32))))
        (type $A 
            (sub $Object
                (struct
                    (field $$tag i32)
                    (field $i (mut i32)))))
        (import "mod1" "a" (global $a (ref $A)))
        (func (export "main") (result i32)
            (i32.add
                (struct.get $A $i
                    (global.get $a))
                (struct.get $A $$tag
                    (global.get $a)))))
    """
let inst2 = instantiateWasm of mod2, { mod1: inst1.exports }
inst2.exports.main() |> print
//│ > 3
```

If a standard inherited class example works, I don't see why `typeinfo` wouldn't work...

b-temirov: I just pushed a new commit experimenting with storing an accessor function to global `typeinfo`, which does work. I'll start working on this new proposed idea and will update soon.

Derppening: This works as well:

```
:silent
fun compileWatToWasm(wat) = 
    let mod = binaryen.parseText of wat
    mod.setFeatures(binaryen.Features.All)
    Predef.js_assert of mod.validate()
    val modBuf = mod.emitBinary()
    mod.dispose()
    modBuf
fun instantiateWasm(wasm, importObj) = 
    assert WebAssembly.validate of wasm
    let mod = new! WebAssembly.Module of wasm
    let inst = new! WebAssembly.Instance of mod, importObj
    inst
let wasm1 = compileWatToWasm of 
    """
    (module
        (type $Typeinfo 
            (sub
                (struct 
                    (field $$tag i32))))
        (type $Object 
            (sub
                (struct
                    (field $typeinfo (ref $Typeinfo)))))
        (type $A 
            (sub $Object
                (struct
                    (field $typeinfo (ref $Typeinfo))
                    (field $i (mut i32)))))
        (global (export "a") (ref $A)
            (struct.new $A
                (struct.new $Typeinfo 
                    (i32.const 1))
                (i32.const 2))))
    """
let inst1 = instantiateWasm of wasm1, {}
let mod2 = compileWatToWasm of 
    """
    (module
        (type $Typeinfo 
            (sub
                (struct 
                    (field $$tag i32))))
        (type $Object 
            (sub
                (struct
                    (field $typeinfo (ref $Typeinfo)))))
        (type $A 
            (sub $Object
                (struct
                    (field $typeinfo (ref $Typeinfo))
                    (field $i (mut i32)))))
        (import "mod1" "a" (global $a (ref $A)))
        (func (export "main") (result i32)
            (i32.add
                (struct.get $Typeinfo $$tag
                    (struct.get $A $typeinfo
                        (global.get $a)))
                (struct.get $A $i
                    (global.get $a)))))
    """
let inst2 = instantiateWasm of mod2, { mod1: inst1.exports }
inst2.exports.main() |> print
//│ > 3
```

I think this is the closest minimal example to your test case? So I don't think we actually need the accessor function. You should spend a bit more time on investigating exporting/importing globals directly.

hkmc2/shared/src/test/mlscript/wasm/ClassInheritance.mls
b-temirov: Should be removed now.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
b-temirov: After some painful merging with `hkmc2`, current changes contain the working version. Please review.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -2023,10 +2302,26 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+        ),
+      )),
+      objectTag = N,
+    ))
+
     ctx.addType(TypeInfo(
       sym = baseObjectSym,
-      StructType(Seq(tagFieldSym -> Field(I32Type, mutable = true, id = "$tag"))),
+      compType = StructType(Seq(
+        typeInfoFieldSym -> Field(RefType(typeInfoBaseTypeIdx, nullable = true), mutable = true, id = "$typeinfo"),
```
Derppening: I feel like I have given this comment before, but is there a case where the `$typeinfo` field is `ref.null`? If not, I don't see why the type needs to be `nullable = true`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -45,25 +45,133 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+  /** Returns `1` when `scrutTypeInfo` is equal to or descends from `targetTypeInfo`, else `0`. */
+  private def isSubtypeByTypeInfo(
+      scrutTypeInfo: Expr,
+      targetTypeInfo: Expr,
+  )(using Ctx, FunctionCtx, Raise): Expr =
+    val currentTmp = mkTempLocal("currentTypeInfo")
+    val targetTmp = mkTempLocal("targetTypeInfo")
+    val resultTmp = mkTempLocal("typeInfoMatch")
+    funcCtx.withLabel(LabelSymbol(N, "typeInfoEnd"), hasContinueLabel = false): endTarget =>
+      funcCtx.withLabel(LabelSymbol(N, "typeInfoLoop"), hasContinueLabel = true): loopTarget =>
```
Derppening: `funcCtx.withLabel` already handles loop-based branching:

```suggestion
    funcCtx.withLabel(LabelSymbol(N, "typeInfo"), hasContinueLabel = true): 
      case LabelTarget(breakLabel, S(continueLabel)) =>
```

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -261,6 +372,45 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
     ordered.toList
   end sortTopLevelClasses
 
+  /** Returns the elaborated source methods for this class. */
+  private def semanticMethodDefs(defn: ClsLikeDefn)(using Raise): List[TermDefinition] =
+    val clsDef = defn.isym.defn.get.asInstanceOf[hkmc2.semantics.ClassLikeDef]
```
Derppening: @LPTK Should this be replaced with pattern matching + `lastWords`?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -261,6 +372,45 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+  /** Computes one derived virtual-table layout for one top-level class. */
+  private def predeclareClassVirtualTable(defn: ClsLikeDefn)(using Ctx, Raise): Unit =
+    val parentVirtualTable = resolveParentSym(defn).flatMap(ctx.getVirtualTable)
+      .getOrElse(Ctx.VirtualTable(Nil, Map.empty))
+    val virtualMethods = ArrayBuf.from(parentVirtualTable.virtualMethods)
+    val virtualMethodSlots = LinkedHashMap.from(parentVirtualTable.virtualMethodSlots)
+
+    semanticMethodDefs(defn).foreach: methodDef =>
+      val slotIdx = virtualMethodSlots.iterator.collectFirst:
+        case (sym, idx) if sym.nme == methodDef.sym.nme => idx
```
Derppening: Matching by symbol name seems pretty fragile. Is there a way we can match by symbol?

hkmc2/shared/src/test/mlscript/wasm/VirtualMethods.mls
```diff
@@ -0,0 +1,176 @@
...
+
+
+class A(val a) with
+  virtual fun f(x) = x + a
+class B(val b) extends A(b) with
+  virtual fun f(x) = x * b
+class C(val c) extends B(c) with
+  fun f(x) = x - c
+C(3).f(5)
+//│ Wasm result:
```
Derppening, line 167: I think a test case where new virtual slots are added will be helpful as well.

Derppening (commented):
I think this is much better and closer to what we want. Please address the comments below and I think the rest LGTM.

One nit/note: After you address a comment, please respond to the feedback indicating that you did so, and preferably also attaching the commit hash you made the change in.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
LPTK: If it's a path that's supposedly impossible to trigger, `asInstanceOf` is clear enough (but would be better accompanied by a comment). If it may be triggered in weird situations (such as incorrect program), this should be a proper user-facing error.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
LPTK: It's not just fragile, it's plain wrong!

LPTK: We should have the symbol in the source IR. We just need to retain it until this point.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
b-temirov: My previous fix got accidentally reverted while merging, restored the fix c57561c.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
b-temirov: Applied the suggested change in c57561c.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
b-temirov: I'm not really sure if that path is impossible to trigger, so I replaced with pattern matching + `lastWords` in c57561c.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
b-temirov: Switched to matching by symbol in c57561c.

hkmc2/shared/src/test/mlscript/wasm/VirtualMethods.mls
b-temirov: Added a new test case in c57561c.

b-temirov:
Accidentally pushed the new commit with faulty commit message, so had to force push with corrected one. Addressed the outstanding comments.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -261,6 +379,62 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
     ordered.toList
   end sortTopLevelClasses
 
+  /** Returns the elaborated semantic class definition for this lowered class. */
+  private def semanticClassDef(defn: ClsLikeDefn)(using Raise): hkmc2.semantics.ClassLikeDef =
+    defn.isym.defn match
+      case S(clsDef: hkmc2.semantics.ClassLikeDef) => clsDef
+      case _ =>
+        lastWords(s"Expected lowered class `${defn.sym}` to retain its semantic class definition")
```
Derppening: Nit: The wording for the error message feels a bit convoluted.

```suggestion
        lastWords(s"Expected definition of class `${defn.sym}` to be present")
```

Derppening (approved):
One nit, otherwise LGTM. Thanks!

PR #478, opened 2026-04-24
wasm: Refactor `WatBuilder` to use per-index scopes

Derppening:
Each type of index now has its own index scope.

This PR has the side effect of making symbols mandatory for everything.


hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -87,8 +89,10 @@ final case class SessionGlobal(
   *   Additional symbols that should resolve to this class binding.
   */
 final case class SessionClass(
-    sym: BlockMemberSymbol,
-    typeInfo: TypeInfo,
+    sym: Symbol,
```
LPTK: Any reason to change from `BlockMemberSymbol`?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -458,68 +431,70 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
-        )
-        ctx.addFunctionImport(
-          S(func.sym),
-          WasmImport(
-            func.moduleName,
-            func.exportName,
-            ExternType.Func(SymIdx(funcName), TypeUse(typeIdx)),
-          ),
-        )
+        // If the function symbol comes from a class or module, generate a TempSymbol to avoid symbol collision with
```
LPTK, line 445: Won't they live in difference namespaces/scopes?

hkmc2/shared/src/test/mlscript/wasm/Basics.mls
```diff
@@ -275,26 +275,26 @@ O.y
...
 //│       (return
 //│         (local.get $this))))
-//│   (func $start1 (type $start)
+//│   (func $start (type $start)
 //│     (block
 //│       (global.set $O$inst
 //│         (ref.cast (ref null $O)
-//│           (call $O_ctor1)))))
-//│   (func $entry (export "entry") (type $entry1) (result (ref null any))
+//│           (call $O_O)))))
```
LPTK: The previous name seemed a lot more informative. Can we keep the `_ctor` suffix?

LPTK (approved):
Otherwise LGTM.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -2036,7 +2002,7 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
       wd: io.Path,
       sessionImports: Seq[SessionBinding],
       preservedSessionSymbols: Set[Local],
-  )(using Raise, Scope): CompiledWasmModule =
+  )(using Raise): CompiledWasmModule =
     for imprt <- p.imports do
```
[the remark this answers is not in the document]
LPTK: lol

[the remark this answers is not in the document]
Derppening: Copilot is stuck in Scala 2 land :cry:

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -155,22 +160,16 @@ final class SessionExportCtx(
 
   def freshCollector(): SessionExportCtx =
     SessionExportCtx(symbolsToExport, ArrayBuf.empty)
```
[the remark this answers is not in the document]
Derppening: Same here - Won't fix since Scala 3 does not have this issue.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
Derppening: Currently, constructors reuse the same symbol as its class. Therefore, we cannot use the same symbol for both the class (struct type) and the constructor (function type), and a `TempSymbol` is needed here.

LPTK: I still don't see why. We use different scopes for types and functions, don't we?

Derppening: Yes, but functions still need a type:

```
//│   (type $Foo (sub $Object (struct (field $$tag (mut i32)) (field $a (mut (ref null any))))))
//│   (type $Foo_init (func (param $this (ref null any)) (param $a (ref null any)) (result (ref null any))))
//│   (type $Foo_ctor (func (param $a (ref null any)) (result (ref null any))))
```

You can see that we have a type `$Foo` for the class and `$Foo_ctor` for the constructor. If we reuse the same symbol for the type of the class and type of the constructor, they will result in the same name `$Foo`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
```diff
@@ -257,71 +278,95 @@ case class MemUse(memidx: MemIdx) extends ToWat:
...
 /** An element segment entry. */
-sealed abstract class ElemSegment(val id: SymIdx, val elemlist: RefType -> Seq[Expr]) extends ToWat:
+sealed abstract class ElemSegment(
+    val elemlist: RefType -> Seq[Expr],
+    val sym: Symbol,
+    val idPrefix: Opt[Str],
+)(using Ctx, Raise) extends ToWat:
+
+  /** Symbolic identifier for the element segment. */
+  val id = SymIdx(summon[Ctx].dataSegmentScp.allocateOrGetNamePrefixed(sym, idPrefix))
```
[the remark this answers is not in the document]
Derppening: Good catch! Addressed in 160871e.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
Derppening: Seems not, reverted in cfebd41.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
LPTK: Ok, I see.

But I really dislike the special-casing behavior, here, which is just making things harder to test and reason about (is the condition `func.sym.asClsOrMod.isDefined` really appropriate, and will it remain so in the future?) for no reason. Why not always use the `TempSymbol`?

hkmc2/shared/src/test/mlscript/wasm/Basics.mls
Derppening: Added the ability to specify both the prefix and suffix of a WAT identifier in [6fa8a4b](https://github.com/hkust-taco/mlscript/pull/478/commits/6fa8a4b7a6f188ad823f2543840640e0abd214ac).

I'd appreciate some feedback on this change, thanks!

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
Derppening: Good point. I have refactored all function type symbols to be `TempSymbol`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -194,8 +193,18 @@ class FuncInfo(
...
 
+  def this(
+    sym: BlockMemberSymbol | TempSymbol,
+    typeUse: TypeUse,
+    params: Seq[Local -> SymIdx],
+    resultTypes: Seq[Result],
+    locals: Seq[Local -> SymIdx],
+    body: Expr,
+    exportName: Opt[Str],
+  )(using Ctx, Raise) = this(sym, N -> N, typeUse, params, resultTypes, locals, body, exportName)
```
LPTK: Are these 9 lines of code really worth it? How hard is it to just pass `N -> N` at call sites?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
```diff
@@ -254,12 +265,19 @@ end GlobalInfo
...
-  *   Symbolic identifier for the memory.
+  * @param sym
+  *   The source [[Symbol]] which this memory is generated from.
   * @param memType
   *   The type of the memory.
+  * @param wrapId
+  *   An pair of optional strings for adding a prefix and suffix to the generated identifier of this memory.
   */
-class MemInfo(val id: SymIdx, val memType: MemType) extends ToWat:
+class MemInfo(val sym: Symbol, val memType: MemType, val wrapId: Opt[Str] -> Opt[Str])(using Ctx, Raise) extends ToWat:
```
LPTK: We can certainly make it more specific than `Symbol`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
LPTK: Same remarks for all the other special constructors.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
```diff
@@ -278,94 +292,118 @@ case class MemUse(memidx: MemIdx) extends ToWat:
...
-
+      new Active(offset, Seq(bytes), memuse, sym, wrapId)
+    
+    def apply(
+        offset: Expr,
+        bytes: Seq[Str],
+        memuse: Opt[MemUse],
+        sym: Symbol,
+    )(using Ctx, Raise): Active =
+      new Active(offset, bytes, memuse, sym, N -> N)
```
LPTK: The latter `apply` is unused. The former is used ONCE and STILL has a default argument. Both are pure cruft and should be removed, unless I'm missing something.

Can you please try not to pile on tons of cruft like this and try to keep things as minimal as reasonable? (KISS principle)

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
LPTK: In fact, this is probably one of the few cases where a default argument is defensible, as it doesn't affect the semantics (so forgetting it is harmless).

hkmc2/shared/src/test/mlscript/wasm/Basics.mls
LPTK: Feedback granted.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
Derppening: Folded secondary constructors and `apply` functions that wraps `wrapId = N -> N` into default arguments in [beda649](https://github.com/hkust-taco/mlscript/pull/478/commits/beda649046bccd685398899fa906684cb38503f3).

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
Derppening: Removed all redundant `apply` functions and always require `Seq(bytes)` to simplify segment construction in [beda649](https://github.com/hkust-taco/mlscript/pull/478/commits/beda649046bccd685398899fa906684cb38503f3).

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Ctx.scala
Derppening: Made `MemInfo` only accept `TempSymbol` in [23ffee9](https://github.com/hkust-taco/mlscript/pull/478/commits/23ffee914af485bcf82cdc79f0766f5d8cbc9f85).

PR #480, opened 2026-04-25
Fix specialized handling for method with 2 argument

AnsonYeung:
There is a duplicated code path and it caused mismatch in codegen for function with 2 arguments.


hkmc2/shared/src/test/mlscript/backlog/ToTriage.mls
LPTK: This is fixed; pls move it as a test case to another test file.

LPTK (approved):
Oh, nice catch, thanks!

I'd never noticed that those calls that inexplicable missed `safeCall` were those with exactly two arguments due to this...

PR #481, opened 2026-04-25
Add basic intraprocedural flow-sensitive analysis for optimization

LPTK:
For now, this is mostly for propagating transitive variable assignments and folding constants. We can later fully leverage the results to simplify pattern matching expressions, remove dead assignments, etc.


LPTK:
I still have to clean up the OptScratch files, but other than that, the PR is ready to review (not commit by commit, obviously).

CAG2Mark:
I can have a look in the coming day or two.

hkmc2/jvm/src/test/scala/hkmc2/TimeOutTests.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  var testName: Str = "‹unknown›"
+  override def withFixture(test: NoArgTest) =
+    testName = test.name
+    super.withFixture(test)
+  override val defaultTestSignaler: Signaler = new Signaler:
+    @annotation.nowarn("msg=method stop in class Thread is deprecated") def apply(testThread: Thread): Unit =
+      println(s"[${Thread.currentThread().getName}] running ${testThread.getName} has run out of time!")
+      println(s"!! Test $testName at $testThread has run out out time !! stopping..." +
+        "\n\tNote: you can increase this limit by changing DiffTests.TimeLimit")
```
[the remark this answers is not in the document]
LPTK: This is not actually a problem, because ScalaTest creates one instance of the class per test that runs in parallel.

CAG2Mark:
I've just done a high-level review of the new `DataFlowAnalysis` and the changes seem reasonable at first glance. I need to work on some other things, so I will take a closer look at the details tomorrow.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -3,42 +3,105 @@ package codegen
...
+class BlockSimplifier
+    (symbolsToPreserve: Set[Local], tl: TL, printer: Program => Str)
+    (using DebugPrinter, State, Config, Raise, Ctx):
+  import tl.*
   
   
-  private var changed = true
+  val deadBranchRemoval = config.deadBranchRemoval
+  
+  val MaxIterations = 10
```
AnsonYeung, line 30: Why is this variable name capitalized

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -146,7 +209,7 @@ class BlockSimplifier(symbolsToPreserve: Set[Local])(using DebugPrinter, State,
     override def applyValue(v: Value)(k: Value => Block) = v match
       // * Replace with `undefined` those references to local variables that are never assigned
       case Value.Ref(loc, N) if localVars.contains(loc) && !definedVars.contains(loc) =>
-        registerChange
+        registerChange("TODO: description")
```
AnsonYeung: Seems there are many TODOs in registerChange

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -228,9 +291,472 @@ class BlockSimplifier(symbolsToPreserve: Set[Local])(using DebugPrinter, State,
...
+        case Merge(a1, a2) => s"{${a1.toString} | ${a2.toString}}"
+      
+      def merge(that: AssignInfo): AssignInfo =
+        if this is that then this
+        else that match
+          case Unknown => that
+          case Uninitialized => this
+          // case Merge(l, r) => Merge(merge(this, l), r)
+          case _: Assigned | _: Merge => Merge(this, that)
+
```
AnsonYeung, line 392: If `this` is `Unknown | Uninitialized` then a `Merge` will be created

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -228,9 +291,472 @@ class BlockSimplifier(symbolsToPreserve: Set[Local])(using DebugPrinter, State,
...
+    var inDryRun = false // for traversing loop bodies once before actually transforming the program
+    
+    def withFreshAssignedResults[T](thunk: => T): T =
+      val oldAssignedResults = assignedResults
+      assignedResults = emptyAssignedResults
+      val res = thunk
+      assignedResults = oldAssignedResults
+      res
+    
+    val atLabelBegin: MutMap[LabelSymbol, AssignedResults] = MutMap.empty.withDefaultValue(emptyAssignedResults)
```
AnsonYeung, line 418: It seems to me that atLabelBegin is effectively unused, the only usage only put the result back into the same map

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -228,9 +291,472 @@ class BlockSimplifier(symbolsToPreserve: Set[Local])(using DebugPrinter, State,
...
+              false
+            pat match
+            case Case.Lit(lit) => 
+              shapes.contains(lit) && { shapes -= lit; true } || regChange
+            case Case.Cls(sym, _) =>
+              
+              // FIXME: take inheritance into account
+              // FIXME: take lit <: virual-cls into account (such as true <: Bool)
+              
+              shapes.contains(sym) && { shapes -= sym; true } || regChange
```
AnsonYeung:
```suggestion
            @inline def regChange(reason: Str) =
              registerChange(s"Arm ${pat.showDbg} is unreachable. Reason: ${reason}")
              false
            pat match
            case Case.Lit(lit) => 
              shapes.contains(lit) && { shapes -= lit; true } || regChange("Impossible literal")
            case Case.Cls(sym, _) =>
              
              // FIXME: take inheritance into account
              // FIXME: take lit <: virual-cls into account (such as true <: Bool)
              
              shapes.contains(sym) && { shapes -= sym; true } || regChange("Impossible instanceof")
```

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -228,9 +291,472 @@ class BlockSimplifier(symbolsToPreserve: Set[Local])(using DebugPrinter, State,
...
+      case Scoped(syms, body) =>
+        syms.foreach:
+          case sym: LocalVar =>
+            assignedResults += sym -> Uninitialized
+          case _ =>
+        val res = super.applyScopedBlock(b)
+        syms.foreach:
+          case sym: LocalVar =>
+            assignedResults += sym -> Uninitialized
+          case _ =>
```
AnsonYeung, line 672: Seems unnecessary to set it back to uninitialized but it should not be used outside of the scope.

hkmc2/shared/src/test/mlscript/basics/DynamicSelection.mls
```diff
@@ -80,14 +80,11 @@ t.01
 //│ ╔══[WARNING] Pure expression in statement position
 //│ ║  l.79: 	!(1,2)
 //│ ╙──      	  ^
-//│ > try { block$res15 = 2.value; undefined; undefined } catch (e) { console.log('\u200B' + e + '\u200B'); }
-//│ >                     ^^
-//│ ═══[COMPILATION ERROR] [Uncaught SyntaxError] Invalid or unexpected token
```
AnsonYeung, line 85: Is this change intended?

hkmc2/shared/src/test/mlscript/opt/OptScratch.mls
```diff
@@ -0,0 +1,39 @@
+:js
```
AnsonYeung, line 1: Seems there are 3 scratch files left behind

hkmc2/shared/src/test/mlscript/opt/OptScratch2.mls
```diff
@@ -0,0 +1,108 @@
+:js
```
AnsonYeung: scratch file 2

hkmc2/shared/src/test/mlscript/ucs/patterns/Parameters.mls
```diff
@@ -180,7 +180,6 @@ pattern AtPrec =
 
 fun f(x) = if x is str as AtPrec as prec then [str, prec]
 f("+")
-//│ = ["+", 2]
```
AnsonYeung, line 183: This seems like some weird diff

hkmc2/shared/src/test/mlscript/ups/examples/BadRecStackParse.mls
```diff
@@ -58,7 +58,6 @@ parse("a" :: "+" :: "(" :: "x" :: "+" :: "y" :: ")" :: Nil)
 
 // * Notice: wrong assoc.
 parse("a" :: "+" :: "(" :: "b" :: "+" :: "x" :: "+" :: "y" :: ")" :: Nil)
-//│ = {Success: Op(Id("a"), "+", Op(Id("b"), "+", Op(Id("x"), "+", Id("y"))))}
```
AnsonYeung, line 61: weird diff

hkmc2/shared/src/test/mlscript/ups/examples/PrecedenceClimbStackParse.mls
```diff
@@ -22,7 +22,6 @@ pattern AtPrec(str, prec) = str as (
 
 fun f(x) = if x is AtPrec(op, p) then [op, p]
 f("+")
-//│ = ["+", 5]
 f("*")
```
AnsonYeung, line 26: weird

hkmc2/shared/src/test/mlscript/ups/examples/PrecedenceClimbStackParse.mls
```diff
@@ -133,18 +132,17 @@ fun f(x) =
   [str, prec]
 
 f("+")
-//│ = ["+", 5]
```
AnsonYeung, line 136: same as above

hkmc2/shared/src/test/mlscript/ups/RecursiveTransformations.mls
```diff
@@ -51,27 +51,23 @@ pattern MultiSet(pattern Head, pattern Rest) =
...
 arr is MultiSet(1, _)
-//│ = true
 arr is MultiSet(0, _)
-//│ = false
 arr is MultiSet(3, _)
-//│ = true
 arr is MultiSet(3, MultiSet(1, MultiSet(2, Nil)))
-//│ = true
 arr is MultiSet(3, MultiSet(0, _))
 //│ = false
```
AnsonYeung, line 62: seems like ups use this broken style of test a lot which breaks on this pr a lot

AnsonYeung (commented):
I think the loop logic is missing a merge for the loop start.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
LPTK: It's absolutely necessary. That's what prevents it from being picked up by transitive assignment propagation. Should be better documented.

hkmc2/shared/src/test/mlscript/basics/DynamicSelection.mls
LPTK: Yes, it's a drive-by fix.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
LPTK: It's a convention for constants in Scala (where in, eg, Java, one would write `MAX_ITERATIONS`).

hkmc2/shared/src/test/mlscript/ups/RecursiveTransformations.mls
LPTK: This annoying behavior should be fixed by Copilot now: https://github.com/hkust-taco/mlscript/pull/488

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
LPTK: Solved starting from https://github.com/hkust-taco/mlscript/pull/481/changes/4860bb0bb82339346e9c3b90919322f5709b79b1 and follow-on commits.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -488,11 +516,18 @@ class BlockSimplifier
...
       case Break(label) =>
+        // TODO: this is probably only needed when `!inDryRun`?
         atLabelEnd.put(label, merge(assignedResults, atLabelEnd(label)))
-        super.applyBlock(b)
+        makeImpossibleAfter:
+          super.applyBlock(b)
+        
+      case _: Return =>
+        makeImpossibleAfter:
+          super.applyBlock(b)
```
AnsonYeung: We can have this case for `Unreachable` and `Throw` as well

hkmc2/shared/src/test/mlscript/opt/CapturedVariableMutation.mls
```diff
@@ -0,0 +1,38 @@
...
+
+
+:noInline
+let emptyItemCount = 1
+fun next() =
+  let emptyItemCount' = emptyItemCount
+  set emptyItemCount = 0
+  print(emptyItemCount')
+//│ emptyItemCount = 1
+
```
AnsonYeung, line 39: the next() function is not tested here

hkmc2/shared/src/test/mlscript/ups/RecursiveTransformations.mls
```diff
@@ -59,19 +59,20 @@ arr is MultiSet(3, _)
 arr is MultiSet(3, MultiSet(1, MultiSet(2, Nil)))
 //│ = true
 arr is MultiSet(3, MultiSet(0, _))
+arr is MultiSet(3, MultiSet(0, _))
 //│ = false
```
AnsonYeung, line 62:
```suggestion
arr is MultiSet(3, MultiSet(1, MultiSet(2, Nil)))
//│ = true
arr is MultiSet(3, MultiSet(0, _))
//│ = false
```

AnsonYeung (approved):

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
AnsonYeung: Actually Throw and Unreachable may interact with TryFinally so it may cause some problems...

If it's in fact a problem, then a function containing a throw statement could in theory do the same thing, so it's more of a `TryFinally` problem than Throw/Unreachable

LPTK: Good catch! The tests pass with the two additional cases, but I also added a conservative handling of `Try` blocks, as we can't assume that the code in `finally` executes right after the code in the `try` body. Unfortunately, this removes some nice optimizations in the tests, but we could get them back by trying to remove `try`/`finally` when it's not really needed (such as purely local or pure control-flow).

PR #461, opened 2026-04-16
Final adjustments for dynamic staging

ChingLongTin:
Some of the changes here will be covered by a future merge of the `shape` branch.

PR #482, opened 2026-04-29
Support `import "filename" as ModuleName` syntax

NeilKleistGao:
Can be used in #461.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
@@ -1086,7 +1086,7 @@ extends Importer with ucs.SplitElaborator:
             go(sts, Nil, acc)
       case (m @ PrefixApp(Keywrd(Keyword.`import`), arg)) :: sts =>
         reportUnusedAnnotations
-        val pathAndAlias = arg match
+        val pathAndAlias: Opt[(Tree, Opt[Ident])] = arg match
           case InfixApp(pathArg, Keywrd(Keyword.`as`), alias: Ident) => S((pathArg, S(alias)))
           case InfixApp(pathArg, Keywrd(Keyword.`as`), Error()) => N
           case InfixApp(_, Keywrd(Keyword.`as`), badAlias) =>
--- a/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
@@ -1086,15 +1086,25 @@ extends Importer with ucs.SplitElaborator:
             go(sts, Nil, acc)
       case (m @ PrefixApp(Keywrd(Keyword.`import`), arg)) :: sts =>
         reportUnusedAnnotations
-        val (newCtx, newAcc) = arg match
-          case StrLit(path) =>
-            val stmt = importPath(path).withLocOf(m)
+        val parsedImport = arg match
+          case InfixApp(pathArg, Keywrd(Keyword.`as`), alias: Ident) => S((pathArg, S(alias)))
+          case InfixApp(_, Keywrd(Keyword.`as`), badAlias) =>
+            raise(ErrorReport(
+              msg"Expected identifier after 'as' in import statement" ->
+              badAlias.toLoc :: Nil))
+            N
+          case pathArg => S((pathArg, N))
+        val (newCtx, newAcc) = parsedImport match
+          case S((StrLit(path), alias)) =>
+            val stmt = importPath(path, alias).withLocOf(m)
             (ctx + (stmt.sym.nme -> stmt.sym),
-            stmt :: acc)
-          case _ =>
+              stmt :: acc)
+          case S((pathArg, _)) =>
             raise(ErrorReport(
               msg"Expected string literal after 'import' keyword" ->
-              arg.toLoc :: Nil))
+              pathArg.toLoc :: Nil))
+            (ctx, acc)
+          case _ => // errors have been reported above.
             (ctx, acc)
         newCtx.givenIn:
           go(sts, Nil, newAcc)
--- a/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
@@ -1086,7 +1086,7 @@ extends Importer with ucs.SplitElaborator:
             go(sts, Nil, acc)
       case (m @ PrefixApp(Keywrd(Keyword.`import`), arg)) :: sts =>
         reportUnusedAnnotations
-        val parsedImport = arg match
+        val pathAndAlias = arg match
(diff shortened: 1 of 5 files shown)


hkmc2/shared/src/main/scala/hkmc2/syntax/ParseRule.scala
```diff
@@ -387,7 +387,22 @@ class ParseRules(using State):
...
     prefixed(`new!`),
     prefixed(`throw`),
-    prefixed(`import`), // TODO improve – only allow strings
+    keepKw(`import`):
+      ParseRule("'import' keyword"):
+        Expr(
+          ParseRule("'import' path")(
+            keepKw(`as`)(
+              ParseRule("'import' alias"):
+                Expr(ParseRule("'import' alias name")(end(()))):
```
LPTK: This should probably be something like
```suggestion
              ParseRule("ímport'-'as' keywords"):
                Expr(ParseRule("'import' alias")(end(()))):
```
You should add some `:pe` parsing test cases triggering this to make sure it looks right.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1086,15 +1086,25 @@ extends Importer with ucs.SplitElaborator:
             go(sts, Nil, acc)
       case (m @ PrefixApp(Keywrd(Keyword.`import`), arg)) :: sts =>
         reportUnusedAnnotations
-        val (newCtx, newAcc) = arg match
-          case StrLit(path) =>
-            val stmt = importPath(path).withLocOf(m)
+        val parsedImport = arg match
+          case InfixApp(pathArg, Keywrd(Keyword.`as`), alias: Ident) => S((pathArg, S(alias)))
```
LPTK, line 1090: Wait why did you change the parser if this is actually how things parse? Is the parser change irrelevant (just adding dead code)?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1086,15 +1086,25 @@ extends Importer with ucs.SplitElaborator:
...
-            stmt :: acc)
-          case _ =>
+              stmt :: acc)
+          case S((pathArg, _)) =>
             raise(ErrorReport(
               msg"Expected string literal after 'import' keyword" ->
-              arg.toLoc :: Nil))
+              pathArg.toLoc :: Nil))
+            (ctx, acc)
+          case _ => // errors have been reported above.
```
LPTK: Please explicitly list the pattern shapes that flow here.

LPTK (changes requested):
There's redundancy in the logic. Can't you just remove the parser changes? You need to add more negative test cases.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
NeilKleistGao: The parser change is necessary, and this is not a good variable name. I've changed it.

hkmc2/shared/src/main/scala/hkmc2/syntax/ParseRule.scala
```diff
@@ -387,7 +387,22 @@ class ParseRules(using State):
     prefixed(`not`),
     prefixed(`new!`),
     prefixed(`throw`),
-    prefixed(`import`), // TODO improve – only allow strings
+    keepKw(`import`):
+      ParseRule("'import' keyword"):
+        Expr(
+          ParseRule("'import' path")(
+            keepKw(`as`)(
+              ParseRule("ímport'-'as' keywords"):
```
LPTK:
```suggestion
              ParseRule("'import'-'as' keywords"):
```

hkmc2/shared/src/main/scala/hkmc2/syntax/ParseRule.scala
```diff
@@ -387,7 +387,22 @@ class ParseRules(using State):
...
+                Expr(ParseRule("'import' alias")(end(()))):
+                  case (alias, ()) => alias
+            ).map(S(_)),
+            end(N),
+          )
+        ) { case (path, alias) => path -> alias }
+    .map:
+      case (kw, (path, S((asKw, alias)))) => PrefixApp(kw, InfixApp(path, asKw, alias))
```
LPTK: Oh, I see that you're actually creating an `InfixApp` yourself here... Im not sure why this is even necessary. With the right precedences, shouldn't `import X as Y` already parse as `import (X as Y)`, meaning no parse rule actually needs to change?

NeilKleistGao: The original parse rule would yield `InfixApp(PrefixApp(Keywrd(keyword 'import'),StrLit(../../mlscript-compile/Example.mls)),Keywrd(keyword 'as'),Ident(AnotherExample))`, i.e., `(import X) as Y`.

LPTK: As I said, it's a precedence issue... I have pushed the precedence fix for you.

LPTK (approved):
Otherwise LGTM.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1086,15 +1086,25 @@ extends Importer with ucs.SplitElaborator:
...
-            val stmt = importPath(path).withLocOf(m)
+        val pathAndAlias = arg match
+          case InfixApp(pathArg, Keywrd(Keyword.`as`), alias: Ident) => S((pathArg, S(alias)))
+          case InfixApp(_, Keywrd(Keyword.`as`), badAlias) =>
+            raise(ErrorReport(
+              msg"Expected identifier after 'as' in import statement" ->
+              badAlias.toLoc :: Nil))
+            N
+          case pathArg => S((pathArg, N))
+        val (newCtx, newAcc) = pathAndAlias match
```
LPTK, line 1098:
```
match may not be exhaustive.

It would fail on pattern case: mlscript.utils.shorthands.S((hkmc2.syntax.Tree, Some[hkmc2.syntax.Tree.Ident]) | (hkmc2.syntax.Tree,
  None.type)(_, _))
```

LPTK (changes requested):

hkmc2/shared/src/main/scala/hkmc2/syntax/ParseRule.scala
NeilKleistGao: Thanks.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
NeilKleistGao: It should be exhaustive with `pathAndAlias: Opt[(Tree, Opt[Ident])]`.

LPTK (approved):
LGTM, thanks!

PR #483, opened 2026-04-29
Fix missing configs for `Resolver` in compile test

NeilKleistGao:
Continuation of #473

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/MLsCompiler.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/MLsCompiler.scala
@@ -90,7 +90,8 @@ class MLsCompiler
       val elab = Elaborator(etl, wd, newCtx)
       val parsed = mainParse.resultBlk
       val (blk0, _) = elab.importFrom(parsed)
-      val resolver = Resolver(rtl)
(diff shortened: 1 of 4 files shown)


hkmc2/shared/src/main/scala/hkmc2/MLsCompiler.scala
```diff
@@ -90,7 +90,8 @@ class MLsCompiler
       val elab = Elaborator(etl, wd, newCtx)
       val parsed = mainParse.resultBlk
       val (blk0, _) = elab.importFrom(parsed)
-      val resolver = Resolver(rtl)
+      val effectiveConfig = Config.extractConfigFromStats(blk0)
+      val resolver = Resolver(rtl)(using summon[Raise], summon[State], summon[Ctx], effectiveConfig)
```
LPTK: This is an anti-pattern, as it makes it very easy to forget passing the correct `Config` in the below code.

The proper way is to use

```suggestion
      Config.extractConfigFromStats(blk0).givenIn:
        val resolver = Resolver(rtl)
        // ...
```

LPTK (changes requested):

LPTK (approved):
LGTM, thx!

PR #484, opened 2026-05-04
wasm: `binaryen` dependency update, add `WebAssembly.Instance` to `Prelude`

Derppening:
The change to add `WebAssembly.Instance` to `Prelude.mls` is for synchronous instantiation of Wasm modules, which is necessary if we want to test loading and executing arbitrary Wasm modules in `HkScratch.mls`.

Example:

```
:silent
fun compileWatToWasm(wat) = 
    let mod = binaryen.parseText of wat
    mod.setFeatures(binaryen.Features.All)
    Predef.js_assert of mod.validate()
    val modBuf = mod.emitBinary()
    mod.dispose()
    modBuf
fun instantiateWasm(wasm, importObj) = 
    assert WebAssembly.validate of wasm
    let mod = new! WebAssembly.Module of wasm
    // Note: `WebAssembly.Instance` needed here
    let inst = new! WebAssembly.Instance of mod, importObj
    inst
let mod = compileWatToWasm of 
    """
    (module
        (func (export "main") (result i32)
            (i32.const 42)))
    """
let inst = instantiateWasm of mod, {}
inst.exports.main() |> print
//│ > 42
```

No functional changes.

### DIFF
--- a/hkmc2/shared/src/test/mlscript/decls/Prelude.mls
+++ b/hkmc2/shared/src/test/mlscript/decls/Prelude.mls
@@ -216,6 +216,7 @@ declare val Infinity
 declare class Promise
 declare val Promise
 declare object WebAssembly with
+  declare object Instance
   declare class Memory
   declare object Module with
     fun
--- a/hkmc2/shared/src/test/mlscript/wasm/Binaryen.mls
+++ b/hkmc2/shared/src/test/mlscript/wasm/Binaryen.mls
@@ -1,0 +1,99 @@
+:js
+:silent
+
+
+// Helper functions
+import "binaryen"
+fun compileWatToWasm(wat) = 
+  let mod = binaryen.parseText of wat
+  mod.setFeatures(binaryen.Features.All)
+  Predef.js_assert of mod.validate()
+  val modBuf = mod.emitBinary()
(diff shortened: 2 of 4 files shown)


LPTK (approved):
Thanks! But why not include your example as a diff-test file?

Derppening:
> Thanks! But why not include your example as a diff-test file?

Added examples as a diff-test file. One example parses WAT to generate a binary and executes it, the other generates a Wasm binary using Binaryen APIs.

I think the MLscript syntax I used probably isn't the best... I would love some feedback on this.

LPTK:
> I think the MLscript syntax I used probably isn't the best... I would love some feedback on this.

Sure. I pushed what are arguably code style improvements. But TBH this is mostly due to coding style (I prefer not using any unnecessary commas and not using lines that just close brackets), and some might find it harder to read 🤷‍♂️ 

FYI, my hierarchy of bad:

```
// bad
foo(
  a,
  b,
  c
)

// acceptable (but meh)
foo(
  a,
  b,
  c,
)

// nice (but not possible in Scala)
foo(
  a
  b
  c
)

// best (but not possible in Scala)
foo of
  a
  b
  c
```

PR #485, opened 2026-05-06
Add linearity and accumulator tracking in flow analysis

ychenfo:
- Add affinity and accumulator tracking in flow analysis
- Refactor flow analysis strategy types
- Update flow analysis related config
- Pretty print deforest result


hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Deforest.scala
```diff
@@ -7,20 +7,36 @@ import mlscript.utils.*, shorthands.*
...
   given fState: FlowAnalysis.State = constraintSolver.fState
   given eState: Elaborator.State = constraintSolver.eState
   given tl: TraceLogger = constraintSolver.tl
+  given Raise = preAnalyzer.raise
+
+  private def pp(id: CtorDtorId): Str =
+    val scope = Scope.empty(Scope.Cfg.default)
+    given Scope = scope
+    given ShowCfg = ShowCfg.internal
+    given SymbolPrinter = new SymbolPrinter(scope)
```
LPTK: You shouldn't create your own printer; you should use the one provided here

https://github.com/hkust-taco/mlscript/blob/fe64200c0ae211d26140eedfacb60c8bfb2047c1/hkmc2DiffTests/src/test/scala/hkmc2/DiffMaker.scala#L53

ie, just make `pp` accept it implicitly.

This way, the symbols will be printed consistently with things like `:sir` and `:lot`.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -167,43 +196,65 @@ trait StratVar(s: StratVarState):
...
 sealed abstract class ProdStrat
+sealed abstract class ConsStrat
+
+sealed trait StratWithOrigin[A <: OriginId]:
+  def exprId: A
+  def instantiationId: Opt[InstantiationId]
+  def concreteId: ConcreteId[A] = ConcreteId(exprId, instantiationId.get)
+
+sealed trait MarkerProdStrat extends ProdStrat
+sealed trait MarkerConsStrat extends ConsStrat
```
LPTK, line 209: What does this stand for? the name is weird. Would be good to at least have one line of doc.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -167,43 +196,65 @@ trait StratVar(s: StratVarState):
...
 )(
   val params: Ls[ProdStrat],
   val res: ConsStrat
-) extends ConsStrat:
+) extends ConsStrat with StratWithOrigin[ResultId]:
   override def toString(): String =
     s"(${params.map(_.toString()).mkString(", ")}) -> ${res.toString()}"
 
-case object NoCons extends ConsStrat
+case object NoCons extends MarkerConsStrat
```
LPTK: Maybe now is a good time to rename `NoCons` to `UnknownCons` and `NoProd` to `UnknownProd`?

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Deforest.scala
```diff
@@ -120,7 +136,13 @@ object Deforest:
     eState: Elaborator.State,
   ): Program =
     // TODO: handle see through imported modules
-    val flowAnalysisRes = FlowAnalysis(p, mono = cfg.deforest.exists(_.mono))
+    val dCfg = cfg.deforest.get
```
LPTK:
```suggestion
    val dCfg = cfg.deforest.getOrElse(lastWords("deforestation is disabled in Config"))
```

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -12,9 +12,14 @@ import scala.collection.mutable.{Set as MutSet, Map as MutMap, LinkedHashMap, Li
 
 
 object FlowAnalysis:
+  object TraceScope:
+    val nonAffineSyms = "flow-analysis/non-affine"
+    val accumulatorSyms = "flow-analysis/accumulator"
```
LPTK: Capitalize such constants (Scala convention).

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -350,26 +430,80 @@ class FlowPreAnalyzer(val pgrm: Program)(using
...
   end ctxTracker
   
+  private def recordAffinityUse(l: Symbol, disamb: Opt[DefinitionSymbol[?]]): Unit =
+    def isClassCtorSym(sym: Symbol): Bool =
+      sym.asTrm.exists: tSym =>
+        tSym.owner
+          .flatMap(_.asCls)
+          .flatMap(_.defn)
+          .flatMap(_.ctorSym)
+          .exists(_ is tSym)
```
LPTK: Lol that's a really crazy hack. So much work to avoid doing the right thing – which would be to just add that info to the TermSymbol (you could make a subclass of TermSymbol called CtorSymbol).

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -239,12 +290,21 @@ class FlowPreAnalyzer(val pgrm: Program)(using
   given stratVarUidState: Uid.StratVar.State = new Uid.StratVar.State
   import StratVarState.freshVar
   
+  private case class CaptureTrackingInfo(
```
LPTK, line 300: Please document what this does and why it's needed.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -350,26 +430,80 @@ class FlowPreAnalyzer(val pgrm: Program)(using
   
   end ctxTracker
   
+  private def recordAffinityUse(l: Symbol, disamb: Opt[DefinitionSymbol[?]]): Unit =
```
LPTK: Can we make the type of `l` more precise? I'm pretty sure it doesn't need to be so general.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -350,26 +430,80 @@ class FlowPreAnalyzer(val pgrm: Program)(using
...
+    def isClassCtorSym(sym: Symbol): Bool =
+      sym.asTrm.exists: tSym =>
+        tSym.owner
+          .flatMap(_.asCls)
+          .flatMap(_.defn)
+          .flatMap(_.ctorSym)
+          .exists(_ is tSym)
+    val processedSymbol =
+      disamb.getOrElse(l) match
```
LPTK: When does it ever make sense to use `l` instead, here? You should avoid confused logic like this. The way the symbols are used is specific and consistent. The logic for handling symbols from disamb cannot meaningfully be mixed with the logic for the symbol in a Ref.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -428,21 +563,43 @@ class FlowPreAnalyzer(val pgrm: Program)(using
       applyPath(qual); applyPath(fld)
     case p@TrackableFieldSelect(qual, _ -> _) =>
         res.selToCtxOfSel.addOne(p.uid -> ctxTracker.getAllCtx)
-        applyPath(qual)
+        qual match
+          case Value.Ref(l, disamb)
```
LPTK: No need for indent.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -541,7 +703,8 @@ class FlowConstraintsCollector(val preAnalyzer: FlowPreAnalyzer, val mono: Bool)
 
       // compute strat scheme for each scc group
       for groupedFuns <- sccInOrder do
-        (new ConstraintsCollector(Some(funToSccRep(groupedFuns.head).get))).givenIn: cc ?=>
+        val groupRep = funToSccRep(groupedFuns.head).get
+        (new ConstraintsCollector(Some(groupRep))).givenIn: cc ?=>
```
LPTK:
```suggestion
        new ConstraintsCollector(Some(groupRep)).givenIn: cc ?=>
```

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -561,15 +730,25 @@ class FlowConstraintsCollector(val preAnalyzer: FlowPreAnalyzer, val mono: Bool)
       cc.constrain(NoProd, preAnalyzer.res.primitiveStratVar.asConsStrat)
       processBlock(preAnalyzer.pgrm.main)(using cc, NoCons)
 
+      // emit NonAffine for non-affine syms that don't belong to any scc:
```
LPTK: Can you explain the reasoning behind "don't belong to any scc" => "emit NonAffine"?

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -845,85 +1035,133 @@ class FlowConstraintSolver(val collector: FlowConstraintsCollector):
...
       case (p: ProdVar, c: ConsVar) =>
         upperBounds(p.s.uid) ::= c
-        for l <- lowerBounds(p.s.uid) do handle(l, c)
         lowerBounds(c.s.uid) ::= p
+        for l <- lowerBounds(p.s.uid) do handle(l, c)
         for u <- upperBounds(c.s.uid) do handle(p, u)
       case (p: ProdVar, c) =>
         upperBounds(p.s.uid) ::= c
+        c match
+          case PossibleAccumulator(s) if p.s === s =>
```
LPTK: wouldn't this be sufficient?

```suggestion
          case PossibleAccumulator(s) if p.s is s =>
```

hkmc2DiffTests/src/test/scala/hkmc2/MLsDiffMaker.scala
```diff
@@ -123,9 +137,31 @@ abstract class MLsDiffMaker extends DiffMaker:
       rewriteWhileLoops = rewriteWhile.isSet,
       tailRecOpt = !noTailRecOpt.isSet,
       deforest = Opt.when(deforest.isSet):
-        Deforest(
+        val flags = parseFlags(deforest.get)
+        val knownFlags = Set(
```
LPTK: Why would you recreate this Set every time? Move it out...

hkmc2DiffTests/src/test/scala/hkmc2/MLsDiffMaker.scala
```diff
@@ -135,20 +171,32 @@ abstract class MLsDiffMaker extends DiffMaker:
...
           if flags.contains("off") then N
-          else S(DeadParamElim(
-            debug = flags.contains("debug"),
-            mono = !flags.contains("poly")
-          )),
+          else
+            S(DeadParamElim(FlowAnalysisConfig(
+              debug = flags.contains("debug"),
+              mono = !flags.contains("poly"),
+              trackNonAffine = resolveFlag(flags, "trackNonAffine", "noTrackNonAffine", DeadParamElim.default.trackNonAffine),
```
LPTK: Is this really something we want to configure? Does it ever make sense to track this stuff for dead param elimination? Does it make sense for this tracking to be configurable? If not, this does not belong in `Config`.

hkmc2/shared/src/test/mlscript/deforest/todos.mls
```diff
@@ -64,53 +61,24 @@ test(p) + f(p) + test(AA(AA(AA(10)))) + test(B)
 //│   let t, a, arg$AA$0$, arg$AA$0$1, arg$AA$0$2, tmp11, tmp12, tmp13, tmp14, tmp15;
 //│   return x(t, a, arg$AA$0$, arg$AA$0$1, arg$AA$0$2, tmp11, tmp12, tmp13, tmp14, tmp15, t, tmp11, tmp12, tmp13, tmp14, tmp15)
 //│ };
-//│ define test_20$test⁰ as fun test_20$test¹(x) {
```
LPTK, line 67: Do you know why some of these definitions are no longer here?

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Deforest.scala
```diff
@@ -7,20 +7,36 @@ import mlscript.utils.*, shorthands.*
...
+    val scope = Scope.empty(Scope.Cfg.default)
+    given Scope = scope
+    given ShowCfg = ShowCfg.internal
+    given SymbolPrinter = new SymbolPrinter(scope)
+    new codegen.Printer().print(id.exprId.getResult).mkString()
+
+  private def pp(field: SelField): Str =
+    field match
+      case field: TermSymbol => field.nme
+      case idx: Int => idx.toString
```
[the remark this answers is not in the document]
LPTK: Copilot is right, creating a Scope for every single `pp(id)` call makes zero sense...

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Deforest.scala
ychenfo: [Updated this](https://github.com/hkust-taco/mlscript/pull/485/commits/207facec4afc9524e340c017e1199a24cae087b9) to thread the `SymbolPrinter` (needed by the code printer) through lowering / flow analysis into DeforestFusionSolver, so deforest debug printing no longer creates its own `SymbolPrinter` and in diff tests we now should use the same `dbgPrinter: SymbolPrinter` provided by `MlsDiffMaker`.

For `MLsCompiler.scala`, I moved the `SymbolPrinter` and trace logger setup into `compileModule` because constructing the `SymbolPrinter` needs a `Scope`, and `Scope.empty` requires an Elaborator.State.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
ychenfo: Done in https://github.com/hkust-taco/mlscript/pull/485/commits/0a73ebe4078160ca9531dc605de1b2439cf8d9b2

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
ychenfo: Done in https://github.com/hkust-taco/mlscript/commit/0a73ebe4078160ca9531dc605de1b2439cf8d9b2

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
ychenfo: Ah, this was a stale wip note to myself and the wording may be misleading, it's removed now. By this comment I only meant that the correct `NonAffine` constraints should be added to the correct collector, and the logic is not "doesn’t belong to an scc, therefore emit NonAffine". The symbol is already classified as non-affine by pre-analysis, and this code here is only about placing the non-affine constraints: in mono mode there are no per-scc collectors, so the global collector gets the relevant constraints; in poly mode, scc-owned symbols are handled in their scc strat scheme, and the global collector handles the remaining symbols.

hkmc2DiffTests/src/test/scala/hkmc2/MLsDiffMaker.scala
ychenfo: I see, indeed. I have changed them to be non-configurable for DPE.

hkmc2/shared/src/test/mlscript/deforest/todos.mls
ychenfo: The gone definitions are polymorphic instantiations of functions that used to contain deforestation rewritings. In this example since `p` is not used affinely, some deforestation matches are gone and then we no longer create that much polymorphic instantiations.

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Deforest.scala
```diff
@@ -22,11 +22,10 @@ class DeforestFusionSolver(val constraintSolver: FlowConstraintSolver)(using val
   given tl: TraceLogger = constraintSolver.tl
   given Raise = preAnalyzer.raise
 
+  private given Scope = Scope.empty(Scope.Cfg.default)
```
LPTK: Why do you need a scope?

ychenfo: Because the printer below needs a Scope: the `codegen.Printer.print` method takes a `using Scope` parameter.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
LPTK: Thanks for the explanation. Could you document that explanation in the code, at an appropriate and if provide visible place?

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Deforest.scala
LPTK: Ah, but this scope is used to show the top-level definitions. I think it's important that it correspond to the `MLsDiffMaker.scala`, so you should thread it through.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
LPTK: PS: Should be `ClassCtorSymbol`, to avoid clashes; don't forget to update `ClsLikeDefn` to use that.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
LPTK: Shpuld probably be `BlockLocalSymbol | TermSymbol`.

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Deforest.scala
ychenfo: I noticed that the `SymbolPrinter` in `MLsDiffMaker.scala` already stores the debug scope created in `MLsDiffMaker.scala`, so in the latest changes in `Deforest.scala` I just used the `Scope` in `SymbolPrinter`.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -350,26 +445,62 @@ class FlowPreAnalyzer(val pgrm: Program)(using
   
   end ctxTracker
   
+  private def recordAffinityUse(s: BlockLocalSymbol | TermSymbol): Unit =
+    s match
+    case _: ClassCtorSymbol => ()
+    case _ => currentAffinityCount(s) = currentAffinityCount(s) + 1
```
LPTK:
```suggestion
    case _ => currentAffinityCount(s) += 1
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -325,6 +325,15 @@ class TermSymbol(val k: TermDefKind, val owner: Opt[InnerSymbol], val id: Tree.I
   
   def subst(using sub: SymbolSubst): TermSymbol = sub.mapTermSym(this)
 
+
+class ClassCtorSymbol(
+  k: TermDefKind,
+  owner: Opt[InnerSymbol],
```
LPTK: It should always be `S[ClassLikeSymbol]` for class constructors, no?

LPTK (approved):
Otherwise LGTM.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
ychenfo: Ah good point, and after checking again I currently further tighten it to `S[ClassSymbol]`. If I didn't miss anything, `ClassCtorSymbol` should only be needed for real classes; Patterns, objects/modules, and companion modules have no ClassCtorSymbols.

LPTK: Ah, that seems to be true, objects have a ctor, but no ctor symbol...

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
ychenfo: Sure, I have added the explanation here: https://github.com/hkust-taco/mlscript/blob/977f3afaf51dc58db17bf058e786489bd7f75066/hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala#L735-L739

PR #486, opened 2026-05-10
Deduplicate UCS splits using `LetSplit` and `UseSplit`

chengluyu:


LPTK:
Pls fix the conflicts.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -135,39 +216,40 @@ class Normalization(lowering: Lowering)(using tl: TL)(using Raise, Ctx, State) e
...
+              S(aliasBindings(pattern, thatPattern)(rec(continuation).getOrElse(continuation) ++ rec(tail).getOrElse(tail)))
             else if thatPattern <:< pattern then
               log(s"Case 1.1.2: $pattern <:< $thatPattern")
-              pattern.markAsRefined; split.copy(tail = rec(tail))
+              pattern.markAsRefined
+              rec(tail).map(newTail => split.copy(tail = newTail))
             else if split.isFallback then
               log(s"Case 1.1.3: $pattern is unrelated with $thatPattern")
-              rec(tail)
+              S(rec(tail).getOrElse(tail))
```
LPTK: Please respond to my original comment on this!

https://github.com/hkust-taco/mlscript/pull/457#discussion_r3199677193

hkmc2/shared/src/test/mlscript/ucs/normalization/Deduplication.mls
```diff
@@ -508,14 +587,129 @@ foo(A, A, A)
...
+//│     set tmp = 2;
+//│     end
+//│ set y¹ = tmp;
+//│ end
 //│ —————————————————| Output |—————————————————————————————————————————————————————————————————————————
 //│ y = 1
 
+fun pred(x) = true
 
+:sir
```
LPTK: I feel like there's already too much IR output in this file. Isn't the `:ucs` output already sufficient for many of these test cases? Please see if you can drop some `:sir`s here.

hkmc2/shared/src/test/mlscript/ucs/normalization/ExcessiveDeduplication.mls
```diff
@@ -47,66 +46,51 @@ if
 //│ = 0
 
 
-// * Currently, dedup' is implemented by hash-consing after the fact, which doesn't work when bindings are involved;
-// * it fails here:
+// * The last split is `y is [b] then ...`. It was duplicated because it serves
```
LPTK: This comment doesn't make sense. People reading it won't know that "was" refers to a _previous_ state of the codebase!

Does this make more sense?
```suggestion
// * The last split is `y is [b] then ...`. It used to be duplicated because it serves
```

hkmc2/shared/src/test/mlscript/ucs/normalization/ExcessiveDeduplication.mls
LPTK:
```suggestion
// * We never try to deduplicate things that are already duplicated in the source.
// * (A previous version of the UCS was using hash-consing to do so, which was a bad idea.)
```

hkmc2/shared/src/test/mlscript/ucs/normalization/LetSplitJoinPoints.mls
```diff
@@ -0,0 +1,417 @@
...
+//│ set tmp2 = B¹(2);
+//│ f3¹(tmp2)
+//│ —————————————————| Output |—————————————————————————————————————————————————————————————————————————
+//│ = "B"
+
+
+// === Join point cases ===
+
+
+// * Join point: non-trivial Else alternative on different scrutinee
```
LPTK: You haven't addressed _any_ of my previous comments...

See https://github.com/hkust-taco/mlscript/pull/457#discussion_r3199708410 and the others.

hkmc2/shared/src/test/mlscript/ucs/normalization/ExcessiveDeduplication.mls
chengluyu: Applied in `ef5c68414`.

hkmc2/shared/src/test/mlscript/ucs/normalization/ExcessiveDeduplication.mls
chengluyu: Applied in `a2658fa2d`.

hkmc2/shared/src/test/mlscript/ucs/normalization/Deduplication.mls
chengluyu: Fixed in 7f2bc86a9. I dropped `:sir` from three cases that already had `:ucs normalized`, and I also removed a duplicate test block.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
chengluyu: Replied in the original PR.

hkmc2/shared/src/test/mlscript/ucs/normalization/LetSplitJoinPoints.mls
chengluyu: Replied in the original PR.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -21,22 +20,17 @@ class Normalization(lowering: Lowering)(using tl: TL)(using Raise, Ctx, State) e
...
         log("tail is discarded")
         these
       else (these match
         case Split.Cons(head, tail) => Split.Cons(head, tail ++ those)
         case Split.Let(name, term, tail) => Split.Let(name, term, tail ++ those)
-        case Split.Else(_) /* impossible */ | Split.End => those)
+        case Split.Else(_) => softAssert(false); those
+        case Split.End => those
+        case Split.LetSplit(sym, tail) => Split.LetSplit(sym, tail ++ those)
+        case Split.UseSplit(_) => these) // UseSplit is terminal; fullness determined by referenced body
```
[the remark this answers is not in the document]
chengluyu: In practice every `LetSplit` body ends up full, because we always append a default else to the outer split and normalization carries that default into inner splits. So `UseSplit.isFull` is true and `those` would be dropped by the first branch anyway. I added a `softAssert` in `f276d09f6` to lock this in.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -135,39 +207,41 @@ class Normalization(lowering: Lowering)(using tl: TL)(using Raise, Ctx, State) e
...
+              pattern.markAsRefined
+              rec(tail).map(newTail => split.copy(tail = newTail))
             else thatPattern match
             case thatPattern: FlatPattern.Record =>
               log(s"Case 1.1.4: $thatPattern is a record")
               // we can use information if pattern is itself a record, or if it is a constructor with arguments
               val simplifiedRecord = thatPattern assuming pattern
               if simplifiedRecord.entries.isEmpty then
-                tail
+                S(tail)
```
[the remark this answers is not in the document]
chengluyu: Fixed in `79422a69e`. The real cause was the inconsistency between `assuming` and `cmpareCasePattern`.

1. Method `assuming` filtered record fields against all class parameters, no matter they are `val` or not..
2. But `compareCasePattern` only looked at `val` params.

I just made `assuming` use the same `val` check, so the empty-entries branch only fires when it really should.

hkmc2/shared/src/main/scala/hkmc2/semantics/Split.scala
```diff
@@ -31,6 +36,8 @@ enum Split extends AutoLocated with ProductWithTail:
     case Let(sym, term, tail) => Let(sym, term.mkClone, tail.mkClone)
     case Else(default) => Else(default.mkClone)
     case End => End
+    case LetSplit(sym, tail) => LetSplit(sym, tail.mkClone)
+    case UseSplit(sym) => UseSplit(sym)
```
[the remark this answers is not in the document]
chengluyu: I tried to write a test that triggers this but couldn't. The bug needs two `LetSplit`s with the same symbol where one sits inside the other's body. `inlineUseSplit` replaces a `UseSplit(sym)` with a copy of the body, and `UseSplit` is a leaf node with no children, so the copies don't end up inside one another. They appear at separate positions in the surrounding tree, like sibling branches of a match. Lowering walks one branch all the way down before starting the next, so by the time the second copy writes `sym.label`, the first copy's reads are already done.

If you (@copilot) find a case that actually triggers the bug, then I'd be glad to revisit it.

hkmc2/shared/src/main/scala/hkmc2/semantics/Split.scala
```diff
@@ -47,31 +54,52 @@ enum Split extends AutoLocated with ProductWithTail:
       case Cons(head, tail) => Cons(head, tail.duplicate)
       case Let(name, term, tail) => Let(name, term, tail.duplicate)
       case Else(default) => Else(default)
-      case End => End).setDuplicated
+      case End => End
+      case LetSplit(sym, tail) => LetSplit(sym, tail.duplicate)
+      case UseSplit(sym) => UseSplit(sym)).setDuplicated
```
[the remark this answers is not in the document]
chengluyu: See https://github.com/hkust-taco/mlscript/pull/486#discussion_r3294314012.

hkmc2/shared/src/main/scala/hkmc2/semantics/Split.scala
```diff
@@ -112,6 +156,8 @@ extension (split: Split)
       case Split.Let(name, term, tail) => Split.Let(name, term, tail ~~: fallback)
       case Split.Else(_) => lastWords("impossible since split is not full")
       case Split.End => fallback
+      case Split.LetSplit(sym, tail) => Split.LetSplit(sym, tail ~~: fallback)
+      case Split.UseSplit(_) => split // UseSplit is terminal; the referenced body determines fullness
```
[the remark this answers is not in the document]
chengluyu: See https://github.com/hkust-taco/mlscript/pull/486#discussion_r3294313987.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -1035,6 +1035,14 @@ class Lowering()(using Config, TL, Raise, State, Ctx, SymbolPrinter):
...
       Assign(l, r, setupTerm("Else", Value.Ref(l) :: Nil)(k))
     case Split.End => setupTerm("End", Nil)(k)
+    case Split.LetSplit(sym, tail) =>
+      quoteSplit(sym.body): r1 =>
+        val l1 = loweringCtx.registerTempSymbol(N)
+        blockBuilder.assign(l1, r1)
+          .chain(b => quoteSplit(tail)(r2 => Assign(loweringCtx.registerTempSymbol(N), r2, b)))
+          .rest(setupTerm("LetSplit", Value.Ref(l1) :: Nil)(k))
+    case Split.UseSplit(sym) =>
+      setupTerm("UseSplit", Nil)(k)
```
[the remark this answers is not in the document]
chengluyu: Fixed in `ad2c7ac54`.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
```diff
@@ -78,25 +72,96 @@ class Normalization(lowering: Lowering)(using tl: TL)(using Raise, Ctx, State) e
...
+      specialize(split, mode, scrutinee, pattern).getOrElse(split)
+
+  /** Replace all `UseSplit(sym)` references in `split` with a duplicate of `body`. */
+  private def inlineUseSplit(split: Split, sym: SplitSymbol, body: Split): Split = split match
+    case Split.Cons(Branch(scrut, pat, cons), tail) =>
+      Split.Cons(Branch(scrut, pat, inlineUseSplit(cons, sym, body)), inlineUseSplit(tail, sym, body))
+    case Split.Let(v, rhs, tail) => Split.Let(v, rhs, inlineUseSplit(tail, sym, body))
+    case Split.Else(_) | Split.End => split
+    case Split.LetSplit(s, tail) => Split.LetSplit(s, inlineUseSplit(tail, sym, body))
+    case Split.UseSplit(s) => if s eq sym then body.duplicate else split
```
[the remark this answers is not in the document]
chengluyu: See https://github.com/hkust-taco/mlscript/pull/486#discussion_r3294314012.

hkmc2/shared/src/main/scala/hkmc2/semantics/ucs/Normalization.scala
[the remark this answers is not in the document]
chengluyu: We should issue a warning for cases like this, but not in this PR...

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -1035,14 +1035,12 @@ class Lowering()(using Config, TL, Raise, State, Ctx, SymbolPrinter):
...
     case Split.End => setupTerm("End", Nil)(k)
-    case Split.LetSplit(sym, tail) =>
-      quoteSplit(sym.body): r1 =>
-        val l1 = loweringCtx.registerTempSymbol(N)
-        blockBuilder.assign(l1, r1)
-          .chain(b => quoteSplit(tail)(r2 => Assign(loweringCtx.registerTempSymbol(N), r2, b)))
-          .rest(setupTerm("LetSplit", Value.Ref(l1) :: Nil)(k))
-    case Split.UseSplit(sym) =>
-      setupTerm("UseSplit", Nil)(k)
+    // The runtime `Term.Split` only has `Cons/Let/Else/End` constructors, so
```
LPTK: What is the "runtime `Term.Split`"? 🤔

LPTK: Ah, I guess this is referring to the reflection Block IR representation implemented in MLscript. Instead of "flattening" splits (which I guess just duplicates code exponentially?!), _clearly_ the correct aproach is to add the missign constructors there. This whole fix feels extremely sloppy (including the comments formatted to end at 80 columns). Should I ake this PR as `slopfest` too?

chengluyu: I fixed this using Claude Code but this fix is approved and reviewed by me (related conversation messages are attached below). I didn't really want to touch this part much so I chose the dumb approach.

image
image

Many commits and fixes in this PR are the result of several rounds of discussion with Claude Opus 4.7 (Max reasoning effort). These are subjective choices I have made deliberately, rather than being part of a fully automated “YOLO mode”. But still feel free to mark it as slopfest.

LPTK: Then can you tell little Claudy and its max reasoning effort to deal with this thing in the "non dumb" way?

hkmc2/shared/src/test/mlscript/ucs/normalization/Deduplication.mls
```diff
@@ -8,191 +8,166 @@ let z = 0
...
-:sjs
+:sir
 if x is
   0 and y is
     0 and z is
-      0 then "000-------------------------------------------------------------------------"
-      1 then "001-------------------------------------------------------------------------"
-    1 then "01-------------------------------------------------------------------------"
-  1 then "1-------------------------------------------------------------------------"
-  else "-------------------------------------------------------------------------"
```
LPTK, line 76: I disagree with these changes. Passing small strings to `id` does not make the term big. Notice the comment above. Why make this change?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
chengluyu: Fixed in f697b15ec9588fd3564401f5f1bec3c5b7339769.

hkmc2/shared/src/test/mlscript/ucs/normalization/Deduplication.mls
chengluyu: I made the changes because I wanted it to be consistent with other tests. Adding more dashes didn't make it more visible. Fixed in 10fe30d8ffdf97dee77c27dedc07150520e42555.

PR #491, opened 2026-05-16
Refactor `Value.Ref` into `Value.{SimpleRef,MemberRef,InnerRef}`

Derppening:
TODO:

- [x] Migrate `case Value.Ref` into `apply` and `unapply` functions


Derppening:
[034](https://github.com/hkust-taco/mlscript/actions/runs/25954060741/job/76297959177?pr=491#step:6:1035)

```
[error] -- Error: /home/runner/work/mlscript/mlscript/hkmc2DiffTests/src/test/scala/hkmc2/JSBackendDiffMaker.scala:293:38 
[error] 293 |                    case sym => Value.Ref(sym, N),
[error]     |                                ^^^^^^^^^
[error]     |object Ref in object Value is deprecated: Use Value.SimpleRef, Value.MemberRef, or Value.This instead.
[error] one error found
[error] (hkmc2DiffTests / Test / compileIncremental) Compilation failed
```

I have no idea what's up with the CI failure. It is only deprecated but it's treated as an error for some reason.

Derppening:
@LPTK You can take a look at the changes first. I will mark this as ready for review once the TODO is also addressed.

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Rewrite.scala
```diff
@@ -261,7 +261,9 @@ class DeforestRewriter(val solver: DeforestFusionSolver)(using Raise):
       
       override def applyValue(v: Value): Unit =
         v match
-        case Value.Ref(l, disamb) if !inCtx(l) && l.asClsLike.isEmpty => freeVars.add(l)
+        case Value.SimpleRef(l) if !inCtx(l) && l.asClsLike.isEmpty => freeVars.add(l)
```
LPTK: I don't think `asClsLike` can be `true` for `SimpleRef`. Just look at its impl.

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Rewrite.scala
```diff
@@ -261,7 +261,9 @@ class DeforestRewriter(val solver: DeforestFusionSolver)(using Raise):
...
+        case Value.MemberRef(bms, _) if !inCtx(bms) && bms.asClsLike.isEmpty => freeVars.add(bms)
+        case Value.InnerRef(l) if !inCtx(l) && l.asClsLike.isEmpty => freeVars.add(l)
```
LPTK: I think this will never meaningfully match. `this` references are almost always to a class-ike thing. @ychenfo what's the intent, here?

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Rewrite.scala
```diff
@@ -353,11 +355,16 @@ class DeforestRewriter(val solver: DeforestFusionSolver)(using Raise):
     
     extension (a: Symbol) def toValueRef =
       a match
-      case bms: BlockMemberSymbol => Value.Ref(bms, bms.tsym)
-      case _ => Value.Ref(a, N)
+      case bms: BlockMemberSymbol => Value.MemberRef(bms, bms.defaultDisamb.get)
+      case sym: TempSymbol => Value.SimpleRef(sym)
+      case sym: VarSymbol => Value.SimpleRef(sym)
+      case sym: (LocalSymbol | BuiltinSymbol) => Value.SimpleRef(sym)
+      case sym: InnerSymbol => Value.InnerRef(sym)
```
LPTK: This jungle of cases is almost certainly unnecessary. Check what type of symbol flows here and split/refine the type of this extension method, isntead of using pattern matching. Perhaps @ychenfo can help.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -1012,7 +1028,7 @@ class FlowConstraintsCollector(
               case Select(p, _) => cc.constrain(processResult(p), UnknownCons)
               case _ => ()
             generatedProdVars(sym).asProdStrat
-          case _: Value.Ref => lastWords("already handled in `RefLike` case")
+          case _: (Value.SimpleRef | Value.MemberRef) => lastWords("already handled in `RefLike` case")
```
LPTK: I see that this `Value.SimpleRef | Value.MemberRef` pattern seems common enough to make them extend a base `Value.Ref` sealed trait that could be used here instead.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -97,10 +99,13 @@ object RefLike:
...
         cls
   
-  def unapply(p: Value.Ref | Select)(using Elaborator.State): Opt[Symbol] =
+  def unapply(p: Value.SimpleRef | Value.MemberRef | Select)(using Elaborator.State): Opt[Symbol] =
     p match
-      case Value.Ref(l, disamb) =>
-        val sym = disamb.getOrElse(l)
+      case Value.SimpleRef(sym) =>
+        classCtorSymbol(sym) orElse S(sym)
+        classCtorSymbol(sym) orElse S(sym)
```
LPTK: Duplicated line...

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -97,10 +99,13 @@ object RefLike:
       yield
         cls
   
-  def unapply(p: Value.Ref | Select)(using Elaborator.State): Opt[Symbol] =
+  def unapply(p: Value.SimpleRef | Value.MemberRef | Select)(using Elaborator.State): Opt[Symbol] =
```
LPTK: It seems this extractor is only used to extract refs to ctors and functions. It should have a much more specific name and return type.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
LPTK: Again, this `orElse`/`getOrElse` slop is wrong and should be removed.

hkmc2/shared/src/main/scala/hkmc2/codegen/llir/Builder.scala
LPTK, line 1: Ouch... There is no need to spend time updating this legacy class. TBH we should probably just remove it from the project.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1224,21 +1224,37 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
-              case N => getVar(l, r.toLoc)
+            ctx.getFunc(bms) match
+              case S(funcIdx) => ref.func(funcIdx, RefType(ctx.getFuncTypeUse_!(bms).typeIdx, nullable = false))
+              case N => getVar(bms, r.toLoc)
+    case Value.InnerRef(sym) =>
+      singletonInfoFor(sym) match
+        case S(info) => singletonGlobalGet(info)
+        case N =>
+          ctx.getFunc(sym) match
+            case S(funcIdx) => ref.func(funcIdx, RefType(ctx.getFuncTypeUse_!(sym).typeIdx, nullable = false))
```
LPTK: How can an inner ref (ie `this`) refer to a function?!

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1224,21 +1224,37 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
         operands = Seq(ref.i31(i32.const(lit.offset)), ref.i31(i32.const(lit.byteLen))),
         returnTypes = Seq(Result(RefType.anyref)),
       )
-    case Value.Ref(l, disamb) =>
-      if (l is State.unitSymbol) || disamb.contains(State.unitSymbol) then
+    case Value.SimpleRef(l) =>
+      if (l is State.unitSymbol) then
```
LPTK: `unitSymbol` is a `ModuleOrObjectSymbol`, so this is dead code.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -920,25 +931,52 @@ case class Select(qual: Path, name: Tree.Ident)(val symbol: Opt[DefinitionSymbol
...
-   */
-  case Ref(l: Local, disamb: Opt[DefinitionSymbol[?]])
-  case This(sym: InnerSymbol) // TODO rm – just use Ref
+    * @param disamb The symbol disambiguating the definition that the reference refers to.
+    */
+  case MemberRef(bms: BlockMemberSymbol, disamb: DefinitionSymbol[?])
+  case This(sym: InnerSymbol)
+  /** Reference to an [[InnerSymbol]] accessed by its bound variable name from outside its scope,
+    * as opposed to [[This]] which is a `findThis_!`-based "current object" reference. */
+  case InnerRef(sym: InnerSymbol)
```
LPTK: This doesn't make sense. We never refer to these symbols directly from outside the scope. `This` and `MemberRef` should be enough.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -920,25 +931,52 @@ case class Select(qual: Path, name: Tree.Ident)(val symbol: Opt[DefinitionSymbol
...
     case _ => ""
 
 object Value:
+  @deprecated("Use Value.SimpleRef, Value.MemberRef, Value.This, or Value.InnerRef instead.")
   object Ref:
+    def apply(l: Local, disamb: Opt[DefinitionSymbol[?]])(using State): Value = 
+      l match
+        case l: (LocalSymbol | BuiltinSymbol) => Value.SimpleRef(l)
+        case sym: TopLevelSymbol if sym === State.globalThisSymbol => Value.This(sym)
+        case tls: TopLevelSymbol => Value.This(tls)
```
LPTK: ??

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -972,6 +1010,18 @@ extension (k: Block => Block)
...
 extension (l: Local)
-  def asPath: Path = Value.Ref(l, N)
+  def asPath(using State): Path = l match 
+    case l: (LocalSymbol | BuiltinSymbol) => Value.SimpleRef(l)
+    case sym: TopLevelSymbol if sym === State.globalThisSymbol => Value.This(sym)
+    case tls: TopLevelSymbol => Value.This(tls)
+    case bms: BlockMemberSymbol => Value.MemberRef(bms, bms.defaultDisamb.getOrElse(lastWords(s"Cannot disambiguate overloaded member symbol ${bms.nme}: no disambiguation provided")))
+    case sym: InnerSymbol => Value.InnerRef(sym)
+    case _: NoSymbol => lastWords("NoSymbol should not be used as a Path/Value")
+    case sym => lastWords(s"$sym (of type ${sym.getClass.getSimpleName}) cannot be converted to a Path/Value")
```
LPTK: This looks like it's mostly copy-pasted from the above. Do NOT duplicate logic like that.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
LPTK, line 1: This file should not have such extensive changes. What's going on? Pls minimize the diff.

hkmc2DiffTests/src/test/scala/hkmc2/JSBackendDiffMaker.scala
```diff
@@ -281,8 +281,16 @@ abstract class JSBackendDiffMaker extends MLsDiffMaker:
...
-                (Arg(N, Value.Ref(sym, N)) :: Nil) ne_:: Nil)(true, false, false),
+                Value.SimpleRef(Elaborator.State.runtimeSymbol).selSN("printRaw"),
+                (Arg(
+                  N,
+                  sym match
+                    case sym: TempSymbol => Value.SimpleRef(sym)
+                    case sym: VarSymbol => Value.SimpleRef(sym)
+                    case sym: BlockMemberSymbol => Value.MemberRef(sym, sym.defaultDisamb.get)
+                    case sym: (LocalSymbol | BuiltinSymbol) => Value.SimpleRef(sym)
+                    case sym: InnerSymbol => Value.InnerRef(sym)
```
LPTK: Again some duplicated (and suspicious) logic.

hkmc2/shared/src/test/mlscript/deforest/module.mls
```diff
@@ -1,6 +1,5 @@
 :js
 
-
```
LPTK, line 3: 👎

hkmc2/shared/src/main/scala/hkmc2/codegen/SymbolRefresher.scala
```diff
@@ -202,22 +202,39 @@ class SymbolRefresher(existingMapping: Map[Symbol, Symbol])(using State) extends
     case _ => super.applyPath(p)(k)
 
   override def applyValue(v: Value)(k: Value => Block): Block = v match
-    case Value.Ref(l, x) =>
+    case Value.SimpleRef(l) =>
       mapping.get(l) match
         case None => super.applyValue(v)(k)
         case Some(newBms: BlockMemberSymbol) =>
```
LPTK: Will the mapping ever map a `SimpleRef` symbol to a `BlockMemberSymbol`?! 🤔

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -579,13 +579,27 @@ class Lowering()(using Config, TL, Raise, State, Ctx, SymbolPrinter):
...
-              Value.Ref(bs, disamb).withLocOf(ref), Nil ne_:: Nil
+              Value.MemberRef(bs, disamb.orElse(bs.defaultDisamb).get).withLocOf(ref), Nil ne_:: Nil
             )(isMlsFun = true, true, annots.contains(Annot.TailCall)))
       case S(_) => ()
       case N => () // TODO panic here; can only lower refs to elab'd symbols
     case _ => ()
     warnStmt
-    k(loweringCtx(Value.Ref(sym, disamb).withLocOf(ref)))
+    (sym, disamb) match
+      case (sym: TopLevelSymbol, _) =>
```
LPTK: Useless case.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -579,13 +579,27 @@ class Lowering()(using Config, TL, Raise, State, Ctx, SymbolPrinter):
...
+        k(loweringCtx(Value.This(sym).withLocOf(ref)))
+      case (sym: TempSymbol, _) =>
+        k(loweringCtx(Value.SimpleRef(sym).withLocOf(ref)))
+      case (sym: VarSymbol, _) =>
+        k(loweringCtx(Value.SimpleRef(sym).withLocOf(ref)))
```
LPTK: Collapse using `sym: (TempSymbol | VarSymbol)`

LPTK (changes requested):
I feel like this should have been marked as a draft and that you didn't properly review the diff. Given the amount of duplicated logic, a lot of it looks like thoughtless slop 😕

PS: thoughtless slop is perfectly fine as long as the PR is marked as Draft and you eventually clean it up 😉

Derppening:
Sorry for all the redundant logic - I will convert this as a draft to clean this up first, then when it's ready I will re-request for your review.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -113,17 +102,39 @@ class JSBuilder(using Config, TL, State, Ctx) extends CodeBuilder:
...
     case Value.SimpleRef(l: BuiltinSymbol) =>
       if l.nullary then l.nme
       else errExpr(msg"Illegal reference to builtin symbol '${l.nme}'")
-    case Value.SimpleRef(l) => getVar(l, r.toLoc)
-    case Value.InnerRef(sym) => getVar(sym, r.toLoc)
+    case Value.SimpleRef(l: semantics.TermSymbol) =>
+      l.owner match
+      case S(owner) =>
+        val isPrivateField =
+          (l.k is syntax.LetBind) && !owner.isInstanceOf[semantics.TopLevelSymbol]
```
LPTK: Let's add a TODO to use the other PR's new `isPrivate` method, once that one is merged.

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Rewrite.scala
Derppening: Addressed in 604cffec07508f157ea83c3ab8bb1fb43f0d3a2a.

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Rewrite.scala
Derppening: Addressed in 954266ec819b5cebcf9ae87b5343cc2033d7e183 - Replaced with a call to `asPath`, since they effectively do the same thing.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
Derppening: Removed in e8c903dd162c9af29664aa561f4d4f0e5ed764fd.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
Derppening: Removed and simplified the logic of `getVar` in fdcfef67b845da14a18d97f936ed3e66b72d2baa.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
Derppening: Removed in fdcfef67b845da14a18d97f936ed3e66b72d2baa.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
Derppening: Removed and replaced with `Value.This` in e8c903dd162c9af29664aa561f4d4f0e5ed764fd.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
Derppening: Addressed in 0a1c572f982837c622590fa5d1ee44c658e78103.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
Derppening: Addressed in 954266ec819b5cebcf9ae87b5343cc2033d7e183 - Replaced with a call to `Value.Ref.apply`, which will eventually be inlined into this function when the `apply` function is removed.

hkmc2DiffTests/src/test/scala/hkmc2/JSBackendDiffMaker.scala
Derppening: Addressed in 954266ec819b5cebcf9ae87b5343cc2033d7e183 - Replaced with a call to asPath.

hkmc2/shared/src/test/mlscript/deforest/module.mls
Derppening: Restored in 9f6e89e1a16622335944b552aaeb3610ed91e3c8.

hkmc2/shared/src/main/scala/hkmc2/codegen/SymbolRefresher.scala
Derppening: Simplified in 76d1636bbc042d555fde10187e23d85c4adfed79.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
Derppening: Addressed in e2e899a5e4c6523ecead279642d2ab02b01e9815.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
Derppening: Simplified in 0738f095113b191ed22c817c49071d7abf94381d - These are now collapsed into `LocalSymbol | BuiltinSymbol`, since `TempSymbol and `VarSymbol` are both `LocalSymbol`s.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -995,16 +995,19 @@ extension (k: Block => Block)
 def blockBuilder: Block => Block = identity
 
 extension (bms: BlockMemberSymbol)
-  /** Best-effort default disambiguation for a `BlockMemberSymbol` reference. */
-  def defaultDisamb: Opt[DefinitionSymbol[?]] =
-    bms.asTrm.orElse(bms.asClsOrMod).orElse(bms.asPat)
+  /** The principal disambiguation for a [[`BlockMemberSymbol`]] reference, or `N` if the BMS does not resolve to a
```
LPTK: When do you ever need this, actually? In principle, all correct disambs should already be resolved.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
Derppening: Done in 414f30b0d40702b842a0ea8bd9887db65e8df885.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
Derppening: The only two locations that require it is:

- `ref` in `Lowering.scala`: Not all `BlockMemberSymbol`s have a `disamb`.
- `Local.asPath` (extension method in `Block.scala`): We need a `DefinitionSymbol` for an instance of `Value.MemberRef` to be constructed.

Do you have a suggestion on how I should handle these cases?

Derppening: > The only two locations that require it is:
> 
>     * `ref` in `Lowering.scala`: Not all `BlockMemberSymbol`s have a `disamb`.
> 
>     * `Local.asPath` (extension method in `Block.scala`): We need a `DefinitionSymbol` for an instance of `Value.MemberRef` to be constructed.
> 
> 
> Do you have a suggestion on how I should handle these cases?

I guess `Local.asPath` can be trivially fixed by adding a `disamb` parameter? But this would also make other symbols require an argument (`N`), is this a good tradeoff?

Derppening: > The only two locations that require it is:
> 
>     * `ref` in `Lowering.scala`: Not all `BlockMemberSymbol`s have a `disamb`.
> 
>     * `Local.asPath` (extension method in `Block.scala`): We need a `DefinitionSymbol` for an instance of `Value.MemberRef` to be constructed.
> 
> 
> Do you have a suggestion on how I should handle these cases?

As for `ref` in `Lowering.scala`, there are various examples that fail:

```fs
@bufferable
class Oops with
  fun hi = print("hi")


Oops.hi(buf, obj2)
//│ /!!!\ Uncaught error: java.lang.Exception: Internal Error: No disamb available: member:Oops
//│ 	at: mlscript.utils.package$.lastWords(package.scala:239)
//│ 	at: hkmc2.codegen.Lowering.ref$$anonfun$8(Lowering.scala:592)
//│ 	at: scala.Option.getOrElse(Option.scala:203)
//│ 	at: hkmc2.codegen.Lowering.ref(Lowering.scala:592)
//│ 	at: hkmc2.codegen.Lowering.term(Lowering.scala:675)
//│ 	at: hkmc2.codegen.Lowering.blockImpl$1(Lowering.scala:196)
//│ 	at: hkmc2.codegen.Lowering.block(Lowering.scala:382)
//│ 	at: hkmc2.codegen.Lowering.$anonfun$59(Lowering.scala:1236)
//│ 	at: hkmc2.codegen.Lowering.inScopedBlock(Lowering.scala:1303)
//│ 	at: hkmc2.codegen.Lowering.program(Lowering.scala:1236)
```

LPTK: Well, `Lowering` should always produce IR code where the disamb is present. In principle, it picks that up from info produced by Resolver. But if using this function is easier, I'm fine with it. The resolver is very messy (and will be completely changed as part of the frontend rewrite anyway).

LPTK: As for `Local.asPath`, I feel like this method doesn't make much sense in its current form. Do we really need such a megamorphic method on Local (ie Symbol), or can we just use more targeted variations?

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -702,67 +692,66 @@ class BlockSimplifier
...
-            
-            val l = analyzeValues(a1)
-            if l.isEmpty && litValue === false then
-              emptyHanded = true
-              analyzeAssignments(a2)
-              Set.empty
-            else l & analyzeValues(a2)
-    
       v match
-
```
LPTK: 👎

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -702,67 +692,66 @@ class BlockSimplifier
...
       v match
-            
       case Value.SimpleRef(loc: LocalVar) if !inDryRun && !capturedVars(loc) =>
         
         val rs = assignedResults(loc)
-        // log(s"SimpleRef ${loc.showDbg} ${rs} ${localVars(loc)} ${capturedVars(loc)}")
-        
+        // log(s"Ref ${loc.showDbg} ${rs} ${localVars(loc)} ${capturedVars(loc)}")
+
+        def analyzeAssignments(asst: AssignInfo): Unit =
```
LPTK, line 735: Pls try to minimize the diff by not moving this around. Or was there a good reason for the move?

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
Derppening: Addressed in a73a3d900090ee9610d7c6e53439a2f1d23a385. We now have `Value.RefLike` that represents what the legacy `Value.Ref` represents, and `Value.Ref` that represents only `Value.SimpleRef` and `Value.MemberRef`.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
Derppening: Addressed in a73a3d900090ee9610d7c6e53439a2f1d23a3853 and 7d14fd106e526c2f2c5febdab689844945a6a053 - Split the appropriate functions into `CtorRef`.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
Derppening: Addressed in a73a3d900090ee9610d7c6e53439a2f1d23a3853.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
Derppening: The diff was originally introduced due to the duplicate logic of the match arms, so I hoisted the functions outside of the match expression to ensure different blocks can reuse the same logic. Now that the duplicate logic is simplified and/or removed, the original structure is restored - Hence the diff.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
Derppening: I couldn't find this formatting change :thinking: Either way, I manually inspected the changes and reverted more whitespace changes.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
Derppening: Addressed in a73a3d900090ee9610d7c6e53439a2f1d23a3853. The `Value.{RefLike,Ref}` has the side effect of simplifying a lot of redundant logic.

Derppening:
Current TODOs:

- [x] Address all `TODO(Derppening)` comments introduced in this PR
- [x] Inspect whitespace changes and revert where necessary
- [x] Inspect whether any other `case Value.SimpleRef | Value.MemberRef` can be collapsed, or helper methods can be added for ergonomics
- [x] Ensure all newly added functions are documented
- [x] Try to convert all `asPath` into `Value.*` constructors
- [x] Investigate `asPrincipal` as a replacement for `principalDisamb`
- [x] Investigate `orElseDisamb` - Can we remove it?

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
Derppening: So I just realized that we already have a `Symbol.asPrincipal` to obtain the principal definition, so my `principalDisamb` function is not necessary anymore (1e36b3f4e5c35ba5fb3ba763c542c84e429f6f50). I will still look into refactoring `asPath` to see if we can fully move away from this.

Derppening: c2cffa971da1ba2a0fdbb1d34360254fc7ceb3a6 completed the refactoring of all `asPath` locations that can be trivially converted into `Value.{SimpleRef,MemberRef,This}` constructors. The rest of the call sites (13 in total AFAICT) accepts a `Symbol` at the call site and cannot be trivially replaced.

Let me know if and how you want me to handle the remaining cases, or whether the refactor to use `asPrinicipal` is a good enough solution for now.

hkmc2/shared/src/main/scala/hkmc2/codegen/llir/Builder.scala
Derppening: Are there any actionable tasks regarding this that I can/should make in this PR? Or should we convert this into an issue to be done in a separate branch?

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -326,7 +326,7 @@ class JSBuilder(using Config, TL, State, Ctx) extends CodeBuilder:
     
     @tailrec
     private def unapplyImpl(
-      b: Block,
+      b: Block,
```
LPTK: Don't revert _these_...

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -795,7 +795,7 @@ class BlockSimplifier
         case false =>
           vars.minByOption(_.symbol.uid) match
           case N => k(v)
-          case S(v2) =>
+          case S(v2) =>
```
LPTK: Idem.

hkmc2/shared/src/main/scala/hkmc2/codegen/FirstClassFunctionTransformer.scala
```diff
@@ -58,7 +58,7 @@ class FirstClassFunctionTransformer(using Elaborator.State, Elaborator.Ctx, Rais
         raise(ErrorReport(msg"Cannot determine if ${sel.name.name} is a function." -> sel.toLoc :: Nil,
           source = Diagnostic.Source.Compilation))
         k(p)
-    case _ => k(p)
+    case _ => k(p)
```
LPTK: Idem.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -135,7 +135,7 @@ class BlockSimplifier
               usedVars += loc
             case _ =>
           super.applyPath(p)
-        
+
```
LPTK: 👎

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
Derppening: Addressed in ca35733cf5ea478b92da6edeaeb47445767f8319.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
Derppening: ca35733cf5ea478b92da6edeaeb47445767f8319

hkmc2/shared/src/main/scala/hkmc2/codegen/FirstClassFunctionTransformer.scala
Derppening: ca35733cf5ea478b92da6edeaeb47445767f8319

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
Derppening: ca35733cf5ea478b92da6edeaeb47445767f8319

Derppening:
Most of the changes are in-place. There are several "internal" TODOs ([here](https://github.com/hkust-taco/mlscript/pull/491#issuecomment-4516386241)) that will be addressed - Afterwards I will mark this PR as ready.

> Investigate `orElseDisamb` - Can we remove it?

There are no more callers of `Symbol.orElseDisamb`. Perhaps we can now remove that?

There are also some TODOs that I would like your input on:

- `Block.scala:1027`: The deprecated `Ref.unapply` is unable to distinguish between a "legacy" `Value.This` (that only represents `this` and nothing else) and a "new" `Value.This` (which represents self-refs in general). Does it matter, considering it's just a transitional shim?
- `ReflectionInstrumenter.scala:180`: Same issue here - Now that we expand the definition of `Value.This` to also include self-refs, we need to include `Value.This` when transforming paths. How do we reconcile this expansion of definitions if the legacy `Value.This` is not supported by the instrumenter?

LPTK:
Great, good job!

> There are no more callers of `Symbol.orElseDisamb`. Perhaps we can now remove that?

Yes! Please do.


> * `Block.scala:1027`: The deprecated `Ref.unapply` is unable to distinguish between a "legacy" `Value.This` (that only represents `this` and nothing else) and a "new" `Value.This` (which represents self-refs in general). Does it matter, considering it's just a transitional shim?

Not sure I understand the question. There's only one self reference `This` in my mind.

By the way, this shim for `Ref` is unused now, right? I guess we can leave it in just to make future merges easier.

> * `ReflectionInstrumenter.scala:180`: Same issue here - Now that we expand the definition of `Value.This` to also include self-refs, we need to include `Value.This` when transforming paths. How do we reconcile this expansion of definitions if the legacy `Value.This` is not supported by the instrumenter?

It's okay to break `ReflectionInstrumenter` and add `:fixme`s to the students to address later. Unless the fix is simple enough.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -181,28 +170,48 @@ class JSBuilder(using Config, TL, State, Ctx) extends CodeBuilder:
...
       else errExpr(msg"Illegal reference to builtin symbol '${l.nme}'")
-    case Value.Ref(l, disamb) => l match
-      case l: BlockMemberSymbol if disamb.exists(_.shouldBeLifted) =>
-        doc"${getVar(l, l.toLoc)}.class"
-      case _ =>
-        getVar(l, r.toLoc)
-    case Call(Value.Ref(l: BuiltinSymbol, _), (lhs :: rhs :: Nil) :: Nil) if !l.functionLike =>
+    case Value.SimpleRef(l: semantics.TermSymbol) =>
+      l.owner match
+      case S(owner) =>
```
LPTK: I believe this should now be dead code, since we no longer use plain references for private fields. You can try `lastWords` here.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -350,8 +359,12 @@ class JSBuilder(using Config, TL, State, Ctx) extends CodeBuilder:
     case Assign(l, r, rst) if l is State.noSymbol =>
       doc" # ${result(r)};${returningTerm(rst, endSemi)}"
     case Assign(l, r, rst) =>
-      doc" # ${getVar(l, l.toLoc // TODO: improve location
-        )} = ${result(r)};${returningTerm(rst, endSemi)}"
+      doc" # ${
+          l match
+            case ts: TermSymbol => result(ts.asSimpleRef)
+            case sym: InnerSymbol => lastWords(s"Inner symbol should not be used as the target of an assignment: $sym")
+            case l => result(l.asPath.withLoc(l.toLoc)) // TODO: improve location
```
LPTK: It is wrong to take the loc of a symbol (which might well be in a different file) and use it as the loc of a local piece of code. `N` is better than something wrong.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
Derppening: Replacing the body of the `case S(owner)` arm results in this:

```
[info] - mlscript-compile/TreeTracer *** FAILED ***
[info]   java.lang.Exception: Internal Error: assumption broken: `term:TreeTracer/steps` has owner `class:TreeTracer`
[info] - mlscript-compile/Runtime *** FAILED ***
[info]   java.lang.Exception: Internal Error: assumption broken: `term:Runtime/curEffect` has owner `module:Runtime`
```

Derppening:
> > Block.scala:1027: The deprecated Ref.unapply is unable to distinguish between a "legacy" Value.This (that only represents this and nothing else) and a "new" Value.This (which represents self-refs in general). Does it matter, considering it's just a transitional shim?
> 
> Not sure I understand the question. There's only one self reference This in my mind.

My original concern is:

> In this PR, we converted all instances of `Value.Ref(InnerValue, _)` to be represented by `Value.This` instead. As such, our `Value.Ref.unapply` shim also needs to correctly convert `Value.This(_)` into an `InnerSymbol -> N` (i.e. no `disamb`) to be fully backwards compatible.
> 
> However, the current implementation of `Value.Ref.unapply` does not fully preserve this behavior. Consider the following match expression:
> 
> ```scala
> v match
>   case Value.Ref(s: InnerSymbol, _) => /* ... */
>   case Value.This(s: InnerSymbol) => /* ... */
> ```
>
> In the past, since some `InnerSymbol`s are represented as `Ref`s and others are represented by `Value.This`, the presence of both branches are meaningful. However, with the new `Value.Ref.unapply` shim, we treat `Value.Ref(s: InnerSymbol, _)` as an effectively-unreachable (it always returns `N`), so only the `Value.This` branch will match, and may result in divergent behavior.

But on second thought, I guess it might make more sense to just have `Value.This(InnerSymbol)` unapply to `InnerSymbol -> N`, and assume that if the code based on the legacy `Value.Ref` semantics treats certain kinds of `This` in a special way then it is an error(?)

The change is made in b913336 for your reference.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
Derppening: Addressed in 5125b8769aeccf8e38a620ebc0df191917fd2760. We now pass `N` into `withLoc`.

Derppening:
> > There are no more callers of `Symbol.orElseDisamb`. Perhaps we can now remove that?
> 
> Yes! Please do.

Done in f7e57949aa3109b19b50d7c532c35ba0bff83da3.

> By the way, this shim for `Ref` is unused now, right? I guess we can leave it in just to make future merges easier.

Yes, it is unused, and yes I plan to keep it for future merges.

> > * `ReflectionInstrumenter.scala:180`: Same issue here - Now that we expand the definition of `Value.This` to also include self-refs, we need to include `Value.This` when transforming paths. How do we reconcile this expansion of definitions if the legacy `Value.This` is not supported by the instrumenter?
> 
> It's okay to break `ReflectionInstrumenter` and add `:fixme`s to the students to address later. Unless the fix is simple enough.

Added `:fixme` to failing block staging tests in b1badad4a93f2b1052c3987be40b40719d18405b.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -129,11 +129,13 @@ class BlockSimplifier
...
                 case _ =>
-            case Value.Ref(loc, _) =>
+            case Value.SimpleRef(loc) =>
+              usedVars += loc
+            case Value.MemberRef(loc, _) =>
               usedVars += loc
             case _ =>
           super.applyPath(p)
-
+
```
Derppening, line 138: I hope *adding* indentation isn't as much of an issue as removing them...

LPTK:
> But on second thought, I guess it might make more sense to just have `Value.This(InnerSymbol)` unapply to `InnerSymbol -> N`, and assume that if the code based on the legacy `Value.Ref` semantics treats certain kinds of `This` in a special way then it is an error(?)

Yes, who cares... There was never supposed to be self references with a disamb! We just fix it if it arises. Is `Ref.unapply` even still used at the moment?

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
LPTK: Pasting these four lines of output here is not helpful, as they're completely devoid of context. Pls instead tell me in which cases an owned symbol is ever referred to as a plain reference.

Derppening:
> > But on second thought, I guess it might make more sense to just have `Value.This(InnerSymbol)` unapply to `InnerSymbol -> N`, and assume that if the code based on the legacy `Value.Ref` semantics treats certain kinds of `This` in a special way then it is an error(?)
> 
> Yes, who cares... There was never supposed to be self references with a disamb! We just fix it if it arises. Is `Ref.unapply` even still used at the moment?

No, `Value.Ref.unapply` is unused in this branch.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
Derppening: According to Claude (since I am not familiar with the recent changes):

> An owned `TermSymbol` reaches the `case S(owner)` arm as a plain `SimpleRef` precisely when it is **not** `isPrivate` — `Symbol.scala:319`:
>
>  ```scala
>  def isPrivate: Bool = (k is LetBind) && owner.exists(!_.isInstanceOf[TopLevelSymbol])
>  ```
>
>  `Lowering.privateFieldSelfSelection` rewrites a ref to `Select`/`AssignField` only when `sym.isPrivate`; everything else falls through to `sym.asSimpleRef`. So "no plain refs for private fields" holds for *private user references*, but two producers still land here, one per sub-branch:
>
>  **1. `else` (non-private) branch — emits `this.name` / `Owner.name`.** Only public **`mut val`** fields (`k = MutVal`, never `LetBind`, so never `isPrivate`), read or `set` inside the owner's own methods. These are the `Runtime/curEffect` and `TreeTracer/steps` failures — `lastWords` here breaks them.
>
>  ```fs
>  module Timer with
>    mut val ticks = 0
>    fun tick() = set ticks += 1
>  // inside tick(): Timer.ticks = Timer.ticks + 1   <- else branch
>  ```
>
>  A `let` member does *not* reach the arm: a module/class is not a `TopLevelSymbol`, so `let` is `isPrivate` and is rewritten to a `Select` during lowering (`module M with let secret = 3; fun get() = secret` → `M.#secret`, no warning).
>
>  **2. `if l.isPrivate` branch — emits `this.#name` / `Owner.#name`.** Never reached from user code (private refs become `Select`s), only from JSBuilder's own accessor synthesis in `mkPrivs`, which feeds the *private backing symbol* through `result`:
>
>  ```scala
>  doc"...get ...() { return ${result(letSym.asSimpleRef)}; }"
>  doc"...set ...(value) { ${result(letSym.asSimpleRef)} = value; }"
>  ```
>
>  `letSym = TermSymbol(LetBind, S(isym), …)` is `LetBind` + non-top-level owner ⇒ `isPrivate` ⇒ this branch. Two cases:
>
>  - Getter/setter for every public `mut val` — fires even with zero references (`class Box with mut val v = 0` → `get v(){ return this.#v } set v(value){ this.#v = value }`). This is why a bare `class Foo(x, val y, mut val z)` warns twice with no method touching `z`.
>  - `privateAccessors` Symbol-keyed get/set, generated when a private field must be reached from outside its defining class — i.e. `:lift` floating a nested class out, or `:deforest` hoisting code out of the method. This is what `AccessorBox/value` and `M/_private` in the test diffs actually are: `get [value$accessorSymbol]() { return this.#value; }`.
>
>  **Bottom line:** the arm isn't dead. The `else` branch is the genuine access path for public `mut val` fields. The only reducible part is the `isPrivate` sub-branch — `mkPrivs`/`privateAccessors` construct `SimpleRef`s of private backing symbols and re-decode them here, so those call sites could emit `this.#name` / `[accessorSymbol]` directly, leaving this arm to handle only the public-`mut val` case.

I can quickly work on inlining the `isPrivate` sub-branch to `mkPrivs` and `privateAccessors`.

Derppening: 426856c59fc4f6681c8954272a55695bcc71f0e6 and ce70bd14ee89ef438b030964700ef6585dcf890b addresses this issue. However, I am not sure if this is the "correct" way of fixing this - Please let me know.

hkmc2/shared/src/main/scala/hkmc2/codegen/llir/Builder.scala
LPTK: I'll delete it after your PR is merged.

hkmc2/shared/src/main/scala/hkmc2/codegen/ReflectionInstrumenter.scala
```diff
@@ -300,7 +300,10 @@ class ReflectionInstrumenter(using State, Raise, Ctx) extends BlockTransformer(n
           transformResult(r): y =>
             transformSymbol(ts): xSym =>
               blockCtor("ValueRef", Ls(xSym)): xStaged =>
-                (Assign(ts, xStaged, _)):
+                // * Reflect the binding as the private field assignment it is, so the
+                // * owned field symbol is selected on its owner rather than emitted as a
+                // * plain reference (which would otherwise reach `JSBuilder`'s owned-`SimpleRef` path).
+                ((cont: Block) => AssignField(lhs, nme, xStaged, cont)(S(ts))):
```
LPTK, line 306: Why do we have a `ValueRef` of a private filed in the first place? 🤔

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
LPTK: It seems good enough for now.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
LPTK: Yes, this seems to be the correct fix. Thanks!

LPTK: BTW it seems Claude is just plain wrong about some things. Its own Case 1 example with `Timer` does not seem to generate any such plain reference, except in JSBuilder, which is just Case 2.

hkmc2/shared/src/main/scala/hkmc2/codegen/ReflectionInstrumenter.scala
Derppening: I refactored it to use the newly introduced `ValueSimpleRef` in ce49ad534ae25316580499a3e2978e8e1e6cb7ba.

As for why we have a `ValueRef` (or now a `ValueSimpleRef`) of a private field, I am guessing the TODO comment above may have something to do with this?

> ```scala
>       // TODO: Improve. This is a kludge to allow private field initialization in modules;
>       //    Ideally, we should just properly reflect these as the private field assignments they are
> ```

hkmc2/shared/src/main/scala/hkmc2/codegen/SymbolRefresher.scala
```diff
@@ -202,21 +202,24 @@ class SymbolRefresher(existingMapping: Map[Symbol, Symbol])(using State) extends
...
-              mapping.get(oldDisamb) match
-                case Some(nd: DefinitionSymbol[?]) => Some(nd)
-                case _ => newBms.tsym.orElse(x)
-            case None => newBms.tsym
-          k(Value.Ref(newBms, newDisamb))
-        case Some(newSym) => k(Value.Ref(newSym, N))
+          val newDisamb = disamb match
+            case d => mapping.get(d) match
+              case Some(nd: DefinitionSymbol[?]) => nd
+              case _ => newBms.tsym.getOrElse(disamb)
```
LPTK:
```suggestion
              case N => d
              case N => lastWords("TODO")
```

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -1075,51 +1080,24 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
-                      l.toLoc,
-                  ),
-                  extraInfo = S(
-                    s"Locals: ${(funcCtx.params ++ funcCtx.locals).toString}\nGlobals: ${ctx.getGlobals.toString}",
-                  ),
-                )
-  end getVar
+  def getVar(l: Local, loc: Opt[Loc])(using Ctx, FunctionCtx, Raise): Expr = l match
+    case ts: (semantics.TermSymbol | semantics.InnerSymbol) => 
+      lastWords(s"Symbol `$ts` (${ts.getClass.getSimpleName}) cannot be resolved as a variable")
```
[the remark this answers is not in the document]
Derppening: This path is indeed unreachable and is an internal compiler error, so `lastWords` is appropriate here.

hkmc2/shared/src/test/mlscript/block-staging/StageSymbols.mls
```diff
@@ -1,14 +1,14 @@
...
   class C with
     fun f() = 1
     fun g() = f()
-//│ > fun ctor_() =
-//│ >   class C with
-//│ >     fun f() = 1
-//│ >     fun g() = C.f()
+//│ ═══[COMPILATION ERROR] Value.This not supported in staged module.
+//│ ═══[COMPILATION ERROR] Value.This not supported in staged module.
+//│ ═══[RUNTIME ERROR] Error: MLscript call unexpectedly returned `undefined`, the forbidden value.
```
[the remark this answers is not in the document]
Derppening: As agreed offline, this is fine for now.

hkmc2/shared/src/test/mlscript/block-staging/SymbolRenaming.mls
```diff
@@ -32,12 +32,11 @@ staged module Selection with
...
   class C(val y)
   fun f() = C(1).y
+//│ ═══[COMPILATION ERROR] Value.This not supported in staged module.
 //│ > fun ctor_() =
 //│ >   class C(y)
-//│ > fun f() =
-//│ >   let {tmp}
-//│ >   tmp = DupeClass.C(1)
-//│ >   tmp.y
+//│ ═══[RUNTIME ERROR] Error: MLscript call unexpectedly returned `undefined`, the forbidden value.
```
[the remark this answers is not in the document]
Derppening: See above.

hkmc2/shared/src/test/mlscript/block-staging/Functions.mls
```diff
@@ -89,13 +90,13 @@ staged module ClassInstrumentation with
...
+//│ ═══[COMPILATION ERROR] Value.This not supported in staged module.
+//│ ═══[COMPILATION ERROR] Value.This not supported in staged module.
 //│ > fun ctor_() =
 //│ >   class Inside(a1, b)
 //│ >   class NoArg
 //│ > fun inst1() = new Outside(1)
-//│ > fun inst2() = new ClassInstrumentation.NoArg
-//│ > fun app1() = Outside(1)
-//│ > fun app2() = ClassInstrumentation.Inside(1, 2)
+//│ ═══[RUNTIME ERROR] Error: MLscript call unexpectedly returned `undefined`, the forbidden value.
```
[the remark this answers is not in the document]
Derppening: See above.

hkmc2/shared/src/main/scala/hkmc2/codegen/SymbolRefresher.scala
Derppening: Addressed in 7448167e06459925bfebb3694fe34563830f51b3.

hkmc2/shared/src/main/scala/hkmc2/codegen/SymbolRefresher.scala
```diff
@@ -210,12 +210,11 @@ class SymbolRefresher(existingMapping: Map[Symbol, Symbol])(using State) extends
     case Value.MemberRef(bms, disamb) =>
       mapping.get(bms) match
         case Some(newBms: BlockMemberSymbol) =>
-          val newDisamb = disamb match
-            case d => mapping.get(d) match
-              case Some(nd: DefinitionSymbol[?]) => nd
-              case _ => newBms.tsym.getOrElse(disamb)
+          val newDisamb = mapping.get(disamb) match
+            case Some(nd: DefinitionSymbol[?]) => nd
+            case _ => newBms.tsym.getOrElse(lastWords(s"Expected disamb to be present for newBms: `$newBms`"))
```
LPTK: This logic is just plain wrong. You should never have to try to access `newBms.tsym`. How do you even know that it's the correct disamb? If the original disamb was the class symbol, and the mapping does not map that class symbol to something else, here you will revert to the term symbol instead of the original class symbol?!

PR #492, opened 2026-05-20
Overhaul handling of private fields

LPTK:
### Change 1: Represent private field accesses as actual field accesses

Previously, private fields were treated, in the IR, as thought hey were plain variables. This required special handling of them as "non-local variables" in BlockSimplifier. It also caused the lifter to create unnecessary capture classes.

But semantically, these should really be viewed as field reads and writes. This PR effects that change.

~~Note that a lot of the special-cases introduced by this PR to preserve existing semantics (of things like the lifter and deforestation) will then be removed by the follow-up PR https://github.com/hkust-taco/mlscript/pull/494, which implements our plan to automatically create private field accessors when needed.~~ [now merged into this PR]

### Change 2: Generate accessors for extruded private fields

References to private fields sometimes have to get extruded from their scope, due to things like lifting, deforestation, and (eventually) method inlining.

This PR allows the IR to refer to a class' private fields directly. When targeting JS, we now generates appropriate accessors to compile these private field accesses. We only generate those accessors that are needed for the generated code to work.

### Other changes

We now also eliminate unused private fields as part of the simple DCE pass.

---

Co-coded with Codex.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
@@ -304,7 +304,7 @@ class BlockSimplifier
     //    Note that the capturing definitions won't see the assignments of the captured variable anyway
     //    because that variable will be treated as unknown, since nested definitions start from an empty environment.
     
-    val unstableRefs = MutSet.empty[Local]
+    // val unstableRefs = MutSet.empty[Local]
     // ^ We currently represent private fields using plain TermSymbol references,
     //    so these have to be special-cased as 'unstable'.
     //    TODO: represent them as fields, as they should be!
@@ -326,14 +326,14 @@ class BlockSimplifier
           case _ =>
           super.applyDefn(defn)
         
-        // This override can go once we get rid of `unstableRefs`
-        override def applyBlock(b: Block): Unit =
-          b match
-          case Assign(_: LocalVar, _, _) =>
-          case Assign(lhs, _, _) =>
-            unstableRefs += lhs
-          case _ =>
-          super.applyBlock(b)
+        // // This override can go once we get rid of `unstableRefs`
+        // override def applyBlock(b: Block): Unit =
+        //   b match
+        //   case Assign(_: LocalVar, _, _) =>
+        //   case Assign(lhs, _, _) =>
+        //     unstableRefs += lhs
+        //   case _ =>
+        //   super.applyBlock(b)
 
         override def applyLam(lam: Lambda): Unit =
           capturedVars ++= lam.freeVars.iterator.collect { case v: LocalVar => v }
@@ -476,8 +476,8 @@ class BlockSimplifier
             else
               val rhs2 = assignedResults(sym)
               S(r -> rhs2)
-          case r @ Value.Ref(sym, _) if unstableRefs(sym) =>
-            N
+          // case r @ Value.Ref(sym, _) if unstableRefs(sym) =>
+          //   N
           case r @ Value.Ref(sym, _) =>
             S(r -> Unknown)
           case _ => N
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
@@ -125,6 +125,12 @@ class BlockSimplifier
         
         override def applyPath(p: Path): Unit =
           p match
+            case sel: Select =>
+              sel.symbol.foreach:
+                case ts: TermSymbol if (ts.k is syntax.LetBind)
+                    && ts.owner.exists(!_.isInstanceOf[TopLevelSymbol]) =>
+                  usedVars += ts
+                case _ =>
             case Value.Ref(loc, _) =>
               usedVars += loc
             case _ =>
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
@@ -115,22 +115,8 @@ class BlockSimplifier
         
         applyProgram(prog)
         
-        override def applyDefn(defn: Defn): Unit =
-          defn match
-          case cls: ClsLikeDefn =>
-            localVars ++= cls.privateFields
-            cls.companion.foreach(localVars ++= _.privateFields)
-          case _ =>
-          super.applyDefn(defn)
-        
         override def applyPath(p: Path): Unit =
           p match
-            case sel: Select =>
-              sel.symbol.foreach:
-                case ts: TermSymbol if (ts.k is syntax.LetBind)
-                    && ts.owner.exists(!_.isInstanceOf[TopLevelSymbol]) =>
-                  usedVars += ts
-                case _ =>
             case Value.Ref(loc, _) =>
               usedVars += loc
             case _ =>
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
@@ -304,12 +304,6 @@ class BlockSimplifier
     //    Note that the capturing definitions won't see the assignments of the captured variable anyway
     //    because that variable will be treated as unknown, since nested definitions start from an empty environment.
     
-    // val unstableRefs = MutSet.empty[Local]
-    // ^ We currently represent private fields using plain TermSymbol references,
-    //    so these have to be special-cased as 'unstable'.
-    //    TODO: represent them as fields, as they should be!
-    //    TODO: then, the symbol type in `Assign` should be tightened to `LocalVar`
-    
     
     def apply(prog: Program): Program =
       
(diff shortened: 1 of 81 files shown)


Derppening:
Acknowledged.

CAG2Mark (commented):
The lifter changes LGTM. I have few small nitpicks but those will be resolved by the private field accessors PR anyway.

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
```diff
@@ -273,7 +291,11 @@ class UsedVarAnalyzer(b: Block, scopeData: ScopeData)(using State):
...
       case _: ScopedObject.Companion | _: ScopedObject.Class =>
-        val (a, b) = s.children.map: c =>
+        val classCaptureChildren = s.children.filterNot: c =>
+            c.obj match
+              case ScopedObject.Func(_, S(MethodKind.ClsMethod | MethodKind.ObjMethod)) => true
+              case _ => false
+        val (a, b) = classCaptureChildren.map: c =>
             val acc = accessMap(c.obj.toInfo)
             val accAll = accessMapWithIgnored(c.obj.toInfo)
             (accAll.mutated, acc.accessed)
```
CAG2Mark: Maybe this will make the logic a bit easier to follow.
```suggestion
        val classCaptureChildren = s.children.iterator.map(_.obj).collect:
          case c @ ScopedObject.Func(_, S(MethodKind.ClsMethod | MethodKind.ObjMethod)) => c
        val (a, b) = classCaptureChildren.map: c =>
            val acc = accessMap(c.toInfo)
            val accAll = accessMapWithIgnored(c.toInfo)
            (accAll.mutated, acc.accessed)
          .toList.unzip
```

CAG2Mark:
ah, I was a bit tired and realized I inverted the logic in my comment :upside_down_face: we should keep using the filter.

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
```diff
@@ -273,11 +277,13 @@ class UsedVarAnalyzer(b: Block, scopeData: ScopeData)(using State):
     // In a class, all variables that are mutated by a child scope and accessed by a lifted class must be captured
     val additional = s.obj match
       case _: ScopedObject.Companion | _: ScopedObject.Class =>
-        val (a, b) = s.children.map: c =>
-            val acc = accessMap(c.obj.toInfo)
-            val accAll = accessMapWithIgnored(c.obj.toInfo)
+        def classCaptureChildren = s.children.iterator.map(_.obj).collect:
+          case c @ ScopedObject.Func(_, S(MethodKind.ClsMethod | MethodKind.ObjMethod)) => c
```
CAG2Mark:
```suggestion
        def classCaptureChildren = s.children.iterator.map(_.obj).filter:
          case c @ ScopedObject.Func(_, S(MethodKind.ClsMethod | MethodKind.ObjMethod)) => false
          case _ => true
```

LPTK:
> ah, I was a bit tired and realized I inverted the logic in my comment 🙃 we should keep using the filter.

Huh, why does this incrrect logic result in no changes in the tests, which are all green? Can you find a way to properly test these paths in our test suite?

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
LPTK: Another broken suggestion 😛

CAG2Mark:
Oh, you merged that PR that generates accessors for private fields, which basically makes that change do nothing.

I can clean up the lifter logic a bit then to remove the now unused stuff.

LPTK:
> Oh, you merged that PR that generates accessors for private fields, which basically makes that change do nothing.
> 
> I can clean up the lifter logic a bit then to remove the now unused stuff.

Ah, ok, great; please do that.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -67,12 +69,94 @@ class JSBuilder(using Config, TL, State, Ctx) extends CodeBuilder:
...
+  private def isLexicallyInside(owner: InnerSymbol)(using Scope): Bool =
+    @tailrec
+    def go(scope: Scope): Bool =
+      scope.curThis match
+      case S(S(sym)) if sym is owner => true
+      case _ =>
+        scope.parent match
+        case S(parent) => go(parent)
+        case N => false
+    go(scope)
```
LPTK: Instead of repeatedly traversing the stack of scopes looking for the owner, why not add a `lazy val inScopeOwners: Set[InnerSymbol]` to `Scope` and just check that instead?

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -67,12 +69,94 @@ class JSBuilder(using Config, TL, State, Ctx) extends CodeBuilder:
...
+
+  private def privateAccessorSymbol(ts: semantics.TermSymbol): semantics.TempSymbol =
+    privateAccessorSymbols.getOrElseUpdate(ts, semantics.TempSymbol(N, "accessorSymbol$"))
```
LPTK: Please include the name of the field in the name of the accessor, for easier debugging.
```suggestion
    privateAccessorSymbols.getOrElseUpdate(ts, semantics.TempSymbol(N, ts.name + "$accessorSymbol"))
```

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -67,12 +69,94 @@ class JSBuilder(using Config, TL, State, Ctx) extends CodeBuilder:
...
+    def go(scope: Scope): Bool =
+      scope.curThis match
+      case S(S(sym)) if sym is owner => true
+      case _ =>
+        scope.parent match
+        case S(parent) => go(parent)
+        case N => false
+    go(scope)
+
+  private def privateAccessorSymbol(ts: semantics.TermSymbol): semantics.TempSymbol =
```
LPTK: A more adequate name would be `getPrivateAccessorSymbol`.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -67,12 +69,94 @@ class JSBuilder(using Config, TL, State, Ctx) extends CodeBuilder:
...
+    privateAccessorSymbols.getOrElseUpdate(ts, semantics.TempSymbol(N, "accessorSymbol$"))
+
+  private def privateAccessorName(ts: semantics.TermSymbol, loc: Opt[Loc])(using Raise, Scope): Document =
+    doc"${scope.lookup_!(privateAccessorSymbol(ts), loc)}"
+
+  private def privateFieldSelect(ts: semantics.TermSymbol, loc: Opt[Loc])(using Raise, Scope): Opt[Document] =
```
LPTK: A more adequate name would be `selectPrivateField`.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -67,12 +69,94 @@ class JSBuilder(using Config, TL, State, Ctx) extends CodeBuilder:
...
+    ts.owner.collect:
+      case owner if ts.isPrivate =>
+        if isLexicallyInside(owner)
+        then doc".#${owner.privatesScope.lookup_!(ts, loc)}"
+        else doc"[${privateAccessorName(ts, loc)}]"
+
+  private def privateAccessorDecls(using Raise, Scope): Document =
```
LPTK: This is used only one; just inline it.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -160,14 +244,18 @@ class JSBuilder(using Config, TL, State, Ctx) extends CodeBuilder:
...
       val name = id.name
-      doc"${resultQual(qual)}${
+      val fieldDoc = field.getOrElse {
         if isValidFieldName(name)
         then doc".$name"
         else name.toIntOption match
           case S(index) => doc"[$index]"
           case N => doc"[${makeStringLiteral(name)}]"
-      }${dotClass}"
+      }
```
```suggestion
      val fieldDoc = field.getOrElse:
        if isValidFieldName(name)
        then doc".$name"
        else name.toIntOption match
          case S(index) => doc"[$index]"
          case N => doc"[${makeStringLiteral(name)}]"
```

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -142,7 +147,10 @@ class BlockSimplifier
               definedVars += lhs
             case _ =>
           super.applyBlock(b)
-      
+
+      unusedPrivateVars =
```
LPTK: This should be a `val` defined here, not a mutable variable.

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
```diff
@@ -60,6 +60,10 @@ class UsedVarAnalyzer(b: Block, scopeData: ScopeData)(using State):
           accessed.mutated.add(lhs)
           applyResult(rhs)
           applyBlock(rest)
+        case AssignField(lhs, _, rhs, rest) =>
+          applyPath(lhs)
+          applyResult(rhs)
+          applyBlock(rest)
```
LPTK: Is this override necessary?

LPTK (commented):
@copilot Please address these review comments.

CAG2Mark:
See my branch for the lifter changes. https://github.com/CAG2Mark/mlscript/tree/private-fields-overhaul

PR #493, opened 2026-05-20
Eta expansion

ychenfo:
This PR adds affine function annotations and enables eta expansion for affine functions and lambdas. Also, deforestation now does not generate one lambda for each constructor call — instead constructor calls are rewritten into partially applied functions which generates affine branch body functions that can get eta-expanded.


hkmc2/shared/src/main/scala/hkmc2/codegen/EtaExpansion.scala
```diff
@@ -0,0 +1,247 @@
...
+    cfg: Config,
+    tl: TL,
+    raise: Raise,
+    eState: Elaborator.State,
+    symbolPrinter: SymbolPrinter,
+  ): Program =
+    (cfg.etaExpansion, cfg.deforest) match
+    case (N, N) => p
+    case (S(_), _) | (N, S(_)) =>
+      val eCfg = cfg.etaExpansion.getOrElse(hkmc2.Config.EtaExpansion.default)
```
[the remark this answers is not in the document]
ychenfo: Good point, I have fixed this at https://github.com/hkust-taco/mlscript/pull/493/commits/0bc21dc1df566dca94192ca69adfe0c1cad3cbd3 by setting eta expansion always on by default and allowing disabling it when deforest is on.

hkmc2DiffTests/src/test/scala/hkmc2/MLsDiffMaker.scala
```diff
@@ -462,4 +462,3 @@ abstract class MLsDiffMaker extends DiffMaker:
           doc" #{ ${trm.showTopLevel(using flowScp)} #} \nwhere #{ ${floan.showFlows(using flowScp)} #} ".mkString()
     
   
-
```
LPTK, line 453: 👎

LPTK:
Pls add a short PR description.

LPTK:
@ychenfo Please also merge with the main branch. I tried doing so, but it raised some new failures.

Note that the main branch has implemented unsaturated call propagation!

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -25,6 +25,7 @@ enum Annot extends AutoLocated:
   case TailRec
   case TailCall
   case Config(modify: hkmc2.Config => hkmc2.Config)
+  case Affine(whichParamList: Int)
```
LPTK, line 33: This needs some documentation!!

hkmc2/shared/src/main/scala/hkmc2/codegen/LambdaRewriter.scala
```diff
@@ -15,17 +15,18 @@ object LambdaRewriter:
...
-          val lamDefn = FunDefn.withFreshSymbol(N, sym, params :: Nil, body)(N, annotations = Nil)
+          val lam2 = super.applyLam(lam)
+          val Lambda(params, body) = lam2
+          val lamDefn = FunDefn.withFreshSymbol(N, sym, params :: Nil, body)(N, annotations = lam2.annot)
           Scoped(Set.single(sym), Define(lamDefn, k(lamDefn.asPath)))
         case _ => super.applyResult(r)(k)
       
       override def applyBlock(b: Block): Block = b match
-        case Assign(lhs, Lambda(params, body), rest) if !lhs.isInstanceOf[TempSymbol] =>
+        case Assign(lhs, lam@Lambda(params, body), rest) if !lhs.isInstanceOf[TempSymbol] =>
```
LPTK: Please use spaces arount `@`, here and everywhere. This is quite hard to read!
```suggestion
        case Assign(lhs, lam @ Lambda(params, body), rest) if !lhs.isInstanceOf[TempSymbol] =>
```

LPTK (approved):
Otherwise LGTM!

hkmc2/shared/src/main/scala/hkmc2/codegen/LambdaRewriter.scala
ychenfo: Done in https://github.com/hkust-taco/mlscript/pull/493/commits/37478274fcc8825858a9469a1cf13b8207aeebdf

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
ychenfo: Done in https://github.com/hkust-taco/mlscript/pull/493/commits/37478274fcc8825858a9469a1cf13b8207aeebdf

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -28,6 +28,8 @@ enum Annot extends AutoLocated:
   case TailCall
   case Inline
   case Config(modify: hkmc2.Config => hkmc2.Config)
+  // marks if a function or lambda is affine, i.e. called at most once.
+  // for functions with multiple parameter lists, `whichParamList` is the zero-based parameter-list index.
```
LPTK, line 32: The index of what??

Also, shouldn't we call this property "one-shot", following the standard terminology?

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Rewrite.scala
```diff
@@ -180,8 +180,10 @@ class DeforestRewriter(val solver: DeforestFusionSolver)(using Raise):
             val branchName = whichBranch.fold("_dflt")(c => s"_${c.ctorClsName}")
             val scrutName = dest._1.getReferredSym.nme
             val branchFnNme = s"${dest.instId.mkFunName}$$$scrutName$branchName"
+            val owner = pre.res.matchScrutToCtxOfMatch(dest._1).collectFirst:
```
LPTK, line 183: Note: `matchScrutToCtxOfMatch` and all the other vals/vars defined alongside it will eventually need to be properly documented, similar to how the lifter is.

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Rewrite.scala
```diff
@@ -560,15 +583,26 @@ class DeforestRewriter(val solver: DeforestFusionSolver)(using Raise):
             case Some(rewrittenBody) =>
               FunDefn(fun.owner, fun.sym, fun.dSym, fun.params, rewrittenBody)(fun.configOverride, fun.annotations)
             case None => super.applyFunDefn(fun)
+        override def applyClsLikeDefn(defn: ClsLikeDefn)(k: Defn => Block): Block =
+          super.applyClsLikeDefn(defn): transformed =>
+            transformed match
+            case cls: ClsLikeDefn =>
+              newClassMethods.get(cls.isym) match
+              case Some(methods) =>
+                k(cls.copy(methods = cls.methods ++ methods)(cls.configOverride, cls.annotations))
```
LPTK, line 592: Actually, can you explain why we're adding methods, here? I have no idea what's going on 🤔 The important high-level aspects of the code in general needs to be documented.

LPTK:
I'm merging now because this is quite imporant and urgent, but these comments need to be addressed, eventually.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
ychenfo: Addressed in https://github.com/hkust-taco/mlscript/pull/499/changes/35b4866176feded1ddb10489424ba1e3e1bba487#diff-3adee776a5273d49f25c89e6d862306fa4dca581514de0321b57169cfcc4d319R32

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Rewrite.scala
ychenfo: When rewriting class methods, the generated branch-body functions may include selections of the class fields so making those branch-body functions top-level does not work because it loses access to those fields, therefore in this patch I added them as methods instead.

I have also added comments related to this in the new PR https://github.com/hkust-taco/mlscript/pull/499.

LPTK: But did you pay attention to my recent changes to private fields? You can now access them from the outside. Would things be much simplified if we removed this method-insertion logic? If so, we should probably do it, as the deforestation code is already complex and hard to maintain enough as it is.

PR #494, opened 2026-05-20
Generate accessors for external private fields

LPTK:
References to private fields sometimes have to get extruded from their scope, due to things like lifting, deforestation, and (eventually) method inlining.

This PR allows the IR to refer to a class' private fields directly. When targeting JS, we now generates appropriate accessors to compile these private field accesses. We only generate those accessors that are needed for the generated code to work.

We also now eliminate unused private fields as part of the simple DCE pass.


LPTK:
Since the making of this PR ended up finding and fixing other places where private fields were wrongly still accessed as local variables, I'll just roll it into https://github.com/hkust-taco/mlscript/pull/492 instead of merging it separately.

PR #495, opened 2026-05-20
Implement inline wrappers

AnsonYeung:
Add a new annotation `@inline` on functions that
1. Block inlining inside the body of the function (unless the functions considered for inlining are themselves `@inline`)
2. Make the function itself always eligible for inlining in terms of bypassing size checks.

Also move all annotations to be under the `annotations` virtual module.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
@@ -610,8 +610,6 @@ sealed abstract class Defn:
         -- auxParams.flatMap(_.paramSyms)
   
 
-// NOTE: Setting isTailRec to false does not affect whether the function is optimized.
-// It only affects whether a warning is thrown if the function is not actually tailrec.
 final case class FunDefn(
     owner: Opt[InnerSymbol],
     sym: BlockMemberSymbol,
@@ -625,6 +623,7 @@ final case class FunDefn(
   val innerSym = N
   val asPath = Value.Ref(sym, S(dSym))
   lazy val forceTailRec: Bool = annotations.contains(Annot.TailRec)
+  lazy val inline: Bool = annotations.contains(Annot.Inline)
   lazy val visibility: Visibility = annotations.collectFirst:
     case Annot.Modifier(Keyword.`private`) => Visibility.Private
     case Annot.Modifier(Keyword.`public`) => Visibility.Public
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
@@ -622,7 +622,7 @@ final case class FunDefn(
 ) extends Defn:
   val innerSym = N
   val asPath = Value.Ref(sym, S(dSym))
-  lazy val forceTailRec: Bool = annotations.contains(Annot.TailRec)
+  lazy val tailRec: Bool = annotations.contains(Annot.TailRec)
   lazy val inline: Bool = annotations.contains(Annot.Inline)
   lazy val visibility: Visibility = annotations.collectFirst:
     case Annot.Modifier(Keyword.`private`) => Visibility.Private
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
@@ -881,7 +881,7 @@ class BlockSimplifier
         
         // Whether this function can be inlined without causing any code duplication,
         // i.e. the original definition can be removed and there is only one usage.
-        def canBeInlineEliminated =
+        def canBeInlineEliminated: Bool =
           isPrivate && !isMethod && useCount <= 1 && !disallowElimination && !isLoopBreaker
           // false
         
@@ -889,6 +889,8 @@ class BlockSimplifier
           if isLoopBreaker then return false
           // method requires the capturing of `this`, which is not supported currently.
           if isMethod then return false
+          // If the definition is marked with inline, we should inline it regardless of the size of the body.
+          if defn.inline then return true
           val threshold = summon[Config.Inliner].inlineThreshold
           newBlk.size <= threshold || canBeInlineEliminated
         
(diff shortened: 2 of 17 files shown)


hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -941,6 +944,9 @@ class BlockSimplifier
         
         override def applyResult(r: Result): Unit = r match
           case c @ Call(TermSymbolPath(ts), argss) =>
+            if currentContext.hasInlineAnnot then
+              // Not eligible for inline elimination
+              hasNakedRef(ts) = true
```
LPTK: Shouldn't `hasNakedRef` be renamed to something more descriptive, such as `ineligibleForInlining`, with a comment documenting when something is not eligible?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -1307,12 +1307,17 @@ class Lowering()(using Config, TL, Raise, State, Ctx, SymbolPrinter):
       msg match
         case S(value) => WarningReport(value -> annot.toLoc :: Nil)
         case N => WarningReport(msg"This annotation has no effect." -> annot.toLoc :: Nil)
-    annotations.foreach:
+    annotations.foreach: a =>
+      a match
```
```suggestion
    annotations.foreach:
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -1307,12 +1307,17 @@ class Lowering()(using Config, TL, Raise, State, Ctx, SymbolPrinter):
...
       case Annot.Untyped => ()
-      case a @ Annot.TailRec =>
+      case Annot.TailRec | Annot.Inline =>
+        val annot = a match
+          case Annot.TailRec => "@tailrec"
+          case Annot.Inline => "@inline"
```
LPTK: It looks like this probably raises an exhaustiveness warning, which should be fixed. The `a @` form probably doesn't. If it does, I can suggest other fixes.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -625,6 +623,7 @@ final case class FunDefn(
   val innerSym = N
   val asPath = Value.Ref(sym, S(dSym))
   lazy val forceTailRec: Bool = annotations.contains(Annot.TailRec)
```
```suggestion
  lazy val tailRec: Bool = annotations.contains(Annot.TailRec)
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -22,8 +22,10 @@ enum Annot extends AutoLocated:
   case Untyped
   case Modifier(mod: Keyword)
   case Trm(trm: Term)
+  // This is for generation of warnings and does not affect whether a function is optimized or not.
```
```suggestion
  // NOTE: The presence of TailRec and TailCall annotations does not affect whether a function is optimized or not;
  // it only affects whether a warning is thrown if the function/call is not actually tail-recursive.
```

hkmc2/shared/src/test/mlscript/codegen/CurriedFunctions.mls
```diff
@@ -45,10 +45,12 @@ foo(1)(2)(3)
...
 //│ define lambda$⁰ as fun lambda$¹(x, y)(z) {
-//│   let inlinedVal, tmp;
+//│   return lambda²(x, y, z)
+//│ };
+//│ define lambda¹ as fun lambda²(x, y, z) {
+//│   let tmp;
 //│   set tmp = +⁰(x, y);
-//│   set inlinedVal = +⁰(tmp, z);
-//│   return inlinedVal
+//│   return +⁰(tmp, z)
```
LPTK, line 53: I guess this regression can't really be dealt with easily. A bit annoying.

LPTK: We'd probably need to have a second inlining phase that disregards the `@inline` annotations of non-exported functions and still inline inside them.

hkmc2/shared/src/test/mlscript/codegen/Inliner.mls
```diff
@@ -197,6 +197,64 @@ f()
...
+//│   tmp8 = a + b;
+//│   return tmp8 + c3
+//│ };
+//│ f10 = function f(a) { return (b) => { return (c3) => { return f11(a, b, c3) } } };
+//│ 6
+//│ —————————————————| Output |—————————————————————————————————————————————————————————————————————————
+//│ = 6
+
+// lifter wrapper uses inline annotation under the hood
+// currently no inlining happens because multi-call isn't recognized
```
LPTK, line 223: Right. Reconstructing multicalls should be done by DataFlowAnalysis soon.

hkmc2/shared/src/test/mlscript/invalml/InvalMLPrelude.mls
```diff
@@ -1,16 +1,11 @@
 :ignore
```
LPTK: Can this be removed, now?

LPTK (approved):
Great, thanks. That was fast!

LGTM otherwise.

hkmc2/shared/src/test/mlscript/codegen/CurriedFunctions.mls
AnsonYeung: yeah idk about that. For non-exported `@inline` definition they are hopefully all inlined at call-site and this function will be gone, provided no funny recursion is going on. For exported `@inline` def maybe we can still inline inside `@inline` body with a even stricter quota, the lifter could generate wrapper for small functions too.

LPTK: Ah, so you're suggesting that we just use two different size thresholds – one for general inlining, and one for inlining inside `@inline` functions, right?

AnsonYeung: the second threshold is also for deciding whether to apply the wrapper-worker transformation so that instead of doing the transformation for tiny function we just mark the function as @inline, which will automatically do the uncurrying (and inlining).

LPTK: Indeed, good point. Would you like to try and add this new threshold and a worker-wrapper transformation pass? The pass should probably just run once, before block IR simplification.

PR #497, opened 2026-05-21
Add small threshold and fix issue in size calculation

AnsonYeung:
There was an issue in size calculation, which causes Define node to be inlined very aggressively. This PR fix this issue.

Also add a secondary inlining threshold for functions with `@inline` annotation, allowing some inlining to happen inside it.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
@@ -114,13 +114,15 @@ sealed abstract class Block extends Product:
     case Scoped(syms, body) => body.definedVars ++ syms
   
   lazy val size: Int = this match
-    case _: Return | _: Throw | _: End | _: Break | _: Continue | _: Unreachable => 1
+    case Return(r: Result, _) => 1 + r.size
+    case Throw(r: Result) => 1 + r.size
+    case _: End | _: Break | _: Continue | _: Unreachable => 1
     case Begin(sub, rst) => sub.size + rst.size
-    case Assign(_, _, rst) => 1 + rst.size
-    case AssignField(_, _, _, rst) => 1 + rst.size
-    case AssignDynField(_, _, _, _, rst) => 1 + rst.size
-    case Match(_, arms, dflt, rst) =>
-      1 + arms.map(_._2.size).sum + dflt.map(_.size).getOrElse(0) + rst.size
+    case Assign(_, r, rst) => 1 + r.size + rst.size
+    case AssignField(p, _, r, rst) => 1 + p.size + r.size + rst.size
+    case AssignDynField(p, fld, _, r, rst) => 1 + p.size + fld.size + r.size + rst.size
+    case Match(p, arms, dflt, rst) =>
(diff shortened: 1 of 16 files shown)


hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -608,6 +608,12 @@ sealed abstract class Defn:
       preCtor.freeVarsLLIR
         ++ ctor.freeVarsLLIR ++ methods.flatMap(_.freeVarsLLIR) ++ stat.iterator.flatMap(_.freeVarsLLIR)
         -- auxParams.flatMap(_.paramSyms)
+
+  lazy val size: Int = this match
+    case FunDefn(_, _, _, params, body) => 1 + body.size
+    case ValDefn(_, _, rhs) => 1
```
LPTK, line 616: The size of `val` definitions doesn't take into account the size of their RHS? What if the RHS is enormous?

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -608,6 +608,12 @@ sealed abstract class Defn:
...
+    case ClsLikeDefn(_, _, _, _, _, paramsOpt, auxParams, parentSym, methods, privateFields, publicFields, preCtor, ctor, stat, bufferable) =>
```
LPTK, line 617: You probably need to also take into account the "static part" `stat` (ie the companion module).

LPTK (approved):
Otherwise LGTM.

I'm surprised that this does not cause more diffs in the tests. Is that legit? 🤔

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
AnsonYeung: assign also doesn't, I think the assumption was `Result` is "small enough".
```scala
    case Assign(_, _, rst) => 1 + rst.size
```

AnsonYeung:
I'm surprised as well, I don't know why it didn't inline many stuffs with it set to 10. One possibility is the small diff test block separate the function definition to prevent inlining and there is no opportunity in the runtime file

hkmc2/shared/src/test/mlscript/codegen/Inliner.mls
LPTK, line 203: Looks like this comment is now wrong.

LPTK:
> One possibility is the small diff test block separate the function definition to prevent inlining and there is no opportunity in the runtime file

True, that's plausible. We don't yet have the infrastructure to inline across compilation units (but soon will).

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
LPTK: Ah, indeed. It would be safer to give all results size 1 except for lamdas and string literals (for the latter, we could use something like length/4).

hkmc2/shared/src/test/mlscript/codegen/Inliner.mls
AnsonYeung: i'll fix this test instead to demonstrate the different inlining limit

PR #498, opened 2026-05-22
Add worker wrapper transformation for uncurrying optimization

LPTK:
Co-coded with Codex. Codex summary below.

---

**Summary**

This PR adds a new WorkerWrapper optimization pass for hkmc2. The pass rewrites eligible curried functions into an inline wrapper plus a flattened worker function, allowing saturated calls to be inlined into direct worker calls with a more efficient calling convention.

The pass currently focuses on plain top-level curried functions. For very small function bodies, it skips worker creation and marks the original function @inline directly. The pass is also disabled when inlining is disabled, since wrappers rely on the inliner to be erased profitably.

**Details**

- Added hkmc2.codegen.WorkerWrapper.
- Wired the pass into the diff-test backend pipeline and MLsCompiler.
- Marked generated wrappers as @inline.
- Avoided worker generation for bodies below the alternative inlining threshold.
- Gated worker-wrapper execution on config.inlining.
- Adjusted inliner recursion handling so inline wrappers in recursive worker cycles can still expose their worker calls.
- Updated loop-breaker selection to prefer non-inline definitions, keeping wrappers inlineable while workers break recursive expansion.
- Updated golden outputs for affected codegen, deforest, handlers, dead-param-elim, and worker-wrapper tests.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
@@ -1027,15 +1027,20 @@ class BlockSimplifier
                   matchAllArgs(call.argss, map(sym).defn.params).isEmpty
                 caller.foreach: caller =>
                   edges.append((caller, sym))
-          
+
+          def pickLoopBreaker(sccComp: Ls[TermSymbol]): TermSymbol =
+            sccComp.minBy: sym =>
+              (if map(sym).defn.inline then 1 else 0, sym.uid)
+
           @tailrec
           def assignLoopBreakers(): Unit =
             val sccs = partitionScc(edges.filterNot((from, to) => map(to).isLoopBreaker), map.keys)
             if sccs.forall(_.sizeIs == 1) then return
             sccs.foreach: sccComp =>
               if sccComp.sizeIs > 1 then
-                // TODO: Score computation
-                map(sccComp.minBy(_.uid)).isLoopBreaker = true
+                // Prefer breaking cycles at non-inline definitions so tiny wrappers
+                // can still disappear while their workers stop recursive expansion.
+                map(pickLoopBreaker(sccComp)).isLoopBreaker = true
             assignLoopBreakers()
           edges.foreach: (from, to) =>
             if from === to then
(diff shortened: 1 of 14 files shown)


LPTK:
@AnsonYeung Before the last commit, there was a place that wasn't fully optimized. It was due to the handling of recursive functions by the inliner. Codex made a nontrivial change there, so it would be best if you could check it. Codex comment:

---

For the return ping⁵(tmp)(unused) case: it was not just threshold logic. The inliner was reaching ping again while computing ping through the recursive worker/wrapper cycle, seeing its optimized body as “in progress”, and declining to inline it. I adjusted BlockSimplifier.scala so inline wrappers can use their original body in that specific in-progress case, while loop breakers still prevent real recursive expansion. I also made SCC loop breaking prefer non-@inline definitions, so wrappers are not chosen as the thing that blocks inlining when a worker can carry the cycle instead.

The recursive golden now goes through ping$worker, and DPE removes the dead unused parameter from both workers.

AnsonYeung:
The change is not necessary, it should instead move the `isLoopBreaker` logic instead to prevent recursive expansion. Loop breaker should make the expansion non-recursive.

AnsonYeung:
Also, I think following may not uncurry properly.
```fs
fun f(a)(b) = g(a)(b)
fun g(a)(b) = f(a)(b)
```
It may be fine, I think it's hard for the function to be small enough to fit in the alternative inlining threshold and be meaningfully recursive.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -1119,13 +1140,7 @@ class BlockSimplifier
...
         override def applyResult(r: Result)(k: Result => Block): Block = r match
           case c @ Call(TermSymbolPath(ts), argss) if m.contains(ts) && argss.nonEmpty =>
-            newFunctionBody.get(ts)
-            .getOrElse:
-              newFunctionBody(ts) = N
-              val newBdy = enterFunBlock(m(ts).defn.inline, applyBlock(m(ts).defn.body))
-              newFunctionBody(ts) = S(newBdy)
-              S(newBdy)
-            .fold(super.applyResult(r)(k)): blk =>
+            inlineCandidateBody(ts).fold(super.applyResult(r)(k)): blk =>
```
AnsonYeung: Something here like `if m(ts).isLoopBreaker then return super.applyResult(r)(k)`

PR #499, opened 2026-05-24
Deforest rewriting improvements

ychenfo:
This PR fixes a deforestation bug where the rewriter wrongly turns implicit returns to explicit returns. Turning implicit returns to explicit ones is needed for branch-body and the continuation "rest" functions whose bodies are originated from top level blocks, but wrong for rewriting blocks like class constructors: it could produce explicit `return super()` and made code after `super()` unreachable, which causes bugs. This is now fixed.

This PR also avoids traversing the program when there is no fusion, and adds some high level comments for the rewriting process.

---

### TODO here and in following PRs

- [x] Temporarily revert the use of `irDefn` to re-expose the bug
- [x] Update `classCtorSymbol` to use `ClassCtorSymbol`
- [x] Remove the special method-body deforestation logic (even tually, we'll do most optimization at the top level); also, this logic would likely break when we make the analysis understand methods
- [ ] Fix this – only support it if doing so does not incur significant extra complexity
    ```fs
    module Test with
      module Test2 with
        let x = 1
        fun foo(a) = if a is
          A(y) then x + y
    Test.Test2.foo(A(2))
    ```
    From the meeting discussion:
    * reconstruct paths based on owner chains that are rooted in toplevel symbols
    * this also changes in `Rewrite`'s handling of FVs
- [ ] As part of the compiler cache's Artifact computation, store the polymorphic analysis types of each top-level-like definition
- [ ] Scope patmat branches, now that we use proper LetSplit and shouldn't share variables across splits as before


hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -28,8 +28,8 @@ enum Annot extends AutoLocated:
   case TailCall
   case Inline
   case Config(modify: hkmc2.Config => hkmc2.Config)
-  // marks if a function or lambda is affine, i.e. called at most once.
-  // for functions with multiple parameter lists, `whichParamList` is the zero-based parameter-list index.
+  // marks if a function or lambda is one-shot, i.e. called at most once.
+  // `whichParamList` is the zero-based index of the parameter list for functions with multiple parameter lists.
```
LPTK: I still don't get it, from this comment. Why is there a parameter list index?!

LPTK:
> Turning implicit returns to explicit ones is needed for branch-body and the continuation "rest" functions whose bodies are originated from top level blocks, but wrong for rewriting blocks like class constructors: it could produce explicit `return super()` and made code after `super()` unreachable, which causes bugs. This is now fixed.

Wow... If I'd known the so-called "implicit returns" would cause such problems, I would have pushed for removing them earlier. They're mostly a legacy piece of design that should probably go. They should certainly not be the cause of such extra complexity.

LPTK:
> I'd known the so-called "implicit returns" would cause such problems, I would have pushed for removing them earlier.

Ok, done. And I've merged this branch with the help of Codex, so please do review your own diff, making sure all the important relevant changes are still in.

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Rewrite.scala
```diff
@@ -359,247 +356,298 @@ class DeforestRewriter(val solver: DeforestFusionSolver)(using Raise):
...
+            returnsFromCurrentBlock = saved
+            res
+          
+          def applyBlockReturning(b: Block): Block =
+            withReturnValues(enabled = true):
+              applyBlock(b)
+          
+          private def tailResult(res: Result): Block =
+            if returnsFromCurrentBlock then Return(res)
+            else Assign(eState.noSymbol, res, End())
```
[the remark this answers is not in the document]
LPTK: Note to @ychenfo: it's probably fine to allow `return` in constructors (?).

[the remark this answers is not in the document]
ychenfo: Indeed, and this logic is removed now because it's no longer needed with implicit returns removed.

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Rewrite.scala
```diff
@@ -359,247 +356,298 @@ class DeforestRewriter(val solver: DeforestFusionSolver)(using Raise):
...
+        class Rewriter(instId: InstantiationId) extends BlockTransformer(_symSubst):
+          extension (resId: ResultId) def concreteId = ConcreteId(resId, instId)
+          private var returnsFromCurrentBlock = true
+          
+          private inline def withReturnValues[A](enabled: Bool)(body: => A): A =
+            val saved = returnsFromCurrentBlock
+            returnsFromCurrentBlock = enabled
+            val res = body
+            returnsFromCurrentBlock = saved
+            res
```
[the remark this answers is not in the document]
ychenfo: I think this is a good point, but this code is removed now as it's no longer needed with implicit returns removed.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -143,10 +134,20 @@ object TrackableSelect:
     case _ => N
 
 object CtorRef:
-  def unapply(s: Path)(using Elaborator.State): Option[ClassSymbol | ModuleOrObjectSymbol] =
-    s match
-      case RefLike(s) => s.asCls orElse s.asObj
-      case _ => None
+  /** Resolves an object reference or a class-ctor `TermSymbol` to its corresponding class/object symbol. */
+  private def classCtorSymbol(sym: Symbol)(using Elaborator.State): Opt[ClassSymbol | ModuleOrObjectSymbol] =
+    sym.asObj orElse
```
LPTK: BTW I don't think this really make sense. I see this is used for `CtorRef` in
```scala
    case Instantiate(_, CtorRef(ctor), argss) => Some(ctor -> argss.flatten)
    case Call(CtorRef(ctor), argss) => Some(ctor -> argss.flatten)
    case CtorRef(ctor) if ctor.asObj.isDefined => Some(ctor -> Nil)
```
But it never makes sense to `Instantiate` or `Call` an object! Also, It never makes sense to use a plain class reference as if it was a constructed object!

Please pay attention to this and separate the two logical situations (object ref vs class construction). Mixing everything into a big bag of symbols is exactly why so many problems slip through that are hard to debug.

hkmc2/shared/src/test/mlscript/deforest/relaxedPrograms.mls
```diff
@@ -28,6 +28,7 @@ store.foo(AA(4))
 :deforest
 :expect 5
 :noInline
+:fixme
```
LPTK, line 31: You should add the logic to treat self-references as free variables, right?

Will you fix this as part of this PR, or a follow-up? I'm fine with both options.

hkmc2/shared/src/test/mlscript/deforest/relaxedPrograms.mls
```diff
@@ -53,6 +57,7 @@ store.foo(AA(4))
 
 :deforest
 :expect 3
+:fixme
```
LPTK:
```suggestion
:fixme // requires the block above to succeed
```

LPTK (changes requested):
Is the PR otherwise ready to merge?

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
ychenfo: Thanks I see. I tried to disallow `Instantiate` or `Call` of an object, but it seems that currently there are [some places](https://github.com/hkust-taco/mlscript/blob/3391d08a6c85ac462d7d1678120817aec71c271a/hkmc2/shared/src/test/mlscript/wasm/Binaryen.mls#L16-L17) that instantiate [objects](https://github.com/hkust-taco/mlscript/blob/3391d08a6c85ac462d7d1678120817aec71c271a/hkmc2/shared/src/test/mlscript/decls/Prelude.mls#L214-L216). I am not sure — should I wait until #505 is merged to fix this here?

hkmc2/shared/src/test/mlscript/deforest/relaxedPrograms.mls
ychenfo: > You should add the logic to treat self-references as free variables, right?

Yes.

> Will you fix this as part of this PR, or a follow-up? I'm fine with both options.

I currently plan to fix this in a follow-up, maybe it's also better to fix this after #505 is merged?

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
LPTK: Well, here you can see that _obviously_ this part of the Prelude is not right... I've pushed a commit fixing it.

LPTK:
@ychenfo Pls fix the conflicts again.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
ychenfo: Thanks. I've pushed the fix to not conflate object ref vs class construction here. `Prelude.Symbol` also needs to be adjusted because it's callable, and currently I change
```
declare object Symbol with
  ...
```

to

```
declare class Symbol
declare module Symbol with
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
ychenfo: I've tried to further clarify this comment in https://github.com/hkust-taco/mlscript/pull/499/commits/b4e9fb6970df306e56e488bd5215eace22ba802d

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -133,29 +134,49 @@ object TrackableSelect:
...
+    case Value.MemberRef(_, disamb) => classCtorSymbol(disamb)
+    case Value.This(sym) => classCtorSymbol(sym)
     case s: Select => s.symbol.flatMap(classCtorSymbol)
     case _ => N
 
-object CtorCall:
-  def unapply(r: Result)(using Elaborator.State): Option[(ClassSymbol | ModuleOrObjectSymbol | Int) -> Ls[Arg]] =
+object ClassSymRef:
+  def unapply(p: Path)(using Elaborator.State): Opt[ClassSymbol] = p match
+    case Value.SimpleRef(sym) => sym.asCls
```
LPTK: This does not make sense. Classes are never acessed as simple references.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -133,29 +134,49 @@ object TrackableSelect:
...
+    case Value.MemberRef(_, disamb) => disamb.asCls
+    case Value.This(sym) => sym.asCls
```
LPTK: This does not make sense. Have you ever seen an instance being instanciated as `this(args)`?! Please try to think a bit harder about the _semantics_ of the code, instead of randomly calling methods that happen to give you the right type...

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -133,29 +134,49 @@ object TrackableSelect:
...
-    case Value.This(sym) => classCtorSymbol(sym) orElse sym.asCls
+    case Value.MemberRef(_, disamb) => classCtorSymbol(disamb)
+    case Value.This(sym) => classCtorSymbol(sym)
     case s: Select => s.symbol.flatMap(classCtorSymbol)
     case _ => N
 
-object CtorCall:
-  def unapply(r: Result)(using Elaborator.State): Option[(ClassSymbol | ModuleOrObjectSymbol | Int) -> Ls[Arg]] =
+object ClassSymRef:
+  def unapply(p: Path)(using Elaborator.State): Opt[ClassSymbol] = p match
```
LPTK: It hink the whole method could just be implemented as a `collect` over a `targetSymbol` call. Also, it could be made a generic helper at the end of `Block.scala`

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -133,29 +134,49 @@ object TrackableSelect:
...
+    case Value.SimpleRef(sym) => sym.asCls
+    case Value.MemberRef(_, disamb) => disamb.asCls
+    case Value.This(sym) => sym.asCls
+    case s: Select => s.symbol.flatMap(_.asCls)
+    case _ => N
+
+object ObjectRef:
```
LPTK: All the same remarks as above apply, here.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -133,29 +134,49 @@ object TrackableSelect:
...
+  def unapply(p: Path)(using Elaborator.State): Opt[ModuleOrObjectSymbol] = p match
+    case Value.SimpleRef(sym) => sym.asObj
+    case Value.MemberRef(_, disamb) => disamb.asObj
+    case Value.This(sym) => sym.asObj
+    case s: Select => s.symbol.flatMap(_.asObj)
+    case _ => N
+
+object CtorProducer:
+  /** Extracts result forms that produce a concrete `Ctor` flow strategy. */
+  def unapply(r: Result)(using Elaborator.State, Raise): Opt[CtorCls -> Ls[Arg]] =
```
LPTK: Why is this _**extractor**_ raising errors?!

If these errors can happen, they should be caught earlier on – in a place that always runs, not just when some optimizations are enabled. Here, you should just softAsser that the condition never happens.

hkmc2/shared/src/test/mlscript/backlog/ToTriage.mls
```diff
@@ -124,6 +124,18 @@ let c = new Cls(123)
...
+//│ ╙──       	            ^^^^^^^
+//│ ╔══[COMPILATION ERROR] Expected a class; found object.
+//│ ║  l.122: 	let c = new Cls(123)
+//│ ╙──       	            ^^^^^^^
+//│ ╔══[COMPILATION ERROR] Expected a class; found object.
+//│ ║  l.122: 	let c = new Cls(123)
+//│ ╙──       	            ^^^^^^^
+//│ ╔══[COMPILATION ERROR] Expected a class; found object.
+//│ ║  l.122: 	let c = new Cls(123)
+//│ ╙──       	            ^^^^^^^
```
LPTK: You don't think it's strange to report the very same error many times?

LPTK (changes requested):

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -28,8 +28,13 @@ enum Annot extends AutoLocated:
...
   case Config(modify: hkmc2.Config => hkmc2.Config)
-  // marks if a function or lambda is one-shot, i.e. called at most once.
-  // `whichParamList` is the zero-based index of the parameter list for functions with multiple parameter lists.
+  // Marks if a function or lambda is one-shot, i.e. called at most once.
+  // Functions with multiple parameter lists are considered here as a chain of
+  // function values. `whichParamList` is the zero-based index of the parameter
+  // list whose corresponding function value is one-shot.
+  // For example, on `fun f(a)(b)`,
+  // `Affine(0)` says that `f` is one-shot;
+  // `Affine(1)` says that each function value produced by `f(a)` is one-shot.
```
LPTK: Can't both be true at the same time? This is what originally confused me. You write "**the** parameter list [...]" as if there had to be a unique one, which I don't think is true.

Please clarify this, and if the above is true, explain your design choice of only remembering one (or fix it).

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
LPTK: I believe this one can also be simplified, right? In fact, `TrackedSymOf` itself should probably be removed, as it follows the old "mash everything together blindly" approach.

ychenfo: Indeed, I removed `TrackedSymOf` at https://github.com/hkust-taco/mlscript/pull/499/commits/7e953964c06a9afb425cbc394fb30e16bf11ac21.

LPTK (approved):
Net-negative PR despite added tests & documentation – nice!

PR #500, opened 2026-05-26
Flatten class parameter lists and implement `constructor(...)` class syntax

LPTK:
Classes now must have their parameter lists flattened to exactly one auxiliary parameter list before they are codegen'ed. This is a global assumtion that we assume all compilation units satisfy, so we can just flatten all `Instantiate` unconditionally.

We also add the explicit auxiliary `constructor` syntax, which was needed to avoid having to make special cases for buffered classes.


CAG2Mark:
BTW, spreads are broken:
```fs
class A(...x)(y) with
  val a = x
  val b = y
let x = new A(1, 2, 3)(4)
x.a
//│ FAILURE: Unexpected internal error
//│ FAILURE LOCATION: softTODO (Elaborator.scala:1560)
//│ ╔══[INTERNAL ERROR] Compiler reached an unsupported state: mismatched param list lengths List() vs List(term:A/x)
//│ ╟── The compilation result may be incorrect.
//│ ╙── This is a known compiler limitation; if it is a blocker for you, please report it to the maintainers.
//│ = 1
//│ x = A(_)
```

LPTK:
@CAG2Mark Indeed, that's a known problem that's not new to this PR (it's been in `ToTriage.mls` for a long time).

LPTK:
So, the merge broke compilation, but this was not found because the CI inexplicably failed to start. Also, Copilot's review inexplicably enountered an error and failed. After that, the CI was inexplicably broken, failing to fetch a working GitHib Action URL even after a few retries. That's like three GitHub bugs right upon opening a new PR. Just Microslop Things 👍

LPTK:
@copilot Please implement a clean compilation failure when spreads are used in class parameters.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -423,9 +423,17 @@ class JSBuilder(using Config, TL, State, Ctx) extends CodeBuilder:
           case ClsLikeDefn(ownr, isym, sym, ctorSym, kind, paramsOpt, auxParams, par, mtds,
               privFlds, pubFlds, preCtor, ctor, modo, bufferable)
           =>
-            val clsParams = paramsOpt.fold(Nil)(_.paramSyms)
-            val ctorParams = clsParams.map(p => p -> scope.allocateName(p))
-            val ctorAuxParams = auxParams.map(ps => ps.params.map(p => p.sym -> scope.allocateName(p.sym)))
+            val backendParamLists = paramsOpt.toList ::: auxParams
```
LPTK: We don't want to allow several possible configurations. After the param-list flattening pass, all classes must have **exactly one auxiliary parameter list and no main parameter list**. Let's document this, enforce it here (and in WatBuilder), and simplify the logic correspondingly.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -423,9 +423,17 @@ class JSBuilder(using Config, TL, State, Ctx) extends CodeBuilder:
...
+            val backendParamList = backendParamLists match
+              case Nil => N
+              case paramList :: Nil => S(paramList)
+              case _ => lastWords(s"JSBuilder expected flattened constructor parameter lists for class ${sym.nme}")
+            val ctorParams = backendParamList.fold(Nil)(_.paramSyms.map(p => p -> scope.allocateName(p)))
+            val metadataParamsOpt = isym.defn.flatMap(_.paramsOpt).orElse(paramsOpt)
+            val sourceParamsOpt = if bufferable.isEmpty then metadataParamsOpt else paramsOpt
```
LPTK: Why is there this weird special-case for bufferable, here? We should make the bufferable transform produce a class with the expected IR shape, instead of special-casing it in JSBuilder!

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -524,25 +532,8 @@ class JSBuilder(using Config, TL, State, Ctx) extends CodeBuilder:
...
-              val pss = newCtorAuxParams.map(_.map(_._2))
-              val newCtorCode = doc"$ctorCode # return this;"
-              val ctorBraced = doc"${ braced(newCtorCode) }"
-              val funBod = pss.foldRight(ctorBraced):
-                case (psDoc, doc) => doc"(${psDoc.mkDocument(", ")}) => $doc"
-              doc" # return $funBod"
-            
             val ctorBod = {{
-                val extraPath = if paramsOpt.isDefined then ".class" else ""
+                val extraPath = if sourceParamsOpt.isDefined then ".class" else ""
```
LPTK: Actually, couldn't/shouldn't this use the `shouldBeLifted` extension helper?

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -608,17 +599,17 @@ class JSBuilder(using Config, TL, State, Ctx) extends CodeBuilder:
               case N =>
                 doc"$freezeDefns(${clsJS});"
             else
-              val paramsAll = paramsOpt match
-                case None => auxParams
-                case Some(value) => value :: auxParams
+              val sourceParamsAll = sourceParamsOpt.toList ::: sourceAuxParams
```
LPTK: Why are we looking at the source parameters, here? It seems this will do the wrong thing if there's a mismatch between the two (which there could be).

LPTK:
(For the record, the Copilot follow-up is happening at https://github.com/hkust-taco/mlscript/pull/501)

LPTK:
So... This was a bit of a yak shave.

To remove the ad-hoc `bufferable` special-cases from JSBuilder, we had to force buffered classes to use normal constructors, rather than class parameters.

So, we needed to implement normal class constructors, as in

```fs
class Foo with
  constructor(x, y)
  fun getX = x
  fun getY = y

new Foo(1, 2).getX
```

I also cleaned up, improved, and consolidated lots of tangentially related logic.

hkmc2/shared/src/main/scala/hkmc2/codegen/ClassParamFlattener.scala
```diff
@@ -0,0 +1,100 @@
...
+      case Nil => PlainParamList(Nil)
+      case single :: Nil => single
+      case _ => flattenParamLists(paramss)
+    cls.copy(paramsOpt = N, auxParams = flatAux :: Nil)(
+      cls.configOverride,
+      cls.annotations,
+    )
+  
+  private def classPathFor(fun: Path, cls: ClassSymbol): Opt[Path] = fun match
+    case Value.Ref(l, _) => S(Value.Ref(l, S(cls)))
```
CAG2Mark: `Value.Ref` is deprecated, pls use `RefLike` and `Value.MemberRef`

hkmc2/shared/src/main/scala/hkmc2/codegen/ClassParamFlattener.scala
```diff
@@ -0,0 +1,100 @@
...
+    case Select(qual, name) => S(Select(qual, name)(S(cls)))
+    case _ => N
+
+  private def saturatedCurriedClassCall(fun: Path, argss: NELs[Ls[Arg]]): Opt[Path] =
+    fun.targetSymbol.collect:
+      case sym: TermSymbol => sym
+    .flatMap: sym =>
+      sym.defn.flatMap: td =>
+        td.companionClass.flatMap: cls =>
```
CAG2Mark: I'm not really understanding the logic here. Why do we need the companion class? It seems that `td.sym.asCls.flatMap` also passes all the tests.

hkmc2/shared/src/test/mlscript/codegen/CurriedClasses.mls
CAG2Mark: We should remove this comment now.

CAG2Mark (commented):
I'm not that familiar with the front-end stuff so I've mainly looked at the middle-end and tests.

LPTK:
@CAG2Mark Thanks for the review!

LPTK:
Turns out there were two serious regressions hidden in the large diff-tests diff... Now fixed, also leading to a removed `:fixme`!

PR #503, opened 2026-05-27
Fix all Scala warnings in HKMC2

Derppening:
Most warnings are mechanically addressable by either using alternative paradigms or removing redundant code.

The parts of this PR which require more scrutiny are:

- 90d09f5c542cd91b758aba8e906e9207cfb28751: Adding `@unchecked` to match arms, which may require more scrutiny since these will not be caught at runtime
- ab1ea34e1786c311819d394eac929d82e4308fa8: Implement missing match arms by either (a) adding the appropriate logic (e.g. in `showPlain`), (b) adding a `TODO(...)`, or (c) adding a `lastWords` or equivalent - Whether my additions are appropriate needs to be looked into.

Most difftests are unchanged - Some `MatchError`s are changed into actual compilation errors.


hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -1008,7 +1008,7 @@ case class Select(qual: Path, name: Tree.Ident)(val symbol: Opt[DefinitionSymbol
 case class DynSelect(qual: Path, fld: Path, arrayIdx: Bool) extends Path
 
 enum Value extends Path with ProductWithExtraInfo:
-  case SimpleRef(sym: LocalVarSymbol | BuiltinSymbol)
+  case SimpleRef(sym: LocalVarSymbol | BuiltinSymbol | SplitSymbol)
```
LPTK: I think this was reverted but... ouch!!

Derppening:
I think this PR should be ready for review. All warnings have been addressed.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
Derppening: Yes it was reverted, the new `SplitSymbol` doesn't cleanly fit into any of the `Refs` :cry: Luckily the symbol tightening changes removed the need for this.

LPTK (approved):
LGTM now, thanks!

PR #505, opened 2026-05-29
Tighten symbol types throughout the codebase

LPTK:
Well... this took me way longer than I would have thought.

We've had a bad habit of just using `Symbol` (or its infamous alias `Local`) everywhere, which led to nobody really knowing whta kind of symbol was really valid where, which in turn led to a proliferation of hacks, workarounds, outright violations, and needlessly defensive programming.

This PR is the first step towards pulling us out of this quagmire. I managed to tighten the symbol types of most things, although some "looser" types are still floating around in some parts, such as `ValueSymbol`. Ideally, we'd remove most uses of this type.

The lifter in particular caused me immense amounts of pain, as I was trying to figure out just _**what**_ kind of symbol was allowed where and how everything somehow manages to magically fall back into place after traversing all these crazy mappings. I think there is a large potential for simplification here; I tried to achieve it for a while, but failed miserably. So, for now it uses quite loose type _and_ a hack which is real code smell:
```scala
      // Locals introduced by this object
      /* // -- This is the old definition: --
      val fromThisObj = node.localsWithoutBms
        .map: s =>
          s -> s.asLocalPath
      */ // -- The new definition now needs a hack to work, which we should remove: --
      val fromThisObj: Map[ScopedOrInnerSymbol, LocalPath] = node.localsWithoutBms
        .flatMap: s =>
          s match
            case s: BlockMemberSymbol =>
              // * This use of `s.asPrincipal` is incorrect – we can't just assume this is the correct disambiguation!
              S(s -> LocalPath.BmsRef(s, s.asPrincipal.getOrElse:
                lastWords(s"Cannot resolve overloaded member symbol ${s.nme}: no principal disambiguation found")
              ))
              // N  // * can't simply do this, as it makes `Debugging.mls` and `Token.mls` fail to compile
            case s: LocalPathSymbol => S(s -> s.asLocalPath)
            case s: InnerSymbol => S(s -> LocalPath.ThisPath(s))
        .toMap
```
There is still something funny going on, which causes everything to comes crashing down if I try to improve this part. Hopefully @CAG2Mark can help with this (in a follow-up PR).

Oh, and I probably wasted much more time than I saved by trying to use Codex and Copilot to help with the refactoring. I think this is a good example of a task at which they are completely useless with: redesigning some complicated mechanisms that sit outside of their usual training data. Their "fixes" were always worthless workarounds that only piled more technical debt on the existing mess!


Derppening:
> Copilot encountered an error and was unable to review this pull request. You can try again by re-requesting a review.

If Copilot won't do it then I'll ask Claude to do it then! (I'll separately go through the changes manually - Just treat this as what Copilot was *supposed* to do). I also explicitly told Claude to ignore the LLIR changes - It's just a simple removal so I don't think there's much to review there.

## Code review

Reviewed locally on `LPTK/tighten-symbol-types`, focusing on the symbol-type tightening (I skipped the LLIR/C++ backend removal). Overall this is a solid, well-motivated change — replacing the `type Local = Symbol` catch-all with precise unions so each IR position only admits the symbols that can legitimately appear there. A few things below; one I'd treat as a blocker.

### Correctness

**1. `freeVars` is non-exhaustive for an `Assign` with a `TermSymbol` LHS — latent `MatchError`.** (`Block.scala`)

`Assign.lhs: AssignableSymbol = LocalVarSymbol | TermSymbol | NoSymbol`. `definedVars` handles all three:

```scala
case Assign(_: NoSymbol, r, rst)       => rst.definedVars
case Assign(l: TermSymbol, r, rst)     => rst.definedVars + l   // ← treated as reachable
case Assign(l: LocalVarSymbol, r, rst) => rst.definedVars + l
```

but `freeVars` only covers two:

```scala
case Assign(_: NoSymbol, rhs, rest)     => rhs.freeVars ++ rest.freeVars
case Assign(lhs: FreeSymbol, rhs, rest) => Set.single(lhs) ++ ...
```

`FreeSymbol = SimpleSymbol | BlockMemberSymbol | LabelSymbol` does **not** include `TermSymbol`, and there's no catch-all, so an `Assign(termSym, …)` falls through to a `MatchError`. `defineSymbol` does construct exactly that shape (`case sym: TermSymbol => sym.owner match … case N => Assign(sym, rhs, rest)`). The pre-PR code added every LHS to `freeVars` unconditionally, and `ValDefn.freeVars` still adds its ownerless `sym`, so dropping the ownerless-`TermSymbol` case here is a semantic change too, not just a partiality.

Please either add an explicit `Assign(l: TermSymbol, …)` case to `freeVars` (mirroring `definedVars`), or — if `Assign(termSym, …)` genuinely can't reach `freeVars` — drop the `TermSymbol` case from `definedVars` and document why. As written the two are inconsistent.

**2. Acknowledged Lifter hack.** In `pathsFromThisObj`, the new `BlockMemberSymbol` branch uses `s.asPrincipal` with a self-flagged comment that the disambiguation is "incorrect", kept only because `N` breaks `Debugging.mls`/`Token.mls`. Fine to land as flagged debt, but it'd be better tracked by an issue than an inline comment, since it's a known-incorrect disambiguation that happens to pass.

Other correctness-sensitive bits look right:
- `Lifter` `capturePath` changing from `captureSym.asSimpleRef` to `Select(Value.This(sym), captureSym.id)(S(captureSym))` is a genuine fix — the capture is a field, so it must be selected on `this`.
- Splitting `addExtraSyms(…, define: Bool)` into `addExtraSyms` (Scoped local) + `initCaptureField` (AssignField on owner) is correct and removes a boolean-blind parameter.
- `quoteSplit` threading `splitTmps: Map[SplitSymbol, TempSymbol]` correctly works around `SplitSymbol` no longer being assignable.
- `applyAssignLhs` / `applySimpleSymbol` / `applyImportSymbol` and the `SymbolRefresher` matches are exhaustive over their unions, so the removed `lastWords` catch-alls are justified.

### Code quality / conventions

**3. Duplicated type aliases across packages.** `SimpleSymbol` is defined twice, identically — in `semantics.Symbol.scala` and `codegen.Block.scala`. The Assign-LHS union is defined twice under different names: `semantics.Assignable` vs `codegen.AssignableSymbol` (both `LocalVarSymbol | TermSymbol | NoSymbol`), and both names are actively used across `Block`/`Lowering`/`BlockTransformer`. Please consolidate to one `SimpleSymbol` and one name for the assignable union, defined once (in `semantics`) and imported — the duplicates invite drift.

**4. Snake_case leak.** `SymbolRefresher.scala` renames `newS` → `new_s`, the only snake_case identifier in the touched code. Suggest keeping it `newS`.

**5. Minor.** `assignSymbol` (`Lowering.scala`) duplicates the MutVal/LetBind→`AssignField` logic across the `TermSymbol` and `BlockMemberSymbol`→`tsym` branches; could be factored, though it reads fine as-is.

### Behavioral changes worth a second look

- **Instances no longer members + `Ins` lowering removed**: `using`/instance defs now get `owner = N` and are no longer lowered via `ValDefn`. Covered by `NestedUsingInstances.mls` and the `softAssert(defn.owner.isEmpty)` in `Resolver` — looks intentional, just flagging that a semantics change rides along here.
- **`ref` to an owned `TermSymbol` now always selects on `this`** (old `privateFieldSelfSelection` removed), which is the source of the golden churn in `codegen/*.mls`, `Primes.mls`, etc. — worth a careful pass to confirm none regressed.
- **`mut val x = 1` at top level is now an error** ("Mutable value definitions must have an owner"). Tested; presumably intended.

### Tests

Good coverage — `BadAssignments.mls` exercises the fun/val/mut-val assignment errors with the new messages, and `NestedUsingInstances.mls` covers the instance reshaping including a `:lift :effectHandlers` case.

### Verdict

Strong, coherent tightening. I'd block on **#1** (the `freeVars`/`definedVars` `TermSymbol` asymmetry — a real partiality contradicting the `definedVars` case) and fold the duplicate aliases (**#3**) since they're trivially unifiable. The Lifter `asPrincipal` hack (**#2**) is fine to ship as flagged debt with a tracking issue.

LPTK:
Impressive review by the slop-machine!

Though, of course, it missed the forest for the tree, it made me realize `TermSymbol`s have nothing to do in `Assign`'s `lhs`.

It also got something plain wrong: `Ins` definitions _are_ still lowered to `ValDefn`, though through a different path.

LPTK:
PS: Feels funny to be complimented on "strong, coherent tightening" by the weak, incoherent code generator.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -588,12 +613,14 @@ sealed abstract class Defn:
...
-  // * At some point we'll want to make `Local` more specific than `Symbol` to express this
-  // * in the type system.
-  lazy val freeVars: Set[Local] = this match
+  lazy val freeVars: Set[FreeSymbol] = this match
     case FunDefn(own, sym, dSym, params, body) =>
       body.freeVars -- params.flatMap(_.paramSyms) ++ sym.optionIf(own.isEmpty)
-    case ValDefn(tsym, sym, rhs) => rhs.freeVars ++ sym.optionIf(tsym.owner.isEmpty)
+    case vd @ ValDefn(tsym, sym, rhs) =>
+      rhs.freeVars ++
+        // * Value definitions without an owner are simialr to local variables;
```
CAG2Mark:
```suggestion
        // * Value definitions without an owner are similar to local variables;
```

CAG2Mark:
Oh boy...
image

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -490,7 +510,8 @@ class Lifter(topLevelBlk: Block)(using State, Raise, Config):
           if (sub2 is sub) && (fin2 is fin) && (rst2 is rst) then rewritten else TryBlock(sub2, fin2, rst2)
         
         // Assignment to variables
-        case Assign(lhs, rhs, rest) => ctx.symbolsMap.get(lhs) match
+        case Assign(_: NoSymbol, _, _) => super.applyBlock(rewritten)
+        case Assign(lhs: ValueSymbol, rhs, rest) => ctx.symbolsMap.get(lhs) match
```
CAG2Mark:
```suggestion
        case Assign(lhs: LocalVarSymbol, rhs, rest) => ctx.symbolsMap.get(lhs) match
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -635,34 +656,41 @@ class Lifter(topLevelBlk: Block)(using State, Raise, Config):
...
       else
-        if define then Scoped(objSyms.toSet, b) else b
+        Scoped(objSyms.toSet, b)
+    
+    protected final def initCaptureField(b: Block, captureSym: TermSymbol): Block =
+      if hasCapture then
+        captureSym.owner match
+        case S(owner) => AssignField(owner.asThis, captureSym.id, instantiateCapture, b)(S(captureSym))
+        case N => lastWords(s"tried to assign lifted capture to ownerless field ${captureSym.nme}")
+      else b
```
CAG2Mark: This should be moved to the `ClsLikeRewrittenScope` mixin.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -678,26 +706,39 @@ class Lifter(topLevelBlk: Block)(using State, Raise, Config):
...
+      val fromThisObj: Map[ScopedOrInnerSymbol, LocalPath] = node.localsWithoutBms
+        .flatMap: s =>
+          s match
+            case s: BlockMemberSymbol =>
+              // * This use of `s.asPrincipal` is incorrect – we can't just assume this is the correct disambiguation!
+              // softTODO(false, s"disambiguation symbol for ${s.nme} picked arbitrarily") // * uncomment to see where this happens; commented by default because too noisy
+              S(s -> LocalPath.BmsRef(s, s.asPrincipal.getOrElse:
+                lastWords(s"Cannot resolve overloaded member symbol ${s.nme}: no principal disambiguation found")
+              ))
+              // N  // * can't simply do this, as it makes `Debugging.mls` and `Token.mls` fail to compile
```
CAG2Mark, line 720: `localsWithoutBms` should remove all BMS's referring to nested definitions. However, value definitions are still included, which is the reason why returning `N` here breaks certain things. Returning `N` causes the symbols introduced by value definitions to _not_ be considered introduced by this scoped object, as they are primarily referenced using their `BlockMemberSymbol`s.

It might be better to create a `LocalPath` case that refers to value definitions specifically.

LPTK:
@CAG2Mark Nothing too serious actually. I had a branch with much more extensive changes, but couldn't go the last mile, so I just backtracked.

hkmc2/shared/src/main/scala/hkmc2/codegen/StackSafeTransform.scala
```diff
@@ -73,8 +74,8 @@ class StackSafeTransform(depthLimit: Int, paths: HandlerPaths, stackSafetyMap: S
           if usesStack(r) then
             super.applyResult(r): r =>
               lhs match
-                case _: NoSymbol => blockBuilder.assign(lhs, r).rest(applyBlock(rest))
-                case _ => extract(r, false, _ => applyBlock(rest), lhs, curDepth)
+              case _: NoSymbol => blockBuilder.assign(lhs, r).rest(applyBlock(rest))
```
CAG2Mark: Wait, is this logic correct? `extract` also inserts the statement that resets the stack depth, and I do not think we can simply exclude it here.

I guess I missed this when reviewing the other PR, whoops

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: I'll let you pls deal with the lifter changes if you don't mind, as you know best.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: Why does this only appear to cause problems with `using` instance definitions (which are lowered to `ValDefn`) and not other `val`s?

(One thing to note is that the owner `using` defs is always `None`.)

hkmc2/shared/src/main/scala/hkmc2/codegen/StackSafeTransform.scala
LPTK: I guess you're asking @AnsonYeung about the logic of the original code?

CAG2Mark: No, the NoSymbol case was added in the Value.Ref PR.

AnsonYeung: Image
I think it has to do with the new LocalVarSymbol based on surrounding comments.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: It seems nested `val` bindings are desugared to `let` bindings.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: See https://github.com/CAG2Mark/mlscript/commit/df44c3aad2be87ce0a79eae7ddec77c1c390d985

edit: also see the commit after it

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: Do you mean JS-codegen outputs let bindings for them? That's at the very end of the pipeline, so the opposite of desugaring, and that shouldn't affect the lifter. And in principle, again, `using` instances should just behave the same.

hkmc2/shared/src/main/scala/hkmc2/codegen/StackSafeTransform.scala
LPTK: Then it's @Derppening who made the change, but I guess it should be easy to fix.

LPTK: I pushed a fix, but it would be best to have at least one regression test that would have caught this @AnsonYeung.

AnsonYeung: This will only cause the stack effect to be thrown more often than necessary

LPTK: By "this", do you mean the original omission or my fix? Sorry, don't have time to look into the logic.

AnsonYeung: I mean the original omission. The fix is fine.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
CAG2Mark: OK I see, I got confused by that. So it turns out returning `N` there does actually cause problems, but we don't have a test for it:
```fs
:noInline
:lift
:sir
fun f =
  val y = 0
  fun g = y
  g
//│ /!!!\ Uncaught error: java.util.NoSuchElementException: key not found: member:y
//│ 	at: scala.collection.immutable.Map$EmptyMap$.apply(Map.scala:252)
//│ 	at: scala.collection.immutable.Map$EmptyMap$.apply(Map.scala:252)
//│ 	at: hkmc2.Lifter$LiftedScope.$anonfun$20(Lifter.scala:850)
//│ 	at: scala.collection.immutable.List.map(List.scala:240)
//│ 	at: hkmc2.Lifter$LiftedScope.formatArgs(Lifter.scala:850)
//│ 	at: hkmc2.Lifter$LiftedFunc.rewriteCall(Lifter.scala:1072)
//│ 	at: hkmc2.Lifter$$anon$3.applyResult$$anonfun$3(Lifter.scala:369)
//│ 	at: hkmc2.codegen.BlockTransformer.applyArgss$$anonfun$2(BlockTransformer.scala:143)
//│ 	at: hkmc2.codegen.BlockTransformer.rec$1$$anonfun$1$$anonfun$1(BlockTransformer.scala:320)
//│ 	at: hkmc2.codegen.BlockTransformer.rec$1(BlockTransformer.scala:318)
```

LPTK: Let me know when I should merge your fix, or if you want to do it in a separate PR.

LPTK: > It might be better to create a LocalPath case that refers to value definitions specifically.

By the way, this seems unprincipled to me. There's nothing special about `val`s, here. Functions etc. are referred to in the same way. If we had an annotation or another reason not to lift a particular function, I guess we'd get the very same problem. So, the fix shouldn't be about `val`, but about how the lifter tracks these things.

By the way, this makes me realize something. There are situations where we might want to lift a definition but not its companion (other definition of another kind with the same name), eg a class can be lifted even though its companion module might not be. So, the lifter should identify things by both bms and definition symbol and should ideally be able to split a group of overloaded defs that used to share a single bms, but won't in the transformed result. Not necessary to do now of course, just food for thought to figure out the correct design.

CAG2Mark: IIRC unlifted definitions are tracked separately (see `defnPathsFromThisObj`).

CAG2Mark: Anyway, since this might involve a bigger refactor, I'd rather do this in a separate PR.

PR #507, opened 2026-06-02
Implement the case-of-case optimization

LPTK:
This PR adds a `CaseOfCase` (COC) optimization pass to `BlockSimplifier`. It simplifies a match whose scrutinee was produced by an earlier match assigning known constructor/literal shapes, allowing nested matches such as chained `Option.map` calls to fuse into a single producer match when doing so is safe.

The implementation also:

- runs DCE to a local fixed point before later optimization passes, so unused inliner labels do not block COC;
- factors constructor-shape detection for calls vs instantiations;
- keeps duplication controlled by the existing inlining threshold;
- refreshes copied consumer branches to preserve symbol hygiene;
- adds regression coverage for full fusion, partial-known producers, prefix dependency safety, large-branch duplication limits, and infinite inlining behavior.

Most other changes are updated golden outputs from the new optimization behavior.


hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -46,17 +47,34 @@ class BlockSimplifier
...
       changed ||= vp.changed
       if vp.changed then log("▶ VP:\n" + printRes)
       
       summon[Config].inlining.foreach: cfg =>
+        
+        // * Runs after DCE so that unused labels from inlining and already removed
+        val coc = new CaseOfCase(using cfg)
+        res = coc.applyProgram(res)
+        changed ||= coc.changed
+        if coc.changed then log("▶ COC:\n" + printRes)
```
AnsonYeung:
```suggestion
        // * Runs after DCE so that unused labels from inlining are already removed
        val coc = new CaseOfCase(using cfg)
        res = coc.applyProgram(res)
        changed ||= coc.changed
        if coc.changed then log("▶ COC:\n" + printRes)
```

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -952,6 +991,177 @@ class BlockSimplifier
...
+    override def applyBlock(b: Block): Block = super.applyBlock(b) match
+      case m @ Match(scrut, arms, dflt,
+        TrivialStatementsAndMatch(k,
+          consumer @ Match(Value.SimpleRef(target: LocalVarSymbol), _, _, consumerRest)))
+      =>
+        
+        val prefix = k.fold[Block](End())(_(End()))
+        
+        val producerDefinedVars: Set[Symbol] = arms.iterator.flatMap(_._2.definedVars).toSet
+          ++ dflt.iterator.flatMap(_.definedVars)
```
AnsonYeung, line 1114:
```scala
// * Note: it seems most historical uses of `definedVars` would be better removed,
// * now that we properly put everything in proper Scoped blocks;
// * and `definedVars` itself should be removed.
```
We currently have this comment for `definedVars`

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -952,6 +991,177 @@ class BlockSimplifier
...
+        case _: End => k(shape)
+        case Assign(`target`, rhs, rest) => loop(rest, getShape(rhs))(k)
+        case Assign(_, _, rest) => loop(rest, shape)(k)
+        case AssignField(_, _, _, rest) => loop(rest, shape)(k)
+        case AssignDynField(_, _, _, _, rest) => loop(rest, shape)(k)
+        case Define(_, rest) => loop(rest, shape)(k)
+        case Scoped(_, body) => loop(body, shape)(k)
+        case Begin(sub, rest) => loop(sub, shape)(loop(rest, _)(k))
+        case _ => N
+      loop(body, N)(identity)
```
AnsonYeung, line 1040: This ignores possible side effects by Call that modifies the variable
```
:noOpt
:soir
fun foo(x) =
  let y
  let f = hide(w => set y = w)
  if x then
    set y = "ok"
    f("nope")
  else
    set y = "nope"
  if y is "ok" then
    true
  else
    false
foo(true)
```

LPTK: Good catch! In fact, we should probbaly not try to reason about captured variables. Avoiding any possible side effect would also work, and be easier, as we don't have the "captured variable" analysis result within this separate transformer (and it would be better if we didn't have to recompute it).

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
LPTK: Good point. I've updated the comment, as this is one rare legitimate use of this definition.

AnsonYeung: should it be `freeVars` instead?

AnsonYeung: I guess not because freeVars counts read as well. I forgot about that

PR #509, opened 2026-06-03
Make lifting the default

AnsonYeung:


hkmc2/shared/src/test/mlscript/wasm/ClassMethods.mls
```diff
@@ -146,7 +146,7 @@ class Box(val x) with
     fun inc() = Box(x + 1).get()
 Box(2).inc()
 //│ Wasm result:
-//│ = 2
+//│ = 3
```
LPTK, line 149: This is quite concerning.

LPTK: (We found the source, and this will be fixed later by @CAG2Mark.)

hkmc2/shared/src/test/mlscript-compile/Predef.mjs
```diff
@@ -5,7 +5,46 @@ import RuntimeJS from "./RuntimeJS.mjs";
 import Runtime from "./Runtime.mjs";
 import Rendering from "./Rendering.mjs";
 import Term from "./Term.mjs";
-let Predef1;
+let Predef1, lambda, lambda1, lambda$, lambda$1, lambda$2;
+lambda$2 = (undefined, function (Predef2) {
```
LPTK, line 9: @CAG2Mark I notice there's a strange capture of a module that should already be available (as `Predef1`, here). Is this known? Ideally should be improved.

CAG2Mark: I think this is known. Assuming we forbid nested modules in the future or desugar them, this can indeed be improved

LPTK: How is that related to nested modules? 🤔

CAG2Mark: If a class or function is lifted out of a nested module (by being inside a definition in that nested module), the lifter just passes that module instance by default, even if that module is at the top level.

LPTK: A module at the top level is not "nested" in any sense of the term. What's a non nested module to you?

CAG2Mark: ah I meant this is true for all modules in general, but the reason I do this is because of nested modules

LPTK: We should have special handling of modules accessible from the top level, as discussed on Discord. Anyway, we'll do this in a follow up PR.

PR #510, opened 2026-06-03
Make two fixes to FlowAnalysis/DPE

LPTK:
We hit two FlowAnalysis bugs resulting in DPE misoptimizing code, which I asked Codex to fix.

While the provenance of these bugs is now clear, I don't think the fix of the "`FunRef`" one is optimal, so @ychenfo please fix the fix by pushing further commits to this PR.

The bug related to propagating `UnknownCons` to the captures of functions constrained by `UnknownCons` came up when we tried to lift everything by default (it made `Iter.mls` miscompile).

### DIFF
--- a/hkmc2/shared/src/test/mlscript/codegen/ConfigDirective.mls
+++ b/hkmc2/shared/src/test/mlscript/codegen/ConfigDirective.mls
@@ -161,6 +161,7 @@ print(42)
 private fun deadParamConfig(a) = 0
 deadParamConfig(42)
 //│ dead-param-elim > >>> dead-param-elim results >>>
+//│ dead-param-elim > prodfun deadParamConfig#0 @ <root> -> eliminable: {0}
 //│ dead-param-elim > <<< dead-param-elim results <<<
 //│ = 0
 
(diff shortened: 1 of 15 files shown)


hkmc2/shared/src/test/mlscript/codegen/ConfigDirective.mls
```diff
@@ -161,7 +161,6 @@ print(42)
 private fun deadParamConfig(a) = 0
 deadParamConfig(42)
 //│ dead-param-elim > >>> dead-param-elim results >>>
-//│ dead-param-elim > prodfun deadParamConfig#0 @ <root> -> eliminable: {0}
```
LPTK, line 164: @ychenfo I wonder why this optimization opportunity is now lost. (This diff occurred when fixing the lambda-capture problem.)

ychenfo: In this test, stack safety is actually on, so when the `‹stack safe body›` function capturing `deadParamConfig` function gets constrained with `UnknownCons` (for being passed to `runtime⁰.runStackSafe﹖(100, ‹stack safe body›)`), `deadParamConfig` is also constrained with `UnknownCons` and therefore there is no eliminable parameters.

Once the propagation of `UnknownCons` to the captures of functions constrained by `UnknownCons` is turned off the optimization opportunity is back.

LPTK (commented):
LGTM, thanks!

PR #511, opened 2026-06-04
Allow multiple parameter lists in class extensions

CAG2Mark:
Support the following:
```fs
class A(x)(y)

new A(1)(2) with
  fun foo = 0

class B extends A(0)(0)
```

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/ClassParamFlattener.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/ClassParamFlattener.scala
@@ -69,7 +69,7 @@ class ClassParamFlattener(using State) extends BlockTransformer(SymbolSubst.Id):
       case defn => k(defn)
   
   override def applyResult(r: Result)(k: Result => Block): Block = r match
-    case c @ Call(r @ Value.RefLike(sym), argss) if sym === State.superSymbol =>
+    case c @ Call(r @ Value.RefLike(State.superSymbol), argss) =>
       applyArgss(argss): argss2 =>
         val flatArgss =
           if argss2.lengthCompare(1) > 0 then argss2.flatten ne_:: Nil
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/ClassParamFlattener.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/ClassParamFlattener.scala
@@ -6,6 +6,8 @@ import hkmc2.utils.*
 
 import semantics.*
 
+import hkmc2.semantics.Elaborator.State
+
 
 /** Flattens class constructor parameter lists in the Block IR.
   *
@@ -22,7 +24,7 @@ import semantics.*
   * Argument spreads are intentionally preserved as `Arg`s while only the
   * surrounding argument-list boundary is removed.
   */
-class ClassParamFlattener extends BlockTransformer(SymbolSubst.Id):
+class ClassParamFlattener(using State) extends BlockTransformer(SymbolSubst.Id):
   
   private def flattenParamLists(paramss: Ls[ParamList]): ParamList =
     val flags = paramss.headOption.fold(ParamListFlags.empty)(_.flags)
@@ -67,6 +69,14 @@ class ClassParamFlattener extends BlockTransformer(SymbolSubst.Id):
       case defn => k(defn)
   
   override def applyResult(r: Result)(k: Result => Block): Block = r match
+    case c @ Call(r @ Value.RefLike(sym), argss) if sym === State.superSymbol =>
+      applyArgss(argss): argss2 =>
+        val flatArgss =
+          if argss2.lengthCompare(1) > 0 then argss2.flatten ne_:: Nil
+          else argss2
+        k:
+          if flatArgss is argss then c
+          else Call(r, flatArgss)(c.isMlsFun, c.mayRaiseEffects, c.explicitTailCall).withLocOf(c)
     case call @ Call(fun, argss) =>
       saturatedCurriedClassCall(fun, argss) match
       case S(cls) =>
@@ -93,8 +103,8 @@ class ClassParamFlattener extends BlockTransformer(SymbolSubst.Id):
 end ClassParamFlattener
 
 object ClassParamFlattener:
-  def apply(program: Program): Program =
+  def apply(program: Program)(using State): Program =
     new ClassParamFlattener().applyProgram(program)
   
-  def apply(block: Block): Block =
+  def apply(block: Block)(using State): Block =
     new ClassParamFlattener().applyBlock(block)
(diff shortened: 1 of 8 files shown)


CAG2Mark:
Some things:
- I don't understand why the parameter lists in the last test case of `ObjectExtensions.mls` is non-empty.
- Was there a  bug in `CrossBranchCapture.mls`? It was extending the `Vec` class without providing the argument. I suppose that `n` is just the length of the vector, so I've filled it in with what I think is the intended code.

LPTK:
> * I don't understand why the parameter lists in the last test case of `ObjectExtensions.mls` is non-empty.

You can see what happened with `:pt`:

```fs
:pt
:fixme // Support?
object Bar extends! Foo!class
//│ ———————————————| Parsed tree |——————————————————————————————————————————————————————————————————————
//│ TypeDef:
//│   k = Obj
//│   head = InfixApp:
//│     lhs = Ident of "Bar"
//│     kw = Keywrd of keyword 'extends'
//│     rhs = App:
//│       lhs = Ident of "!"
//│       rhs = Tup of Ls of 
//│         DynAccess:
//│           obj = Ident of "Foo"
//│           fld = Ident of "class"
//│ ╔══[COMPILATION ERROR] Expected a statically known class; found reference.
//│ ║  l.47: 	object Bar extends! Foo!class
//│ ║        	                  ^
//│ ╙── The 'new' keyword requires a statically known class; use the 'new!' operator for dynamic instantiation.
```

This is parsed as a prefix use of `!` (quite surprising, but it makes sense).


> * Was there a  bug in `CrossBranchCapture.mls`? It was extending the `Vec` class without providing the argument. I suppose that `n` is just the length of the vector, so I've filled it in with what I think is the intended code.

Yeah, seems like this was just a silly omission.

hkmc2/shared/src/main/scala/hkmc2/codegen/ClassParamFlattener.scala
```diff
@@ -67,6 +69,14 @@ class ClassParamFlattener extends BlockTransformer(SymbolSubst.Id):
       case defn => k(defn)
   
   override def applyResult(r: Result)(k: Result => Block): Block = r match
+    case c @ Call(r @ Value.RefLike(sym), argss) if sym === State.superSymbol =>
```
LPTK:
```suggestion
    case c @ Call(r @ Value.RefLike(State.superSymbol), argss) =>
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -384,12 +379,40 @@ class Lowering()(using Config, TL, Raise, State, Ctx, SymbolPrinter):
...
     blockImpl(imps ::: funs ::: rest, res)
   
+  def getCtorParamLists(cls: Path): Ls[ParamList] =
+    cls.targetSymbol.flatMap: sym =>
+      sym.asClsOrMod.flatMap(_.defn) orElse
+      sym.asTrm.flatMap(_.owner).flatMap(_.asDefnSym.defn)
+    .fold(Nil: Ls[ParamList]): clsDef =>
+      clsDef.paramsOpt.toList ::: clsDef.auxParams
+  
   // * Lowers the `super(...)` call we get from the `extends C(...)` syntax
```
```suggestion
  // * Lowers the `super(...)(...)...` call we get from the `extends C(...)(...)...` syntax
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -384,12 +379,40 @@ class Lowering()(using Config, TL, Raise, State, Ctx, SymbolPrinter):
...
-  def lowerSuperCtorCall(fr: Path, isMlsFun: Bool, isTailCall: Bool, arg: Opt[Term], loc: Opt[Loc])(k: Result => Block)(using LoweringCtx): Block =
-    arg match
-    case S(arg) =>
-      lowerArgs(arg)(as => k(Call(fr, as ne_:: Nil)(isMlsFun, true, isTailCall).withLoc(loc)))
-    case N =>
+  def lowerSuperCtorCall(parentClsPth: Path, fr: Path, isMlsFun: Bool, isTailCall: Bool, args: List[Term], loc: Opt[Loc])(k: Result => Block)(using LoweringCtx): Block =
+    val ctorParamLists = getCtorParamLists(parentClsPth)
+    args match
+    case _ :: _ =>
+      def zipArgs(remainingParamss: Ls[ParamList], args: Ls[Term], acc: Ls[Ls[Arg]]): Block = (remainingParamss, args) match
```
LPTK, line 394: We now have this complicated logic to lower things carefully for `super`. Is there no way to share it with the `new` case with some helper function?

hkmc2/shared/src/test/mlscript/codegen/CurriedClassInheritance.mls
```diff
@@ -3,9 +3,10 @@
 
 class Foo(val x)(val y)
 
-
+:todo
```
LPTK:
```suggestion

:e
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -384,12 +379,40 @@ class Lowering()(using Config, TL, Raise, State, Ctx, SymbolPrinter):
...
+          if remainingArgs.isEmpty then
+            raise(ErrorReport(
+              msg"Too many parameter lists for parent class" -> loc :: Nil,
+              source = Diagnostic.Source.Compilation))
+          lowerArgs(arg): as =>
+            zipArgs(Nil, remainingArgs, as ne_:: acc)
+        case (remainingParamss2, Nil) =>
+          if !remainingParamss2.isEmpty then
+            raise(ErrorReport(
+              msg"Extending a partially applied class is not yet supported" -> loc :: Nil,
```
```suggestion
              msg"Extending a partially applied class is not supported" -> loc :: Nil,
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -384,12 +379,40 @@ class Lowering()(using Config, TL, Raise, State, Ctx, SymbolPrinter):
...
+              source = Diagnostic.Source.Compilation))
+          k(Call(fr, acc.reverse.ne_!)(isMlsFun, true, isTailCall).withLoc(loc))
+      zipArgs(ctorParamLists, args, Nil)
+    case Nil =>
+      if !ctorParamLists.isEmpty then
+        raise(ErrorReport(
+          msg"Extending a partially applied class is not yet supported" -> loc :: Nil,
```
```suggestion
          msg"Extending a partially applied class is not supported" -> loc :: Nil,
```

LPTK (approved):
Otherwise LGTM. Nice fixes, thanks!

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
CAG2Mark: I tried refactoring it and it ended up quite messy

PR #512, opened 2026-06-05
Lifter Cleanup and Fixes

CAG2Mark:
- Refactor `RefOfBms` into `RefOfDefn` (it no longer checks for a `BlockMemberSymbol` at all)
- Remove all `BlockMemberSymbol` indices in the lifter
- Remove the `isPrincipal` hack by including value definitions in the scope tree
- Pass required variables into nested superclasses
- Fix the weird `Box` bug. The proper fix is at https://github.com/hkust-taco/mlscript/pull/512/changes#diff-8ed0a2cb83cde5929bf5559470e91620ab33f866549491a6be5d96ac5c7b4ec2R752; Codex's fix from the meeting was more of a band-aid solution. We do in fact want to entirely rewrite some selections and not just the head, i.e. when selecting a lifted nested class defined in another class.


CAG2Mark:
Note: The changes this time are quite simple and should be fairly painless to review.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
LPTK: This comment (and probably others as well) mentions BMS, but I guess that is no longer up to date?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -334,102 +328,109 @@ class Lifter(topLevelBlk: Block)(using State, Raise, Config):
...
             case c: ScopedObject.Companion =>
-              ctx.symbolsMap.get(c.clsBody.isym).map(_.read)
+              ctx.symbolsMap.get(c.clsBody.isym).map(_.read) // Reference to an unlifted module
             case _ => N
 
         override def applyResult(r: Result)(k: Result => Block): Block =
           r match
-          // if possible, directly rewrite the call using the efficient version
-          case c @ Call(RefOfBms(l, S(d), _), argss) =>
+          case c @ Call(Value.RefLike(sym), argss) if sym === State.superSymbol => superClass match
```
```suggestion
          case c @ Call(Value.RefLike(State.superSymbol), argss) => superClass match
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -707,17 +708,11 @@ class Lifter(topLevelBlk: Block)(using State, Raise, Config):
       val fromThisObj = node.localsWithoutBms
         .map: s =>
           s -> s.asLocalPath
-      */ // -- The new definition now needs a hack to work, which we should remove: --
+      */
```
LPTK: I think we can remove the commented old version, right?

LPTK (approved):
Great, thanks!

LPTK:
> Codex's fix from the meeting was more of a band-aid solution

Yeah, unfortunately, that's the case 80% of the time. As I said many times, it's useful to diagnose problems, but pretty worthless at fixing them properly, since it usually misses the big picture.

PR #513, opened 2026-06-05
Improve Tuple.get handling

AnsonYeung:
There is currently a bug in FlowAnalysis that yield an assertion error. When `runtime.Tuple.get(x, 0)` is encounted, all enclosing match are searched for a `Match(Value.SimpleRef(x), ... :: Case.Arr(<len>) -> <inner> :: ..., ...)` so that the length can be determined. However not all scrutinee shape were handled properly. When `Select` is used as scrutinee, FlowAnalysis will crash.

It is also unclear whether FlowAnalysis correctly deal with IR that reassigns `x` after the match is performed.

This PR both remove the producer of the IR above and also make FlowAnalaysis to not crash by the IR above. The producer of the IR is HandlerLowering where Tuple.get is split to another block which does not have a corresponding enclosing match of the tuple and there is an enclosing match that uses a path (`runtime.curEffect`) in the scrutinee.


ychenfo:
Thanks! I don't see any problem with the changes to flow analysis itself. ~~but with those changes now in place, do we still need to modify HandlerLowering? If we don't, will that cause further issues in flow analysis?~~

> It is also unclear whether FlowAnalysis correctly deal with IR that reassigns x after the match is performed.

I can't see an immediate problem now, is there a concrete example that's concerning? Thanks!

hkmc2/shared/src/test/mlscript/dead-param-elim/basic.mls
```diff
@@ -250,6 +250,15 @@ f(Pair(1, 2))
 //│ = 1
 
 
+// TODO: This regression test should have used runtime.Tuple.get but it's not resolved currently.
+// The usage of runtime.Tuple.get(c, 1) will cause all enclosing match to be searched, but
+// not all scrutinee shapes of match are correctly handled.
```
LPTK: Is this TODO still current? I don't actually understand what it means. Please clarify it, if it's still current.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -107,17 +109,17 @@ object PossibleTrackableTupleSelect:
   def unapply(s: Result)(using eState: Elaborator.State): Opt[Value.SimpleRef -> Int] =
     s match
     case Call(
-      Select(Select(Value.SimpleRef(runtimeSym), Tree.Ident("Tuple")), Tree.Ident("get")),
+      p,
       (Arg(N, ref@Value.SimpleRef(scrut)) :: Arg(N, Value.Lit(Tree.IntLit(n))) :: Nil) :: Nil
-    ) if runtimeSym is eState.runtimeSymbol => S(ref -> n.toInt)
+    ) if p.targetSymbol === S(eState.tupleGetSymbol) => S(ref -> n.toInt)
```
LPTK:
```suggestion
    ) if p.targetSymbol.contains(eState.tupleGetSymbol) => S(ref -> n.toInt)
```

hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
```diff
@@ -173,9 +173,16 @@ class HandlerLowering(paths: HandlerPaths, opt: EffectHandlers)(using TL, Raise,
...
+  val knownSymbolsExcludedFromEffect = Set[DefinitionSymbol[?]](State.tupleGetSymbol, State.tupleSliceSymbol)
+
   object EffectfulResult:
     def unapply(r: Result) = r match
-      case c: Call if c.mayRaiseEffects => S(r)
+      case c: Call if c.mayRaiseEffects =>
+        if c.fun.targetSymbol.exists(knownSymbolsExcludedFromEffect.contains(_)) then
+          N
+        else
+          S(r)
```
LPTK: Instead of special-casing the symbol here, why not fix this at the source by making `mayRaiseEffects` `false` for such calls in the first place? (In Lowering)

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -107,17 +109,17 @@ object PossibleTrackableTupleSelect:
...
     case _ => N
 
 object TrackableSelect:
   def unapply(s: Result)(using pre: FlowPreAnalyzer, eState: Elaborator.State): Opt[(from: Path, field: SelField, owner: CtorCls)] =
     given fState: FlowAnalysis.State = pre.fState
     s match
     case sel@PossibleTrackableTupleSelect((ref@Value.SimpleRef(scrut)) -> ith) =>
-      pre.res.getEnclosingMatchesForSel(sel.uid).find(_._1.getReferredSym is scrut).flatMap:
+      pre.res.getEnclosingMatchesForSel(sel.uid).find(_._1.getReferredSymOpt === S(scrut)).flatMap:
```
LPTK:
```suggestion
      pre.res.getEnclosingMatchesForSel(sel.uid).find(_._1.getReferredSymOpt.contains(scrut)).flatMap:
```

LPTK (approved):
Otherwise LGTM.

AnsonYeung:
> Thanks! I don't see any problem with the changes to flow analysis itself. ~but with those changes now in place, do we still need to modify HandlerLowering? If we don't, will that cause further issues in flow analysis?~
> 
> > It is also unclear whether FlowAnalysis correctly deal with IR that reassigns x after the match is performed.
> 
> I can't see an immediate problem now, is there a concrete example that's concerning? Thanks!

It may retrieve a wrong length from a wrong enclosing match, which I don't know what the length would be used for.

hkmc2/shared/src/test/mlscript/dead-param-elim/basic.mls
AnsonYeung: Yes, writing this in the code directly will result in error of `Tuple` field mising and if I import `Runtime` directly then it will resolve Tuple.get to a different symbol.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -349,6 +349,8 @@ class TermSymbol(val k: TermDefKind, val owner: Opt[InnerSymbol], val id: Tree.I
   def isPrivate: Bool = (k is LetBind) && owner.exists(!_.isInstanceOf[TopLevelSymbol])
   
   def subst(using sub: SymbolSubst): TermSymbol = sub.mapTermSym(this)
+  var _mayRaiseEffects = true
```
LPTK: This is pretty bad! We do _not_ want to add yet another piece of mutable data just for our own convenience – it's pure technical debt that will bite us back one day or another.

The principled approach is just to add an annotation: https://github.com/hkust-taco/mlscript/pull/513/changes/059034e24764ad8357f83597d7c30f2a7e0b7e9d

hkmc2/shared/src/test/mlscript/dead-param-elim/basic.mls
LPTK: I hope you realize that your documentation is basically impossible to understand on its own: it's both too vague and incorrect/mistaken. It's too vague because I had no idea what you meant by "This regression test should have used [...]" as it's given without any explanation, and it's incorrect/mistaken because `runtime` is not meant to be a name bound for programmers to use, so of course it will not resolve, and this is not a problem – the real problem is that the explicitly imported version resolves to a different symbol.

Please carefully consider my fix: https://github.com/hkust-taco/mlscript/pull/513/changes/79ac1bbc6bff4306736a8f9dd5c3fc78429292fa

LPTK:
> > I can't see an immediate problem now, is there a concrete example that's concerning? Thanks!
> 
> It may retrieve a wrong length from a wrong enclosing match, which I don't know what the length would be used for.

@ychenfo Have you considered this? Actually, is reassigning the scrutinee a general problem that affects all deforestation code?

hkmc2/shared/src/test/mlscript/dead-param-elim/basic.mls
AnsonYeung: It's not a problem that `runtime.Tuple.get` does not resolve. But it is a problem that the regression test does not actually directly test the intended code path, but rather relying on HandlerLowering to emit IR of a certain shape. The comment is clear from the PR context as we want to test `runtime.Tuple.get` being moved out of the corresponding Match to test whether FlowAnalysis can handle it. (and the PR can be retrieved easily from git blame of the comment)

In fact it is not a problem that the get symbol cannot be constructed by the frontend, but it just mean that this regression test itself is fragile.

LPTK: > rather relying on HandlerLowering to emit IR of a certain shape

Notice that the vast majority of our tests are like that. Yes, it has downsides, and it would be better to test things a bit more directly. (Besides, this is not rlying on HandlerLowering to produce the `Tuple.get`.)

> The comment is clear from the PR context

You have to remember that this context is essentially lost after the PR is merged. Nobody wants to have to investigate the PR a commit was from just to understand an unclear comment!! Especially since this may require traversing several levels of history blame. And even with the PR context, I found this very unclear. And I'm by far the person who knows most about this codebase, so...

PR #515, opened 2026-06-06
Do not scope top-level symbols exported by diff-test block

LPTK:
We used to always place the symbols exported by a diff-test block into a top-level `Scoped` block, which was semantically wrong (as these symbols are not truly locally scoped; they are visible by later blocks).

This PR removes the strange behaviod, allowing some simplifications. Further simplifications are probably possible in optimization passes (maybe they no longer need `symbolsToPreserve`), but this will be investigated later.

During this change, I discovered that Lumberhack actually was rewriting these fakely-local symbols as though they were truly local, and this was relied on in the tests. To preserve the test behaviors, I had to make many top-level `let` bindings local by adding an `in` clause to them.


LPTK:
Approved by @ychenfo!

PR #516, opened 2026-06-06
Make `NoSymbol` a plain `object`

Derppening:
This PR simplifies `NoSymbol` by removing its dependency on `State` and converting it into a plain singleton `object`. The legacy `State.noSymbol` is now deprecated in favor of directly using `NoSymbol`.

No functional changes.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
@@ -45,7 +45,7 @@ case class Program(
 type SimpleSymbol = LocalVarSymbol | BuiltinSymbol
 
 /** Symbol that can be used as the left-hand side of an `Assign`. */
-type Assignable = LocalVarSymbol | NoSymbol.type
+type Assignable = LocalVarSymbol | NoSymbol
 
 /** Symbols that `Scoped` introduces as block-local bindings.
   * This deliberately excludes things like `TermSymbol`s, which never need to be scoped.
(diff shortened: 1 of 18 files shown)


hkmc2/shared/src/main/scala/hkmc2/utils/utils.scala
```diff
@@ -92,7 +92,7 @@ class DebugPrinter:
       case codegen.Scoped(syms, body) =>
         val symsStr = "{" + syms.toArray.sortBy(_.uid).map(_.showAsPlain).mkString(", ") + "}"
         s"Scoped(syms = $symsStr): \n" + s"body = ${printProduct(false, body)}".indent("  ")
-      
+
```
LPTK: Nooo the PR was almost perfect up until the very last diff 😅

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -45,7 +45,7 @@ case class Program(
 type SimpleSymbol = LocalVarSymbol | BuiltinSymbol
 
 /** Symbol that can be used as the left-hand side of an `Assign`. */
-type Assignable = LocalVarSymbol | NoSymbol
+type Assignable = LocalVarSymbol | NoSymbol.type
```
LPTK: To avoid having to type `.type`, let's add a companion `type NoSymbol = NoSymbol.type` to `NoSymbol`.

LPTK (approved):
Otherwise LGTM, thanks!

hkmc2/shared/src/main/scala/hkmc2/utils/utils.scala
Derppening: Oops... Fixed in 44a09bfb56e906af13ecf1302af03e1f6b5024ce.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
Derppening: Done in ae73f8b1864b3b3c8a44529057f08a991b33f02d.

PR #517, opened 2026-06-09
Improve the `Begin` smart constructor and fix two lifter bugs

LPTK:
I made the smart constructor of `Begin` smart***er***, but this resulted in `Rendering.mls` miscompiling.

I used Codex in Plan mode to find the issue. After fumbling a bit, it found the true root causes and fixed them.

This is a best-case scenario for Codex use: it saved me a lot of time and did _not_ produce a slew of low-quality slop 👍


CAG2Mark (approved):
Thanks, LGTM. Not sure how I missed that.

BTW would it be a good idea to simplify `Begin` even more? i.e. if the first block in a `Begin` ends with an `End`, we can replace the tail of the first block with the second block, and then we can avoid a `Begin` block entirely. And if the first block is abortive, we can discard the second block entirely.

LPTK:
> BTW would it be a good idea to simplify `Begin` even more?

Indeed, there were a few more cases that were not included and that now are 👍

(Note that we already check `isAbortive`.)

CAG2Mark:
I was thinking of something like
```scala
var mappedTail = false
val mapped = sub.mapTail:
  case End(_) => mappedTail = true; rest
  case b => b
if mappedTail then mapped else Begin(sub, rest)
```
It seems the `mapTail` function no longer exists though

PR #518, opened 2026-06-10
Add metadata wrapper for `Call` and `Instantiate` IR nodes + fix ClassCtorSymbol logic

NeilKleistGao:
- Add `CallMetadata` and `InstantiateMetadata`
- Add annotation list for metadata
- Retrieve `explicitTailCall` from annotations


hkmc2/shared/src/main/scala/hkmc2/codegen/BlockTransformer.scala
```diff
@@ -349,4 +349,3 @@ class BlockTransformerShallow(subst: SymbolSubst) extends BlockTransformer(subst
 // to traverse sub-blocks while using this class to perform more complicated transformations on the blocks themselves.
 class BlockDataTransformer(subst: SymbolSubst) extends BlockTransformerShallow(subst):
   override def applySubBlock(b: Block): Block = b
-
```
LPTK, line 352: 👎

LPTK (approved):
Otherwise LGTM. As discussed in the meeting, it's a bit weird to have bespoke metadata classes for `Call` and `Instantiate` but not for definitions, but this is fine for now.

PR #520, opened 2026-06-11
Fix issue with HandlerLowering not instrumenting class inside modules

AnsonYeung:
Previously, `doNotInstrumentTopLevelModCtor` will skip all module constructor, even if they contain a inner class. This behaviour is now fixed. The instrumentation should only be disabled for actual code inside the constructor and not inner class or functions, as those functions and class could potentially be used later.

From the config description, this is mainly used for resolving a cyclic dependency problem while preserving stack safety checks.
```scala
// Skips instrumenting module constructors, this can be used when the file is statically known to not
// raise any effect and cannot use the Runtime.mls module during module construction due to cyclic dependency.
// One specific scenario is Rendering.mls, which Runtime.mls depends on, and hence using stack safety will
// reference Runtime.mls during construction of the Rendering module, causing a cyclic dependency error.
doNotInstrumentTopLevelModCtor: Bool = false,
```

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/HandlerLowering.scala
@@ -585,8 +585,7 @@ class HandlerLowering(paths: HandlerPaths, opt: EffectHandlers)(using TL, Raise,
             // TODO: Companion's ctor is more well behaved so it is possible to handle it
             // However, JSBuilder inserts extra statements between preCtor and ctor and it's not possible to replicate the exact behavior
             // without many special handling.
-            val newCtor = if opt.doNotInstrumentTopLevelModCtor && !h.innerDefIsTrulyNested then bod.ctor else
-              translateCtorLike(bod.ctor, bod.isym.asThis, true)
+            val newCtor = translateCtorLike(bod.ctor, bod.isym.asThis, true)
             tl.log(s"companion name: ${bod.isym.nme}")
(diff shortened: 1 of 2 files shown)


AnsonYeung:
It seems that the issue of the cyclic dependency came up yet again after this fix, and we might need a better fix for the cyclic dependency problem.

LPTK:
> It seems that the issue of the cyclic dependency came up yet again after this fix, and we might need a better fix for the cyclic dependency problem.

Can you please be more specific? I have a fix in mind for the general problem and would like to see if it would work.

AnsonYeung:
I had found out a fix for this now. Apparently it's just a matter of removing the import of `Predef.mls` from several files. `Runtime.mls` should not import from `Predef.mls` directly or indirectly.

Some of the stuff defined in `Predef.mls` runs actual code and HandlerLowering cannot statically determine whether they will raise effect and as a result instrument them. When `Runtime.mls` import `Predef.mls` transitively, `Predef.mls` will also import `Runtime.mls` which will only get a placeholder due to cyclic import. If the code for initialization actually uses this runtime placeholder, it will crash with the runtime cyclic dependency error.

AnsonYeung:
more specifically, it's the effect checking code that uses the `runtime`, during the construction of `Refl` super call. We can simply not insert the check but I think in general some modules may want to do some precomputation, in which case it will cause a problem because the stack safety instrumentation will try to access runtime and fail the initialization due to cyclic import.

The changes I made to my handler-benchmark-patch branch: remove Predef.mls import from Option.mls, Stack.mls, Iter.mls and change `let x() = ...` into `fun f() = ...` as the former one will emit lambdas, causes lifter to emit some code that seems to cause usage of `runtime`.

PR #523, opened 2026-06-16
codegen/wasm: Make codegen APIs multi-value aware

Derppening:
Reduces the need to emit a lot of `nop` and `block`s.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
@@ -414,12 +414,14 @@ case class FoldedInstr(
   /** Returns the result type of this instruction if this instruction only has 0-1 result values. */
   def resultType: Opt[Type] = resultTypes match
     case Seq() => N
-    case ty :: Seq() => S(ty)
-    case _ => lastWords(s"resultType_! called on instruction with multi-value result type: $this")
+    case Seq(ty) => S(ty)
+    case tys => 
+      lastWords(s"resultType called on instruction `$mnemonic` with multi-value result type: ${tys.map(ty => doc"`${ty.toWat}`").mkDocument(doc"[", doc", ", doc"]").mkString()}")
 
   /** Returns the singular result type of this instruction, otherwise throws an exception. */
   def resultType_! : Type = resultType.getOrElse:
-    lastWords(s"resultType_! called on instruction with a non-unique result type: $this")
+    lastWords:
+      s"resultType_! called on instruction `$mnemonic` with non-unique result type(s): ${resultTypes.map(_.toWat).mkDocument(doc"[", doc", ", doc"]").mkString()}"
 
   def toWat: Document = doc"($mnemonic${
       instrargs.map: a =>
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
@@ -416,12 +416,13 @@ case class FoldedInstr(
     case Seq() => N
     case Seq(ty) => S(ty)
     case tys => 
-      lastWords(s"resultType called on instruction `$mnemonic` with multi-value result type: ${tys.map(ty => doc"`${ty.toWat}`").mkDocument(doc"[", doc", ", doc"]").mkString()}")
+      lastWords:
+        s"resultType called on instruction `$mnemonic` with multi-value result type: ${tys.map(ty => doc"`${ty.toWat}`").mkDocument(doc"[", doc", ", doc"]").mkString()}"
 
   /** Returns the singular result type of this instruction, otherwise throws an exception. */
   def resultType_! : Type = resultType.getOrElse:
     lastWords:
-      s"resultType_! called on instruction `$mnemonic` with non-unique result type(s): ${resultTypes.map(_.toWat).mkDocument(doc"[", doc", ", doc"]").mkString()}"
+      s"resultType_! called on instruction `$mnemonic` with no result type: ${resultTypes.map(_.toWat).mkDocument(doc"[", doc", ", doc"]").mkString()}"
 
   def toWat: Document = doc"($mnemonic${
       instrargs.map: a =>
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
@@ -411,7 +411,23 @@ case class FoldedInstr(
     resultTypes: Seq[Type],
 ) extends Instruction:
 
-  /** Returns the result type of this instruction if this instruction only has 0-1 result values. */
+  /** Returns the result type of this instruction if this instruction has exactly 0-1 result values.
+    *
+    * This is provided as a shorthand since the majority of instruction builders (in [[Instructions]]) are designed to
+    * take one value per operand argument, and either place no value (represented by `N`) or one value (represented by
+    * `S(ty)`) on the stack.
+    * 
+    * The use of these APIs with multi-value instructions, for instance:
+    *
+    * ```scala
+    * i32.add(
+    *   call(returnTypes = Seq(Result(I32Type), Result(I32Type)), /* ... */),
+    *   i32.const(1),
+    * )
+    * ```
+    *
+    * is not supported and is therefore rejected by this function, which throws an exception.
+    */
   def resultType: Opt[Type] = resultTypes match
     case Seq() => N
     case Seq(ty) => S(ty)
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
@@ -419,11 +419,6 @@ case class FoldedInstr(
       lastWords:
         s"resultType called on instruction `$mnemonic` with multi-value result type: ${tys.map(ty => doc"`${ty.toWat}`").mkDocument(doc"[", doc", ", doc"]").mkString()}"
 
-  /** Returns the singular result type of this instruction, otherwise throws an exception. */
-  def resultType_! : Type = resultType.getOrElse:
-    lastWords:
-      s"resultType_! called on instruction `$mnemonic` with no result type: ${resultTypes.map(_.toWat).mkDocument(doc"[", doc", ", doc"]").mkString()}"
-
   def toWat: Document = doc"($mnemonic${
       instrargs.map: a =>
         a match
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
@@ -983,7 +983,7 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
       exprs: Seq[Expr],
       isAbortive: Bool,
   )(using Ctx, FunctionCtx, Raise, SessionExportCtx): Seq[Expr] =
-    if exprs.flatMap(_.resultTypes).isEmpty && !isAbortive then
+    if exprs.forall(_.resultTypes.isEmpty) && !isAbortive then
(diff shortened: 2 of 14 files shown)


hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -948,42 +961,32 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
                   returnTypes = Seq(Result(RefType.anyref)),
                 )
-                blockInstr(
-                  label = N,
-                  children = Seq(asStatement(prefixWat), drop(superCall)),
-                  resultTypes = Seq.empty,
-                )
-              case N =>
-                nop
+                prefixWat.toSeq :+ drop(superCall)
```
LPTK: I'd rather make the type `Vector` to make sure this `:+` has the right complexity.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -948,42 +961,32 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
       isAbortive: Bool,
-  )(using Ctx, FunctionCtx, Raise, SessionExportCtx): Expr =
-    if expr.resultTypes.isEmpty && !isAbortive then
-      blockInstr(
-        label = N,
-        children = Seq(expr, result(State.unitBlockMemberSymbol.asMemberRef(State.unitSymbol))),
-        resultTypes = Seq(Result(RefType.anyref)),
-      )
+  )(using Ctx, FunctionCtx, Raise, SessionExportCtx): Seq[Expr] =
+    if exprs.flatMap(_.resultTypes).isEmpty && !isAbortive then
```
LPTK: Don't construct unnecessary intermediate collections!

```suggestion
    if exprs.forall(_.resultTypes.isEmpty) && !isAbortive then
```

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -948,42 +961,32 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
...
+      exprs :+ result(State.unitBlockMemberSymbol.asMemberRef(State.unitSymbol))
```
LPTK: Same remark: better make this explicitly typed as a Vector.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
```diff
@@ -2111,24 +2071,28 @@ class WatBuilder(using TraceLogger, State) extends CodeBuilder:
         def getScrutExpr: Expr =
           scrutLocalResult.fold(result(scrut))(getLocalAnyref)
 
-        def assignTailResult(target: LocalIdx, expr: Expr): Expr =
-          if expr.isControlTransfer then expr
+        def assignTailResult(target: LocalIdx, exprs: Seq[Expr]): Seq[Expr] =
+          if exprs.isEmpty || exprs.last.isControlTransfer then exprs
```
```suggestion
          if exprs.lastOption.forall(_.isControlTransfer) then exprs
```

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
LPTK: Same remarks in other places where things are appended or concatenated.

hkmc2/shared/src/test/mlscript/wasm/ControlFlow.mls
```diff
@@ -18,40 +18,9 @@ let i = 0 in
 //│   (type $Object (sub (struct (field $$typeinfo (mut (ref $TypeInfoBase))))))
 //│   (type $lt_impl (func (param $lhs (ref null any)) (param $rhs (ref null any)) (result (ref null any))))
 //│   (type $plus_impl (func (param $lhs (ref null any)) (param $rhs (ref null any)) (result (ref null any))))
-//│   (type $Unit_typeinfo (sub $TypeInfoBase (struct (field $$tag i32) (field $$parent (ref null $TypeInfoBase)))))
-//│   (type $Unit (sub $Object (struct (field $$typeinfo (mut (ref $TypeInfoBase))))))
-//│   (type $Unit_init (func (param $this (ref null any)) (result (ref null any))))
-//│   (type $Unit_ctor (func (result (ref null any))))
 //│   (type $entry (func (result (ref null any))))
-//│   (type $start (func))
```
LPTK, line 26: I wonder why all these are deleted, now. Is it because we no longer emit `Unit` in some cases?

LPTK (approved):
LGTM otherwise. Nice little improvement!

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
Derppening: Addressed in 01b5dc954c124c6ce7f311a381f8079d3a422947.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
Derppening: Addressed in 7f66e9069d511126a1803e07649ed053ec111935. `toSeq`s that appear as an intermediate collection is now converted into `toVector`.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
Derppening: Applied in 3fc7e430c9e677949696cc9df71bb781b96ab2b0.

hkmc2/shared/src/test/mlscript/wasm/ControlFlow.mls
Derppening: Yes, some redundant `Unit`s are no longer emitted.

We have a function `assignTailResult` in `WatBuilder` that sets the result value of a `match` expression to `Unit` if the match does not result in a value, and part of the heuristic is to check whether the last instruction is a control transfer (and if it is, the `local.set` to `$Unit$inst` is redundant since it is unreachable). That check used to always fail because the control transfer instruction is always buried within layers of nested blocks.

Now that we flatten the expression tree, the control transfer instruction is "bubbled" to the top-level, and thus the check now correctly fires and omits the redundant setting of `Unit`.

LPTK: What do you mean by "we flatten the expression tree"?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
LPTK: Don't forget to also update the _types_. The `Seq` type in general is pretty useless as it's too loose. I prefer using `List` when things are build in stack-like manner and no repeated concatenation/appension is needed, and `Vector` when it is (like here). `Seq` is ok fine if you're doing neither.

hkmc2/shared/src/test/mlscript/wasm/ControlFlow.mls
Derppening: What I was referring to is that Wasm expressions within (deeply) nested `block`s are now "flattened" into the `Seq[Expr]` that this PR converts into.

For instance, this pattern in the current version:

```wat
//│                       (then
//│                         (block $arm
//│                           (block
//│                             (block
//│                               (local.set $tmp
//│                                 (call $plus_impl
//│                                   (local.get $i)
//│                                   (ref.i31
//│                                     (i32.const 1))))
//│                               (block
//│                                 (local.set $i
//│                                   (local.get $tmp))
//│                                 (br $lbl_cont)))        ;; <--
//│                             (local.set $matchRes
//│                               (global.get $Unit$inst)))
//│                           (br $match))))
```

is flattened into:

```wat
//│                   (then
//│                     (block $arm
//│                       (local.set $tmp
//│                         (call $plus_impl
//│                           (local.get $i)
//│                           (ref.i31
//│                             (i32.const 1))))
//│                       (local.set $i
//│                         (local.get $tmp))
//│                       (br $lbl_cont)         ;; <--
//│                       (br $match)))))
```

As described, `assignTailResult` function checks if the last expression (indicated by `<--`) is a control transfer instruction. In the old version, the last expression is a `block` that returns no value, and thus the setting of `$matchRes` to `$Unit$inst` is emitted. With this PR, blocks are now created only when it is needed, and thus `assignTailResult` now sees that the last expression *is* a control transfer, and thus omits the `$Unit$inst` assignment.

(I notice that `br $match` is also redundant here. Let me patch this as well.)

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/WatBuilder.scala
Derppening: Applied your suggestion in 9f4226a55eb95066e5fdceeeb6f229e797618a9b. Let me know if there are other amendments I should make here.

hkmc2/shared/src/test/mlscript/wasm/ControlFlow.mls
LPTK: Oh, ok, that's good!

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
```diff
@@ -414,12 +414,14 @@ case class FoldedInstr(
   /** Returns the result type of this instruction if this instruction only has 0-1 result values. */
   def resultType: Opt[Type] = resultTypes match
```
LPTK, line 431: This function ~~never returns `N` and instead~~ duplicates the `resultType_!` crash!

Derppening: I don't understand the issue here. `resultType` is used to retrieve the (optionally) singular result type of the expression - It returns `S` if the expression returns something and `N` if it doesn't. 

The crash message also differs very slightly which is why both are needed. I updated the wording in 1bcb925 to clarify the cause of the error.

LPTK: I just don't uynderstand why the two situations "zero field" and "strictly more than one field" are handled differently. Is there any reason that users of this function should always be expected to know that a `resultType` will never have more than one result, whereas we provide special support for the "zero" case?

Derppening: Addressed in c68b10f78a2ef29ffe81b0e53bb1e8fa7283cdf8. `resultType_!` is removed entirely since there are only two use sites.

hkmc2/shared/src/test/mlscript/wasm/ControlFlow.mls
LPTK: > (I notice that br $match is also redundant here. Let me patch this as well.)

Will you do this here, or in a separate PR?

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
LPTK: This doesn't really address my question, though. Could you please briefly document why it's useful to have a function that returns N when the sequence is empty and throws when it's bigger than 1?

hkmc2/shared/src/test/mlscript/wasm/ControlFlow.mls
Derppening: Done in 5a866909e8de1b6f772e9ec759872dc63b7ffbf3, and while I was at it I also landed some simplifications in 4c4fa2a95b97976ffe62e9d67c8ece1bc03c7ddb.

hkmc2/shared/src/main/scala/hkmc2/codegen/wasm/text/Wasm.scala
Derppening: After several revisions, I decided on b85ea8ebc84517846033fc2a05567848439ce8f2. Please let me know if this better addresses your question.

PR #524, opened 2026-06-16
Fix tailrec miscompilation with spreads

CAG2Mark:
Fixes a bug where the following
```fs
fun f(a, ...z) = if z is
  [] then a
  [x, ...xs] then f(a + 1, ...xs)
```
would have the spread parameter changed into a normal parameter by the tailrec optimizer.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
@@ -508,9 +508,13 @@ class TailRecOpt(using State, TL, Raise):
       if funs.length === 1 then initial
       else Param.simple(curIdSym) :: initial
     
+    val newParamLists =
+      if funs.length === 1 && funs.head.params.size <= 1 then funs.head.params
+      else PlainParamList(params) :: Nil
+    
     val loopDefn = FunDefn(
       owner, bms, dSym,
-      PlainParamList(params) :: Nil,
+      newParamLists,
       loop)(N, annotations = Nil) // Q: maybe should be Private?
     
     if funs.size === 1 then
--- a/hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
@@ -264,7 +264,13 @@ class TailRecOpt(using State, TL, Raise):
 
     val maxParamLen = maxInt(funs, paramsLen)
     val paramSyms =
-        if funs.length === 1 then (getParamSyms(funs.head))
+        if funs.length === 1 then
+          val syms = getParamSyms(funs.head)
+          if funs.head.params.length === 1 then syms
+          else
+            // Duplicate the params for the internal loop defn (see the doc at the 
+            // end of this function), but preserve the names.
+            syms.map(v => VarSymbol(Tree.Ident(v.id.name)))
         else
           for i <- 0 until maxParamLen yield VarSymbol(Tree.Ident("param" + i))
       .toList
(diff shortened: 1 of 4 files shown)


CAG2Mark:
Argh, it seems like https://github.com/hkust-taco/mlscript/pull/470 introduced a bug where the internal loop (for functions with multiple parameter lists) uses the same symbols as the original function and causes the symbols to be duplicated. I should also fix that

hkmc2/shared/src/test/mlscript/tailrec/MultiArgLists.mls
```diff
@@ -194,3 +194,38 @@ fun g(n)(m) =
...
+//│         set tmp = +⁰(a, 1);
+//│         set a = tmp;
+//│         set z = xs;
+//│         continue loopLabel
+//│       else
+//│         throw new globalThis⁰.Error⁰("match error")
+//│     end
+//│   end
+//│ };
+//│ define f² as fun f³(a)(...z) { return f$tailrec⁰(a, z) };
```
LPTK, line 227: I wonder why this one isn't annotated as `@inline`. Such "wrapper" functions should be.

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
```diff
@@ -508,9 +508,13 @@ class TailRecOpt(using State, TL, Raise):
       if funs.length === 1 then initial
       else Param.simple(curIdSym) :: initial
     
+    val newParamLists =
+      if funs.length === 1 && funs.head.params.size <= 1 then funs.head.params
```
LPTK: There are many `funs.length === 1` and `funs.size === 1` expressions here and above (incl around L267). Each incurs a traversal of the whole list. Would be much better to compute this once and store the result in a descriptively-named variable.

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
LPTK: Side note: better use `.iterator.map`

LPTK: `.iterator.flatMap`

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
LPTK: In general, having so many list-length checks would normally point to fishy logic that could be refactored in a less error-prone design. But I don't have a concrete improvement suggestion right now.

LPTK (approved):
Otherwise LGTM. Thanks for the fixes.

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
CAG2Mark: I've refactored it a bit in the latest commit.

hkmc2/shared/src/test/mlscript/tailrec/MultiArgLists.mls
```diff
@@ -20,7 +20,7 @@ fun loop(x)(y) =
 loop(1000)(0)
 //│ ———————————————| Lowered IR |———————————————————————————————————————————————————————————————————————
 //│ let loop$tailrec;
-//│ @private
+//│ @inline @private
```
LPTK: I think you made the wrong function `@inline`. This function should _not_ be `@inline`! Only wrappers like `define f² as fun f³(a)(...z) { return f$tailrec⁰(a, z) }` should be.

CAG2Mark: Oh, I see.

CAG2Mark:
Actually I guess if `f` is inline, then the internal loop definition should be made inline too

PR #525, opened 2026-06-17
Improve compilation tests

AnsonYeung:
Remove extra imports and refactor `let` into `private fun` which will not create a lambda and directly bind as a method instead of a field storing lambda.


LPTK (approved):
LGTM, thx.

PR #527, opened 2026-06-18
Add language config directive support

LPTK:
Adds parsing support for a new `Config.language` field and introduces `#lang(...)` as shorthand for `#config(language: ...)`.

Also includes:

- parsing all available `Language` fields through `#lang` and `Language(...)`
- preserving `versionName` as preset metadata, not a directive-overridable field
- regression coverage for language directives
- a DRY refactor of `ConfigParser` using shared helpers/extractors for named args, calls, optional fields, and config field parsing


LPTK (commented):
I reviewed "my" code and found nothing wrong.

PR #528, opened 2026-06-18
Support symbolic suffix aliases for identifiers

LPTK:
This adds support for Scala-style symbolic suffixes to hkmc2 identifiers, so names like `foo_<:<`, `value_?!`, and `Box_#` are accepted in source where ordinary identifiers already work.

The important semantic bit is that a suffixed identifier keeps a normal canonical name internally: for example, `foo_abc_<:<` is treated like source alias `foo_abc_<:<` for canonical name `foo_abc`.

So a definition like

```
fun foo_abc_<:< = ...
```

is basically syntax sugar for

```
fun (foo_abc_<:<) foo_abc = ...
```

Summary:

- Extend lexing so alphanumeric identifiers may include a symbolic suffix after `_`, excluding `.` so selections still parse correctly.
- Split suffixed identifiers into canonical names plus source-only aliases for definitions, params, fields, and members.
- Preserve aliases through symbols, symbol refreshing, member maps, and class/module field elaboration.
- Lower resolved selections and assignments to canonical member names so JS codegen emits normal JS identifiers.
- Add regression coverage for symbolic suffix functions, params, class names, fields, and methods.
- Update golden outputs for cases that now resolve through symbolic aliases, including `|>>` and `::`.


chengluyu (approved):
This is a convenient feature. I didn’t find anything suspicious.

PR #529, opened 2026-06-20
Make `private` members actually private in generated JS code

LPTK:
This PR implements proper codgen for private field and method members (`val` and `fun`) by emitting JS private names, checking private accesses during resolution, and generating accessors for extruded private members. Also fixes the codegen of `return undefined` in MIR by emitting `return runtime.Unit` so that, eg, dead-param elimination can keep removing unused private function parameters safely.

---

Mostly implemented by Codex, with close supervision and manual edits.


flandia (approved):
LGTM!

PR #530, opened 2026-06-23
Share common symbols across files

LPTK:
Previously, each compilation unit created its own Prelude elaboration or inherited one from whoever was importing it and triggering its compilation first. Now, all the compilation units share a single Prelude elaboration, and consequently they all agree on the builtin symbols declared therein.

Moreover, we used to use fake symbols representing definitions from the `Runtime.mls` module. But these fake symbols disagreed with the real ones created when elaborating that module. Now, we instead use the compiler cache to retrieve the actual symbols from the `Runtime.mls` elaboration result – unless we are already in the process of compiling that file, in which case we just look them up in the current elaborated block.

Both of these should greatly simplify the cross-file inlining process being prototyped in https://github.com/LPTK/mlscript/pull/27 (currently, it needs to rely on ad-hoc mappings).

---

Mostly implemented by Codex, with close supervision and manual edits.

### DIFF
--- a/hkmc2/js/src/main/scala/hkmc2/Compiler.scala
+++ b/hkmc2/js/src/main/scala/hkmc2/Compiler.scala
@@ -63,7 +63,8 @@ class Compiler(paths: MLsCompiler.Paths)(using cctx: CompilerCtx):
     perFileDiagnostics
 
 @JSExportTopLevel("Paths")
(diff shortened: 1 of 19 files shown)


chengluyu (approved):
Looks like a nice cleanup. LGTM.

PR #534, opened 2026-06-25
Remove double lowering

AnsonYeung:
Make `:sjs` more accurate. The old code was using a separate compilation pass which does not use the same IR. Now both `:ssjs` and `:sjs` shares the same lowered IR and only differs in the final step.

Notable changes from test:
- implicit `return` are mapped to assignment
- worker wrapper transformation is also applied
- JS output now appears after IR
- Lowered IR no longer contains sanity check (see first commit in this PR)


hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -1048,7 +1048,7 @@ sealed abstract class Path extends TrivialResult:
 /**
  * @param symbol The symbol representing the definition that the selection refers to, if known.
  */
-case class Select(qual: Path, name: Tree.Ident)(val symbol: Opt[DefinitionSymbol[?]]) extends Path with ProductWithExtraInfo:
+case class Select(qual: Path, name: Tree.Ident)(val symbol: Opt[DefinitionSymbol[?]], val sanitize: Boolean = false) extends Path with ProductWithExtraInfo:
```
LPTK: This should probably be
```suggestion
case class Select(qual: Path, name: Tree.Ident)(val symbol: Opt[DefinitionSymbol[?]])(val sanitize: Boolean) extends Path with ProductWithExtraInfo:
```

Note also the lack of default argument.

https://github.com/hkust-taco/mlscript/blob/e0c682ef43fdd3e21b771f1f0d772c6e5ad55134/AGENTS.md#L55-L56

hkmc2/shared/src/test/mlscript/backlog/NonReturningStatements.mls
```diff
@@ -31,7 +31,7 @@ fun foo =
 :sjs
 foo
 //│ ————————————| JS (unsanitized) |————————————————————————————————————————————————————————————————————
-//│ Predef.print(1); Predef.print("..."); return runtime.Unit
+//│ Predef.print(1); Predef.print("..."); block$res6 = runtime.Unit;
```
LPTK: I don't like this change. Because it contains a counter, it's going to create _lots_ of diff churn. It was better to just see a `return`, here. Can't we print the `:sjs` before the return is mapped?

LPTK (changes requested):
Very nice! But I have some minor change requests.

hkmc2/shared/src/test/mlscript/opt/CaseOfCase.mls
```diff
@@ -356,16 +348,14 @@ x => if partlyKnownProducer(x) is
...
-//│       set selRes = x.length﹖;
-//│       do x.length$__checkNotMethod﹖;
-//│       match selRes
-//│         undefined =>
-//│           throw new globalThis⁰.Error﹖("Access to required field 'length' yielded 'undefined'")
+//│       set inlinedVal = x.length﹖;
+//│       match inlinedVal
 //│         0 =>
 //│           return Predef⁰.print⁰("Z")
 //│         1 =>
```
AnsonYeung, line 354: This regression seems to be because selRes was a separated local before and it isn't now. Either way this will not appear after we have dead assignment removal

LPTK (approved):
LGTM, thx!

PR #539, opened 2026-07-01
Compilation Pipeline

AnsonYeung:
Move all lowered IR transformation pass inside a single class, added `preOptimizeHook` for JSBackendDiffMaker to print IR before optimization but after transformations.

### DIFF
--- a/hkmc2/shared/src/main/scala/hkmc2/Config.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/Config.scala
@@ -85,15 +85,7 @@ object Config:
     commentGeneratedCode = false,
     noFreeze = false,
     noModuleCheck = false,
-    optimizer = Optimizer(
-      deforest = N,
-      tailRecOpt = true,
-      inlining = S(Inliner(default.inlineThreshold)),
-      deadBranchRemoval = default.deadBranchRemoval,
-      deadCodeElim = true,
-      dataFlowAnalysis = true,
-      deadParamElim = S(DeadParamElim.default),
-    ),
+    optimizer = Optimizer.FastOpt,
   )
   object default:
     val patMatConsequentSharingThreshold = S(15)
@@ -253,7 +245,7 @@ object Config:
   )
   
   object Optimizer:
-    val noOpt = Optimizer(
+    val NoOpt = Optimizer(
       N,
       false,
       N,
@@ -261,6 +253,16 @@ object Config:
       false,
       false,
       N
+    )
+    
+    val FastOpt = Optimizer(
+      deforest = N,
+      tailRecOpt = true,
+      inlining = S(Inliner(default.inlineThreshold)),
+      deadBranchRemoval = default.deadBranchRemoval,
+      deadCodeElim = true,
+      dataFlowAnalysis = true,
+      deadParamElim = S(DeadParamElim.default),
     )
 
 end Config
@@ -609,7 +611,7 @@ object ConfigParser:
   private def parseField(name: Str, value: Tree)(using Raise): Config => Config = name match
     case "language" => parseLanguageOverride(value)
     case "noOpt" =>
-      parsedField(value)(parseBool)(v => _.mapOptimizer(_ => Optimizer.noOpt))
+      parsedField(value)(parseBool)(v => _.mapOptimizer(_ => Optimizer.NoOpt))
     case "tailRecOpt" =>
       parsedField(value)(parseBool)(v => _.mapOptimizer(_.copy(tailRecOpt = v)))
     case "noFreeze" =>
--- a/hkmc2/shared/src/main/scala/hkmc2/Config.scala
+++ b/hkmc2/shared/src/main/scala/hkmc2/Config.scala
@@ -36,6 +36,7 @@ case class Config(
   funcToCls: Bool,
   commentGeneratedCode: Bool,
   noFreeze: Bool,
+  noOpt: Bool,
   noModuleCheck: Bool,
   deadParamElim: Opt[DeadParamElim],
 ):
@@ -80,6 +81,7 @@ object Config:
     funcToCls = false,
     commentGeneratedCode = false,
     noFreeze = false,
+    noOpt = false,
     noModuleCheck = false,
     deadParamElim = S(DeadParamElim.default)
   )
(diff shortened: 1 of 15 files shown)


hkmc2/shared/src/main/scala/hkmc2/codegen/CompilationPipeline.scala
```diff
@@ -0,0 +1,63 @@
...
+      result = Program(result.imports, Lifter(result.main).transform)
+    
+    result = config.effectHandlers.fold(result): opt =>
+      HandlerLowering(new HandlerPaths, opt).translateProgram(result)
+    
+    result = Program(result.imports, result.main.flattened)
+    
+    result = BufferableTransform().transform(result)
+    
+    // * TODO[Anto]: Can we remove MergeMatchArmTransformer? Seems no longer necessary
```
LPTK: I think this comment is outdated. It turned out `MergeMatchArmTransformer` is actually still quite useful to merge matches.

hkmc2/shared/src/main/scala/hkmc2/Config.scala
```diff
@@ -36,6 +36,7 @@ case class Config(
   funcToCls: Bool,
   commentGeneratedCode: Bool,
   noFreeze: Bool,
+  noOpt: Bool,
```
LPTK: Instead of a Boolean flag, let's have an `Optimizer` case class with one field per kind of optimization (incl. the deforestation), and some presets (similar to the language version presets), such as `NoOpt`, `FastOpt` (the default), and `FullOpt`.

hkmc2DiffTests/src/test/scala/hkmc2/JSBackendDiffMaker.scala
```diff
@@ -121,36 +121,37 @@ abstract class JSBackendDiffMaker extends MLsDiffMaker:
...
+                output(irStr)
+            preOptimize = S(prog)
+            super.preOptimizeHook(prog)
+        customPipeline.run(lowered, print, symbolsToPreserve, dtl)
       
+      // TODO: remove this and make sure all transformer preserve object identity
+      val loweredTransformed = preOptimize.get
       // TODO: Test that transformers retain object identity when there are no changes
-      if (optimized isnt lowered) && (optimized === lowered) then
+      if (optimized isnt loweredTransformed) && (optimized === loweredTransformed) then
```
LPTK: I think this should be checked on every individial optimization pass, with a "hook" similar to `preOptimizeHook`.

LPTK (approved):
Cool, thanks! LGTM otherwise.

hkmc2/shared/src/main/scala/hkmc2/codegen/CompilationPipeline.scala
```diff
@@ -0,0 +1,66 @@
...
+      CompilationPass("ReflectionInstrumenter", ReflectionInstrumenter(using summon).apply),
+      CompilationPass("TailRecOpt", TailRecOpt().transform),
+      CompilationPass("PreOptimizeHook", preOptimizeHook),
+      CompilationPass("WorkerWrapper", WorkerWrapper(symbolsToPreserve, otl, printer)),
+      CompilationPass("BlockSimplifier", BlockSimplifier(symbolsToPreserve, otl, printer).apply),
+      CompilationPass("DeadParamElim", otl.givenIn(DeadParamElim.apply)),
+    )
+    allPasses.foldLeft(prog): (before, pass) =>
+      val after = pass.transform(before)
+      passHook(pass, before, after)
```
LPTK: Is there any advantage to using a concrete list of instances of some one-off case class followed by a foldLeft versus just directly calling a helper function several times?

LPTK: The current way seems both cumbersome and inefficient.

AnsonYeung: The advantage is just to insert between pass instrumentation like object identity check

LPTK: But this can be done with a function as well! What do the case class and intermediate list buy you?

AnsonYeung:
```scala
result = Pass(...).transform(result)
hook("PassName", result)
```
versus
```scala
CompilationPass("PassName", Pass(...).transform(_))
```

I guess not that significant, it's just to print the IR pass name

LPTK: For the record, I prefer using a simple mutable local variable over this.

AnsonYeung: Sure, maybe I gone a bit too far when I saw the repeated `hook()` calls

LPTK: But you don't even need to repeat these calls. You can use a single call to a helper function receiving both the pass name and a lambda. That helper function can both update the mutable variable and call the hook. You shouldn't need two mutable variables either.

Basically, your idea of using lamdas was fine, but the case class and list folding were totally unnecessary.

hkmc2DiffTests/src/test/scala/hkmc2/JSBackendDiffMaker.scala
```diff
@@ -128,7 +128,7 @@ abstract class JSBackendDiffMaker extends MLsDiffMaker:
         val customPipeline = new CompilationPipeline:
           override def passHook(passName: Str, before: Program, after: Program) =
             // TODO: fix these passes
-            val excludedPassNames = Set("Lifter", "Lifter after FirstClassFunctionTransformer")
+            val excludedPassNames = Set.single("Lifter")
```
LPTK: You're building a constant singleton set data structure every time this function is called. Why not lift it out?

LPTK: Also, do we really need a set at this point?

AnsonYeung: No, it was a Set before to exclude multiple passes. I tried fixing the lifter but it was more than just a couple of places to fix

hkmc2/shared/src/main/scala/hkmc2/codegen/CompilationPipeline.scala
AnsonYeung: I don't know why I didn't thought of that in the first place

PR #541, opened 2026-07-04
Make tail-rec optimization improvements

LPTK:
- Run `TailRecOpt` both before and after simplification so annotations are checked early while later optimization opportunities exposed by simplifications can still be rewritten.
- Add a `checkAnnotations` mode to suppress duplicate/spurious annotation diagnostics on the second TailRecOpt run.
- Fix TailRecOpt’s call graph so under-applied curried calls are not treated as executable recursive call edges.
- Add/update regression coverage for multi-argument-list tail recursion, including the partial-application case that previously caused the second pass to reprocess already optimized code.
- Refresh affected golden outputs, including `Runtime.mjs`.

---

Codex was used to help diagnose issues and change code, but the design was mine (the fixes proposed by Codex were all stupid workarounds eschewing the real issue).


hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
```diff
@@ -247,13 +271,13 @@ class TailRecOpt(using State, TL, Raise):
       case c: CallEdge.NormalCall => c.f2 -> c.call
     val nonTailCalls = nonTailCallsLs.toMap
     
-    if nonTailCallsLs.sizeCompare(calls) === 0 then
-      for f <- funs if f.tailRec do
-        raise(WarningReport(msg"This function does not directly self-recurse, but is marked @tailrec." -> f.dSym.toLoc :: Nil))
+    if calls.isEmpty then
```
CAG2Mark, line 274: This changes the semantics of the code. Before, this error message would be raised if there were no tail-calls within the SCC. Now, it is only raised when there are no calls within the SCC at all (i.e., it never self-recurses).

It seems the original condition was mismatched with the original error message. If we change the error message to the original one, then the condition and the reported error will match.

CAG2Mark (approved):
Just one small comment about the error message.

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
LPTK: Can you please commit the optimal changes you think are desired, here?

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
```diff
@@ -247,13 +271,13 @@ class TailRecOpt(using State, TL, Raise):
...
+      for f <- funs if checkAnnotations && f.tailRec do
+        raise(WarningReport(msg"This function is marked @tailrec but has no apparent tail calls." -> f.dSym.toLoc :: Nil))
```
CAG2Mark: Something like this:
```suggestion
        raise(WarningReport(msg"This function is marked @tailrec but has no direct self-recursion." -> f.dSym.toLoc :: Nil))
```

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
CAG2Mark: See my latest comment

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
LPTK: What is "self-recursion"? Do you mean "has no apparent recursive calls" or "recursive call edges"?

CAG2Mark: Something like that. Was not sre how to phrase it

hkmc2/shared/src/test/mlscript/tailrec/Simple.mls
```diff
@@ -26,19 +26,15 @@ module Foo with
...
-//│ f3 = function f(x) {
-//│   loopLabel: while (true) {
-//│     let tmp, tmp1;
-//│     tmp = f3(x);
-//│     tmp1 = tmp + 1;
-//│     x = tmp1;
-//│     continue loopLabel;
-//│   }
-//│ };
+//│ let f3; f3 = function f(x) { let tmp, tmp1; tmp = f3(x); tmp1 = tmp + 1; return f3(tmp1) };
```
LPTK: @CAG2Mark This comment was written by Codex. I think it was post-rationalizing what looks like a regression to me. Looks like it _sneakily_ introduced a little `return (N, funs)` in what were apparently unrelated changes, in b99eb95f08f8f22885d4ceb2fe11a3e48c163e70.

I reverted this change. Do you agree that this function should indeed be optimized?

CAG2Mark: This should definitely be optimized. I don't know what Codex is on about

PR #542, opened 2026-07-12
Fix handler loop bug

AnsonYeung:
There is a typo in replaceStaleLabels. Adds regression test and fixes the issue.


LPTK (approved):
LGTM.

PR #544, opened 2026-08-06
JS Generators

AnsonYeung:
Implements generators with `@generator` annotations and `yield` operators that compiles to JavaScript generators.


hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
```diff
@@ -654,9 +654,9 @@ class TailRecOpt(checkAnnotations: Bool)(using State, TL, Raise):
       case f: FunDefn => L(f)
       case c: ClsLikeDefn => R(c)
       case _ => die // unreachable as floatOutDefns only floats out FunDefns and ClsLikeDefns
-    // Filter out functions that have a @config annotation disabling tailRecOpt
+    // Filter out functions that should not be tail rec optimized
```
LPTK:
```suggestion
    // Filter out functions that should not be tail-rec optimized
```

LPTK (changes requested):
Nice, the changes are pretty minimal!

But you forgot to write any *functional* tests! None of the test case actually runs the defined generator... And `yield*` should be tested properly too.

Also, it's better to separate the regression tests for obviously broken snippets into a separate `BadGenerators.mls`.

hkmc2/shared/src/test/mlscript/codegen/Generators.mls
```diff
@@ -46,11 +46,47 @@ test.next(123)
...
+//│ };
+//│ —————————————————| Output |—————————————————————————————————————————————————————————————————————————
+
+fun printGenerated(generator) =
+  let result = generator.next()
+  while not result.done do
+    print(result.value)
+    set result = generator.next()
+
+printGenerated(traverse([[1, [2, 3]], 4]))
```
LPTK, line 109: Why not also apply this to all generators in the file?

AnsonYeung: The other generator are tested using the js interface directly

LPTK (approved):
LGTM, thanks!

PR #546, opened 2026-08-21
Scc Analyzer and FlowAnalysis Only Register Upper Bounds in `var < var` Case

ychenfo:
Add an SccAnalyzer which allow users to override the following 3 methods:
- `successors`: The succesors of each node
- `handleScc`: Every node reachable from any query is visited at most once, and each SCC is reported exactly once to `handleScc`. By the time `handleScc(scc)` runs, every SCC reachable from `scc` other than `scc` itself has already been handled. `handleScc` must not itself start a query.
- `isHandled`: Whether a node's SCC has already been handled. Must become true for every member of an SCC after `handleScc` is done.

This allows querying the scc from a certain root without calculating the whole graph's scc while keeping the cache of previous queries.

LLM helped a lot to implement the changes in `UsedVarAnalyzer.scala`. I made some manual tweaks and reviewed them carefully, but it's worth paying extra attention to this file.


LPTK (changes requested):
I really don't like that you deleted the existing declarative interfaces for computing SCCs and replaced their uses by ugly, bug-prone imperative traversals. You should have kept the declarative interfaces (possibly with some improvements to avoid intermediate collections), reimplementing them with the new efficient analysis under the hood.

The new SCC analysis is meant to be a low-level interface we reach for in perf-sensitive parts of the compiler, like global flow analysis. There's no need to rewrite things like TailRecOpt with it, unless we find that it's perf-sensitive (unlikely).

This is a prime example of lettign LLMs make the codebase worse by letting it vomit imperative slop that replaces perfectly good declarative code.

ychenfo:
> I really don't like that you deleted the existing declarative interfaces for computing SCCs and replaced their uses by ugly, bug-prone imperative traversals. You shoudl have kep the declarative interfaces (possibly with some improvements to avoid intermediate collections), reimplementing them with the new efficient analysis under the hood.
> 
> The new SCC analysis is meant to be a low-level interface we reach for in perf-sensitive parts of the compiler, like global flow analysis. There's no need to rewrite things like TailRecOpt with it, unless we find that it's perf-sensitive (unlikely).
> 
> This is a prime example of lettign LLMs make the codebase worse by letting it vomit imperative slop that replaces perfectly good declarative code.

Well I did implement very most of the changes by hand and reviewed carefully the changes done by LLM. And I do also provide the declarative interface `SccAnalysis.sccsFrom(succs: A => IterableOnce[A], roots): Ls[Ls[A]]` called by block simplifier.

On the call sites of the original `partitionScc`: the reason I started this bigger refactor is because that I noticed that in all the call sites of the original `partitionScc` the adjacency list information is somehow already there, and in the previous code they all somehow took this round trip of "adjacency list -> edges list -> adjacency list again" due to the `partitionScc` api.

And on the consumer sites: I realized that the consumer sites of the `partitionScc` result in `TailRecOpt` and `UsedVarAnalyzer` also feels less clean to me (without considering performance) and can be improved with the new scc utility:
- `UsedVarAnalyzer` used this hand-written imperative dp on scc DAG which requires the `sccEdges` to be materialized, but the dp can be fused into `handleScc` natually
- `TailRecOpt` used this `-1` which I don't think is ideal: it's an out of range value in the scc id space needed only because the `groupBy` key doubles as a filter. And I think it under filters because non`@tailcall` cross-scc edges stay in the caller's group and are only dropped ~70 lines later in `optScc`'s `scc.calls.filter(c => fSyms.contains(c.f2))`. The new version is a single `if inScc(c.f2)` and it doesn't have to build and consult `sccMap` to see if two nodes in an call edge is in the same scc.

But now I can see that the diff on the consumer site is probably larger than what we want, I will revert the consumption site changes but I still think it's reasonable to remove the old api requiring edges list and use the new one.

LPTK:
> Well I did implement very most of the changes by hand and reviewed carefully the changes done by LLM.

Ah, I see. I think your original comment below made this more ambiguous:

> LLM helped a lot to implement the changes in `UsedVarAnalyzer.scala`. I made some manual tweaks and reviewed them carefully, but it's worth paying extra attention to this file.

where "helped a lot to implement the changes" sounded like "helped to implement most of the changes". These days it's easy to get a bit paranoid...

> And I do also provide the declarative interface `SccAnalysis.sccsFrom(succs: A => IterableOnce[A], roots): Ls[Ls[A]]` called by block simplifier.

Sorry, I had missed that one.

> But now I can see that the diff on the consumer site is probably larger than what we want

Is it, or are you just saying this to agree with me? Because it seems from what you said above that what you implemented already is the cleaner and more optimal approach.

Let's ask @CAG2Mark what he thinks of the changes concerning his code.

CAG2Mark:
Noted

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
```diff
@@ -201,25 +201,13 @@ class UsedVarAnalyzer(b: Block, scopeData: ScopeData)(using State):
...
-    val (rootId, rootElems) = sccs.find:
-        case (id, elems) => elems.contains(rootInfo)
-      .get
-    if rootElems.size != 1 then lastWords("SCC containing root had a degree other than 1.")
+
+    // The out-edges of `src` inside the subtree rooted at `s`: self-edges do not affect this
+    // analysis, and (very important) we only care about edges that flow into the subtree rooted at `s`
+    def refdInSubtree(src: ScopedInfo): Iterator[ScopedInfo] =
+      accessInfoMap(src).refdDefns.iterator.filter: r =>
+        src =/= r && childInfo.contains(r) && r =/= rootInfo
```
CAG2Mark: For readability's sake, maybe add inline comments explaining the purpose of `src =/= r` and `childInfo.contains(r) && r =/= rootInfo` as in the old code.

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
```diff
@@ -228,38 +216,43 @@ class UsedVarAnalyzer(b: Block, scopeData: ScopeData)(using State):
...
-    def removeUnused(m: Map[ScopedInfo, AccessInfo]) = m.map:
-      case k -> v =>
-        val node = scopeData.getNode(k)
-        k -> v.intersectLocals(node.existingVars)
-    
-    val (m1, m2) = (removeUnused(go(true)), removeUnused(go(false)))
+    def removeUnused(dp: collection.Map[ScopedInfo, AccessInfo]): Map[ScopedInfo, AccessInfo] =
+      dp.map: (info, accesses) =>
+        val node = scopeData.getNode(info)
+        info -> accesses.withoutLocals(node.obj.definedLocals).intersectLocals(node.existingVars)
```
CAG2Mark: `node.obj.definedLocals` is already disjoint with `node.existingVars` so I believe the `withoutLocals` is redundant.

CAG2Mark (commented):
Looks good, just some minor comments. I actually do like this interface a bit more than the old interface for the tailrec and lifter usages. It feels a bit less clunky.

I do still think it is useful to assign each SCC a numeric ID that is accessible in the `handleScc` function, but I guess one can implement themselves easily using the interface if needed, and it seems to not be needed (yet).

ychenfo:
> Ah, I see. I think your original comment below made this more ambiguous:
> 
> > LLM helped a lot to implement the changes in UsedVarAnalyzer.scala. I made some manual > > > tweaks and reviewed them carefully, but it's worth paying extra attention to this file.
>
> where "helped a lot to implement the changes" sounded like "helped to implement most of the changes". These days it's easy to get a bit paranoid...


Thanks! And sorry for the ambiguity — I didn't mean to say I threw unchecked LLM code to others to review.

> Is it, or are you just saying this to agree with me? Because it seems from what you said above that what you implemented already is the cleaner and more optimal approach.

Yeah I do think my changes are improvements and I'd appreciate them getting reviewed and merged, but since the refactor diff is larger than what we discussed earlier before I raised this PR I meant that I am happy to separate the scc-result-use-site changes into another one.

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
ychenfo: Oh indeed I tried to comment out the `withoutLocals` and every thing seems fine, I added `withoutLocals` this here to mirror the `sym -> sccAccessInfo(id).withoutLocals(scopeData.getNode(sym).obj.definedLocals)` (line 254) in the original implementation, not sure if I missed anything.

ychenfo: In the previous code commenting out `withoutLocals` also seems fine. I've removed the `withoutLocals` now.

CAG2Mark: It appears the line you should have mirrored was line 260

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -1036,13 +1031,41 @@ class FlowConstraintSolver(val collector: FlowConstraintsCollector):
   val upperBounds = MutMap.empty[StratVarId, Ls[ConsStrat]].withDefaultValue(Nil)
   val lowerBounds = MutMap.empty[StratVarId, Ls[ProdStrat]].withDefaultValue(Nil)
   
+  object AllUpperBounds extends SccAnalysis[StratVarId]:
```
LPTK: Why not make this extend a trait that already defines and maintains the `computed` mutable map for us?

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
```diff
@@ -228,38 +218,43 @@ class UsedVarAnalyzer(b: Block, scopeData: ScopeData)(using State):
...
-      case k -> v =>
-        val node = scopeData.getNode(k)
-        k -> v.intersectLocals(node.existingVars)
-    
-    val (m1, m2) = (removeUnused(go(true)), removeUnused(go(false)))
+    def removeUnused(dp: collection.Map[ScopedInfo, AccessInfo]): Map[ScopedInfo, AccessInfo] =
+      dp.map: (info, accesses) =>
+        val node = scopeData.getNode(info)
+        info -> accesses.intersectLocals(node.existingVars)
+      .toMap
```
LPTK, line 255: What does this `toMap` do? It's already supposed to be a map.

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
```diff
@@ -228,38 +218,43 @@ class UsedVarAnalyzer(b: Block, scopeData: ScopeData)(using State):
...
+
+    val (m1, m2) = (removeUnused(withIgnored), removeUnused(withoutIgnored))
```
LPTK: Use two separate `val`s.

LPTK (approved):
Otherwise LGTM now.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -1036,13 +1031,41 @@ class FlowConstraintSolver(val collector: FlowConstraintsCollector):
...
+
+    override protected def isHandled(v: StratVarId) = computed.contains(v)
+
+    override protected def handleScc(members: Ls[StratVarId], sccId: Int) =
+      val res = MutSet.empty[ConsStrat]
+      for
+        m <- members
+        ub <- upperBounds(m)
+      do ub match
+        case ConsVar(s) => res.addAll(computed.getOrElse(s.uid, Set.empty[ConsStrat]))
```
LPTK: Woudn't
```suggestion
        case ConsVar(s) => res.addAll(computed.getOrElse(s.uid, Nil))
```
also work?

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
LPTK: That trait would have an extra parameter type for the collection type (the value type of the map) and an abstract method to compute that value from an SCC.

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
ychenfo: The `toMap` was to convert `MutMap` to `Map`, but indeed I just realize that using `collection.Map` here can just work.

hkmc2/shared/src/main/scala/utils/SccAnalysis.scala
```diff
@@ -0,0 +1,141 @@
...
+      super.handleScc(members, sccId)
+  
+  trait DefaultCollecting[A] extends SccAnalysis[A]:
+    val collected = ListBuffer.empty[Ls[A]]
+    
+    abstract override protected def handleScc(members: Ls[A], sccId: Int): Unit =
+      collected.addOne(members)
+      super.handleScc(members, sccId)
+  
+  abstract class SccFromSuccFun[A](succs: A => IterableOnce[A]) extends SccAnalysis[A]:
```
LPTK: I don't really like this: the use site is quite awkward, with a lambda in the extends clause, and also it needlessly requires allocating the lambda. What's the point? I don't even think it's shorter.

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
```diff
@@ -156,12 +156,12 @@ class UsedVarAnalyzer(b: Block, scopeData: ScopeData)(using State):
...
+  private def combineInfos(m1: collection.Map[ScopedInfo, AccessInfo], m2: collection.Map[ScopedInfo, AccessInfo]): collection.Map[ScopedInfo, AccessInfo] =
     if m2.size < m1.size then combineInfos(m2, m1)
-    else m1.foldLeft(m2):
-      case (acc, info -> accesses) => m2.get(info) match
-        case Some(value) => acc + (info -> (accesses ++ value))
-        case None => acc + (info -> accesses)
+    else
+      m2 ++
+      m1.map: (k, v) =>
+        k -> m2.get(k).map(_ ++ v).getOrElse(v)
```
LPTK: `map` and `getOrElse` is just `fold`.

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
LPTK: I don't like your solution, which mixes several kinds of maps (could be detrimental to JIT compilation). You could instead just use `dp.iterator.map ... .toMap` to avoid the intermediate map.

hkmc2/shared/src/main/scala/utils/SccAnalysis.scala
```diff
@@ -91,6 +91,22 @@ object SccAnalysis:
   trait DefaultNoopHandling[A] extends SccAnalysis[A]:
     override protected def handleScc(members: List[A], sccId: Int): Unit = ()
   
+  trait DefaultCachingWithComputedValue[A, ComputedValuePerNodeType, ComputedValuePerSccType] extends SccAnalysis[A]:
```
LPTK: That name is incredibly awkward. Let's make it `CachingComputedSccValue`.

Also, let's define a `CachingComputedNodeValue[V] extends CachingComputedSccValue[V, V]` that sends the node to its SCC's mapped value (so this does not have to be done at the current use site).

LPTK: The `Default` prefix is pretty meaningless and can be dropped from the other trait names, too.

hkmc2/shared/src/main/scala/utils/SccAnalysis.scala
ychenfo: Indeed, the use site of this can just override the `successors` method easily. Will remove this.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -1033,11 +1032,13 @@ class FlowConstraintSolver(val collector: FlowConstraintsCollector):
...
   object AllUpperBounds extends
-    SccFromSuccFun[StratVarId](
-      upperBounds(_).iterator.collect:
-        case ConsVar(s) => s.uid)
-    with DefaultNoopHandling[StratVarId]
-    with DefaultCachingWithComputedValue[StratVarId, collection.Set[ConsStrat], collection.Set[ConsStrat]]:
+    SccAnalysis.NoopHandling[StratVarId]
+    with SccAnalysis.CachingComputedNodeValue[StratVarId, collection.Set[ConsStrat]]:
+      
+      override protected def successors(node: StratVarId): IterableOnce[Uid[StratVar]] =
```
LPTK: Do not use `override` just to mean "implements", as it makes it locally unclear whether you are overiding and thus throwing away an existing implementation. This applies to other places of the PR as well.

LPTK:
@ychenfo What do you think about Copilot's (suppressed) comment?

ychenfo:
> LinkedHashMap.last walks the insertion-ordered map, so repeatedly calling it while removing one SCC member makes closeScc quadratic in the SCC size.

Hmm I don't think copilot is correct about this "LinkedHashMap.last walks the insertion-ordered map": `LinkedHashMap` maintains a lastEntry `protected var lastEntry: Entry | Null = null` and overrides the `last` method so `last` is O(1)
```
override def last: (K, V) =
    if (size > 0) (lastEntry.nn.key, lastEntry.nn.value)
    else throw new NoSuchElementException("Cannot call .last on empty LinkedHashMap")
```

But it's true that I couldn't find that the API doc provides a complexity guarantee for `last`... Maybe I should just move away from LinkedHashMap

LPTK:
> But it's true that I couldn't find that the API doc provides a complexity guarantee for `last`... Maybe I should just move away from LinkedHashMap

Does not seem necessary for now, as you just demonstrated that it seems to work as intended.

PR #552, opened 2026-08-25
FlowAnalysis: Remove `cache: HashSet[Constraint]` and Use Object Identity

ychenfo:
This PR removes `cache: HashSet[Constraint]` for constraint solving and checks the cache only when handling constraints related to type variables. Flow analysis now also uses object identities instead of hashing on `case class`es and use hashmaps to store upper/lowerBounds.

---

~~This PR still updates both the upper and lower bounds of a type variable on `var < var` case and hence a draft, will update and un-draft this when #546 is done.~~


LPTK:
@ychenfo The branch now has conflicts.

Pro-tip: you can base a PR on another banch, there's no need to necessarily base it on the main branch.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -151,37 +151,49 @@ object FunRef:
       if (tSym.k is syntax.Fun) && tSym.owner.forall(_.asMod.isDefined) => S(tSym -> qual)
     case _ => N
 
-type StratVarId = Uid[StratVar]
+type StratVarId = Uid[StratVarState]
+
+sealed abstract class ProdStrat
+sealed abstract class ConsStrat
 
 class StratVarState(val uid: StratVarId, val name: Str, val generatedForFun: Opt[TermSymbol]):
```
LPTK: Couldn't we just make this class extend both `ProdStrat` and `ConsStrat`? (One would have to be made a trait.)

ychenfo: Indeed, and this also removes the ProdVar/ConsVar wrapper classes. In the change I pushed I made both of them traits because I don't see a principled way of picking one. Please let me know if it should be in another way, thanks!

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -1023,33 +1023,43 @@ class FlowConstraintSolver(val collector: FlowConstraintsCollector):
   given Raise = preAnalyzer.raise
   
   
-  val ctorDests = LinkedHashMap.empty[ConcreteProducer, Set[ConcreteConsumer | MarkerConsStrat]].withDefaultValue(Set.empty)
-  val dtorSrcs = LinkedHashMap.empty[ConcreteConsumer, Set[ConcreteProducer | MarkerProdStrat]].withDefaultValue(Set.empty)
-  val funDests = LinkedHashMap.empty[ProdFun, Set[ConsFun | MarkerConsStrat]].withDefaultValue(Set.empty)
-  val funSrcs = LinkedHashMap.empty[ConsFun, Set[ProdFun | MarkerProdStrat]].withDefaultValue(Set.empty)
+  val ctorsWithDests = mutable.Buffer.empty[Ctor]
```
LPTK, line 1026: Yes, I like this much better. We can directly store the relevant data in the ctors and use simple buffers to record which ctors are affected.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -153,35 +153,40 @@ object FunRef:
 
 type StratVarId = Uid[StratVar]
```
LPTK, line 154: It looks like this type is now basically only used for logging/debugging (which is fine). In the logging code, there are comparisons like `pAcc.s.uid === stratVar.uid` – couldn't these be changed to just `pAcc.s is stratVar`?

LPTK (approved):
LGTM, thanks!

PR #553, opened 2026-08-27
Various Bug Fixes and Tailrec Cleanup

CAG2Mark:
- Fixed two possible dormant bugs in the handler runtime where the continuation chain was not unlinked/relinked properly. (I only discovered these bugs when re-using the handler runtime for the shadow stack handler experimentation.) CC @AnsonYeung 
- Made the tailrec pass idempotent by rewriting known calls to functions within each SCC into calls to the merged loop definition and added the corresponding tests.
- Cleaned up and refactored some logic in the tailrec pass to make it less error-prone.


CAG2Mark:
Typo in the latest commit, it should be "make worker private". It seems making the worker private actually enables some optimizations.

hkmc2/shared/src/test/mlscript/tailrec/MultiArgLists.mls
```diff
@@ -169,15 +170,16 @@ fun handleEffect(x) =
...
+//│ define handleEffects_handleEffect as fun handleEffects_handleEffect⁰(id) {
 //│   loop loopLabel:
 //│     match id
 //│       0 =>
 //│         set id = 1;
 //│         continue loopLabel
 //│       1 =>
 //│         let tmp;
-//│         set tmp = resume⁰(param0);
+//│         set tmp = resume⁰(undefined);
```
LPTK, line 182: Interesting that it was able to properly figure out the `x` param and `param0` are useless.

Cc @ychenfo

hkmc2/shared/src/test/mlscript/tailrec/MultiArgLists.mls
```diff
@@ -270,30 +271,32 @@ fun f(a)(...z) = if z is
...
-//│         set inlinedVal = z.slice﹖(1, tmp1);
-//│         set tmp = +⁰(a, 1);
-//│         set a = tmp;
-//│         set z = inlinedVal;
-//│         continue loopLabel
-//│       else
-//│         throw new globalThis⁰.Error⁰("match error")
-//│     end
-//│   unreachable /* Rest of abortive labelled block */
+//│ define f² as fun f³(a)(...z) {
```
LPTK: Looks like this function is no longer given an inline wrapper. Any idea why?

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -682,6 +682,7 @@ final case class FunDefn(
   lazy val affineInfo: Ls[Int] =
     annotations.collect:
       case Annot.Affine(whichParamList) => whichParamList
+  lazy val paramSyms: Ls[VarSymbol] = params.flatMap(_.paramSyms)
```
AnsonYeung:
```suggestion
  lazy val allParamSyms: Ls[VarSymbol] = params.flatMap(_.paramSyms)
```

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
```diff
@@ -447,43 +499,13 @@ class TailRecOpt(checkAnnotations: Bool)(using State, TL, Raise):
                   case CallArgsResult.ForceSpread =>
                     // Forcibly spread the args in an array.
                     // Assume the lengths are correct
-                    val paramList = ogParamList.params
-                    val restParam = ogParamList.restParam
-                    
-                    val tupleSym = TempSymbol(N, "argList")
```
AnsonYeung: outdated comment because of moved code

hkmc2/shared/src/test/mlscript/tailrec/MultiArgLists.mls
CAG2Mark: Ah, I think I forgot to add the inline annotation to the wrapper.

CAG2Mark:
It seems the wrappers for the mutually tail-recursive function definitions were never given the inline annotation, so I've added that now

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
```diff
@@ -249,7 +249,55 @@ class TailRecOpt(checkAnnotations: Bool)(using State, TL, Raise):
...
+  /**
+    * Applies a parameter list to an argument list, both of which could contain arbitrary spreads.
+    *
+    * @param plist The parameter list to apply.
+    * @param args The argument list.
+    * @return A tuple `(Block => Block, List[Path])`, where:
+    * - The first parameter contains the code that breaks up the arguments into the shape expected by the parameter list, and
+    *   should be applied to the remaining code;
+    * - The second parameter is a list of paths containing the arguments in order. If there is a spread parameter, it will be
+    *   the last one.
```
AnsonYeung: Do you mean return value instead of parameter

hkmc2/shared/src/test/mlscript-compile/Runtime.mls
```diff
@@ -412,6 +412,10 @@ fun handleEffect(cur) =
     cur.contTrace.lastHandler,
     false
   )
+  
+  if cur.contTrace.last === handlerFrame do set saved.last = saved
+  if cur.contTrace.lastHandler === handlerFrame do set saved.lastHandler = saved
```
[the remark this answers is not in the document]
LPTK: It's true that it would be better to have a regression test for that.

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
```diff
@@ -498,6 +532,40 @@ class TailRecOpt(checkAnnotations: Bool)(using State, TL, Raise):
         // Not a tail call
         case _ => super.applyBlock(b)
       
+      // We explicitly rewrite *all* fully applied calls to functions within the SCC. Tail calls are rewritten using a
+      // `continue`, while other calls are rewritten as a call to the merged function. This is to make the tailrec
+      // optimizer pass idempotent. Without this, the subsequent tailrec optimizer passes could see the merged and wrapper
+      // functions as an SCC and try to rewrite the wrapper's tail call.
+      override def applyResult(r: Result)(k: Result => Block): Block = r match
+        case c @ Call(Value.MemberRef(bms, calleeSym: TermSymbol), args) if hasWrapper => funsMap.get(calleeSym) match
```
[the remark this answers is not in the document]
LPTK: Might want to clarify in a comment that we don't need to do that because the current tailrec optimizer does not look at module methods...

hkmc2/shared/src/test/mlscript-compile/Runtime.mls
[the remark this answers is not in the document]
CAG2Mark: This might not be possible in the current implementation, but I am not sure

hkmc2/shared/src/main/scala/hkmc2/codegen/TailRecOpt.scala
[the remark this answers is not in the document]
CAG2Mark: Actually I think this comment is correct, and we do handle module methods, just not paramless methods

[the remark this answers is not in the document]
CAG2Mark: (Addressed in the latest commit, FYI)

LPTK (approved):
LGTM, thanks!

Issue #559, opened 2026-09-15
Take parameter types from declaration


Derppening:
```
:sir
:siret
fun l: (rsc C) -> C
fun l(x) = x
//│ ———————————————| Lowered IR |———————————————————————————————————————————————————————————————————————
//│ define l⁰ as fun l¹(x: ?): C² { return (x as! C²) }; end
```

Note that the param `x` is typed as `?`, when its declaration(?) is typed as `rsc C`.

LPTK:
Wait – the result type is picked up, but not the parameter types? Weird.

PR #560, opened 2026-09-16
Add `rsc` / `rsc?` resource type modifiers

Derppening:
*This description is generated by Fable 5.1 and edited by me.*

This PR adds the surface syntax for marking a type as a resource and carries that information through elaboration into the erased types of the Block IR.

## Syntax

- `rsc T` marks `T` as a resource; `rsc? T` leaves it to a runtime test.
- `rsc` binds tighter than `|`/`&` (`rsc A | B` is `(rsc A) | B`) but looser than `->`'s LHS (`rsc (A, B) -> C` is a resource *function*, not a function taking a resource). A single resource parameter is written `(rsc A) -> B`, but is rejected for now (see below).
- `new [mut] [rsc] C(...)` creates a resource instance; `mut` and `rsc` may come in either order. `rsc` covers the argument lists (unlike `mut`, which binds tighter than application), so `new rsc C(1)(2)` works. `new rsc?` is an error, since the resource-ness must be known during instantiation. The modifier is split off during desugaring into a `rsc` field of `ProperNew`/`DynamicNew`. Written outside the instantiation (`rsc (new C())`), it is not the instance's modifier and has no effect (a warning), whether or not it would be accepted inside.
- `rsc (x => ...)` and `rsc f(x)` (a partial application) create a resource function value. A function value without a modifier is `rsc?`, so `rsc?` on one has no effect (a warning). On a call that is not known to leave some of its callee's parameter lists unapplied, the modifier has no effect either (a warning).

The modifiers elaborate to `Term.Annotated(Annot.Modifier(kw), t)`; the `Annot.Resource` extractor yields the resource-ness they denote. The two keywords are matched as `Keyword.RscLike`. See `doc/reference.md` § *Resource Types*.

## Erased types

Resource-ness is an `Opt[Bool]` on the new sealed mixin `HasRsc` (`AnyRef`, `ValueLike`, `Unknown`):

| value | meaning |
|---|---|
| `S(true)` | a resource |
| `S(false)` | not a resource (the unannotated default) |
| `N` | `rsc?`: not statically known, needs a runtime test |

`Unknown` becomes a `case class Unknown(rsc)`. `Primitive` and `Incompatible` carry no resource-ness.

- `eraseSign` threads the modifier down to whatever the signature denotes, so it survives unions and `forall`. Members of a union take their own modifier if they have one.
- `lub` keeps agreeing resource-ness and folds disagreement to `N`.
- `needsCast` now checks two dimensions: the class hierarchy as before, then resource-ness iff both sides are `HasRsc`. Equal ⇒ no cast; widening into `N` is free; narrowing out of `N` is a checked cast; `S(true)` vs `S(false)` is an error, reported as an unrelated type (`Cannot use a value of type 'rsc C' at an unrelated type 'C'`).
- Type aliases may write modifiers on their members (`type X = (rsc A) | B`). `resolveTpeSymAlias` now yields `AliasMember(sym, ownRsc)` per member. A member that cannot be resolved (recursion, type parameter, no definition) makes the alias erase to `Unknown` with the join of the members' resource-ness. A member that is a function type, also under a `forall`, resolves to `Function`, so an alias of one erases like the inline type.
- A modifier on a reference to an alias combines with the members' own (`CanonicalErasedValueType.combineRsc`): `rsc` overrides them (`rsc List` where `type List = Cons | Nil`), while `rsc?` and no modifier keep them (`rsc? U` where `type U = rsc A | B` is `rsc A | rsc? B`, erasing to `rsc? lub(A, B)`). A union written in place combines the same way with its members' own modifiers.
- A type-parameter reference with a modifier (`x: rsc A`) erases to `Unknown` under that modifier; an unannotated one erases as before.
- `Instantiate` gains a real `rsc: Bool` field and is typed as `ValueLike(S(rsc), cls)`.
- `Lambda` gains a `rsc: Bool` field and is typed as `Function(S(true))` if set, `Function(N)` otherwise. A resource lambda is not lifted into a function definition during lowering, so the slots it flows into see its resource-ness. `LambdaRewriter` marks the definition it lifts one into with `@rsc` (`FunDefn.rsc`), and FCFT instantiates the wrapper as a resource accordingly.
- `Call` gains a `rsc: Bool` field in its second parameter list, next to `metadata`, so patterns on `Call` are unchanged. `Lowering` sets it only on a call known to be under-applied, which is then typed as `Function(S(true))`; an unmodified one stays `Function(N)`. A call that is exactly or over-applied is never `rsc` (asserted where its type is computed). Passes that rebuild a call keep its flag, except where arguments are appended to it (`EtaExpansion`), since the result is a different value.

**Function signatures carry no resource-ness; function values do.** A signature describes how a definition is called and never types a value, so `ErasedFuncSignature` (formerly `ErasedFuncType`) and its `Signature`/`CanonicalSignature` (formerly `FuncRef`/`CanonicalFuncRef`) have no `rsc` field. A function value is an instance of `Function` and has a resource-ness like any other instance: `rsc` for a resource lambda or partial application, `rsc?` otherwise, including a reference to a definition. A signature is parsed once into its chain of function types (`Elaborator.Sig`). A modifier on a function type that a definition consumes as its own parameter list (head, curried inner, or paramless `declare`) is rejected; one on a function type that types a *value* (a getter's result, a `val`, a parameter) is kept.

**Signatures and value types are kept apart.** `ErasedFuncSignature` (`Signature`, `CanonicalSignature`) is no longer a subtype of the erased value types, whose root is now `ErasedValueType`; the memoized `canonicalize` both share lives in the `Canonicalizable` mixin. A `TermSymbol` stores either a value type or a signature, and a reference to a symbol with a signature is `rsc? Function`. `HasErasedType` and `HasLateInitErasedType` are removed: each symbol kind declares its own `erasedType`, matches on erased types are exhaustive over concrete symbol types, and a late-initialized erasure is set once through a setter that soft-asserts it was unset.

Since any class can be instantiated with `new rsc`, a class's `this` is `rsc?`.

## Validation

Checked in `Resolver` (types) and `Elaborator` (definitions, parameters, type-parameter declarations, `new`). A modifier in a position that is not supported yet is still checked like any other, so it may report more than one error. To reach function-type parameters, `traverseSign` now traverses a tuple type's fields as types rather than as terms; no other difftest output changes.

- `Primitive type 'Int32' cannot be a resource.` (a warning under `rsc?`, which has no effect). Applies through aliases.
- `A type takes at most one resource modifier.` for a modifier written directly within another's reach (e.g. `rsc rsc? T`). A modifier on an alias reference or on a union combines with the members' instead (see above).
- `Every member of this type is a resource, so 'rsc?' has no effect.` (a warning) when `rsc?` keeps member modifiers that already make every member a resource.
- `Resource modifiers on this type are not supported yet.` for a type erasure would drop the modifier from: one with no erased counterpart (a negation, a tuple or a literal type), reached directly or through a union, and `()`, which is never a resource. The check asks erasure itself whether the modifier survives, so it does not duplicate the list of shapes.
- `Resource modifiers inside an intersection type are not supported yet.` An intersection erases to `Unknown`, so the modifier would be silently lost.
- `Resource modifiers inside the parameters of a function type are not supported yet.` A function type erases to `Function` without its parameter types, so `(rsc C) -> D` would silently lose the modifier. This applies wherever the function type is written, including a definition's separately written or `declare`d signature (`fun f: (rsc C) -> D`); once a signature's parameter types reach the definition (#559 for a separately written one), the rejection can be lifted there. A modifier on the function type itself (`rsc (C -> D)`) is supported.
- `Resource modifiers apply to types, not to <kind> definitions.` The definition is kept and compiles without the modifier.
- `Resource modifiers apply to types, not to parameters.` The parameter is kept: `ucs.Normalization` indexes class params positionally under a hard `assert`, so dropping it would crash on a constructor pattern. In a contextual parameter list a bare tree is a type, so `using rsc C` is a resource *type*; only named or spread parameters take the modifier as the parameter's.
- `Resource modifiers apply to function values, not to a function definition's parameter lists.`
- `An instance cannot be 'rsc?': it either is a resource or is not.`
- `'rsc' modifiers on type parameters are not supported yet.` (or `'rsc?'`) on a declaration (`class C[rsc A]`, `class C[out rsc A]`, `fun f[rsc A]`, `[rsc A] -> T`). A rejected declaration still declares the parameter, so its uses resolve. A use (`x: rsc A`) is allowed; a non-resource argument such as `rscIdentity[Int32](1)` is then rejected by the coercion into the `rsc Unknown` slot.

- `This annotation has no effect.` (a warning), as for any other annotation, on a modifier written on an expression that is neither a lambda nor a call: a `new` carries its own (see above), and anything else, e.g. `rsc 1`, takes the modifier as an annotation that does nothing.

All of the above diagnostics are demonstrated via at least one difftest.

Not supported yet, rejected with an error: `new! rsc`, `new rsc` with a refinement, resource instantiation in a staged module, and a resource `Instantiate` or `Call` in the Wasm backend. The JS backend does not handle `Instantiate.rsc` yet (TODO): resources will carry a reference count there too.

## Existing Difftest Changes

- `'Unknown'` becomes `'rsc? Unknown'` in existing error messages, since nothing is known about an unannotated slot. The printer emits `rsc` and `rsc?` to match the surface language spelling.
- `fun test: Function = id` now takes a checked cast, and so do partial applications flowing into a `Function` slot, because function references are `rsc? Function` (and `rsc? Function` to `Function` takes a downcast).
- Passing `this` into a plain class (`C`) slot now inserts a checked cast.
- `Instantiate` dumps show the new `rsc = false` field.
- *(WIP commit, in or out of this PR at the reviewer's call)* A separately written signature is elaborated and resolved once for the declaration and the definition using it, so its diagnostics are no longer reported twice.

## Tests

- `parser/Rsc.mls`: precedence and token fusion.
- `codegen/ResourceTypes.mls`: erasure, joins, casts, `new rsc`, `this`, function types, parameter-list consumption, resource lambdas and partial applications, each shape of a modifier written outside a `new`, and the remaining positions a modifier can be written in: `as`, where it has no effect because no cast is lowered, and `is` patterns and `handle`, which reject it.
- `codegen/ResourceTypeAliases.mls`: modifiers inside aliases, unresolvable members, function-type members, alias-side checks.
- `codegen/ResourceModifierChecks.mls`: every rejection above, including the contextual-parameter-list cases, modifiers inside type arguments (which are accepted, also on an alias's right-hand side that is not a type), function-type parameters, type parameters with and without variance, and the types erasure would drop a modifier from.

No type alias that erases without a diagnostic is left without a use site, so every golden shows either an error or what the alias erases to.

## Known gaps

- Resource-holding classes do not taint their classes (`class Box(x: rsc C)` doesn't imply `Box` must be `rsc`).
- The FCFT eta wrapper is instantiated as a resource only for a lifted resource lambda; for other function values it waits on their resource-ness being resolved (TODO).
- A reference to a function definition lifted from a resource lambda is still `rsc? Function`, as for any symbol with a signature.
- A partial application's `rsc` flag is set from the callee's definition, but its type is computed from the callee's erased signature; without a signature, the flag has no effect and no warning is reported.
- A `let` binding does not keep the erased type of its value, so `let p = rsc f(x)` binds an untyped `p`.
- A separately written signature's parameter types do not yet reach the definition's own parameters (#559).
- The JS backend compiles a resource `Instantiate` or `Lambda` like a non-resource equivalent, with no diagnostic; only the `Instantiate` TODO records this.
- Narrowing out of `rsc?` asks for a checked cast, but neither backend tests resource-ness at runtime: the Wasm `Cast` arm tests only the class, and JS erases casts. This matches `Cast`'s scaladoc, which already says a checked cast generates the same code as an unchecked one.


LPTK:
> **Function types carry no resource-ness.** Whether a closure is a resource depends on what it captures, which is not implemented as part of this PR.

I think that's a misunderstanding of what erased function types represent. They don't represent values... You should have caught this.

> So `FuncRef`/`CanonicalFuncRef` lose their `rsc` field

Nothing should silently "lose" their `rsc` field. Any `rsc` that can't be given proper meaning should yield an error.

LPTK:
> * Type arguments are erased, so `Box[rsc C]` loses the modifier; a primitive type argument also escapes the primitive check.

That's pretty serious. Why would you/Claude think this is acceptable?

LPTK:
> * The FCFT eta wrapper is instantiated as a non-resource with a TODO; it should be a resource iff the wrapped function is, once that is resolved.

What is FCFT?!

LPTK:
> Since any class can be instantiated with `new rsc`, a class's `this` is `rsc?`.

Oof... we didn't think of that.

I'm afraid this is completely unacceptable. We probably can't just add runtime checks on every single use of `this`, can we? Unless maybe we can optimize them adequately...

EDIT: On second thought, I don't know if using `this` as a plain value is so common as to make this truly unacceptable. Provided we don't mess with uses of `this` that are only for calling methods and accessing fields (eg `this.f`), this might be ok.

Derppening:
> I think that's a misunderstanding of what erased function types represent. They don't represent values... You should have caught this.
> 

Doesn't erased function types represent the type of top-level functions, and non-free functions (i.e. lambdas) are represented as a simple (possibly `rsc`-modified) `Function`? If so, I think my edit is just poorly placed; 

> Nothing should silently "lose" their `rsc` field. Any `rsc` that can't be given proper meaning should yield an error.

That's poor wording - it means that `FuncRef` and `CanonicalFuncRef` no longer has a `rsc` modifier on them.

> That's pretty serious. Why would you/Claude think this is acceptable?

I recall us deciding that this (resource-ness on generic type parameters) wouldn't be part of the initial `rsc` implementation - I added "not implemented" errors to all usages that would otherwise silently drop the `rsc` modifier.

The PR description is also updated to reflect the current implementation.

> What is FCFT?!

`FirstClassFunctionTransformer`.

> Oof... we didn't think of that.
> 
> I'm afraid this is completely unacceptable. We probably can't just add runtime checks on every single use of `this`, can we? Unless maybe we can optimize them adequately...
> 
> EDIT: On second thought, I don't know if using `this` as a plain value is so common as to make this truly unacceptable. Provided we don't mess with uses of `this` that are only for calling methods and accessing fields (eg `this.f`), this might be ok.

I agree with your second thought. From what I remember when you first described resource-ness, `rsc C` changes how the function body is compiled (compared to non`-rsc`), so it makes sense for the implicit `this` of class functions to be `rsc? C` - we would just have a runtime check that chooses which logic (RC-based or GC-based) to apply to the current object.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -2201,16 +2205,22 @@ extends Importer:
...
-                  mk(id, S(false))
-                case Modified(Keywrd(Keyword.`out`), id: Ident) =>
-                  mk(id, S(true))
+                  mk(id, vce)
+                case Modified(Keywrd(Keyword.`in`), body) if vce.isEmpty =>
+                  go(body, S(false), rscd)
+                case Modified(Keywrd(Keyword.`out`), body) if vce.isEmpty =>
+                  go(body, S(true), rscd)
+                // * The parameter is still declared, so its uses resolve.
+                case Modified(kw @ Keywrd(Keyword.`rsc` | Keyword.`rsc?`), body) if !rscd =>
```
LPTK: We should add an alias `` Keyword.RscLike = Keyword.`rsc` | Keyword.`rsc?` `` to avoid all this repetition.

hkmc2/shared/src/test/mlscript/codegen/ResourceTypes.mls
```diff
@@ -25,12 +25,20 @@ fun h(x: C) = x
...
 
+:ge
 fun l: (rsc C) -> C
 fun l(x) = x
+//│ ╔══[COMPILATION ERROR] Resource modifiers inside the parameters of a function type are not supported yet.
+//│ ║  l.34: 	fun l: (rsc C) -> C
+//│ ╙──      	            ^
+//│ ╔══[COMPILATION ERROR] Resource modifiers inside the parameters of a function type are not supported yet.
+//│ ║  l.34: 	fun l: (rsc C) -> C
+//│ ╙──      	            ^
```
LPTK: The error is duplicated.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
```diff
@@ -101,6 +101,16 @@ object Annot:
       case Modifier(Keyword.`rsc`) => S(S(true))
       case Modifier(Keyword.`rsc?`) => S(N)
       case _ => N
+    
+    /** The keyword of a resource modifier denoting `rsc`, or `N` for a non-resource. */
+    def keyword(rsc: Opt[Bool]): Opt[Keyword] = rsc match
```
LPTK: This is only used to get back a keyword from an `Opt[Bool]`. But there's no reason to carry an `Opt[Bool]` everywhere; we could just use `Opt[RscLike]`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -1260,24 +1276,51 @@ class Resolver(tl: TraceLogger)
...
    */
   private def checkAllRscModifiers(t: Term): Unit =
     t match
       case Term.Annotated(Annot.Resource(rsc), target) => checkRscModifier(rsc, target)
       case Term.CompType(lhs, rhs, false) => checkRscInIntersection(lhs, rhs)
+      case Term.FunTy(lhs, _, _) => checkRscInFunParams(lhs)
       case _ => ()
-    t.subTerms.foreach(checkAllRscModifiers)
+    t match
+      case Term.TyApp(con: Resolvable, _) => checkAllRscModifiers(con)
```
LPTK: Don't indent this or the ones in the match before.

LPTK: Same goes for all other matches in this PR.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -556,6 +556,7 @@ class Resolver(tl: TraceLogger)
       
       case Term.TyApp(lhs: Resolvable, targs) =>
         resolve(lhs, prefer = prefer, inAppPrefix = inAppPrefix, inCtxPrefix = inCtxPrefix, inTyPrefix = true)
+        checkRscInTyArgs(targs)
```
LPTK: Why is this needed? Does it even make sense?

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
```diff
@@ -1085,16 +1085,23 @@ class Resolver(tl: TraceLogger)
       // respectively.
       case Term.TyApp(con, targs) => 
         traverseSign(con, expect = expect, inAppPrefix = true)
+        checkRscInTyArgs(targs)
```
LPTK: Ditto.

LPTK:
> Doesn't erased function types represent the type of top-level functions, and non-free functions (i.e. lambdas) are represented as a simple (possibly `rsc`-modified) `Function`? If so, I think my edit is just poorly placed;

What does this have to do with "top-level functions" and "non-free functions"? AFAIK, an erased function "type" is really a _signature_ (it should be renamed accordingly to avoid further future confusion), and it never types a value (it's not an `ErasedValueType`). So, the **conjunction** of the two assertions "Function types carry no resource-ness." and "Whether a closure is a resource depends on what it captures" as though the latter had any connection to the former doesn't make sense.

LPTK:
> I recall us deciding that this (resource-ness on generic type parameters) wouldn't be part of the initial `rsc` implementation - I added "not implemented" errors to all usages that would otherwise silently drop the `rsc` modifier.

Yes, resourceness of "generic" type **parameters** is not supported. Why would it not be supported on **references** to said parameters? What's the reason?

LPTK:
> `FirstClassFunctionTransformer`.

> * The FCFT eta wrapper is instantiated as a non-resource with a TODO; it should be a resource iff the wrapped function is, once that is resolved.

You should see that this is a symptom of a bigger problem: lambdas, which are not very different from normal object instances, should clearly have a `rsc` field, at least in the IR. Otherwise, how do you think we will reference-count them? The flag can probably be inferred later, but for now we'll need to just write resource lambdas as `rsc (x => ....)`.

LPTK:
> we would just have a runtime check that chooses which logic (RC-based or GC-based) to apply to the current object

I think you're suggesting to double-compile every single method body (once for the RC branch, once for the GC branch), which is obviosuly unreasonable. The checks will have to be where `this` is used as a value, which is why I wrote what I wrote above.

hkmc2/jvm/src/test/scala/hkmc2/codegen/ErasedTypeRscTests.scala
LPTK, line 1: This entire file is unnecessary slop.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
```diff
@@ -383,7 +383,8 @@ class JSBuilder(using Config, TL, State, Ctx) extends CodeBuilder:
       if ai
       then doc"${resultQual(qual)}.at(${result(fld)})"
       else doc"${result(qual)}[${result(fld)}]"
-    case Instantiate(mut, cls, argss) =>
+    // * JS has no separate layout for resources.
```
LPTK: This comment is wrong. Even in JS, resources will have a refcount.
```suggestion
    // * TODO: handle `rsc`
```

hkmc2/shared/src/main/scala/hkmc2/codegen/ErasedType.scala
```diff
@@ -47,26 +47,34 @@ object ErasedType:
...
+      // * A canonical type with resource-ness takes it from the modifiers written inside the alias referred to, if any,
+      // * and from this reference otherwise.
+      // * The resource-ness of the alias is considered first because a reference states resource-ness only when
+      // * annotated (otherwise it has `rsc = S(false)`), and `Resolver` rejects annotating one whose alias writes its
+      // * own.
+      canon match
+        case h: HasRsc if h.rsc =/= rsc =>
+          assert(CanonicalErasedValueType.resolveTpeSymAlias(sym).exists(_.ownRsc.isDefined),
+            s"the resource-ness of '$canon' must come from this reference ($rsc) or from a modifier inside '$sym'")
+        case _ => ()
```
LPTK: This reads like sloppy nonsense. @Derppening please describe in your own words what this does and why it's justified.

hkmc2/shared/src/main/scala/hkmc2/codegen/ErasedType.scala
```diff
@@ -384,13 +430,25 @@ sealed abstract class ErasedType:
...
+            case _ => ""
+          if tpeSym.asMod.isDefined then s"${rscPrefix}module $name" else s"$rscPrefix$name"
+
+  /** The type of the value this type denotes, i.e. a first-class `Function` for a function type.
+    *
+    * A function type has no resource-ness of its own: whether a function value is a resource depends on what it
+    * captures, which is left to be resolved per reference. Until then, it is `rsc?`.
+    */
+  final def valueType: ErasedValueType = this match
+    case _: ErasedFuncType => ErasedType.Function(N)
```
LPTK: When does this case trigger? Why does it make sense. @Derppening explain in your own words.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -1074,6 +1076,11 @@ extends Importer:
     
     def error = Term.Error().withLocOf(tree)
     
+    /** Splits the body of a `new` into its resource modifier (if any), its class, and its argument lists. */
+    def splitRscNew(c: Tree, args: Ls[Tup]): (Opt[Keywrd[?]], Tree, Ls[Tup]) = c match
```
LPTK: Wouldn't it be better to add a `rsc` field to `ProperNew` and rely on desugaring instead of complicating the elaborator logic?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -2018,8 +2047,16 @@ extends Importer:
                 case _ =>
                   Modulefulness.none
               
-              /** Splits a signature's arrow chain into the parameter lists it describes and the type it returns.
-                * Yields `N` if the signature is not an arrow or if some parameter list's arity cannot be read.
+              /** Whether a signature is a possibly-quantified function type. */
+              def wrapsFunTy(sign: Term): Bool = sign match
```
LPTK: This Boolean-blind check fucntion is fishy as hell. Just properly parse the signature into its components and use the resulting data properly, instead of checking things haphazardly. https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

LPTK (changes requested):

Derppening:
> What does this have to do with "top-level functions" and "non-free functions"? AFAIK, an erased function "type" is really a _signature_ (it should be renamed accordingly to avoid further future confusion), and it never types a value (it's not an `ErasedValueType`). So, the **conjunction** of the two assertions "Function types carry no resource-ness." and "Whether a closure is a resource depends on what it captures" as though the latter had any connection to the former doesn't make sense.

Just to clarify, under the new terminology: function signatures do not have resource-ness, and function values do?

> > * The FCFT eta wrapper is instantiated as a non-resource with a TODO; it should be a resource iff the wrapped function is, once that is resolved.
> 
> You should see that this is a symptom of a bigger problem: lambdas, which are not very different from normal object instances, should clearly have a `rsc` field, at least in the IR. Otherwise, how do you think we will reference-count them? The flag can probably be inferred later, but for now we'll need to just write resource lambdas as `rsc (x => ....)`.

Right now they are erased down to (possibly `rsc`-qualified) `Function`, and I am not sure if this is too coarse to be useful. Given these two points, I wonder if we should have an *actual* `FuncRef` type that represents lambdas (and functions used in value position in-general), which stores its resource-ness (and should be inferred by its captures as you described), and the function signature of its application.

> Yes, resourceness of "generic" type **parameters** is not supported. Why would it not be supported on **references** to said parameters? What's the reason?

Ah. I wasn't thinking clearly when I made the edits last night. Will fix this to only reject generic type parameters.

Derppening:
> I think you're suggesting to double-compile every single method body (once for the RC branch, once for the GC branch), which is obviosuly unreasonable. The checks will have to be where `this` is used as a value, which is why I wrote what I wrote above.

Yes, I agree that the checks should only be performed where `this` is used as a value. 

The "choosing which logic" thing I was referring to is my poor attempt in trying to say that methods are effectively compiled as `fun f = if this.isRsc then ... else ...` (just that these checks will be inserted where needed rather than enclosing the entire function body) - I am *not* trying to suggest we take the double-compilation approach.

LPTK:
> Just to clarify, under the new terminology: function signatures do not have resource-ness, and function values do?

Of course. Function values are just instances of `Function`.

> Right now they are erased down to (possibly `rsc`-qualified) `Function`, and I am not sure if this is too coarse to be useful. Given these two points, I wonder if we should have an _actual_ `FuncRef` type that represents lambdas (and functions used in value position in-general), which stores its resource-ness (and should be inferred by its captures as you described), and the function signature of its application.

I think the current status quo is fine. We do need to add a `rsc` field to lambda terms and correspondingly lambda `Result`s.

Now, the tricky bit is how to deal with partially-applied functions/methods, whose resourceness depends on what they capture (ie, the receiver object and the arguments given so far). The only reasonable solution I see, if we want to make things consistent with the current design, is to add a `rsc` flag to calls themselves... as they may be unsaturated, in which case they denote a function value that may need to be a resource.

So, `Call` nodes should probably get a new field (in a second parameter list, to avoid disruption) that stores its resourceness (yes or no Boolean – no uncertain value, similar to `Instantiate`). The invariant would be that fully starurated `Call`s are never marked as resources – their erased type comes from the result type of the called function, which will carry its own resourceness annotation.

> The "choosing which logic" thing I was referring to is my poor attempt in trying to say that methods are effectively compiled as `fun f = if this.isRsc then ... else ...`

But they are not.

> (just that these checks will be inserted where needed rather than enclosing the entire function body)

Yes.

> I am _not_ trying to suggest we take the double-compilation approach.

The pseudocode you showed **is** double compilation.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
LPTK: I now see that `Opt[Bool]` is used all over the place for resourceness. So, maybe this one is fine, but we probably shouldn't need to recover a keyword instance just to call the diagnostic function – that diagnostic function should instead just take the relevant info directly, and an overlad could conserve the original interface.

hkmc2/shared/src/main/scala/hkmc2/codegen/ErasedType.scala
Derppening: Consider this example:

```
type U = rsc A | B
fun f(x: rsc U) = 
  ...
```

`rsc U` expands to `rsc (rsc A | B)`, which is erased to `rsc (rsc? lub(A, B))`, and this has two possible readings:

- Make it `rsc lub(A, B)`, which discards that `B` should be a non-`rsc` as stated in `U`
- Make it `rsc? lub(A, B)`, which discards that `U` should be a `rsc` as stated in the type annotation of `x`

To me, neither interpretation makes sense, so I decided the best way is just to reject it outright when there are conflicting `rsc` modifiers on the same type (including the lack of `rsc` - which is explicitly a non`-rsc`); the same modifier on the type (`type V = rsc A | rsc B; x: rsc V`) is allowed.

 FYI the resolver already rejects such forms if it is written outright, i.e. `rsc rsc? T` is always an error. More specifically it is stricter than this heuristic - `rsc rsc T` is also an error (only one modifier is allowed).

The phrasing of the assertion should probably improved though.

LPTK: I disagree. It should of course be possible to use `rsc List` where `type List = Cons | Nil`. Asserting `rsc` clearly should override whatever was already in the type.

As for `rsc? U` where `U = rsc A | B`, it should be the same as `rsc A | rsc? B`.

Derppening: > As for `rsc? U` where `U = rsc A | B`, it should be the same as `rsc A | rsc? B`.

Just to confirm, this will ultimately erase to `rsc? lub(A, B)` correct? Since ``` lub(`rsc`, `rsc?`) = `rsc?` ```.

(Or are you suggesting to make unions first-class in `ErasedType`? Unions are currently erased to the `lub` of its members.)

hkmc2/shared/src/main/scala/hkmc2/codegen/ErasedType.scala
Derppening: This is triggered by `ErasedType.erasedValueType`, which returns the equivalent value type `Function` when this erased type is used on a value, for example when a function is passed as an argument or returned (e.g. `fun f = ...; id(f)`).

I suspect `ErasedType.Function` may be confusing in that regard, since it actually refers to `ErasedType.AnyRef(rsc = ..., sym = ctx.builtins.Function)` rather than `ErasedFuncType` (the signature type).

hkmc2/shared/src/main/scala/hkmc2/codegen/ErasedType.scala
LPTK: Right, it would result in `rsc?`.

LPTK:
I just noticed that functions like `def localType` pattern match on the loosely typed `HasErasedType`, which is not sealed, so the pattern match is not even checked for exhaustiveness. This is a real anti-pattern and very sloppy. Please fix it by making matches are exhaustive by construction and don't rely on vague open-ended helper traits.

hkmc2/shared/src/main/scala/hkmc2/codegen/ErasedType.scala
LPTK: > This is triggered by `ErasedType.erasedValueType`

I suppose you mean `HasErasedType#erasedValueType`. I really don't like this loosely typed `HasErasedType`. It's almost impossible to checker whether it's used correctly, as it's infectious, being used all over the place and not sealed.

I am uncomfortable with the logic mixing distinct cases that should never be confused. Lumping function signatures and value types in the same bag and then tryign to tease them apart like in `def valueType` is a recipe for disaster and should never be needed. It can only hide subtle logic bugs.

> I suspect `ErasedType.Function` may be confusing in that regard, since it actually refers to `ErasedType.AnyRef(rsc = ..., sym = ctx.builtins.Function)` rather than `ErasedFuncType` (the signature type).

It will be fine after the renaming of `ErasedFuncType` to `ErasedFuncSignature`.

LPTK: EDIT: `ErasedType.Function` doesn't actually seem to forward to `ErasedType.AnyRef` – why?

Derppening: > EDIT: `ErasedType.Function` doesn't actually seem to forward to `ErasedType.AnyRef` – why?

Because of `Prelude.mls`. From the docs of `ErasedType.ValueLike`:

```scala
    * - This transient type is needed to represent value types before the `Prelude` is fully elaborated. The IR should
    *   always operate on the canonicalized type.
```

It will be converted into `AnyRef(ctx.builtins.Function)` once the type is canonicalized.

LPTK: Ok. Now, on the other question, could you please confirm that the logic is indeed very messy and unnecessarily mixes different things?

Derppening: Yes - I agree that the current `ErasedFuncSignature` has diverged from `ErasedType` so much that they should no longer share a hierarchy. I will address this as well.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
Derppening: Addressed in 822c51d7b59ff30aeacae9155317e8564dbf126c.

hkmc2/shared/src/main/scala/hkmc2/semantics/Term.scala
Derppening: Obsolete in 2c7407ea6604eccb2db0ee39a631ab32f6a710be.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
Derppening: Addressed in 09f881f6bc8271dd59793c720449b93901c53c67.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
Derppening: This was part of me erroneously rejecting `rsc`-modified generic type *arguments*. Removed in 14de3611f9d5bdcab7aef98866dfddc5ec027787.

hkmc2/shared/src/main/scala/hkmc2/semantics/Resolver.scala
Derppening: Obsolete - see above.

hkmc2/jvm/src/test/scala/hkmc2/codegen/ErasedTypeRscTests.scala
Derppening: Removed in a882be4ea6833082471257c270ac9961278b3cc3.

hkmc2/shared/src/main/scala/hkmc2/codegen/js/JSBuilder.scala
Derppening: Fixed the comment in a882be4ea6833082471257c270ac9961278b3cc3.

hkmc2/shared/src/main/scala/hkmc2/codegen/ErasedType.scala
Derppening: Implemented in 798b8ef8356955147cc4dbc942764af20c6580f7. `rsc` now "forcibly" applies to the type it modifies.

hkmc2/shared/src/main/scala/hkmc2/codegen/ErasedType.scala
Derppening: The refactoring is done in 839607880254f84c6a08498767486a7fd1380f14. `ErasedFuncSignature` is fully separated from `ErasedValueType`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
Derppening: Implemented in 12ac9cd2a386d475ec83e6315acf773402ea493b.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
Derppening: Implemented in 1a4e5d1c0b3ebe43542eca08e1f50e81997864c9.

Derppening:
> The flag can probably be inferred later, but for now we'll need to just write resource lambdas as `rsc (x => ....)`.

Implemented `rsc`-modifier on lambdas in c4d8e51cc08a42c2869cb5fcb99539735c34d0d1

> So, `Call` nodes should probably get a new field (in a second parameter list, to avoid disruption) that stores its resourceness (yes or no Boolean – no uncertain value, similar to `Instantiate`). The invariant would be that fully starurated `Call`s are never marked as resources – their erased type comes from the result type of the called function, which will carry its own resourceness annotation.

`rsc`-modifier on partial applications are implemented in e93ad3108b365f17432256a76282a0e3b48162be.

hkmc2/shared/src/test/mlscript/codegen/ResourceTypes.mls
Derppening: Implemented in 1cc917b9d2b5d346b1aaead48511e32ecf5b3dcd by only elaborating the separate signature once (and fixing a `FIXME`), but with the caveat that the declaration **must** appear before the definition (which is how difftests are written right now anyways) - otherwise a variable name that shadows a type will crash the compiler (e.g. `class C(); fun f: C -> C; fun f(C) = ...`).

I am not sure if this fix is acceptable: it seems like an edge case to me and is an improvement over the existing behavior (where the compiler would just crash regardless). At the same time, I also think this could probably be refined and fully addressed with Claude/Codex.

Derppening:
> I just noticed that functions like `def localType` pattern match on the loosely typed `HasErasedType`, which is not sealed, so the pattern match is not even checked for exhaustiveness. This is a real anti-pattern and very sloppy. Please fix it by making matches are exhaustive by construction and don't rely on vague open-ended helper traits.

Done in 7097243a0f0a407fcf9d50d40bff91aac2f6006f.

The PR description is also updated.

doc/reference.md
```diff
@@ -287,6 +287,16 @@ _.f(0, _, 2)     // (x, y) => x.f(0, y, 2)
 {_ + 1}          // x => x + 1
 ```
 
+**Resource lambdas and partial applications** — a function value is `rsc?` unless marked as a resource:
```
LPTK, line 290: If the function value is not created with `rsc`, it should **not** be a resource (it should not be `rsc?` either).

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -1298,8 +1312,9 @@ object Cast:
       case Cast(inner, _, innerCheck) => new Cast(inner, target, check || innerCheck)
       case _ => new Cast(value, target, check)
 
-case class Lambda(params: ParamList, body: Block)(val annot: Ls[Annot]) extends Result:
+case class Lambda(rsc: Bool, params: ParamList, body: Block)(val annot: Ls[Annot]) extends Result:
```
LPTK, line 1315: We placed `rsc` in the second parameter list of `Call`, so for consistency we should do the same for `Lambda`, no?

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -444,35 +444,64 @@ class Lowering()(using Config, TL, Raise, State, Ctx, SymbolPrinter):
...
   
+  /** Whether the resource modifiers in `annots` make a function value (a lambda or a partial application) a resource.
+    * A function value is `rsc?` unless stated otherwise, so `rsc?` is reported as having no effect.
+    */
+  def functionValueRsc(annots: Ls[Annot], loc: Opt[Loc]): Bool =
+    annots.foldLeft(false):
+      // * The annotation does not record where its keyword is written, so the warning points at `loc`.
+      case (rsc, Annot.Resource(N)) =>
+        raise(WarningReport(
+          msg"A function value is 'rsc?' unless stated otherwise, so 'rsc?' has no effect." -> loc :: Nil,
```
LPTK, line 465: No. _Again_, as I said before, it should be treated the same as for `new`.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -444,35 +444,64 @@ class Lowering()(using Config, TL, Raise, State, Ctx, SymbolPrinter):
...
           source = Diagnostic.Source.Compilation))
       // * No arguments to a super ctor means a nullary call, e.g., `extends C` means `extends C()`
-      k(Call(fr, Nil ne_:: Nil)(CallMetadata(isMlsFun, true, Nil)).withLoc(loc))
+      k(Call(fr, Nil ne_:: Nil)(CallMetadata(isMlsFun, true, Nil), rsc = false).withLoc(loc))
   
+  /** Whether the resource modifiers in `annots` make a function value (a lambda or a partial application) a resource.
+    * A function value is `rsc?` unless stated otherwise, so `rsc?` is reported as having no effect.
+    */
+  def functionValueRsc(annots: Ls[Annot], loc: Opt[Loc]): Bool =
+    annots.foldLeft(false):
```
LPTK, line 461: `exists`

hkmc2/shared/src/main/scala/hkmc2/codegen/Lowering.scala
```diff
@@ -444,35 +444,64 @@ class Lowering()(using Config, TL, Raise, State, Ctx, SymbolPrinter):
...
   /** Lower a call with multiple argument lists into `Call` nodes,
     * trying to group as many as possible into a single one
     * when they correspond to parameter lists of the same callee. */
   def lowerMultiCall(fr: Path, isMlsFun: Bool, annotations: Ls[Annot], args: Ls[Term], loc: Opt[Loc])(k: Result => Block)(using LoweringCtx): Block =
+    // * `Elaborator` records the modifier of `rsc f(x)` as an annotation on the call, as for `rsc (x => ...)`.
+    val (rscAnnots, callAnnots) = annotations.partition:
+      case Annot.Resource(_) => true
+      case _ => false
+    // * Only a call known to leave some parameter lists unapplied is a function value that can be a resource.
+    def warnNotPartial(): Unit =
```
LPTK, line 480:
```suggestion
    def warnNotPartial: Unit =
```

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
```diff
@@ -1070,24 +1074,23 @@ sealed abstract class Result extends AutoLocated, HasErasedType:
...
+  /** The [[ErasedValueType]] of this result, or `N` if the erased type is not known. */
+  lazy val erasedType: Opt[ErasedValueType] = this match
     case Value.SimpleRef(sym) => sym match
-      case hasErasedType: HasErasedType => hasErasedType.erasedType
-      case _ => 
-        // * Some symbols may not have an erased type (e.g. `BuiltinSymbol`, where it may represent more than one
-        // * function).
-        N
+      case l: LocalVarSymbol => l.erasedType
+      // * A `BuiltinSymbol` has no erased type, as it may represent more than one function.
```
LPTK, line 1092: Bultin symbol *uses* most definitely have an erased type, so returning `N` here is simply incorrect. What this case indicates is that either
* `SimpleRef` should store the erased type at which the symbol is used.
* We should just use different builtin symbols for different builtin operation arities.

I now think that the latter makes a lot more sense and is the obviously right thing to do, but it will require modifying `Elaborator` a bit. It can be done in a follow-up PR (pls open the issue).

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -546,10 +616,9 @@ class ModuleOrObjectSymbol(val tree: Tree.TypeDef, val id: Tree.Ident)(using Sta
 
 class TypeAliasSymbol(val id: Tree.Ident)(using State)
     extends MemberSymbol
-    with DefinitionSymbol[TypeDef]
-    with HasErasedType:
+    with DefinitionSymbol[TypeDef]:
 
-  override val erasedType: Opt[ErasedType] = S(ErasedType.ValueLike(rsc = S(false), this))
+  val erasedType: Opt[ErasedValueType] = S(ErasedType.ValueLike(rsc = S(false), this))
```
LPTK, line 621: Why does the type alias symbol advertize that it's not a resource and that it's a `ValueLike` of itself? Seems that it would make more sense to have an `ErasedType.Alias` case that doesn't make these incorrect assumptions.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -367,22 +394,64 @@ class TermSymbol(val k: TermDefKind, val owner: Opt[InnerSymbol], val id: Tree.I
...
+    case S(tpe: ErasedValueType) => S(tpe)
     case N => N
 
   def subst(using sub: SymbolSubst): TermSymbol = sub.mapTermSym(this)
   def mayRaiseEffects(using Config) =
     defn.forall(_.mayRaiseEffects)
 
 object TermSymbol:
-  def fromFunBms(b: BlockMemberSymbol, owner: Opt[InnerSymbol], erasedType: Opt[ErasedType])(using State) =
+  def apply(
```
LPTK, line 446: What's the point of this function?!

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -345,11 +365,18 @@ sealed abstract class MemberSymbol(using State) extends Symbol:
...
+    val owner: Opt[InnerSymbol],
+    val id: Tree.Ident,
+    private var _erasure: Opt[ErasedValueType | ErasedFuncSignature],
+)(using State)
     extends MemberSymbol
     with DefinitionSymbol[TermDefinition]
-    with HasOnceMutableErasedType
     with NamedSymbol:
+  
+  private lazy val functionType: ErasedValueType = ErasedType.Function(N)
```
LPTK, line 378: What is such a lazy val returning a constant value doing here?

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
```diff
@@ -225,24 +226,43 @@ class SplitSymbol(val body: Split, name: Str = "split")(using State) extends Loc
...
   override def subst(using s: SymbolSubst): TempSymbol = s.mapTempSym(this)
+  
+  def erasedType: Opt[ErasedValueType] = _erasedType
+  
+  /** Sets the erased type, or raises a soft assertion if it is already set. */
+  def erasedType_=(newType: Opt[ErasedValueType])(using Line, FileName, Raise): Unit =
+    initErasure(this, _erasedType, newType)(_erasedType = _)
+
+object TempSymbol:
+  def apply(trm: Opt[Term], erasedType: Opt[ErasedValueType], dbgNme: Str = "tmp")(using State): TempSymbol =
```
LPTK, line 256: Why?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -64,6 +64,41 @@ object Elaborator:
...
+    * signature elaborated and resolved once for each, reporting every diagnostic in it twice. The signature is not
+    * shared with a definition that has type parameters of its own, as the signature may then refer to them.
+    */
+  def sharedSignatureDefinition(sym: BlockMemberSymbol): Opt[Tree.TermDef] =
+    val hasSeparateSignature = sym.trees.exists:
+      case td: Tree.TermDef => td.rhs.isEmpty && td.annotatedResultType.isDefined && td.paramLists.isEmpty
+      case _ => false
+    if !hasSeparateSignature then N
+    else sym.trees.collectFirst:
+      case td: Tree.TermDef if td.rhs.isDefined && td.annotatedResultType.isEmpty && td.typeParams.isEmpty => td
```
LPTK, line 100: This logic is *very* funky. Why not just a single `collectFirst` call?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -64,6 +64,41 @@ object Elaborator:
...
+        case _ => Sig.Result(sign)
+      go(sign, Nil)
+  
+  /** The definition of `sym` that shares the elaborated signature of its separately written declaration, if any.
+    *
+    * A declaration such as `fun f: Int -> Int` followed by a definition `fun f(x) = x` would otherwise have that
+    * signature elaborated and resolved once for each, reporting every diagnostic in it twice. The signature is not
+    * shared with a definition that has type parameters of its own, as the signature may then refer to them.
+    */
+  def sharedSignatureDefinition(sym: BlockMemberSymbol): Opt[Tree.TermDef] =
```
LPTK, line 94: This should probably be a lazy val in BlockMemberSymbol.

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -2011,13 +2056,20 @@ extends Importer:
...
+              def elabSignature = st.map:
                 // unwrap possible module modifier
                 // e.g, `fun f: module M`
                 //              ^^^^^^
                 case TypeDef(Mod, st, N) => term(st)(using newCtx)
                 case st => term(st)(using newCtx)
+              // * A declaration and the definition that consumes its signature share its elaboration, whichever comes
+              // * first in the block.
+              val sharesSignature = Elaborator.sharedSignatureDefinition(sym).exists: defn =>
+                (defn is td) || td.rhs.isEmpty && td.paramLists.isEmpty && td.annotatedResultType.isDefined
```
LPTK, line 2069: Aren't these things already checked in `sharedSignatureDefinition`?

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -2011,13 +2056,20 @@ extends Importer:
...
+              val s =
+                if sharesSignature then sharedSignatures.getOrElseUpdate(id.name, elabSignature)
```
LPTK, line 2071: This logic is just weird. It looks like it might still elaborate a signature more than once in cases like
```
fun f = 1
fun f: Int
```

hkmc2/shared/src/main/scala/hkmc2/semantics/Elaborator.scala
```diff
@@ -658,6 +693,23 @@ extends Importer:
         N
     case _ => N
   
+  /** Applies the resource modifier `kw`, written at `kwLoc` inside an instantiation (`new rsc C()`),
+    * to the elaborated instantiation `inst`. A modifier has no effect if written outside one (`rsc (new C())`).
```
LPTK, line 697: No effect? Or does it raise an error? (It should.)

hkmc2/shared/src/test/mlscript/codegen/ResourceTypes.mls
LPTK: How would this be acceptable?! Why would this crash the compiler? Compiler crashes are never "acceptable". This is nonsense.

This is **not** a hard problem. Elaborating a signature only once is not difficult. At the very least, just make the elaborator reject signatures defined after definitions. You also don't need to use this hacky hash map. But you should get the scopes right: the fact that the parameter of `f` is called `C` or something else should have absolutely no effect on the result.

hkmc2/shared/src/main/scala/hkmc2/syntax/Tree.scala
```diff
@@ -432,17 +458,37 @@ object Tree:
...
   
+  /** Matches a resource modifier keyword, typed as one. */
+  object RscModifier:
+    def unapply(t: Tree): Opt[Keywrd[Keyword.RscLike]] = t match
+      case kw @ Keywrd(rsc: Keyword.RscLike) => S(new Keywrd[Keyword.RscLike](rsc).withLocOf(kw))
+      case _ => N
+  
+  /** Splits the resource modifier off the body of a `new`.
+    *
+    * `rsc` parses looser than application, and `mut` tighter, so a `mut` written before `rsc` is moved back onto the
```
LPTK, line 469: Any reason `rsc` and `mut` have different precedences, leading to this weird logic?

hkmc2/shared/src/test/mlscript/codegen/ErasedTypes.mls
```diff
@@ -874,18 +874,14 @@ fun badInst() = useInt(new Bar("x"))
...
+// A reference to a `fun` is a `Function` of unresolved resource-ness (`rsc?`), so returning one at `Function`
+// takes a checked cast.
 fun test: Function = id
 //│ ———————————————| Lowered IR |———————————————————————————————————————————————————————————————————————
-//│ define test² as fun test³(): Function⁰ { return id¹ }; end
-
-fun test: Function = new Object
-//│ ———————————————| Lowered IR |———————————————————————————————————————————————————————————————————————
-//│ define test⁴ as fun test⁵(): Function⁰ { return (new globalThis⁰.Object⁰() as! Function⁰) }; end
+//│ define test² as fun test³(): Function⁰ { return (id¹ as! Function⁰) }; end
```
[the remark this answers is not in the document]
LPTK: Yes, why are we removing a test case here?

LPTK:
> * The test comment still describes the old double-elaboration behavior, but `sharedSignatureDefinition` now shares the signature and the golden output reports one diagnostic. Update the comment to match the behavior this regression test now exercises.
>   **hkmc2/shared/src/test/mlscript/codegen/ResourceModifierChecks.mls:738**
> * This comment still says the diagnostic is duplicated, while the shared-signature path and golden output now report it once. Update the explanation so the test documents its current behavior.

These "suppressed comments" do seem relevant.

hkmc2/shared/src/test/mlscript/codegen/ErasedTypes.mls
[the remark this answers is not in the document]
Derppening: Reinstated in e30d6624a5c06ae045b2e71b52568a2af0ee3e05. I think I was just confused on what the test case was showing and thought the newly added test replaces the old one.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
Derppening: Removed `TypeAliasSymbol.erasedType` entirely in 5edf1f3c4c63858738f5025dee33f31f521e7f23 - they can never be used as a value anyways so the field doesn't make sense. In type position they still canonicalize to the LUB of its members so that isn't changed.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
Derppening: Fixed (by inlining to its only use-site) in 53788aa4f45e47ea1f35ecaa1599d3d0ad0cc3a5.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
Derppening: Same reasoning as this comment: https://github.com/hkust-taco/mlscript/pull/560#discussion_r4074490609.

Fixed in 8313923861eee2ff5d6a2fededc2d01f56877e41 by renaming the parameter to `initErasedType`.

hkmc2/shared/src/main/scala/hkmc2/semantics/Symbol.scala
Derppening: My original intention was to be consistent and have all symbols (that has an erased type) take an `erasedType` parameter, but `TermSymbol` (and `TempSymbol`)'s erased type may be late-initialized (and thus its setter needs to be overridden to ensure that the erased type is written exactly once), so I decided the best way was to hide the primary constructor and instead have an `apply` function to match this pattern.

Addressed in 4394a4ee10868d6a3223adb50500d9e5fe92bdff - I just renamed the parameter to `erasure` so it doesn't conflict with `erasedType`.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
Derppening: Done in e38e8f1a13086891afc760ea21748c49c5bbd1fb. There are several minor designs that I made:

- `Result.showDbg` for lambdas display its resource-ness in a second parameter list (e.g. `Lambda(...)(rsc = true)`)
- I also cleaned up some unused patterns in the same commit, e.g. `case Lambda(params, body)` where both destructured fields are unused to `case l: Lambda`

One other thing: `Instantiate` still has its `rsc` in the first paramlist. Should I move it to the second paramlist? If so, should I also move `mut` alongside it?

LPTK: Yes, I think it makes sense to move both.

hkmc2/shared/src/main/scala/hkmc2/codegen/ErasedType.scala
```diff
@@ -581,7 +581,10 @@ extension (s: ValueSymbol | DefinitionSymbol[?])
     */
   def mapErasedValueType(using Raise): Opt[ErasedValueType] = s match
     case l: LocalVarSymbol => l.erasedType
-    case c: (ClassSymbol | ModuleOrObjectSymbol) => c.erasedType
+    // * A class symbol may appear as a value as a reference to `this` in a class, and so carries the erased type of its
+    // * own class, but we don't have enough information to determine its resource-ness.
+    case c: ClassSymbol => S(ErasedType.ValueLike(rsc = N, c))
```
LPTK, line 586: This is extremely fishy. What does "may appear as a value as a reference to `this`" even mean? It's no longer true in the IR.

LPTK: Sounds like this "helpful" extension method just sits at the wrong abstraction level and should be removed.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -30,11 +30,19 @@ object Lifter:
...
+    /** Maps the definition to the erased type of its instances.
+      *
+      * Note that the resource-ness of the returned type is marked as `rsc?` since we currently do not have information
+      * about the resource-ness of lifted classes (although we mark its `Instantiate` as `rsc = false` as a
+      * placeholder).
+      */
     private def instanceType(using Raise): Opt[ErasedValueType] = d.isym match
-      case cls: ClassLikeSymbol => cls.erasedType
+      case cls: ClassSymbol =>
+        // * We can construct the `AnyRef` directly since lifted classes are neither `Anything` nor a primitive, which
```
LPTK, line 41: What? Why aren't lifted classes `Anything`?

hkmc2/shared/src/test/mlscript/codegen/ErasedTypes.mls
```diff
@@ -303,7 +303,7 @@ resetCounter(41)
 //│   }
 //│ };
 //│ define resetCounter⁰ as fun resetCounter¹(x: Int⁰): Int⁰ {
-//│   let resetCounter$cap: Capture$resetCounter⁰;
+//│   let resetCounter$cap: rsc? Capture$resetCounter⁰;
```
LPTK, line 306: I think the lifter should be able to make this more precise (either `rsc` or not, `!rsc`), given that it can just check what's the resourceness of the instantiation (for now always `!rsc`, but later it will be based on the captures).

hkmc2/shared/src/test/mlscript/codegen/ErasedTypes.mls
```diff
@@ -881,6 +881,13 @@ fun test: Function = id
 //│ define test² as fun test³(): Function⁰ { return (id¹ as! Function⁰) }; end
 
 
+// `new Object` creates a non-resource instance of `Object`, so a cast is inserted to coerce it to `Function`.
+// Note that this cast has nothing to do with resource-ness - both `Object` and `Function` are non-resource types.
```
LPTK, line 885: This comment seems completely incongruous.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
Derppening: Lifted classes are subtypes of `Anything`, not `Anything` itself.

I'll reword the comment to actually state that lifted classes are already canonical and therefore can be directly constructed with `ErasedType.AnyRef`.

hkmc2/shared/src/test/mlscript/codegen/ErasedTypes.mls
Derppening: Tried to address it in 045b3af219597ef6df00ff4126a962207d2804e4 - it now describes that a `Cast` node is inserted because `Function <: Object`.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
Derppening: Moved `Instantiate`'s `mut` and `rsc` into the second paramlist in 25a55cae77bfdfd416ba7a447b074176459563e7. One follow-up from this: now that `Instantiate` and `Lambda`'s `rsc` (and `mut` in `Instantiate`'s case) exist in the second param list, it makes sense to also override `equals` for them to also take those fields into comparison, right?

hkmc2/shared/src/main/scala/hkmc2/codegen/ErasedType.scala
Derppening: Removed in b368512281ae2d0cd2f6b93e61c5f67b622ebf87. All its uses are converted so that they only match symbol types that can exist at that location.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
Derppening: Obsolete - removed as part of b368512281ae2d0cd2f6b93e61c5f67b622ebf87.

hkmc2/shared/src/test/mlscript/codegen/ErasedTypes.mls
Derppening: Addressed in 1562237bac1e9ce2e689ca915c843c2e9a98bcbb. The resourceness of the instantiation is now used to determine the resourceness of the lifted class.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
```diff
@@ -693,14 +680,10 @@ class Lifter(topLevelBlk: Block)(using State, Raise, Config):
...
         Instantiate(
-          mut = true,
-          // TODO(Derppening): This needs to be determined from everything that the capture class *captures*, which is
-          //                   the responsibility of the new resolver(?)
-          rsc = false,
           captureClass.sym.asMemberRef(captureClass.isym),
           captureInfo._2.map(
             (sym, _) => sym.asPath.asArg) :: Nil
-        )(InstantiateMetadata.empty)
+        )(InstantiateMetadata.empty, mut = true, rsc = false)
```
LPTK: You removed the TODO, but this isn't fixed!

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Rewrite.scala
```diff
@@ -545,6 +545,16 @@ class DeforestRewriter(val solver: DeforestFusionSolver)(using Raise):
               N, bms, tSym, refreshedParams,
               bodyWithCorrectSymbols)(N, PrivateModifier :: fDefn.annotations)
         end newPolyFuns
+
+        def refreshedFvSymErasedType(s: ValueSymbol): Opt[ErasedValueType] = s match
+          case l: LocalVarSymbol => l.erasedType
+          // * `FreeVarTraverser` drops a `MemberRef`'s `disamb`, so BMSes have no resolvable erased type.'
```
LPTK: This, again, is hinting at a design problem that needs a fix, not a workarounds. Are you really just dropping erased types in cases like this? Because it looks like you are.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
LPTK: Yes, sure, pls do.

hkmc2/shared/src/main/scala/hkmc2/codegen/ErasedType.scala
LPTK: The logic there with all the soft asserts is quite messy.

hkmc2/shared/src/test/mlscript/codegen/ErasedTypes.mls
LPTK: The cast is not inserted because the former is a subtype of the latter, but rather because the latter is NOT a subtype of the former.

Derppening: Reworded it once again in 3837043241e591ed71c1143a6cc3da9e08601010.

hkmc2/shared/src/main/scala/hkmc2/codegen/Lifter.scala
Derppening: Reinstated the TODO with an updated comment in f72b28f69dc3f36eb16375ff6f82325928e59432.

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Rewrite.scala
Derppening: Fixed in 7c0e7728d17aab4aad4fe5a1ffaca2c487a2bd51, `disamb` of BMSes are now preserved by making the rewriter capture references (`Value.Ref`) as opposed to just symbols.

hkmc2/shared/src/main/scala/hkmc2/codegen/ErasedType.scala
Derppening: I tried to clean it up in 355823e57ea1bb4779ac008d21fc981a4bd0150a and a7ec922ab2ff1d02cff4d18cc8412cf3a55ab772.

I didn't get very far with tightening up the symbol types for `reqPassedSyms` and `reqDefns`. Perhaps I can look into those again after addressing the rest of the feedback.

hkmc2/shared/src/main/scala/hkmc2/codegen/Block.scala
Derppening: Done in 5c3913f7909a3e5395de868233c11db3fd6ddde0.

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Rewrite.scala
```diff
@@ -546,15 +546,11 @@ class DeforestRewriter(val solver: DeforestFusionSolver)(using Raise):
...
-          case _ =>
-            softAssert(
-              false,
-              s"Expected refreshed function symbols to contain only LocalVarSymbols or BlockMemberSymbols, got $s")
-            N
+        // * Enforce that all references to the same symbol have the same `disamb`.
+        def checkFvDisambs(fvs: Ls[Value.Ref]): Unit =
+          fvs.groupBy(_.symbol).foreach: (sym, refs) =>
+            softAssert(refs.distinct.sizeIs == 1,
+              s"Free variable $sym is referenced with different disambiguations: ${refs.distinct}")
```
LPTK: What is this for? It makes no sense.

hkmc2/shared/src/main/scala/hkmc2/codegen/ErasedType.scala
LPTK: Sure.

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
```diff
@@ -393,10 +393,13 @@ class UsedVarAnalyzer(b: Block, scopeData: ScopeData)(using State):
...
             for l <- reads do
-              if hasMutator.contains(l) then
-                reqCapture += l
-              if mutated.contains(l) && !linearValueVars.contains(l) then
-                reqCapture += l
+              l match
+                case l: LocalVarSymbol =>
+                  if hasMutator.contains(l) then
+                    reqCapture += l
+                  if mutated.contains(l) && !linearValueVars.contains(l) then
```
LPTK: Why isn't this an `||`?

Derppening: If by "this" you mean the two identical `if` bodies, I am not sure - I just added the `match` and kept the original body.

I'll fold them into a single `if` if you'd like.

hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Rewrite.scala
Derppening: That's my mistake, I trusted Claude that this assertion does something when it fact it doesn't. :man_facepalming: 

Removed in 111f89a47ba3e6cc2912ad4d99bb902a61b519ba.

hkmc2/shared/src/main/scala/hkmc2/codegen/UsedVarAnalyzer.scala
LPTK: Why would you keep them separate? That makes no sense. It duplicates the `reqCapture += l` and means it might be executed twice!

Derppening: Folded into a single `if` in eca0519f28514eeb19d30478117875680bc3b242.

PR #561, opened 2026-09-18
Merged flow based rewriting pass

ychenfo:
- Add the merged pass
- Move the shared logic between deforest and the merged pass into `PolyInstantiationRewrite`
- Add dead ctor elim
- Make eta expansion support polymorphic analysis result and add manual @affine annot for testing
- Update config to use only one `flowBasedOpt`
- Minor fixes on opaque constraints
- Add `:showPipeline` for checking program changes across different pass more easily


hkmc2/shared/src/main/scala/hkmc2/codegen/deforest/Rewrite.scala
ychenfo, line 1: It might be helpful to use "Hide whitespaces" to review the diff of this file because of indentation changes of a large chunk of code.

hkmc2/shared/src/test/mlscript/dead-param-elim/clash.mls
```diff
@@ -104,7 +104,7 @@ private fun g(a, b, c, ...rest) = rest
 (if false then f else g)(1,2,3,4,5)
 //│ dead-param-elim > >>> dead-param-elim results >>>
 //│ dead-param-elim > prodfun f#0 @ <root> -> eliminable: {0, 1}
-//│ dead-param-elim > prodfun g#0 @ <root> -> eliminable: {0, 1, 2}
+//│ dead-param-elim > prodfun g#0 @ <root> -> eliminable: {0, 1}
```
ychenfo: Before this PR, dead param elimination sits after a block simplifier pass, so DPE saw
```
if false then f(1,2,3,4,5)
else g(1,2,3,4,5)
```
and was able to eliminate {0,1,2} from g.

Now DPE sees
```
let tmp
if false then tmp = f; tmp(1,2,3,4,5)
else tmp = g; tmp(1,2,3,4,5)
```
so both `f` and `g` only get {0,1} eliminated.

hkmc2/shared/src/main/scala/hkmc2/codegen/DeadConstructorElim.scala
```diff
@@ -0,0 +1,94 @@
...
+          def onlyStoresParams(b: Block): Bool = b match
+            case AssignField(Value.This(self), _, Value.SimpleRef(p), rest) if self is cls.isym =>
+              onlyStoresParams(rest)
+            case Define(ValDefn(_, _, Value.SimpleRef(p)), rest) =>
+              onlyStoresParams(rest)
+            case End(_) => true
+            case _ => false
+          onlyStoresParams(cls.preCtor) && onlyStoresParams(cls.ctor)
+        end hasSimpleCtor
+        cls.irClsLikeDefn.exists(hasSimpleCtor)
```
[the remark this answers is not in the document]
LPTK: Copilot is right to be concerned about purity, but the solution is to use `Assign.discard` of the non-pure arguments, not to prevent the ctor deletion.

hkmc2/shared/src/main/scala/hkmc2/codegen/PolyInstantiationRewrite.scala
```diff
@@ -0,0 +1,142 @@
...
+  
+  
+  final def apply(): Program =
+    if !solvers.exists(_.hasWorkToDo) then pre.pgrm
+    else
+      val newBody = mkNewProgramBody
+      if newBody is pre.pgrm.main then pre.pgrm
+      else Program(pre.pgrm.imports, newBody)
+  
+  val newPolyFnSyms : LinkedHashMap[InstantiationId, Map[TermSymbol, (BlockMemberSymbol, TermSymbol)]] =
```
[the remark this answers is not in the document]
ychenfo: This is not true, the things that `solvers` refers to are class arguments and are initialized before `newPolyFnSyms` is computed. If copilot was true no test would have passed (we would instead get dereferencing null error all the time).

hkmc2/shared/src/main/scala/hkmc2/codegen/FlowAnalysisBasedRewrite.scala
```diff
@@ -0,0 +1,320 @@
...
+            case arg :: tl =>
+              applyArg(arg): arg2 =>
+                rec(tl, idx + 1, changed || !(arg2 is arg), arg2 :: accRev)
+          rec(args, 0, false, Nil)
+      end rewriteArgs
+      
+      r match
+      case ctorSite@CtorProducer(_, _, _)
+        if deadConstructorElimSolver.deadCtors.contains(ConcreteId(ctorSite.uid, instId)) =>
+        k(Value.Lit(Tree.UnitLit(false)).withLocOf(ctorSite))
```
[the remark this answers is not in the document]
ychenfo: The same issue as the first comment, fixed.

hkmc2/shared/src/main/scala/hkmc2/codegen/FlowAnalysisBasedRewrite.scala
```diff
@@ -0,0 +1,320 @@
...
+      
+      def eliminableParamsOf(targets: EtaTargets): Set[Int] =
+        val arities = targets.prodFuns.map(pf => pf.params.size + pf.restParam.fold(0)(_ => 1))
+        assert(arities.forall(_ === targets.paramCount),
+          s"eta expansion level disagrees with its targets on arity: ${targets.pp} -> $arities")
+        val eliminable = targets.prodFuns.map: target =>
+          deadParamElimSolver.eliminableParams(target.concreteId)
+        assert(eliminable.size <= 1,
+          s"eta expansion targets disagree on eliminable parameters: " +
+          s"${targets.pp} -> $eliminable")
```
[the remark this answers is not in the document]
ychenfo: I wrote this assertion because the invariant should hold since DPE makes sure two functions flowing into the same point should have the same set of eliminable parameters via the web.

But then I found this edge case where two functions flow into the same point that *is never called* so DPE never sees the fact that these two functions flow into the same point, and then this assertion would fail. I fixed this by disallowing eta expansion for things that are never called. Other than this case I believe the assertion should always hold.

```
@affine(0) private fun g1(x) = 0
@affine(0) private fun h1(x) = x
private fun f1(b) = if b then g1 else h1
private fun main1(b) =
  f1(b) // <-- never called
  g1(1) + h1(2)
main1(true)
```

hkmc2/shared/src/main/scala/hkmc2/codegen/DeadConstructorElim.scala
```diff
@@ -0,0 +1,94 @@
...
+          case UnknownCons => true
+          case _: FieldSel | _: Dtor => true
+      .map(_.concreteId)
+      .toSet
+    
+    LinkedHashSet.from:
+      allCtorStrats.iterator
+        .filter(c => isRemovable(c.exprId.getResult))
+        .map(_.concreteId)
+        .filterNot(liveCtorConcreteIds)
```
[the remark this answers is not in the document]
ychenfo: This is left as a TODO in the test, the flow analysis for class constructor bodies that I am working on should resolve the issue in the next PR.

hkmc2/shared/src/main/scala/hkmc2/codegen/DeadConstructorElim.scala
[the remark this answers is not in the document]
ychenfo: Indeed, I have pushed the fix.

hkmc2/shared/src/main/scala/hkmc2/codegen/FlowAnalysisBasedRewrite.scala
```diff
@@ -0,0 +1,323 @@
...
+            case arg :: tl =>
+              applyArg(arg): arg2 =>
+                rec(tl, idx + 1, changed || !(arg2 is arg), arg2 :: accRev)
+          rec(args, 0, false, Nil)
+      end rewriteArgs
+      
+      r match
+      case ctorSite@CtorProducer(_, args, selectedFrom)
+        if deadConstructorElimSolver.deadCtors.contains(ConcreteId(ctorSite.uid, instId)) =>
+        (selectedFrom.toList ::: args.map(_.value))
```
LPTK, line 165: Don't create useless intermediate lists... Use `reverseIterator`, `++ selectedFrom` (without `toList`), and `foldLeft`.

hkmc2/shared/src/main/scala/hkmc2/codegen/DeadConstructorElim.scala
```diff
@@ -0,0 +1,94 @@
...
```
LPTK, line 74: [Astra] **[P1] Preserve arguments observed by constructor bodies**

Discarding a constructor's result does not make its arguments dead. The simple-constructor check protects `Consumer` itself, but not the nested `Observed` allocation below: this becomes `Consumer(undefined)` and throws instead of returning `42`. Constructor-body uses need to constrain the actual arguments; until that flow is modeled, the affected arguments must be retained conservatively. Leaving this as a TODO for a later PR is unsafe while constructor elimination is enabled by default.

Regression: [`consume()`](https://github.com/hkust-taco/mlscript/blob/d05a78f235f31739200602380ab03524e1f4ca24/hkmc2/shared/src/test/mlscript/opt/DeadConstructorElim.mls#L424-L435). The retained test has `:fixme`; omit it to reproduce the failure:

```mlscript
:js
:noInline
:flowBasedOpt mono
:expect 42
data class Observed(x)
class Consumer(x) with
  if x is Observed then () else throw "expected Observed"
fun consume() =
  let unused = Consumer(Observed(1))
  42
consume()
```

The same program returns `42` with `:flowBasedOpt off`.

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -952,19 +976,20 @@ class FlowConstraintsCollector(
           case cls: ClassSymbol =>
             cls.tree.clsParams.size match
             case 1 =>
-              val clsParams = cls.tree.clsParams.head
+              val clsParams = cls.tree.clsParams.head // TODO use irDefn here and elsewhere
               // TODO: properly check the parameter lists, which may change after passes like lifting
               // softTODO(argsStrat.size === clsParams.size, s"mismatched ctor arg and cls param sizes")
-              new Ctor(c.uid, instId)(ctor, clsParams.zip(argsStrat))
+              registerCtor(new Ctor(c.uid, instId)(ctor, clsParams.zip(argsStrat)))
```
LPTK, line 982: [Astra] **[P1] Account for constructor arguments added by lifting**

The source parameter list excludes capture parameters added by Lifter, so `clsParams.zip(argsStrat)` silently drops those actual arguments from the flow graph. Here the captured tuple remains live through `get()`, but its allocation is removed and the generated code calls `new Capturing(0, undefined)`. Use the current IR parameter/field mapping, or conservatively constrain unmatched arguments as live.

Regression: [`capture()`](https://github.com/hkust-taco/mlscript/blob/d05a78f235f31739200602380ab03524e1f4ca24/hkmc2/shared/src/test/mlscript/opt/DeadConstructorElim.mls#L472-L484). Without its `:fixme` marker:

```mlscript
:js
:noInline
:flowBasedOpt mono
:lift
:expect [1, 2]
fun capture() =
  let captured = [1, 2]
  class Capturing(x) with
    fun get() = captured
  let instance = Capturing(0)
  instance.get()
capture()
```

The result is rendered as `()` instead of `[1, 2]`; disabling `flowBasedOpt` restores the expected result.

hkmc2/shared/src/main/scala/hkmc2/codegen/DeadConstructorElim.scala
```diff
@@ -0,0 +1,94 @@
...
+      case _ => false
+    
+    val allCtorStrats = collector.allRealCtors.filter(_.instantiationId.isDefined)
+    
+    val liveCtorConcreteIds = allCtorStrats.iterator
+      .filter: c =>
+        c.dests.exists:
+          case NonAffine | Accumulator => false
+          case UnknownCons => true
+          case _: FieldSel | _: Dtor => true
```
LPTK, line 66: [Astra] **[P1] Treat checked casts as uses of allocations**

This liveness calculation relies on consumer edges, but `FlowAnalysis.processResult` currently handles `Cast(value, _, _)` by returning the operand's flow without recording the runtime check. An allocation used only by an otherwise-unused checked cast is therefore removed. The Wasm program below changes into `undefined as! Box` and traps with `RuntimeError: illegal cast` instead of returning `42`. A checked cast must keep its operand live even when its result is discarded.

Regression: [`first cast() case`](https://github.com/hkust-taco/mlscript/blob/d05a78f235f31739200602380ab03524e1f4ca24/hkmc2/shared/src/test/mlscript/wasm/DeadConstructorElim.mls#L9-L17). A [second case](https://github.com/hkust-taco/mlscript/blob/d05a78f235f31739200602380ab03524e1f4ca24/hkmc2/shared/src/test/mlscript/wasm/DeadConstructorElim.mls#L33-L42) covers the same flow through an intermediate `val`. Without the retained `:fixme` marker:

```mlscript
:global
:wasm
:noInline
:flowBasedOpt mono
data class Box(x)
fun cast() =
  let x = Box(1)
  val unused: Box = x as Object
  42
cast()
```

With `:flowBasedOpt off`, the first case returns `42` successfully.

hkmc2/shared/src/main/scala/hkmc2/codegen/FlowAnalysisBasedRewrite.scala
```diff
@@ -0,0 +1,323 @@
...
+        if targetShape.map(_.paramCount).startsWith(existingShape) then
+          targetShape.drop(existingShape.size).zipWithIndex.map:
+            case (targets, idx) =>
+              val eliminable = eliminableParamsOf(targets)
+              val params = (0 until targets.paramCount).iterator.filterNot(eliminable).map: i =>
+                Param.simple(new VarSymbol(new Tree.Ident(s"eta$$$idx$$$i"), erasedType = N))
+              .toList
+              EtaParamList(
+                ParamList(ParamListFlags.empty, params, N),
+                params.map(p => Arg(N, p.sym.asSimpleRef)),
```
LPTK, line 97: [Astra] **[P2] Preserve variadic parameters during eta expansion**

`EtaTargets` counts a rest parameter as one parameter, but the synthesized list always has `restParam = N` and forwards only ordinary arguments. The newly supported `@affine(1)` annotation therefore allows this wrapper to silently truncate `(1, 2, 3)` to `(1)`. The returned function is called exactly once, so the annotation's one-shot promise is satisfied. Preserve and spread the rest parameter, or exclude variadic targets from eta expansion.

Regression: [`returnRest`](https://github.com/hkust-taco/mlscript/blob/d05a78f235f31739200602380ab03524e1f4ca24/hkmc2/shared/src/test/mlscript/opt/EtaExpansion.mls#L226-L233). Without its `:fixme` marker:

```mlscript
:js
:noInline
:flowBasedOpt mono
:expect [1, 2, 3]
@affine(1) fun rest(a)(...xs) = xs
fun returnRest(a) = rest(a)
returnRest(0)(1, 2, 3)
```

This returns `[1]`; with `:flowBasedOpt off`, it returns `[1, 2, 3]`.

LPTK (changes requested):
[Astra] Changes requested: four reproduced correctness issues remain in constructor elimination and variadic eta expansion. Each inline comment links to its regression test and includes a standalone reproducer. The tests and golden outputs are retained with `:fixme` in commit `d05a78f23`. Validation performed during the review: `ctest` and `hkmc2AllTests/test` passed with the known failures recorded, and each of the four original reproducers succeeded with `flowBasedOpt` disabled.

hkmc2/shared/src/main/scala/hkmc2/codegen/FlowAnalysisBasedRewrite.scala
ychenfo: Fixed in https://github.com/hkust-taco/mlscript/pull/561/commits/543e8e0757f7d708c889f7b795b92bf43479eb1b

hkmc2/shared/src/main/scala/hkmc2/codegen/DeadConstructorElim.scala
ychenfo: Fixed in https://github.com/hkust-taco/mlscript/pull/561/commits/402554c6f30e2b32ddcf03943812aec52334b7a6

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
ychenfo: Will fix in a follow up PR that I am working on to properly track the flow of class constructors. That PR will also get this information properly using the stored lowered IR in a symbol instead of using `tree.clsParams` and blindly doing `zip`.

hkmc2/shared/src/main/scala/hkmc2/codegen/DeadConstructorElim.scala
ychenfo: Will fix in a follow up PR that I am working on to properly track the flow of class constructors.

hkmc2/shared/src/main/scala/hkmc2/codegen/FlowAnalysisBasedRewrite.scala
ychenfo: Fixed in https://github.com/hkust-taco/mlscript/pull/561/commits/376e849afd7213264a347d1bdf21643fa6de824b

ychenfo:
I also came to realize that the side effect of dead-parameter-eliminated args was not taken care of previously, fixes are in https://github.com/hkust-taco/mlscript/pull/561/commits/edc09ad2acff112a6a89d819d3674063aa9fa0d3 and https://github.com/hkust-taco/mlscript/pull/561/commits/0924c23aff8ae23ed7e2b9ed19bcfb20957ec394

PR #562, opened 2026-09-20
Effect Analysis Prototype

flandia:
This artifact is endorsed by Anto


flandia:
wtf Codex stole my authorship

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -60,10 +60,34 @@ object FlowAnalysis:
     mono: Bool,
     nonAffineTracking: Bool,
     accumulatorTracking: Bool,
+  )(using TraceLogger, Elaborator.State, Raise, SymbolPrinter): FlowConstraintSolver =
```
LPTK, line 63: 👎

LPTK:
Thanks!

As discussed,
* let's add a flag to `Assign` nodes to determine their effectfulness. We should make sure not to treat local mutation as a nonlocal side effect;
* do not use an effectful annotation (just use actual effects instead);
* make the optimizer not run when the lifter is disabled, so we don't need to care about nested definitions;
* merge this into the eta expansion pass (later: the **megapass**).

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -922,18 +972,45 @@ class FlowConstraintsCollector(
...
+      def handleCallLike(
+        callExprId: ResultId,
+        f: Path,
+        args: List[Arg],
+        mayRaiseEffects: Bool,
+      ): ProdStrat =
         val fStrat = processResult(f)
         val argsStrat = args.map(a => processResult(a.value))
         if args.exists(_.spread.isDefined) then
+          // TODO: preserve callee-effect precision for spread calls without modeling their argument flow.
```
LPTK, line 999: @ychenfo is this the same bug as discussed?

hkmc2/shared/src/main/scala/hkmc2/codegen/flowAnalysis/FlowAnalysis.scala
```diff
@@ -922,18 +972,45 @@ class FlowConstraintsCollector(
...
       val instId = cc.instId
-      def handleCallLike(callExprId: ResultId, f: Path, args: List[Arg]): ProdStrat =
+      def freshCallEffect(): StratVar =
+        val effect = freshVar("call_effect", cc.forFunGroup)
+        // A call's effect contributes to the enclosing function's latent effect.
+        currentEffect.foreach: enclosingEffect =>
+          cc.constrain(effect, enclosingEffect)
+        effect
+      def constrainConstructorEffect(result: Result): Unit =
+        val mayRaise = result match
```
LPTK, line 984:
```suggestion
        val mayHaveSideEffects = result match
```

hkmc2/shared/src/main/scala/hkmc2/codegen/CompilationPipeline.scala
```diff
@@ -46,6 +46,11 @@ class CompilationPipeline(using Config, Raise, State, Ctx, SymbolPrinter):
       if config.liftDefns.isDefined then
         blockPass(Lifter(_).transform)(prog)
       else prog
+    runPass("EffectAnalysis"): prog =>
```
LPTK, line 49: What si this extra pass for?

PR #563, opened 2026-09-22
Improve memory use in `BlockSimplifier`

NeilKleistGao:
The main reason we have an out-of-memory problem is that we cache local variables for each block:
```scala
def default =
  block.subBlocks.iterator.flatMap(analyze)
```
A local variable with 20 outer block nodes is cached 20 times.
But later we only collect all local variables in the whole program:
```scala
val vp = new DataFlowAnalysis(LocalVars.analyze(res.main))
```
So those caches for intermediate nodes are not used.


hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -127,22 +127,29 @@ class BlockSimplifier
...
+      val locals: MutSet[LocalVar] = MutSet.empty[LocalVar]
+      // * java.util.IdentityHashMap compares keys by **object identiy (eq)**,
+      // * which is faster than doing by structure equality (equals)
+      // * Values in the map must be nullable so we use java.lang.Boolean here.
+      val visited = java.util.Collections.newSetFromMap(
+        new java.util.IdentityHashMap[Block, java.lang.Boolean]())
+      def paramsOf(paramLists: IterableOnce[ParamList]): Unit =
+        locals ++= paramLists.iterator.flatMap(_.paramSyms)
+      def rec(current: Block): Unit =
+        if visited.add(current) then
```
LPTK: Why are we using a `visited` set? Blocks can't be cyclic.

LPTK (changes requested):
Thanks!

If we no longer want to cache this computation, why would we still define it in a `CachedAnalysis`? I guess we may still want to cache it at the top level, but then this should be clearly stated, otherwise there's a risk of confusion from this fringe use of `CachedAnalysis`.

Also, perhaps the actual problem is that the hashmap of `CachedAnalysis` should be a weak hashmap, to avoid retaining old memory? Indeed, this problem is not specific to this analysis. It's exacerbated by the `BlockSimplifier` now being long-lived (which wasn't originally the case).

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
```diff
@@ -127,22 +127,23 @@ class BlockSimplifier
...
-          paramsOf(cd.companion.iterator.flatMap(_.methods).flatMap(_.params)) ++
-          default).toSet
-      case Scoped(syms, rest) =>
-        (rest.analyze.iterator ++ syms.iterator.collect { case v: LocalVar => v }).toSet
-      case _ => default.toSet
+      val locals: MutSet[LocalVar] = MutSet.empty[LocalVar]
+      def paramsOf(paramLists: IterableOnce[ParamList]): Unit =
+        locals ++= paramLists.iterator.flatMap(_.paramSyms)
+      def rec(current: Block): Unit =
+          current match
```
LPTK: There's too much indentation now.

NeilKleistGao:
> If we no longer want to cache this computation, why would we still define it in a CachedAnalysis?

I think we no longer need the cache for `LocalVars`, since its result is consumed immediately.

> Also, perhaps the actual problem is that the hashmap of CachedAnalysis should be a weak hashmap, to avoid retaining old memory?

I asked GPT to profile memory usage for both `IdentityHashMap` and `WeakHashMap`. It seems that there is no significant improvement if we use a weak hash map. Furthermore, `java.util.WeakHashMap` is not supported by the current version of `Scala.js`.  `hkmc2JS/fullLinkJS` will report the following problem (meanwhile, the main tests can pass normally):

> Referring to non-existent class java.util.WeakHashMap

Another proposed change is to make objects like `PostCondAnalysis` classes and only use them in `JSBuilder` :
```scala
class JSBuilder(...) extends CodeBuilder:
  import JSBuilder.*
  private given PostCondAnalysis = new PostCondAnalysis
```
So that cached entries can be freed when they are no longer used.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
NeilKleistGao: It is redundant here. Already removed.

hkmc2/shared/src/main/scala/hkmc2/codegen/BlockSimplifier.scala
NeilKleistGao: Fixed.

LPTK:
> I think we no longer need the cache for `LocalVars`, since its result is consumed immediately.

Ok, but then there's no reason to still leave it in an object.

> Another proposed change is to make objects like `PostCondAnalysis` classes and only use them in `JSBuilder` 

Wow!! I had no idea about this obvious memory leak (`PostCondAnalysis` keepign all results for the entirety of the application's run). Cc @CAG2Mark. Let's make it a class, yes.

CAG2Mark:
> > I think we no longer need the cache for `LocalVars`, since its result is consumed immediately.
> 
> Ok, but then there's no reason to still leave it in an object.
> 
> > Another proposed change is to make objects like `PostCondAnalysis` classes and only use them in `JSBuilder` 
> 
> Wow!! I had no idea about this obvious memory leak (`PostCondAnalysis` keepign all results for the entirety of the application's run). Cc @CAG2Mark. Let's make it a class, yes.

Wow how did I not notice that 😅 

Yes let's do that

NeilKleistGao:
> > I think we no longer need the cache for `LocalVars`, since its result is consumed immediately.
> 
> Ok, but then there's no reason to still leave it in an object.
> 
> > Another proposed change is to make objects like `PostCondAnalysis` classes and only use them in `JSBuilder`
> 
> Wow!! I had no idea about this obvious memory leak (`PostCondAnalysis` keeping all results for the entirety of the application's run). Cc @CAG2Mark. Let's make it a class, yes.

Both are done.

LPTK:
Dude. The object was useless, but instead of removing it, you made it a useless class instead. Can you try to think your changes through a bit more? @NeilKleistGao

hkmc2/shared/src/main/scala/hkmc2/codegen/SpecializedSwitch.scala
```diff
@@ -160,8 +160,13 @@ private object PostCondAnalysisImpl extends CachedAnalysis[Block, PostCondRes]:
       case f: FunDefn => analyze(rest)
     case b: BlockTail => PostCondRes.empty.copy(isAbortive = b.isAbortive)
 
-private object PostCondAnalysis extends CachedAnalysis[Block, Map[ValueSymbol, Literal]]:
-  override def analyzeUncached(b: Block): Map[ValueSymbol, Literal] = PostCondAnalysisImpl.analyze(b).varsMap
+end PostCondAnalysisImpl
+
+private[codegen] final class PostCondAnalysis extends CachedAnalysis[Block, Map[ValueSymbol, Literal]]:
+  private val impl = new PostCondAnalysisImpl
```
LPTK: Why? What does this separation achieve?

LPTK: To be more specific, I don't think we need two layers of caching here. But maybe there's a good reason for it? Cc @CAG2Mark

CAG2Mark: If I recall it was just to discard some data the user of the analysis might not find relevant, but it's probably not that important

NeilKleistGao: Then I guess I can merge these two classes, and I think I should cache `PostCondRes` instead of `Map[ValueSymbol, Literal]`?

LPTK: You should cache whatever it is that is being recursively computed, and probably no need to cache more.

LPTK (approved):
LGTM, thanks!

